PR #1, opened 2015-11-18
Validator Cleanup

abrown-sg:
* Started process of moving validators out to their own file. This will likely include others down the road.

* Fixed logic error that has been preventing trailing whitespace and newline validators from running at all.

* Added tests.

### DIFF
--- a/molecule/core.py
+++ b/molecule/core.py
@@ -28,7 +28,6 @@ import sys
 import termios
 from subprocess import CalledProcessError
 
-import molecule.validators as validators
 import prettytable
 import sh
 import vagrant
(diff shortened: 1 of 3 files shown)


molecule/core.py
```diff
@@ -28,6 +28,7 @@
 import termios
 from subprocess import CalledProcessError
 
+import molecule.validators as validators
```
retr0h: Need to fix this import :)

retr0h:
:+1: 

This is exactly the kind of cleanup I like to see.  Nice work breaking it out and adding tests.  I assume at some point we will want to allow for custom validators.  Similar to how giftwrap handled custom [builders](https://github.com/blueboxgroup/giftwrap/commit/f2d79a0db952280bde4910cbcea00dfc1be40e70).

rgreinho:
:+1: especially for the tests :)

PR #5, opened 2015-11-18
No need for test constants

retr0h:


tests/test_molecule.py
```diff
@@ -72,8 +72,10 @@ def setUp(self):
 
     def test_parse_provisioning_output_failure_00(self):
         res = self.molecule._parse_provisioning_output(TestStringMethods.OUTPUT_MIXED_FAILED)
```
abrown-sg: Could probably move these constants too?

abrown-sg:
:ship:

PR #6, opened 2015-11-18
Use deep merge for config dictionaries

abrown-sg:
* Created new file for "utilities" which will be used more extensively in future changes.

* Switched to using deep dictionary merging for combining config values. Previously, update() would clobber nested settings.

* Tests!

### DIFF
--- a/molecule/core.py
+++ b/molecule/core.py
@@ -36,8 +36,8 @@ from colorama import Fore
 from jinja2 import Environment
 from jinja2 import PackageLoader
 
-import molecule.validators as validators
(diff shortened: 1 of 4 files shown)


rgreinho:
@abrown-sg Can you wait before merging please? I wrote a very good deep merge function, but it is more complex than yours and I'd like to compare.

rgreinho:
I might be biased but :+1: :P

rgreinho:
Ah true, it's because the tests come from one of my old projects and I was using ``unittest``.

retr0h:
:+1:

PR #10, opened 2015-11-20
Utilities cleanup

abrown-sg:
This is a start, there's probably more that can be moved as we clean things up.

* Fixed utility test for deep_merge that was failing
* Made print_line two different functions for stdout and stderr
* Updated print functions to be Python 3 ready
* Moved template creation into a generic function
* Test all the (moved) things


molecule/utilities.py
```diff
@@ -18,7 +18,13 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
+from __future__ import print_function
```
abrown-sg: Python wants __future__ imports to come first.

molecule/templates/test.j2
```diff
@@ -0,0 +1,2 @@
+this is a {{ test }}
+
```
retr0h: Not really a fan of this.  I know we need to test it, but test files shouldn't live along side real files.

abrown-sg: I can make the test write out a test.j2 to /tmp or something.

abrown-sg: Actually maybe I can't. The Jinja2 library expects a module name and a path.

abrown-sg: So PackageLoader won't let us break out of the molecule package. There are other loaders, like FileSystemLoader that will take a full path to a template, but that doesn't really test how we're using things. I don't really see a good option to having the test template coexist.

abrown-sg: I could pass in another directory under molecule/molecule though. So instead of having it in molecule/templates it could be in like, molecule/tests or testdata or whatever.

tests/test_utilities.py
```diff
@@ -18,7 +18,12 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
+import StringIO
+
+import binascii
+import os
 import testtools
+from mock import patch
```
rgreinho: Import order: 
* ``os`` before ``StringIO``
* ``mock`` before ``testtools ``

retr0h:
:+1:

abrown-sg:
One sec, one last fix

abrown-sg:
Alright, @rgreinho found a better fix for that failing test.

retr0h:
:+1:

PR #12, opened 2015-11-24
Config Cleanup

abrown-sg:
* Updated format of config.yml and molecule.yml so they use the same data structure for easier merging. In general it's more clear and easy to understand

* Defaults are now loaded from a defaults file (YAML) rather than a giant hash. Maintaining data in two formats was getting tiresome.

* Decoupled main() from init() in Molecule core to make future tests easier. 

* Removed mock from existing tests that no longer require it now that main() is decoupled.

* Moved all config handling to an external class. Greatly simplified all logic.

* Added tests for new config class.

* Cleaned up all messages using format() to have consistent syntax.

* Fixed status command to not fire unless a vagrantfile is present since it was triggering vagrant errors.

* Renamed _init_new_role() to init() to be consistent with other commands

* Fixed incorrect messaging in _print_valid_providers()

* Fixed edge case in vagrantfile template to make sure we always have default cpus/memory set for virtualbox instances

* Leveraged new config flexibility to clean up old hack for `molecule init`


molecule/config.py
```diff
@@ -0,0 +1,106 @@
...
+        self.config['molecule']['state_file'] = '/'.join([self.config['molecule']['molecule_dir'], self.config[
+            'molecule']['state_file']])
+        self.config['molecule']['vagrantfile_file'] = '/'.join([self.config['molecule']['molecule_dir'], self.config[
+            'molecule']['vagrantfile_file']])
+        self.config['molecule']['rakefile_file'] = '/'.join([self.config['molecule']['molecule_dir'], self.config[
+            'molecule']['rakefile_file']])
+        self.config['ansible']['config_file'] = '/'.join([self.config['molecule']['molecule_dir'], self.config[
+            'ansible']['config_file']])
+        self.config['ansible']['inventory_file'] = '/'.join([self.config['molecule']['molecule_dir'], self.config[
+            'ansible']['inventory_file']])
```
rgreinho: If you are building paths, you must use ``os.path.join`` - https://docs.python.org/2/library/os.path.html.

rgreinho:
90% of the time, when we exit from an exception, we exit a function like that:
```
print('ERROR: {}'.format(e))
sys.exit(e.exit_code)
```

Maybe we could have a utility function for that instead (I know it is just 2 lines, but still, just in case).

abrown-sg:
I agree we could move that into a utility function. I think I'd like to do that as part of additional utility cleanup since the scope of the PR is getting a bit out of hand already.

rgreinho:
:+1:

retr0h:
:+1:

PR #14, opened 2015-11-25
More RTD fixes

abrown-sg:
More fixes to keep theme loading from happening when being processed by RTD.


retr0h:
Seems logical :+1:

PR #18, opened 2015-11-25
Merged provisioners into core

abrown-sg:
This looks like a huge change, but it's mostly moving logic from one file to another.

* Moved logic from provisioners to core. This was in place for a time when we supported using vagrant or ansible-playbook to provision instances, and that's no longer the case.

* Collapsed a few methods whose existence was to support vagrant provisioners which has since been removed.

* Fixed tests and CLI to no longer reference provisioners.


molecule/core.py
```diff
@@ -207,23 +207,115 @@ def _sigwinch_passthrough(self, sig, data):
...
+                                 kwargs=kwargs)
+
+    def _format_instance_name(self, name):
+        for instance in self._config.config['vagrant']['instances']:
+            if instance['name'] == name:
+                if 'options' in instance and instance['options'] is not None:
+                    if 'append_platform_to_hostname' in instance['options']:
+                        if not instance['options']['append_platform_to_hostname']:
+                            return name
+        return name + '-' + self._env['MOLECULE_PLATFORM']
```
rgreinho: This function is a bit confusing :/ 6 levels of nesting.

molecule/core.py
```diff
@@ -207,23 +207,115 @@ def _sigwinch_passthrough(self, sig, data):
...
+                                              ssh['User'])
+
+        # get a list of all groups and hosts in those groups
+        groups = {}
+        for instance in self._config.config['vagrant']['instances']:
+            if 'ansible_groups' in instance:
+                for group in instance['ansible_groups']:
+                    if group not in groups:
+                        groups[group] = []
+                    groups[group].append(instance['name'])
```
rgreinho: 5 levels of nesting as well.

rgreinho:
:+1: 
If you could write some UTs for the functions where I commented, I could try to see if I can find a way to refactor them a little. Or if you can give me an example of input and output, I'll write the UTs, that would work for me too :)

abrown-sg:
I'm gonna tackle this cleanup in a different way.

PR #19, opened 2015-11-26
Remove old wrapper methods

abrown-sg:
These can be rolled into the actual methods now that we're not trying to support multiple provisioners.


rgreinho:
:+1: :smile:

PR #23, opened 2015-11-30
removed relative paths from having base dir added

ephur:
Also catch exceptions and print warnings, since sometiems it may be desirable for molecule to not write to a file

### DIFF
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -95,14 +95,10 @@ class Config(object):
         :return: None
         """
         vals = [
-            ('molecule', 'state_file'),
-            ('molecule', 'vagrantfile_file'),
-            ('molecule', 'rakefile_file'),
-            ('ansible', 'config_file'),
-            ('ansible', 'inventory_file')
+            ('molecule', 'state_file'), ('molecule', 'vagrantfile_file'), ('molecule', 'rakefile_file'),
+            ('ansible', 'config_file'), ('ansible', 'inventory_file')
         ]
 
         for k, v in vals:
             if not self.config[k][v].startswith("../"):
-                self.config[k][v] = os.path.join(self.config['molecule']['molecule_dir'],
-                                                            self.config[k][v])
+                self.config[k][v] = os.path.join(self.config['molecule']['molecule_dir'], self.config[k][v])
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -94,11 +94,12 @@ class Config(object):
 
         :return: None
         """
-        vals = [
+        values_to_update = [
             ('molecule', 'state_file'), ('molecule', 'vagrantfile_file'), ('molecule', 'rakefile_file'),
             ('ansible', 'config_file'), ('ansible', 'inventory_file')
         ]
 
-        for k, v in vals:
-            if not self.config[k][v].startswith("../"):
-                self.config[k][v] = os.path.join(self.config['molecule']['molecule_dir'], self.config[k][v])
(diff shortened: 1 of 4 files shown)


molecule/utilities.py
```diff
@@ -90,8 +90,11 @@ def write_file(filename, content):
     :param content: what gets written into the file
     :return: None
     """
-    with open(filename, 'w') as f:
-        f.write(content)
+    try:
+        with open(filename, 'w') as f:
+            f.write(content)
+    except Exception as e:
+        raise
```
rgreinho: What the point of catching the exception to re-raise it right away?

molecule/provisioners.py
```diff
@@ -168,12 +171,17 @@ def converge(self, idempotent=False):
...
+                print "Playbook: %s\nArgs%s\nkwargs%s" % (playbook, args, kwargs)
                 output = sh.ansible_playbook(playbook, *args, **kwargs)
                 return output
             except sh.ErrorReturnCode as e:
                 print('ERROR: {}'.format(e))
                 sys.exit(e.exit_code)
         try:
+            print "Executing ansible_playbook"
+            print "Playbook: %s\nArgs%s\nkwargs%s" % (playbook, args, kwargs)
+            # kwargs['inventory_file'] = '../../inventory_fusion'
```
rgreinho: Remove comment.

molecule/config.py
```diff
@@ -94,13 +94,15 @@ def build_easy_paths(self):
...
-                                                             self.config['molecule']['state_file'])
-        self.config['molecule']['vagrantfile_file'] = os.path.join(self.config['molecule']['molecule_dir'],
-                                                                   self.config['molecule']['vagrantfile_file'])
-        self.config['molecule']['rakefile_file'] = os.path.join(self.config['molecule']['molecule_dir'],
-                                                                self.config['molecule']['rakefile_file'])
-        self.config['ansible']['config_file'] = os.path.join(self.config['molecule']['molecule_dir'],
-                                                             self.config['ansible']['config_file'])
-        self.config['ansible']['inventory_file'] = os.path.join(self.config['molecule']['molecule_dir'],
-                                                                self.config['ansible']['inventory_file'])
+        vals = [
```
rgreinho: Can we name ``vals`` to something else? At least ``values``?

molecule/config.py
```diff
@@ -94,13 +94,15 @@ def build_easy_paths(self):
...
+            ('molecule', 'state_file'),
+            ('molecule', 'vagrantfile_file'),
+            ('molecule', 'rakefile_file'),
+            ('ansible', 'config_file'),
+            ('ansible', 'inventory_file')
+        ]
+
+        for k, v in vals:
```
rgreinho: In that case it is not ``key`` and ``value`` anymore, so I would replace ``k`` and ``v`` with more meaningful variable names.

molecule/utilities.py
ephur: Force of habit -- will remove this try/catch.

ephur:
Addressed the comments above. On the variable naming for key/value I changed it to section and item, because it reminds me of ini file format, where molecule or Ansible is a section, and config_file is an item.  Also I had the try/catch because often times in the past I've caught the list of excpetions I am 'expecting' and raise exceptions of a different type, so I am just used to that pattern but it certainly didn't add any value here. Removed.

retr0h:
+ 1

rgreinho:
:+1:

PR #25, opened 2015-12-01
Fixed inventory_file bug

abrown-sg:
* Slightly cleaner and more complete fix to an issue where specifying an inventory file wasn't working
* Molecule sets defaults for inventory_file and config_file that are used unless optionally overridden in ansible config block of molecule.yml.
* Cleaned up error message from previous commit to be consistent with existing style
* Updated docs, tests


abrown-sg:
yapf 0.6.2 here

retr0h:
> yapf 0.6.2 here

Yeah, same here.  @ephur must have an old version.  We need to figure out how to better ensure our contributors are using the same versions of this library.

retr0h:
:+1:

abrown-sg:
Yeah. The `molecule` section is the default now for the config_file and inventory_file that it writes. Those defaults will be used unless overridden in the `ansible` section now.

abrown-sg:
So now, anything specified in the `ansible` section won't get molecule_dir appended to it. This opens the door for specifying absolute paths. I included config_dir in this change since it would have the same issue inventory_dir did previously.

PR #26, opened 2015-12-01
Cleanup and test format_instance_name

abrown-sg:
* Moved function out of provisioners and into utilities
* Refactored to be more clear
* Added tests

### DIFF
--- a/molecule/utilities.py
+++ b/molecule/utilities.py
@@ -108,6 +108,7 @@ def format_instance_name(name, platform, instances):
     for instance in instances:
         if instance['name'] == name:
             working_instance = instance
+            break
 
     # no instance found with given name
     if working_instance is None:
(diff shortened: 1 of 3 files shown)


retr0h:
Seems logical +1

molecule/utilities.py
```diff
@@ -94,6 +94,37 @@ def write_file(filename, content):
...
+    :param platform: the current molecule platform in use
+    :param instances: the current molecule instances dict in use
+    :return: formatted instance name if found in instances, otherwise None
+    """
+    working_instance = None
+
+    # search instances for given name
+    for instance in instances:
+        if instance['name'] == name:
+            working_instance = instance
```
rgreinho: Could you break as soon as you found a match? or do you want the last one?

rgreinho:
:+1: So much clearer! And the UTs help a lot. Great job!

molecule/utilities.py
abrown-sg: Good call. Added a break.

PR #27, opened 2015-12-01
Move config logic into config module

abrown-sg:
The previous change put this logic into provisioners, however, that doesn't work in all cases. It needs to be in the config object so the changes are updated when molecule files are merged in. That way, the values are available throughout the script.

This opens the door for updating testinfra to use the inventory file we specify.


retr0h:
:+1:

rgreinho:
:+1:

PR #32, opened 2015-12-03
releng docs

dhutty:


docs/source/development.rst
```diff
@@ -0,0 +1,75 @@
...
+Release
+^^^^^^^
+
+Tag the release and push to github.com
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+  .. code-block:: bash
+
+      $ git tag -a 1.0.5 -m "Version 1.0.5"
+      $ git push origin 1.0.5
```
rgreinho: We do not need the dollar sign. It prevent copy/paste. The rendering makes sure that the reader know that these are command lines.

docs/source/development.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Upload to `PyPI`_
```
rgreinho: I'm, not sure this section is relevant in the public documentation. It looks like an internal thing to me. I think @abrown-sg wrote a wiki page about this.

dhutty:
@rgreinho this is the rework of adam's wiki page and yes I think it makes more sense to keep it here rather than hidden in a private wiki. As for the '$', I rather agree but was merely being consistent with other blocks.

docs/source/development.rst
retr0h: We are using `$` for this project.  All of the existing documentation follows this pattern.

docs/source/development.rst
```diff
@@ -0,0 +1,75 @@
...
+* Install `twine`_ using `pip`.
+* Upload to  `PyPI`_.
+
+      .. code-block:: bash
+
+         cd /path/to/molecule
+         rm -rf dist/
+         python setup.py sdist bdist_wheel
+         twine upload dist/*
+
```
retr0h: The existing documentation prefixes commands with `$`.  We should follow that.

docs/source/development.rst
retr0h: Yah, the reason I had @abrown-sg put this on our internal wiki, was for the sharing of the password information.

dhutty: Yes, I have updated that page to point to this for instructions and provided info about how to get the passwords, having shared them in normal channels

PR #34, opened 2015-12-07
Cleaned up error message on bad login

abrown-sg:
* Error message wasn't formatted properly when logging into a non-existent host.
* Removed check to make sure host is passed since docopt takes care of that for us.


retr0h:
Much nicer...

:+1:

PR #35, opened 2015-12-08
Fix global vagrant arguments

bborysenko:
Fix typo in vagrantfile template


abrown-sg:
Ah, yeah, good catch. None of those vagrant raw config args were getting written at all. I'm surprised we didn't notice sooner! :+1:

retr0h:
:+1:

PR #36, opened 2015-12-08
Move commands into a separate module

abrown-sg:
This looks like a huge change, but in fact it's just moving all external commands into a separate module.

The idea here is that we will be able to subclass and change how commands are run. This will be important when implementing the 'metal' provider, which will allow use to use molecule to deploy to existing systems. In that case, we'll want to bypass the create portion of a converge.

Support for this has been stubbed out in this PR, you can see in the MetalCommands class.

Added docstrings for all moved methods.

### DIFF
--- a/molecule/cli.py
+++ b/molecule/cli.py
@@ -59,7 +59,7 @@ from docopt import docopt
 from docopt import DocoptExit
 
 import molecule
-from commands import Commands
+from molecule.commands import Commands
 
 
 class CLI(object):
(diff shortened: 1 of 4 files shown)


molecule/cli.py
```diff
@@ -59,7 +59,7 @@
 from docopt import DocoptExit
 
 import molecule
-from provisioners import Ansible
+from commands import Commands
```
rgreinho: The imports should start from the top of the package:
```
from molecule.commands import Commands
```

rgreinho:
:+1:

PR #39, opened 2015-12-08
Moved remaining provisioners logic to core, added docstrings

abrown-sg:
With the commands moving to their own class, there wasn't much sense having core logic spread into the provisioners file.

Added docstrings.

Started process of updating changelog during development to better track changes. Will update version once we decide on a new release.


CHANGELOG.rst
```diff
@@ -1,5 +1,10 @@
 CHANGELOG for molecule
 ======================
+1.0.7.dev1
+----
```
retr0h: Do we really need a changelog entry for this?  Seems this work can go out as 1.1.0 once  the 'static' provider is done, no?  It's all prep work for that.

rgreinho: if you have ``dev1`` at the end of the version, it means there was an issue with tagging/packaging.

molecule/commands.py
```diff
@@ -32,7 +32,7 @@
 
 import molecule.utilities as utilities
 import molecule.validators as validators
-from provisioners import Ansible
+from molecule.core import Molecule
```
rgreinho: Bump it 2 lines higher, 'c' > 'u'.

abrown-sg:
Eh, I was just thinking of a way to document changes along the way. I'll axe it.

molecule/commands.py
abrown-sg: John asked me to keep froms below imports.

retr0h: And yes I know this is holly war ground for that particular topic. :P

rgreinho: Haha, not so much holly war though: http://docs.openstack.org/developer/hacking/#imports

abrown-sg:
@retr0h @rgreinho So, :+1:  :-1:? :)

rgreinho:
:+1: I'm not going to hold you for an alphabet problem, but I still think we should follow the OpenStack guidelines.

PR #40, opened 2015-12-09
Support for static inventory

abrown-sg:
* Enables a mode in molecule where existing inventory can be provisioned using `molecule converge` and all instance management functions will be ignored.

* Opens the door for a future --fast option to `molecule converge` that is 20x faster since it can skip machine creation.

* Fixes a broken test from a previous PR. :blush:


docs/source/conf.py
```diff
@@ -65,9 +65,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '1.0.6'
+version = '1.1.0'
 # The full version, including alpha/beta/rc tags.
-release = '1.0.6'
+release = '1.1.0'
```
retr0h: We should prob update this to use `molecule.__version__`.

CHANGELOG.rst
```diff
@@ -1,5 +1,11 @@
 CHANGELOG for molecule
 ======================
+1.1.0
+----
+
+* Added support for static inventory where molecule can manage existing sites, not just vagrant instances.
+
+
```
retr0h: Did we have any other fixes come in before this minor work?

abrown-sg: Just some housekeeping moving lingering methods in provisioners to core.

docs/source/conf.py
abrown-sg: I think that will break readthedocs since it uses this string for version info but isn't molecule aware.

retr0h:
:1 once we can remove the hard coding of the version in the sphinx documentation generation.

abrown-sg:
You're supposed to specify release and version. Usually they're the same. This just copies the value of one to the other.

This isn't working yet. Still struggling with getting it to build in readthedocs. I think it might need a pip_requirements so it installs molecule. Had to stop messing with it before I got it working.  Will revisit in the AM.

abrown-sg:
Had to update readthedocs to use a virtualenv and the specify a requirements.txt file. It will now pull the package down from pip and use its version.

http://molecule.readthedocs.org/en/static/

@retr0h Ready for another look.

retr0h:
Ok, looks good now.

:+1:

PR #41, opened 2015-12-10
name instances by rolename

dhutty:


molecule/templates/molecule.yml.j2
```diff
@@ -107,7 +107,7 @@ vagrant:
 #        cpus: 2
 
   instances:
-    - name: vagrant-01
+    - name: {{ role }}-01
```
abrown-sg: Won't we end up with instances named, like cisco.horizon-01?

retr0h: For the template it's prob not a bad idea, since it can be changed by the end-user.  It's just a default.

abrown-sg:
LGTM

molecule/templates/molecule.yml.j2
dhutty: Not so important for laptop-based testing of a couple of roles, but will make a big difference to being to clean up OpenStack instances created by Jenkins

PR #43, opened 2015-12-15
add more plugin path directives to the generated ansible.cfg

dhutty:


rgreinho:
I think it might be the same for the callback plugins. We do not have any, but I am planning to write one for having a nice idempotence test, and it should not be enabled all the time.

retr0h:
> I think it might be the same for the callback plugins. We do not have any, but I am planning to write one for having a nice idempotence test, and it should not be enabled all the time.

The default ansible.cfg should be very generic, and if something specific is needed we investigate as we see fit.  At the moment one could accomplish this PR though the following...

```ansible:
  raw_env_vars:
    ANSIBLE_VARS_PLUGINS: ../../../plugins/vars
    ANSIBLE_ROLES_PATH: ../../../roles
    ANSIBLE_LIBRARY: ../../../library
    ANSIBLE_FILTER_PLUGINS: ../../../plugins/filters
    ANSIBLE_LOOKUP_PLUGINS: ../../../plugins/lookups
```

retr0h:
We probably want to develop a way to roll custom templates.  I could see us wanting an ansible.cfg across all our roles, that varies quite a bit from the defaults shipped with molecule.

retr0h:
@dhutty What if we put an ansible.cfg in ansible-systems which has our use cases, and leverage molecule's ansible_config_template?  Seems like a better option since this PR is rather tailored for us.

https://github.com/metacloud/molecule/blob/master/molecule/conf/defaults.yml#L37

dhutty:
sounds reasonable

PR #44, opened 2015-12-15
Print vm_name during serverspec run

bborysenko:
Just to have more clear serverspec output with several nodes.


abrown-sg:
I'm good with this. :+1:

retr0h:
:+1: 

I was hoping there was a way to get serverspec or rake to do this for us, but I can't seem to find anything on this topic.  We could probably accomplish this by adding `host_inventory['hostname']` as a part of the `describe` to each test, but that seems overkill.

Really wish serverspec did this for us.  Seems like a bug IMO.

PR #45, opened 2015-12-16
add --fast to molecule destroy

dhutty:
Skips vagrant's halt() method, goes straight to destroy()

### DIFF
--- a/molecule/cli.py
+++ b/molecule/cli.py
@@ -24,7 +24,7 @@ Usage:
   molecule idempotence [--platform=<platform>] [--provider=<provider>] [--debug]
   molecule test        [--platform=<platform>] [--provider=<provider>] [--debug]
   molecule verify      [--platform=<platform>] [--provider=<provider>] [--debug]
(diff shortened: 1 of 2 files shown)


rgreinho:
:+1: But should not we always go directly do destroy? What is the point of halting a machine that we are going to delete?

abrown-sg:
It's possible to add hooks to halt, like deregistering from Satellite.

I'm working on a change that will remove the --fast flag from converge and make it automatic using the state file. Not that it affects this work, just an FYI.

retr0h:
> But should not we always go directly do destroy? What is the point of halting a machine that we are going to delete?

For our particular use case.  We call halt first, because we want Vagrant to issue a shutdown to the VM.  This allows the shutdown scripts to execute in the guest OS.  We have custom systemd unit files when then unregister the system out of satellite.

I don't have a problem with this change because the upstream community is very likely not following our use case.  However, we do need to ensure our engineers do not follow this workflow when shutting down their VMs.  We will eat up all the available satellite licenses.

retr0h:
No longer valid with  #47.

PR #46, opened 2015-12-16
Genericized molecule's ansible.cfg

retr0h:
We specified `lookup_plugins` and `filter_plugins` for our own
use cases.  This project is upstreamed, and our needs should be
handled in house.  We are specifying `roles_path` and `library`,
since roles can ship a library/ directory.  If this is not good
enough, others can leverage `ansible_config_template` or `raw_env_vars`
as noted in #43.


abrown-sg:
:+1:

PR #47, opened 2015-12-16
Updated state file logic, converge fast

abrown-sg:
* Cleaned up logic around dealing with state file so that it's decoupled from its previous use case and more generally available throughout molecule.

* Removed redundant logic when checking state for various keys.

* State file now tracks whether or not instances are created. 

* Leverage state file info for automatic fast converging.

* Removed superfluous --fast flag from converge.

* Provider and platform should now be properly tracked in state file. There was a bug with this previously.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,12 @@
 CHANGELOG for molecule
 ======================
+1.1.1
+----
+
+* Cleaned up state file management logic to be more concise, functional for other purposes.
+* Removed --fast flag from converge in favor of using state file for fast converging.
+* Instance hostname is now printed during serverspec runs.
+
 1.1.0
 ----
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.rst
```diff
@@ -1,5 +1,12 @@
 CHANGELOG for molecule
 ======================
+1.1.1
+----
+
+* Cleaned up state file management logic to be more concise, functional for other purposes.
+* Removed --fast flag from converge in favor of using state file for fast converging.
+* Instance hostname is now printed during serverspec runs.
+
```
retr0h: I would prefer we not release this yet, until we fix up the ansible.cfg template bug, so others can override the template with a path to an alternate template.  As it stands today, if we release this, we will break our ansible-systems testing.

abrown-sg:
@retr0h Changelog reverted.

retr0h:
:+1:

PR #48, opened 2015-12-16
Support absolute paths to templates

abrown-sg:
Previously, if a user tried to specify a path to a template file, it would fail because the template loader was only able to look within the package.  This adds support for looking on the filesystem then in the package.

### DIFF
--- a/molecule/utilities.py
+++ b/molecule/utilities.py
@@ -88,7 +88,7 @@ def write_template(src, dest, kwargs={}, _module='molecule', _dir='templates'):
         print('\n{}Unable to locate template file: {}{}'.format(Fore.RED, src, Fore.RESET))
         sys.exit(1)
 
-    # look for template in filesystem, them molecule package
+    # look for template in filesystem, then molecule package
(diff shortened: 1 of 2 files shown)


retr0h:
:+1: 

This looks like a good solution!

molecule/utilities.py
```diff
@@ -76,8 +80,19 @@ def write_template(src, dest, kwargs={}, _module='molecule', _dir='templates'):
...
-    template = env.get_template(src)
+    path = os.path.dirname(src)
+    filename = os.path.basename(src)
+
+    # template file doesn't exist
+    if path and not os.path.isfile(src):
+        print('\n{}Unable to locate template file: {}{}'.format(Fore.RED, src, Fore.RESET))
+        sys.exit(1)
+
+    # look for template in filesystem, them molecule package
```
rgreinho: them -> then

rgreinho:
:+1:  I did not know about the Choiceloader, that's pretty cool!

retr0h:
> I did not know about the Choiceloader, that's pretty cool!

Yeah, me either.  Pretty sweet.

PR #49, opened 2015-12-17
Added .cache to ignore_paths

retr0h:
When using testinfra this directory is created, and will fail
validators.


abrown-sg:
:+1:

PR #51, opened 2015-12-17
Improved Debug Output

abrown-sg:
* Passing --debug to molecule now displays config, environment (separated into ansible and other groups) and ansible-playbook command.

* Refactored converge code slightly to remove duplication. Multiple return values weren't being used and were unnecessary.

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -157,13 +157,14 @@ class BaseCommands(object):
             kwargs['_env']['ANSIBLE_FORCE_COLOR'] = 'false'
 
         try:
-            output = sh.ansible_playbook(playbook, *args, **kwargs)
+            ansible = sh.ansible_playbook.bake(playbook, *args, **kwargs)
             if self.molecule._args['--debug']:
                 ansible_env = {k: v for (k, v) in kwargs['_env'].items() if 'ANSIBLE' in k}
                 other_env = {k: v for (k, v) in kwargs['_env'].items() if 'ANSIBLE' not in k}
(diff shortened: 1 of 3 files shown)


molecule/commands.py
```diff
@@ -154,15 +155,16 @@ def converge(self, idempotent=False, create_instances=True, create_inventory=Tru
...
-                output = sh.ansible_playbook(playbook, *args, **kwargs)
-                return output
-            except sh.ErrorReturnCode as e:
-                print('ERROR: {}'.format(e))
-                sys.exit(e.exit_code)
+
         try:
             output = sh.ansible_playbook(playbook, *args, **kwargs)
-            return output.exit_code
+            if self.molecule._args['--debug']:
```
retr0h: Can we report the output prior to execution.  Would be nice to see what we will be doing vs waiting to converge to display this.

molecule/utilities.py
```diff
@@ -25,6 +25,8 @@
 import sys
 
 from colorama import Fore
+from colorama import Back
```
rgreinho: Put ``Back`` before  ``Fore``

rgreinho:
Question: Does ``init(autoreset=True)`` work in our use cases? If no, disregard. If yes, why not using it?

molecule/commands.py
abrown-sg: Done!

molecule/utilities.py
abrown-sg: Fixed!

abrown-sg:
Good find with colorama! I'll take a look at doing that in a future update.

retr0h:
:+1:

PR #55, opened 2015-12-19
Don't generate a __pycache__ directory

retr0h:
We gain no real benefit to generating python byte code, it only
causes problems with linters.  Rather than ignoring this directory
in the linter, we shouldn't generate it at all.


abrown-sg:
:+1:

timgclark:
:+1:

PR #56, opened 2015-12-19
Sanitize the generated hostname

retr0h:
Since we take the role name provided by `molecule init`, and pass that
to Vagrant, we can generate invalid hosnames[1].  I am going to assume
roles are generally named in such a way that they are valid Vagrant
hostnames[1].  However, roles tend to use underscore and dashes
interchangeably.  Also, when using roles from galaxy they contain
`namespace.rolename`.  Linux will treat these hostnames PS1 as
`namespace` which is just lame.

[1] The hostname set for the VM should only contain letters, numbers,
    hyphens or dots. It cannot start with a hyphen or dot.

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -20,6 +20,7 @@
 
 import os
 import pexpect
+import re
 import signal
 import sys
 from subprocess import CalledProcessError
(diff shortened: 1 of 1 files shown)


rgreinho:
:+1:

PR #59, opened 2015-12-21
Destroy not resetting converged state

abrown-sg:
When destroying instances, the state file wasn't updating to show the converge state as false.  This would lead to a scenario where the generated inventory file wasn't being re-created during a new converge. This is especially problematic if you change what groups your instances are in.


retr0h:
:+1:

PR #60, opened 2015-12-21
Retr0h/improved linting

retr0h:


molecule/validators.py
```diff
@@ -35,14 +35,19 @@ def check_trailing_cruft(ignore_paths=[]):
     filenames = []
     pruned_filenames = []
     found_error = False
+    valid_extensions = ['py', 'yml', 'rb']
```
abrown-sg: Should this be a config option?

abrown-sg:
Looks good. This probably addresses, more or less, CIA-133.

molecule/validators.py
retr0h: Thought about it, but not really sure it mattes.  Was mostly trying to keep the config from growing excessively, and couldn't think of a good reason people would be changing this.  I assume we can make it a config option if it displeases people ;)  This was me asserting some conventions w/o the configuration :)

rgreinho:
Just my 2 cents here, but a comprehension would be nicer here I think:
```
filenames.append([os.path.join(root, name) for name in files if name.split(os.extsep)[-1] in valid_extensions])
```

This is just a matter of opinion, so I give :+1:  anyway

retr0h:
> Just my 2 cents here, but a comprehension would be nicer here I think:

I agree.  I try to use list comprehension wherever possible.  In this case I thought it was doing a bit much, but after looking at it more, I like it. Updated PR.

rgreinho:
:+1:

PR #62, opened 2015-12-22
Fixed names for pypi

abrown-sg:
Pypi's "Requires Distributions" section is case sensitive when it comes to package names. It will only hyperlink to projects if the name and case matches.

See prettytable and jinja2 at: https://pypi.python.org/pypi/molecule.


rgreinho:
:+1:

PR #65, opened 2015-12-22
Add Travis CI support

rgreinho:
* Parallelize Travis-ci tox tests
* Add Travis-ci badge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: python
+env:
+  matrix:
+    - TOXENV=flake8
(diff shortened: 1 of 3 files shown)


rgreinho:
Nope, it's just that if in the future we add support for 3.5, it looks nicer to have it at the end after 2.7 :)

retr0h:
> Nope, it's just that if in the future we add support for 3.5, it looks nicer to have it at the end after 2.7 :)

Makes sense to me!

:+1:

PR #67, opened 2015-12-27
Update idempotence test

rgreinho:
The idempotence test will now try to leverage the idempotence ansible
callback plugin to determine whether the role is idempotent and if not, 
display the tasks that are not.

In case the callback plugin is not found, the current behavior is not
altered and only a success or failure message is displayed to the user.

The ansible idempotence callback plugin was included into the molecule
package.

Add unit tests for the idempotence tests.

Update setup.cfg.

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -159,6 +159,11 @@ class BaseCommands(object):
             kwargs.pop('_err', None)
             kwargs['_env']['ANSIBLE_NOCOLOR'] = 'true'
             kwargs['_env']['ANSIBLE_FORCE_COLOR'] = 'false'
+
+            # Save the previous callback plugin if any.
+            callback_plugin = kwargs.get('_env', {}).get('ANSIBLE_CALLBACK_PLUGINS', None)
+
+            # Set the idempotence plugin.
             kwargs['_env']['ANSIBLE_CALLBACK_PLUGINS'] = os.path.join(
                 sys.prefix, 'share/molecule/ansible/plugins/callback/idempotence')
 
@@ -176,6 +181,10 @@ class BaseCommands(object):
             if not self.molecule._state.get('converged'):
                 self.molecule._state['converged'] = True
                 self.molecule._write_state_file()
+
+            if idempotent:
+                # Reset the callback plugin to the previous value.
+                kwargs['_env']['ANSIBLE_CALLBACK_PLUGINS'] = callback_plugin
 
             return output
         except sh.ErrorReturnCode as e:
@@ -199,11 +208,15 @@ class BaseCommands(object):
 
         # Display the details of the idempotence test.
         if changed_tasks:
-            print('{}Idempotence test failed because of the following tasks{}:'.format(Fore.RED, Fore.RESET))
-            print('{}\t{}{}'.format(Fore.RED, '\n'.join(changed_tasks), Fore.RESET))
+            print('{}Idempotence test failed because of the following tasks:{}'.format(Fore.RED, Fore.RESET))
+            print('{}{}{}'.format(Fore.RED, '\n'.join(changed_tasks), Fore.RESET))
         else:
             # But in case the idempotence callback plugin was not found, we just display an error message.
             print('{}Idempotence test failed.{}'.format(Fore.RED, Fore.RESET))
+            warning_msg = "The idempotence plugin was not found or did not provide the required information. " \
+                "Therefore the failure details cannot be displayed."
+
+            print('{}{}{}'.format(Fore.YELLOW, warning_msg, Fore.RESET))
         sys.exit(1)
 
     def verify(self):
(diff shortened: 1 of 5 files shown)


rgreinho:
Build link (passing): https://travis-ci.org/rgreinho/molecule/builds/99019517

molecule/commands.py
```diff
@@ -159,6 +159,8 @@ def converge(self, idempotent=False, create_instances=True, create_inventory=Tru
             kwargs.pop('_err', None)
             kwargs['_env']['ANSIBLE_NOCOLOR'] = 'true'
             kwargs['_env']['ANSIBLE_FORCE_COLOR'] = 'false'
+            kwargs['_env']['ANSIBLE_CALLBACK_PLUGINS'] = os.path.join(
+                sys.prefix, 'share/molecule/ansible/plugins/callback/idempotence')
```
retr0h: By setting it here, it looks like it negates the ability for us to override it in molecule.yml's [raw_env_vars](
https://github.com/metacloud/molecule/blob/master/docs/source/usage.rst#ansible
) section?

rgreinho: That's the point. The plugin really has only one usage, which is to make it easy to identify the "changed" tasks from the playbook output. If the plugin is replaced, the idempotence fucntion may not work as expected. 

As a compromise I could read from the config file and if no value is specified, I would set that one as default. But again, I'm not sure this should be a value that should be configured.

retr0h: > That's the point.

Thats kind of the problem though.  People who use `molecule` may be developing a role which uses a callback plugin.  If we are setting `ANSIBLE_CALLBACK_PLUGINS` for idempotence checks, then the user may not be able to test their callback plugin.  If we do allow them to override it, then idempotence checks will not work.

bborysenko: Although it's a great idea to use callback plugin, I'm agree with @retr0h. [kitchen-ansiblepush](https://github.com/ahelal/kitchen-ansiblepush) uses [something similar](https://github.com/ahelal/kitchen-ansiblepush/blob/master/lib/kitchen/provisioner/ansible_push.rb#L222) for testing idempotency, and they clearly point to that this might cause conflict.

Perhaps the better method is to not use environment variable, but instead add `callback_plugins` to `ansible.cfg.j2` template and allow user to override it in some way?

abrown-sg: It could be set as a default in ansible.cfg.j2 as a raw_env_var under the ansible block and then overridden by a user if they set their own value for that environment var in their molecule.yml.  The idempotence command could check to see if the plugin is available, and if not, print a warning about falling back to the method of parsing the output and not displaying the tasks.

rgreinho: I don't think this is true. A callback plugin only adds information to the ansible output, so it cannot really conflict with anything. 

In the case of my plugin, it adds lines that are easy to identify with a regex to the playbook summary. This summary is used by the updated idempotence function and if the information is not found, it falls back to the previous behaviour.

My point is that for a test tool, you do not want the user to the change the output. How often do you want to change the output of ``Test Infra``, ``Testtools``, or ``py.test``? 

I think this is necessary, because at the moment, the idempotence test says only ``PASSED`` or ``FAILED``, and in case of failure we do not know directly which tasks or handlers caused the failure. We have to replay a ``molecule converge`` to get the necessary information to fix the idempotence. That's not really a user friendly behaviour, and that's what I am trying to address.

rgreinho: @abrown-sg The updated idempotence test, already falls back to the default behaviour. It does not display a warning though. This is a good idea. 

Again, I think the user should not have the control of the ansible output for the tests.

abrown-sg: But what will a user do if they want to add their own callback plugin? Presumably this value can accept a list of paths like most other path config options?

rgreinho:
@abrown-sg I added a warning message to inform the user that the idempotence plugin was not found or did not return the expected values, and therefore no details could be provided for the idempotence test.

Interesting fact, if you define several callback plugin paths, Ansible will agregate the output of all the plugins.

Again, the users can add their own callback plugins if they want, but they should not be used during the tests. Moreover, the idempotence test does not display any ansible output, so I really don't see why we should use a plugin specified by a user for this test. Can you foresee a use case, where it would make sense, for the idempotence test, to use your own callback plugin?

rgreinho:
@retr0h Deal! I'll update the PR ASAP.

retr0h:
@bborysenko how do you feel about this?

bborysenko:
:+1:

abrown-sg:
:+1:

retr0h:
@abrown-sg I don't believe @rgreinho made the updates I was asking for yet.

retr0h:
@rgreinho Could you be sure to get what we talked about in, now that this PR is merged?

abrown-sg:
@retr0h @rgreinho Sorry about that, I misunderstood. I'll drop 1.2.1 once we get Remy's changes in.

rgreinho:
@retr0h @abrown-sg Sorry guys, my holidays schedule is significantly slower that usually :) In spite of that, PR #73 brings the changes that @retr0h wanted to see in place.

PR #69, opened 2015-12-28
Path to templates can be relative to user home dir

bborysenko:
To be able set path to templates in ` ~/.config/molecule/config.yml` like:

```
---
molecule:
  ansible_config_template: ~/.config/molecule/ansible.cfg.j2
```


abrown-sg:
:+1:

PR #73, opened 2015-12-29
Append idempotence callback plugin path

rgreinho:
During the idempotence test, the path of the idempotence plugin,
provided by molecule is appended to the existing one, if any, instead of
overwriting it.


retr0h:
:+1:

PR #74, opened 2015-12-29
Update requirements

rgreinho:
* Update requirements and docs-requirements.
* Add a requires.io badge to ensure requirements are always up to date


retr0h:
:+1:

PR #78, opened 2016-01-02
Fix idempotence plugin

rgreinho:
The idempotence plugin was crashing in some cases when the 'changed'
value could not be read from the 'res' dictionary.

Also, the changed task list contained duplicates in some cases. When a task
was executed several time because of a loop (using with_items for
instance), the task was being displayed once per item. Replacing the
list by a set solved this issue.


retr0h:
+1

Issue #76, opened 2015-12-31
Proposal: leveraging DocOpt subcommands


rgreinho:
I wrote a tutorial here (https://github.com/rgreinho/docopt-subcommands-example) which explains how to use DocOpt subcommands in a console application. 

I think we should leverage this in order to be able to easily add options to the main command or the subcommands in the future.

link: PR #75

retr0h:
Totally agree.

/cc @bborysenko @abrown-sg

rgreinho:
@retr0h Even with a way better command dispatcher to get rid of the ``if ... elif ... else`` that you dislike :)

https://github.com/rgreinho/docopt-subcommands-example#find-the-command-to-execute
https://github.com/rgreinho/docopt-subcommands-example/blob/master/main.py#L37-L43

retr0h:
> @retr0h Even with a way better command dispatcher to get rid of the if ... elif ... else that you dislike :)

![so sexy](http://gif.co/sas5.gif)

abrown-sg:
The challenge we're facing when trying to implement this (which may be a DocOpt bug) is that long arguments don't seem to work in both forms.  For example:

`--platform=ubuntu` works
`--platform ubuntu` fails

The expected behavior is that both will work in DocOpt, however we've been unable to find a workaround with Remy's subcommand pattern that support this.

abrown-sg:
#81 implements as much of this as possible, however, by moving commands into individual classes. This is a win regardless since it's a significant simplification over what we were using previously. It will also be easy to drop in DocOpt subcommand support once the bug is resolved.

rgreinho:
I filed an issue upstream: https://github.com/docopt/docopt/issues/305.

rgreinho:
@abrown-sg There was a mistake in my tutorial (that I fixed), and now the parameters for the arguments work both with and without the equal sign.

Basically there need to be an ellipsis ('``...``') after the args (looks like ``[<args>...]``), so that the program can have multiple values. The issue was that ``--platform ubuntu`` was counted has 2 arguments, but as the ellipsis was not specified, only one argument was allowed.

If you can confirm that it works for your branch as well (I don't think you pushed it, so I could not try myself), then I'll close the issue that I filled upstream.

PR #81, opened 2016-01-06
CIA-270: Refactor Commands

abrown-sg:
This looks like a huge change, but it's mostly moving code around and removing duplication.  I was unable to use Remy's DocOpt pattern mentioned in #76 due to an unresolved DocOpt bug. However this work makes it very easy to switch to his pattern at a later time.

* Refactored commands.py to use a separate class for each command.

* No longer subclassing all commands for static inventory. Removed the redefinition of many methods.

* Cleaned up DocOpt spacing in cli.py to be consistent.

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -149,7 +149,8 @@ class Converge(AbstractCommand):
             create_inventory = False
 
         if create_instances and not idempotent:
-            Create(self.args, self.molecule).execute()
+            c = Create(self.args, self.molecule)
+            c.execute()
 
         if create_inventory:
             self.molecule._create_inventory_file()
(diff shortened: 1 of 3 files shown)


molecule/commands.py
```diff
@@ -146,8 +144,12 @@ def converge(self, idempotent=False, create_instances=True, create_inventory=Tru
         if self.molecule._state.get('converged'):
             create_inventory = False
 
+        if self.static:
+            create_instances = False
+            create_inventory = False
+
         if create_instances and not idempotent:
-            self.create()
+            Create(self.args, self.molecule).execute()
```
rgreinho: Not a blocker, just my 2 cents:  
You used this pattern in a lot of places, but I prefer to see that split in 2 lines, one for creating the object, one for calling the function.

rgreinho:
I like it! Way clearer/simpler! :+1:

abrown-sg:
@rgreinho No problem, I updated to break those all out into multiple lines.

retr0h:
+1 Yeah, deff a good cleanup.

PR #86, opened 2016-01-07
CIA-282: Remove hardcoded serverspec dir from rakefile template

### LINKED ISSUE #85: Search in test/integration/default/ for tests
Test-kitchen searches for tests in `test/integration/default/`, see here:
http://kitchen.ci/docs/getting-started/writing-test

It'd be nice if molecule could search there, too, so existing tests do not have to be changed (we test puppet/ansible/chef with the same tests, that reside in the aformentioned directory).

Maybe its also possible to let the user specify where to search for tests?

abrown-sg:
* Fixes #85.
* Passes serverspec_dir through to rakefile template and uses that config value.

### DIFF
--- a/molecule/core.py
+++ b/molecule/core.py
@@ -259,7 +259,8 @@ class Molecule(object):
         # rakefile
         kwargs = {
             'molecule_file': self._config.config['molecule']['molecule_file'],
-            'current_platform': self._env['MOLECULE_PLATFORM']
+            'current_platform': self._env['MOLECULE_PLATFORM'],
+            'serverspec_dir': self._config.config['molecule']['serverspec_dir']
         }
         utilities.write_template(self._config.config['molecule']['rakefile_template'],
                                  self._config.config['molecule']['rakefile_file'],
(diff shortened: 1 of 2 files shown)


rgreinho:
Question: What if ``serverspec_dir`` is not specified? During the rakefile file generation as the path would start with ``/``. Maybe we should have a default value for it?

abrown-sg:
@rgreinho we set a default in molecule's config. Is that enough?

https://github.com/metacloud/molecule/blob/master/molecule/conf/defaults.yml#L13

retr0h:
> Question: What if serverspec_dir is not specified? During the rakefile file generation as the path would start with /. Maybe we should have a default value for it?

The default is [`spec`](https://github.com/metacloud/molecule/blob/60c55bf5b0bb3d3c44ed7d5f284db96b50a97257/molecule/conf/defaults.yml#L13).  You can override this as @abrown-sg pointed out.

retr0h:
:+1:

rgreinho:
That's perfect :) :+1: too

Issue #84, opened 2016-01-07
Trailing whitespace in molecule.yml after molecule init


rndmh3ro:
molecule.yml is the default one created with `molecule init`
```
(molecule)user $  molecule verify
Trailing whitespace found in ./molecule.yml on lines: 100
```
At `# box_version:`
```
  # molecule's --platform option will look for these names
  platforms:
    - name: trusty64
      box: trusty64
      box_url: https://vagrantcloud.com/ubuntu/boxes/trusty64/versions/14.04/providers/virtualbox.box
      # Optional item, box_version - support for vagrant versioning
      # box_version: 
```

abrown-sg:
Fixed by #87. Thanks for the report!

PR #87, opened 2016-01-07
CIA-283: Molecule init creates files with trailing whitespace

abrown-sg:
* Addresses #84 
* Adds a default value to box_version for the `init` command so that the template isn't rendered with a trailing whitespace. Also avoids referencing an undefined variable.
* box_version is still commented out by default.
* The file meta/main.yml is still created with trailing whitespace that won't pass `molecule verify`. However, this file is created by the underlying call to `ansible-galaxy init`. At this time, we'd prefer not to overwrite that file with our own.


retr0h:
:+1:

rokka-n:
I wonder what part of the system doesn't like whitespaces and who released it.
So everybody is dealing now with a "whitespaces" issue.
What an incredible waste of time...

And verify still broken in 1.4.1

PR #88, opened 2016-01-07
Fix callback_plugin path

rgreinho:
When concatenating several paths, they must be separated using a colon
``(':')`` sign.

Fixes metacloud/molecule#82


abrown-sg:
:+1:

PR #89, opened 2016-01-08
Fix ssh_config issue

srsakhamuri:
Use correct vagrant config function for logging in and writing
ssh_config file


abrown-sg:
:+1:

PR #91, opened 2016-01-08
CIA-291: Moved rubocop from core to validators

abrown-sg:
* Moved rubocop out of core and into validators
* sh is no longer required in core


retr0h:
Tests! :)

abrown-sg:
Gonna do some more work on this to move rake, too. Will also work on tests.

PR #92, opened 2016-01-12
CIA-291: Move validators out of core/comands

abrown-sg:
* Moved rubocop from core to validators
* sh is no longer used in core
* Moved rake from commands to validators
* Moved testinfra from commands to validators
* Added tests


retr0h:
:+1: 

Much better tests now that we are mocking what was called vs setting up temporary directories.

rgreinho:
:+1:

PR #94, opened 2016-01-12
Vagrant halt/destroy needs a vagrantfile

abrown-sg:
If a machine hasn't been created, and we call `molecule destroy`, an error will be produced since a vagrantfile isn't written. This becomes a problem with `molecule test` where typically the first command called is `destroy`, and often a VM hasn't been created yet.


retr0h:
:+1:

srsakhamuri:
LGTM

PR #95, opened 2016-01-14
Fix init command

abrown-sg:
1.2.3 introduced a bug that prevents `init` from working properly.

The `init` command is generally not called in a directory where molecule.yml is present, since it's used to create a new role, so we have to bypass all the logic around loading config files. With the move of provisioners to its own class, one of these checks was missing.


rgreinho:
Seems straightforward :+1:

PR #96, opened 2016-01-22
Remove useless call to "vagrant halt"

0xDEC0DE:
"vagrant destroy --force" will implicitly halt the VM, so this call
to "vagrant halt" is unnecessary (and slows down "molecule destroy")


retr0h:
The call to halt is a tricky situation.  We have opensourced molecule, however, the halt call is special to "us" b/c we want the guest OS to execute it's shutdown scripts prior to termination.  The reason being, our VMs are RHEL based, and ship with a shutdown script to deregister themselves from satellite.

We may not care about this any longer, since we have a cronjob to cleanup stale systems in satellite.

/cc @alop

alop:
This is correct, we have a cronjob named orbitaldecay that prunes stale subscriptions, so it's not unreasonable to skip the shutdown. That's the initial reason why it's there. Since we've mitigated the cause, I suppose it's safe to yank the halt, as this is intended to be a testing framework

alop:
The correct answer would be for vagrant to have a cleanup section for ansible, as today it only does cleanup on chef.

abrown-sg:
Keep in mind that removing this line as is will reintroduce the bug where calling destroy when a VM isn't running (like we often do with molecule test) will blow up. The call in the _halt method to force the creation of an inventory file would need to be moved to the _destroy method.

retr0h:
I think we want to remove halt.  That will speed things up.  We just need to implement the vagrant triggers plugin, to execute custom callbacks

PR #100, opened 2016-01-26
Disable the vagrant-vbguest plugin if it's installed

0xDEC0DE:
This plugin is useful most of the time, but on RHEL VMs, it tries
to install packages before the system is registered.

Disable it.

### DIFF
--- a/molecule/templates/vagrantfile.j2
+++ b/molecule/templates/vagrantfile.j2
@@ -40,6 +40,10 @@ Vagrant.configure('2') do |config|
     vb.linked_clone = true if Vagrant::VERSION =~ /^1.8/
     {%- endif %}
   end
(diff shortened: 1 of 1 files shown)


molecule/templates/vagrantfile.j2
```diff
@@ -40,6 +40,10 @@ Vagrant.configure('2') do |config|
     vb.linked_clone = true if Vagrant::VERSION =~ /^1.8/
     {%- endif %}
   end
+  # This plugin tries to install packages before the system is registered.  Disable it.
+  if Vagrant.has_plugin?('vagrant-vbguest')
+    config.vbguest.auto_update = false
```
rgreinho: Could you parameterize the value (``false`` or ``true``)? If I test against Ubuntu or Centos, it may be useful to be able to use it.

PR #102, opened 2016-01-29
CIA-347: Remove halt, use vagrant-triggers

abrown-sg:
* Remove call to `vagrant halt` when doing a `molecule destroy`
* `molecule destroy` will no longer attempt to `vagrant destroy` instances not in the running state
* Added very basic support to default vagrantfile template for vagrant-triggers plugin
* This will be released as 1.3.0

### DIFF
--- a/molecule/provisioners.py
+++ b/molecule/provisioners.py
@@ -139,14 +139,6 @@ class BaseProvisioner(object):
         return
 
     @abc.abstractmethod
-    def halt(self):
-        """
-        Halts the VMs
-        :return:
-        """
-        return
-
-    @abc.abstractmethod
     def status(self):
         """
         Returns the running status of the VMs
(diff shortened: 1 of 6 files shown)


retr0h:
:+1:

molecule/provisioners.py
```diff
@@ -274,13 +266,12 @@ def up(self, no_provision=True):
...
 
     def destroy(self):
-        self._vagrant.halt()
-        self._vagrant.destroy()
-        os.remove(self.m._config.config['molecule']['vagrantfile_file'])
-
-    def halt(self):
         self._write_vagrant_file()
-        self._vagrant.halt()
+        for status in self._vagrant.status():
```
retr0h: Would be nice if we can look at the state file's status vs another heavy call to `vagrant status`.

abrown-sg:
@retr0h Good call, quite a bit faster!

retr0h:
Much better :+1:

rgreinho:
Yeah :+1:

Issue #71, opened 2015-12-28
Install role dependencies with ansible-galaxy?


bborysenko:
Hi,

Could you please give piece of advice about the right way to test ansible role that has some dependencies with molecule? For the moment I use a bit different  directory layout and override some molecule defaults:

```
$ tree .
...
├── tests
│   ├── playbook.yml       <-- playbook to execute during `molecule converge`
│   ├── requirements.yml   <-- requirements for `ansible-galaxy`
│   ├── roles              <-- extra roles installed with `ansible-galaxy`
│   │   └── yum-epel
│   │   └── ...
│   ├── serverspec         <-- `serverspec_dir`
│   │   └── ...
...

$ cat molecule.yml
---
molecule:
  serverspec_dir: tests/serverspec
  testinfra_dir: tests/testinfra
  ansible_config_template: /Users/bbo/.config/molecule/ansible.cfg.j2
ansible:
    playbook: tests/playbook.yml
...

$ cat /Users/bbo/.config/molecule/ansible.cfg.j2
[defaults]

roles_path = tests/roles:../
library    = library
```

And before `molecule converge` I manually run:

```
ansible-galaxy install -r tests/requirements.yml -p tests/roles
```

But I guess that it should be some better way to do it. It will be great to have configuration option with path to a file with requirements for `ansible-galaxy` and install them automatically before `molecule converge`.

retr0h:
@bborysenko I've not given this piece too much thought.  My initial expectation would have people configure roles_path and library to their liking, pretty much like you have done.  This is how we do it as well. I'm kinda curious to see how people want to tackle dependencies, and go from there.  Cool?

abrown-sg:
Fixed by #105.

PR #105, opened 2016-02-02
Automatically install role dependencies

bborysenko:
This is PR for #71. 

The main goal is to simplify the testing of roles that have some dependencies. For the moment I'm using a simple wrapper around molecule to install extra roles with `ansible-galaxy`. But would like to use molecule only for that.

Changes in brief:

- update `ansible.cfg.j2` template to set `roles_path` to something like `.molecule/roles:../`
- add `requirements_file` option to `ansible` section of `molecule.yml`
- if `requirements_file` is set, then role dependencies will be automatically installed with `ansible-galaxy` during `molecule converge` to first directory specified in `roles_path`.

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -34,6 +34,7 @@ from jinja2 import PackageLoader
 
 import molecule.utilities as utilities
 import molecule.validators as validators
+from molecule.ansible_galaxy_install import AnsibleGalaxyInstall
 from molecule.ansible_playbook import AnsiblePlaybook
 from molecule.core import Molecule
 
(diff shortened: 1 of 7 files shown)


retr0h:
:100: 

This is a pretty clean solution.

molecule/commands.py
```diff
@@ -35,6 +35,7 @@
 import molecule.utilities as utilities
 import molecule.validators as validators
 from molecule.ansible_playbook import AnsiblePlaybook
+from molecule.ansible_galaxy_install import AnsibleGalaxyInstall
```
retr0h: This should go above `AnsiblePlaybook` (alphabetical imports)

bborysenko:
@retr0h thanks, fixed

abrown-sg:
:+1:

PR #107, opened 2016-02-02
Path concatenation with os.path.join

bborysenko:
Just change the method to concatenate path, nothing specific.


rgreinho:
:+1:

retr0h:
Nice cleanup :+1:

bborysenko:
Thanks

Issue #106, opened 2016-02-02
Idempotence test fails if ansible_managed is used


bborysenko:
I prefer to add [`ansible_managed`](http://docs.ansible.com/ansible/intro_configuration.html#ansible-managed) at the top of any template inside my roles. But since by default it includes time stamps, idempotence test fails most times. For the moment I just use a custom `ansible.cfg` with own `ansible_managed`. But it forced me to add `ansible.cfg` to each role, and I feel like that is not the best solution.

The other way is to use a [`DEFAULT_ROLES_PATH`](https://github.com/ansible/ansible/blob/devel/lib/ansible/constants.py#L134) environment variable and add it to `raw_env_vars` sections of `molecule.yml`. Well it's a better, but still forced me to keep in mind to add this line to molecule config of each role.

What are you thinking about adding to default `ansible.cfg.j2` template something like (without time stamps):

```
ansible_managed = Ansible managed
```

Other way is add some addition setting to `molecule.yml.j2`, but I guess that it's not likely that someone really needs to override it. There are no more use cases to override it as far I know.

retr0h:
Yeah, we also use ansible_managed in our templates, which is hard coded to be static.  I actually would agree with you, we should set `ansible_managed` to something static in ansible.cfg.j2.

:+1:

abrown-sg:
Fixed by #108

PR #108, opened 2016-02-03
Add ansible_managed to ansible config template

bborysenko:
PR for #106.


dhutty:
I think that since this is closer to the default that ships with Ansible, it would be better.

```ansible_managed = Ansible managed: Do NOT edit this file manually!```

(Not that it matters what *we* do, but by way of example, we actually have something that greps for 'Ansible managed' at the beginning of the file|line)

bborysenko:
Done.

abrown-sg:
:+1:

PR #112, opened 2016-02-04
76/docopt subcommands

### LINKED ISSUE #76: Proposal: leveraging DocOpt subcommands

abrown-sg:
* Fully implements Remy's DocOpt subcommand pattern. Fixes #76.
* Simplifies CLI logic around allowed commands.
* Uses get() to access CLI args throughout since they're not longer guaranteed to be present.
* Fixed a bug where `molecule test --platform xxx` wouldn't preserve platform in state file throughout entire run.
* Fixed a bug where specifying an invalid platform and provider simultaneously would blow up with stack trace.
* Updated test to look for different SystemExit response code when using DocOpt's built-in --version.

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -159,12 +159,12 @@ class Converge(AbstractCommand):
     Provisions all instances defined in molecule.yml.
 
     Usage:
-        converge [--platform=<platform>] [--provider=<provider>] [--tags=<tag1,tag2>] [--debug]
+        converge [--platform=<platform>] [--provider=<provider>] [--tags=<tags>...] [--debug]
 
     Options:
         --platform=<platform>  specify a platform
         --provider=<provider>  specify a provider
-        --tags=<tag1,tag2>     comma separated list of ansible tags to target
+        --tags=<tags>          comma separated list of ansible tags to target
         --debug                get more detail
     """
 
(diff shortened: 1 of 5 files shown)


retr0h:
:+1: looks like a good cleanup.  I'm still an argparse fan tho :p

molecule/commands.py
```diff
@@ -133,16 +155,26 @@ def execute(self):
 
 
 class Converge(AbstractCommand):
+    """
+    Provisions all instances defined in molecule.yml.
+
+    Usage:
+        converge [--platform=<platform>] [--provider=<provider>] [--tags=<tag1,tag2>] [--debug]
```
rgreinho: I would prefer to see ``[--tag=<tag>...]`` to handle multiple tags.

rgreinho:
I definitely like it better like that! :) :+1: 

@retr0h out of here! :stuck_out_tongue_winking_eye:

abrown-sg:
@rgreinho @retr0h Updated. --tag and --tags work without defining --tag as an alias. And I'm not sure I want to change the name of that flag this far into the life of molecule.

abrown-sg:
For now I think we should stay with the existing style of handling tags since it matches what ansible-playbook uses.

PR #113, opened 2016-02-05
Add auto-login support

rgreinho:
If there is only one running host the ``login`` command does not need an
argument, it can detect it automatically. This simplifies the use of
this command.


retr0h:
:+1:

Issue #114, opened 2016-02-08
VirtualBox: enable promiscuous mode on guests


0xDEC0DE:
When standing up OpenStack with molecule + vagrant, it is not possible to pass traffic to VMs, since doing so relies on the network interfaces being set to promiscuous mode, and VirtualBox blocks promiscuous mode on VMs by default.

This can be worked around in the `molecule.yml` with the following:

    raw_config_args:
      - "vm.provider :virtualbox do |vb| vb.customize ['modifyvm', :id, '--nicpromisc1', 'allow-all', '--nicpromisc2', 'allow-all', '--nicpromisc3', 'allow-all'] end"

...but the VirtualBox provider should set that on all guests by default.

conorsch:
Seems like #9 foresaw problems like this. Wouldn't it be better to support OpenStack as a separate backend provider?

retr0h:
I'm going to close this out for now.

Mostly b/c we are no longer testing openstack with vagrant so this feature isn't something we need any longer.

PR #115, opened 2016-02-09
VirtualBox: set all interfaces to promiscuous mode

0xDEC0DE:
Promiscuous mode is blocked on VMs by default on VirtualBox; this
makes things not work as expected.  Enable promiscuous mode on all
private_network interfaces

Fixes Issue #114

### DIFF
--- a/molecule/templates/vagrantfile.j2
+++ b/molecule/templates/vagrantfile.j2
@@ -118,6 +118,10 @@ Vagrant.configure('2') do |config|
     c.vm.hostname = '{{ instance.vm_name }}'
     {%- for interface in instance.interfaces %}
     c.vm.network '{{ interface.network_name }}', type: '{{ interface.type }}', auto_config: {{ interface.auto_config|lower }}
+    # VBox disables promiscuous mode by default.  Enable it.
+    c.vm.provider :virtualbox do |vb|
+      vb.customize ['modifyvm', :id, "--nicpromisc{{ 1 + loop.index }}", 'allow-all']
+    end
     {%- endfor %}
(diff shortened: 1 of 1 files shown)


retr0h:
Is this something we should do by default or have an option for?

bigjools:
I can't see how having the VM blocking the OS from enabling promiscuous mode can help anything. LGTM.

chrisplo:
Have you tried the less drastic [hostonly](https://www.virtualbox.org/manual/ch06.html#network_hostonly) mode for that interface that isn't passing traffic?

0xDEC0DE:
Yes.  It has no effect.  Enabling promiscuous mode allows the MCP to pass traffic to the instances.

conorsch:
It doesn't make sense to enable promiscuous mode by default. Based on #114, you seem to need it enabled for OpenStack support, but that doesn't warrant changing the default, including VirtualBox via Vagrant. Perhaps there's a way to enable it conditionally based on provider, which might be a good compromise.

retr0h:
I like this idea, but it should be optional.

0xDEC0DE:
OK, in what circumstances would you want to block guests from enabling promiscuous mode, then?

retr0h:
> OK, in what circumstances would you want to block guests from enabling promiscuous mode, then?

It should be optional.  The circumstance would be to not setup promiscuous by default, and have a flag which turns it on.

retr0h:
Needs a rebase.  Am happy to merge it once we are able to configure promiscuous enabled/disabled through molecule.yml.

PR #120, opened 2016-02-15
More Verify Visibility

abrown-sg:
* Add more messaging so it's apparent what steps are being run during `molecule verify`
* Add line breaks between steps to help separate them visually
* Standardize color (magenta) for verify messaging

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -261,7 +261,7 @@ class Idempotence(AbstractCommand):
         if self.static:
             self.disabled('idempotence')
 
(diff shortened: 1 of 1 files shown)


rgreinho:
:+1: 

But the magenta color is close to red, don't you think it could be associated with an error instead of an information?

abrown-sg:
It's hard to say with terminal colors because people can configure their terminal programs to render colors however they want. My magenta and red look way different personally.  As far as defaults go, I think the two colors are different enough not to be an issue.
ubuntu-demo

PR #121, opened 2016-02-16
Reworked the core config handling

retr0h:
Broke out the guts of config into anyconfig[1].  Trying to encapsulate config
into more of a model, where the config object handles the semantics of the
config.  This is the first step to reworking the rest of the module.

[1] https://github.com/ssato/python-anyconfig


retr0h:
*WIP* Still need to perform further testing.

PR #122, opened 2016-02-17
Extra line break in `molecule status` for better visibility

abrown-sg:
Before:
screen shot 2016-02-17 at 10 42 35 am

After:
screen shot 2016-02-17 at 10 42 43 am

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -19,6 +19,7 @@
 #  THE SOFTWARE.
 
 from __future__ import print_function
+
 import os
 import pexpect
 import re
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -269,7 +269,7 @@ class Idempotence(AbstractCommand):
 
         if idempotent:
             print('{}Idempotence test passed.{}'.format(Fore.GREEN, Fore.RESET))
-            print
+            print()
             return
 
         # Display the details of the idempotence test.
@@ -329,7 +329,7 @@ class Verify(AbstractCommand):
                 msg = '\n{}Executing testinfra tests found in {}/.{}'
                 print(msg.format(Fore.MAGENTA, serverspec_dir, Fore.RESET))
                 validators.testinfra(inventory_file, **kwargs)
-                print
+                print()
             else:
                 msg = '{}No testinfra tests found in {}/.\n{}'
                 print(msg.format(Fore.YELLOW, testinfra_dir, Fore.RESET))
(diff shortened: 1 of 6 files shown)


molecule/commands.py
```diff
@@ -436,6 +436,7 @@ def execute(self):
         print(x)
         print
         self.molecule._print_valid_platforms()
+        print
```
rgreinho: Just to be consistent, can we use the print function ``print()`` instead of the keyword. Same goes for line 437.

rgreinho:
:+1: after my remarks.

abrown-sg:
@rgreinho Cleaned up the use of print keyword for this change and the last. Interestingly, `print()` will be treated like the keyword and will literally print '()' to the screen. Had to print a blank string to get the intended behavior.

rgreinho:
@abrown-sg Good to know. It is a python 2 specific behavior though. Maybe a python bug?

```
Python 3.5.1 (default, Dec  7 2015, 21:59:08)
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.1.76)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> print()

>>>
```

abrown-sg:
@rgreinho Yeah, looks like a behavior of 2.X.  Should we use print_function from __future__?

```
>>> print()
()
>>> from __future__ import print_function
>>> print()

```

rgreinho:
Sure, that does not sound like a bad idea to me.

abrown-sg:
@rgreinho @retr0h I went ahead and cleaned up the usage of print() everywhere and standardized on using the __future__ version. We were already doing this in utilities. May as well make it consistent.  __future__ imports have to come first in the file.

rgreinho:
Big :+1: !

retr0h:
Why don't we just have `utilities.print()` that implements future print vs having to import all over the place?

rgreinho:
Hummm... because you did not propose this great idea before? :)

abrown-sg:
@retr0h @rgreinho It doesn't look like you can redefine the print statement as a function in Python 2 because it's a language construct. Attempting to do `def print(line):...` in utilities ends in tears. I think to do what @retr0h suggested we'd have to name it something else. Not sure I prefer utilities.my_print() everywhere over just living with the import and using print(). Or doing away with the import and keeping the 2.x behavior until we make a real push toward 3.x compatibility.

abrown-sg:
Disregard that, looks like it's possible with the __future__ stuff. That said, is using utilities.print() everywhere instead of print() cleaner and more clear than a single import line?

rgreinho:
Right, it is kind of the the same. I'm fine with the ``__future__`` import.

abrown-sg:
@rgreinho @retr0h Threw another quick fix into this since it's related.

rgreinho:
:+1:

PR #124, opened 2016-02-17
Fix annoying whitespace in meta/main.yml

cimarron-pistoncloud:
This is a minor change to fix some annoying whitespace left in `meta/main.yml` by ansible-galaxy init
when running `molecule init`.

Please see following gist for before/after behavior and output from `tox`.
https://gist.github.com/cimarron-pistoncloud/ebd7e61de92bc6e14cbe

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -515,6 +515,16 @@ class Init(AbstractCommand):
         init <role>
     """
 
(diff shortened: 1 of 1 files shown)


abrown-sg:
I like this. Clears up a regular annoyance and potentially gives new users a better first impression of molecule. :+1:

PR #130, opened 2016-02-29
Converge before create with tags

abrown-sg:
Because of how commands can call other commands, we need to be aware of how CLI args are also passed around. In this case, passing --tags to converge would fail because if it called create with that argument, DocOpt would error and exit.

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -99,7 +99,7 @@ class Create(AbstractCommand):
     Creates all instances defined in molecule.yml.
 
     Usage:
-        create [--platform=<platform>] [--provider=<provider>] [--debug]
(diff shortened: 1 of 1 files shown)


retr0h:
:+1:

rgreinho:
Just as curiosity, would not it be cleaner to clean the call to ``create` by removing the tags? Cause the tags don't really make sense to me in the create call.

abrown-sg:
@rgreinho @retr0h Please review again. Updated with Remy's feedback.

Since it may be a little confusing, command_args is a list where commands are followed by their arguments, hence deleting them together as a slice after finding position.

abrown-sg:
@rgreinho @retr0h How's this look?

PR #132, opened 2016-03-01
Destroy flexibility for molecule test

### LINKED ISSUE #109: `molecule test` interrupted if one of step failed.
Just realized that actually it's not possible to use just `molecule test` on CI environment. Since we use `sys.exit` [here](https://github.com/metacloud/molecule/blob/master/molecule/ansible_playbook.py#L156) and [here](https://github.com/metacloud/molecule/blob/master/molecule/commands.py#L279), if `converge` or `verify` step is failed we stuck with a VM that need to be destroed manually. 

As workaround I run molecule on CI server like this: 

```
molecule test || molecule destroy
```

But it will be great if molecule will take care about that point itself.

abrown-sg:
* Fixes #109 
* Makes `molecule test` more flexible for CI/CD use by giving it destroy options
* Fixes an edge case where calling `molecule idempotence` on a playbook with errors would print a bunch of broken provisioning output. Now it exits gracefully.
* Makes argument sanitizing cleaner by moving it to a utility function (with test)
* Commands now have multiple return values for capturing exit_code and stdout.
* Commands can now optionally exit during execution error or return the exit status for handling elsewhere. Necessary for supporting commands called directly and programmatically during `molecule test`.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -137,7 +137,6 @@ Here is a commented example:
           - converge
           - idempotence
           - verify
-          - destroy
 
       init:
         # default platform to populate when doing `molecule init`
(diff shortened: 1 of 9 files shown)


retr0h:
:+1:

PR #138, opened 2016-03-10
Config options for default_provider and default_platform

abrown-sg:
These options are not required and are off by default, which will retain the existing behavior of picking the first platform and provider from the list. However, it's possible for users to override this behavior by specifying values in their `~/.config/molecule/config.yml`.


dhutty:
:+1:  Very good!

conorsch:
Sounds great! Thanks for updating the docs too.

PR #139, opened 2016-03-21
Allow partial hostnames with `molecule login`

0xDEC0DE:
Since users typically hate typing any more than they absolutely
have to, allow them to provide partial hostnames to `molecule login`,
and have it do the right thing.


rgreinho:
:+1:

charlieb00:
I also think this change is :+1:

conorsch:
Please yes.

retr0h:
:+1:

PR #143, opened 2016-03-31
Pass testinfra_dir to testinfra

retr0h:
If molecule is configured with `testinfra_dir` in molecule.yml,
it does not use this directory when executing tests.  Updated
molecule to do so.


abrown-sg:
LGTM!

PR #147, opened 2016-04-01
Change default for append_platform_to_hostname

abrown-sg:
Default to False since this seems to be by far the more common use case. Keeps us from having to put it in every single molecule.yml.


retr0h:
:+1:

PR #153, opened 2016-04-19
Remove testinfra as dependency

pcastellazzi:
testinfra is not required to run molecule or its tests. It should not be a dependency. A suggested/recommended package on a package manager that support it may be.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,4 +8,3 @@ PrettyTable==0.7.2
 python-vagrant==0.5.10
 PyYAML==3.11
 sh==1.11
-testinfra
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -5,3 +5,4 @@ testtools
 tox
 wheel
 yapf
+testinfra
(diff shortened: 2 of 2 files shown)


retr0h:
👎 Testinfra is a requirement if one uses testinfra.  I would rather not require additional instructions to install testinfra.

pcastellazzi:
My thinking is the target audience for molecule is versed enough on installing stuff for this to be a problem and since this may cause a conflicts with another tool is best not make it an explicit dependency.

After reading molecule's code a little bit more a more general approach can be made with some extra configuration options, something like:

```yaml
test_framework:
  - path: tests
    command: testinfra tests  ....
  - path: spec
    command: rake spec ...
```

If understood correctly molecule is only shelling out the external tools. I think this would do better than the previous way and allows for more testing frameworks, like BATS, to also be used.

retr0h:
Let me ask a different question.  Why does it matter?  What is the problem this solves?

pcastellazzi:
Please read my comments on #154. The same logic applies here. I will close this PR for now, because we are having the same discussion in two places.

PR #154, opened 2016-04-19
Remove support for opinionated tools

pcastellazzi:
The kind of checks done by some of the validators are more suited to be run from an external tool like [ansible-lint](https://github.com/willthames/ansible-lint).

I think limiting or forcing a tool is not a good idea. For one user one tool can be great, for other a problem. Also this can be easily done with minimal shell scripting.


retr0h:
👎 I'm going to disagree with this PR.  I might be okay with this PR if the custom whitespace validators were removed, and replaced with `ansible-lint`, much like we do with `serverspec`.  

However, I'm not okay with removing all of the validators all together.  Unfortunately, this is a somewhat opinionated test harness.

* People shoudln't write serverspec tests which do not follow some form of best practices.  We chose serverspec.
* People shouldn't write ansible code with trailing whitespace and newlines, that is just sloppy.

The goal of this project is to maintain consistent code.  I'm not willing to remove the consistency.

pcastellazzi:
After thinking a little bit more about the problem i think all the tools should be external, and molecule should provide a generic framework to execute other tools. This PR and #153 are the first steps in that direction. I understand your point about consistent code, but i think what "consistent code" means should be left to the user because different teams have different rules for it. i.e: for you trailing newlines are an issue, but for me they are not, i think a final new line at the end of each file, before EOF should be mandatory to keep POSIX tools happy.

PR #157, opened 2016-05-12
Fix interface definitions in usage docs

DavidWittman:
The usage documentation had `auto_config` listed twice for the first interface, and not at all for the second (which results in an error). I fixed the documentation, but it might also be worth looking into defaulting `auto_config` to `true` in the Vagrantfile template.


retr0h:
👍 Awesome, thanks @DavidWittman!

PR #160, opened 2016-05-13
Pinned to YAPF 0.8.0

retr0h:
We run into issues where older versions of YAPF format the files
differently.  We should all be using the same version.  Also, I
can't stand the 120 line length any longer.  Lets just stick with
what we do everywhere else.

### DIFF
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,3 +1,2 @@
 [style]
 based_on_style = pep8
-column_limit = 120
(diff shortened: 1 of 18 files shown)


abrown-sg:
I prefer 120 personally, but I understand the reason for wanting 80 and can live with it to make retr0h's life better. :)

alop:
looks good to me! 👍 for being consistent...

PR #162, opened 2016-05-14
Run tests in parallel

retr0h:
Testinfra is fairly slow when using the ansible connection backend.
Parallel tests help some.  Although, I'm thinking there is some
fact gathering that is the culprit.


abrown-sg:
👍

alop:
Why stop at 3? According to the docs, n isn't tied to number of hosts, rather, number of tests

alop:
👍 BTW

retr0h:
> Why stop at 3? According to the docs, n isn't tied to number of hosts, rather, number of tests

Yea, we could prob bump this to the number.  I'll need to do some testing to see the difference between 3, 5, 8, 13...

PR #163, opened 2016-05-17
Vagrantfile Template Bug Fix

sebinjohn:
This PR addresses one bug fix and one improvement to the vagrantfile template. 
### Bug
When key values are specified under `provider.options` contains boolean values, the following error occurs

$> molecule create
There was an error loading a Vagrantfile. The file being loaded
and the error message are shown below. This is usually caused by
a syntax error.

Path: <provider config: virtualbox>
Line number: 7
Message: NameError: uninitialized constant False
ERROR: Command '['/usr/local/bin/vagrant', 'up', '--no-provision']' returned non-zero exit status 1

Eg:
```
  providers:
    - name: virtualbox
      type: virtualbox
      options:
        linked_clone: false
```

### Improvement
For the above example when the bug is fixed `vb.linked_clone` will be set twice as shown below
```
config.vm.provider 'virtualbox' do |vb|
    vb.cpus = 2
    vb.linked_clone = false
    vb.memory = 512
    vb.linked_clone = true if Vagrant::VERSION =~ /^1.8/
  end
```

### DIFF
--- a/molecule/templates/vagrantfile.j2
+++ b/molecule/templates/vagrantfile.j2
@@ -40,8 +40,12 @@ Vagrant.configure('2') do |config|
     {%- if provider.type is defined and provider.type == 'virtualbox' and provider.name == current_provider %}
   config.vm.provider 'virtualbox' do |vb|
     {%- if provider.options %}
-      {%- for k, v in provider.options.iteritems()|sort %}
+      {%- for k, v in provider.options.iteritems() | sort %}
+        {%- if v == true or v == false %}
+    vb.{{ k }} = {{ v | lower }}
+        {%- else %}
     vb.{{ k }} = {{ v }}
+        {%- endif %}
       {%- endfor %}
       {%- if not provider.options.memory %}
     vb.memory = 512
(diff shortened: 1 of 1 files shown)


molecule/templates/vagrantfile.j2
```diff
@@ -40,16 +40,20 @@ Vagrant.configure('2') do |config|
     {%- if provider.type is defined and provider.type == 'virtualbox' and provider.name == current_provider %}
   config.vm.provider 'virtualbox' do |vb|
     {%- if provider.options %}
-      {%- for k, v in provider.options.iteritems()|sort %}
+      {%- for k, v in provider.options.iteritems() | sort %}
+        {%- if v == true or v == false %}
```
dhutty: Perhaps use "... == 'true' or v == 'false'" to make it clear that you are doing a *string* comparison?

sebinjohn: Actually, I am doing a boolean comparison here.
for an input like
```
  providers:
    - name: virtualbox
      type: virtualbox
      options:
        linked_clone: false
```
in molecule.yml, the value of `v` will be `False`, which is a jinja boolean value. So I am comparing it with jinja `true` and `false`

retr0h:
👍

PR #168, opened 2016-05-30
Testinfra default

retr0h:
Yes, we are being prescriptive with the defaults we generate.

### Testing a fresh molecule init

```
Executing testinfra tests found in tests/.
platform darwin -- Python 2.7.11, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
rootdir: /private/tmp/foobar, inifile:
plugins: xdist-1.14, testinfra-1.1.2
gw0 [1] / gw1 [1] / gw2 [1]
scheduling tests via LoadScheduling
.
=========================== 1 passed in 1.70 seconds ===========================

No serverspec tests found in spec/.
```

### Testing an existing role with serverspec tests

```
No testinfra tests found in tests/.

Executing rubocop on *.rb files found in spec/.
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.3-compliant syntax, but you are running 2.2.0.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
Inspecting 3 files
...

3 files inspected, no offenses detected

Executing serverspec tests found in spec/.
*** Run serverspec on demo-01 instance ***
/Users/jodewey/.rvm/rubies/ruby-2.2.0/bin/ruby -I/Users/jodewey/.rvm/gems/ruby-2.2.0/gems/rspec-support-3.4.1/lib:/Users/jodewey/.rvm/gems/ruby-2.2.0/gems/rspec-core-3.4.1/lib /Users/jodewey/.rvm/gems/ruby-2.2.0/gems/rspec-core-3.4.1/exe/rspec --pattern spec/\*_spec.rb,spec/demo-01/\*_spec.rb,spec/hosts/demo-01/\*_spec.rb,spec/demo/\*_spec.rb,spec/groups/demo/\*_spec.rb

File "/etc/molecule"
  should be a file
  should be owned by "root"
  should be grouped into "root"
  should be mode 644
  should contain "molecule test file"

Finished in 0.52921 seconds (files took 0.66566 seconds to load)
5 examples, 0 failures
```


abrown-sg:
👍

test-requirements.txt
```diff
@@ -4,4 +4,4 @@ testrepository
 testtools
 tox
 wheel
-yapf==0.8.0
+yapf==0.9.0
```
rgreinho: Just an idea, but as we are always using the latest version of yapf, maybe we do not need to pin it.

rgreinho:
👍  (just a minor remark regarding yapf).

test-requirements.txt
retr0h: > Just an idea, but as we are always using the latest version of yapf, maybe we do not need to pin it.

Yeah, I actually moved to pinning a couple revs ago, b/c YAPF moves so fast, depending on the version of YAPF you are using, you will get different results.  Felt it was safer to pin, so we all format exactly the same way.

PR #171, opened 2016-06-07
Add support for group_vars

rgreinho:
It happens that for testing some roles in a specific context, you want
to use group_vars instead of maintaining copies of the variables in the
molecule playbook.

This patch adds a configuration option in the molecule section allowing
you to specify the path of the group_vars to use.


dhutty:
docs

dhutty:
👍

PR #173, opened 2016-06-08
Rubocop can't share kwargs with testinfra

abrown-sg:
* Gave rubocop/rake its own kwargs since it takes different options than testinfra
* Removed superfluous addition of env to testinfra kwargs since it's now set by ansible-playbook

At some point we should probably consider using a base class for validators to avoid inconsistent args between them.


abrown-sg:
I think @kireledan will want to look at this one. Also, if someone can help me test that it doesn't break docker, that would be great. I'm not setup to use it yet. @retr0h

retr0h:
This looks fine for now.  We can refactor into ABC in the future.  We need to cleanup some things before then anyways.

PR #174, opened 2016-06-09
Reworked imports and ran YAPF

retr0h:
Before reworking the validator wanted to cleanup the classes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,6 @@
 *.egg-info/
 *.py[cod]
 .eggs/
-.idea/
 .tox/
-build/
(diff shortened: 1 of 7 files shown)


timgclark:
👍 for readability

abrown-sg:
Seems like we should enable this for more cleanup:

If you find yourself repeatedly sending reset sequences to turn off color changes at the end of every print, then init(autoreset=True) will automate that:

```
from colorama import init
init(autoreset=True)
print(Fore.RED + 'some red text')
print('automatically back to default color again')
```

PR #175, opened 2016-06-09
Added sudo option for testinfra and some failure handling

kireledan:
This is a small PR to address some small issues I had while working with molecule. 

- By default, testinfra will not run **without** sudo and the ```--sudo``` flag can be passed into ```molecule verify``` or ```molecule test``` to run testinfra with sudo. 

- Testinfra will no longer run if no test files were found (sometimes I didn't need any test and it isn't good for any CI system when it returns a failure code when it didn't actually fail)

- Added a conditional to the rakefile because if no vagrant instances are defined you get a null error. Wanted to give a clearer error message. 

- Moved the ```serverspec_arg``` and ```testinfra_arg``` to be properties in the base class (because they were implemented as properties in the non base classes. 

The reason I wanted to easily be able to toggle sudo is because if you run testinfra on a docker image that has ```Defaults requiretty``` in the ```/etc/sudoers``` file. You're out of luck - testinfra throws a tantrum because it doesn't allocate a pseudo tty.

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -33,6 +33,7 @@ import prettytable
 import sh
 import yaml
 from docopt import docopt
+import glob
 
 import molecule.utilities as utilities
 import molecule.validators as validators
@@ -353,12 +354,13 @@ class Verify(AbstractCommand):
     Performs verification steps on running instances.
 
     Usage:
-        verify [--platform=<platform>] [--provider=<provider>] [--debug]
+        verify [--platform=<platform>] [--provider=<provider>] [--debug] [--sudo]
 
     Options:
         --platform=<platform>  specify a platform
         --provider=<provider>  specify a provider
         --debug                get more detail
+        --sudo                 runs tests with sudo
     """
 
     def execute(self, exit=True):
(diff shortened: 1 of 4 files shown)


molecule/commands.py
```diff
@@ -442,26 +446,32 @@ class Test(AbstractCommand):
...
     """
 
     def execute(self):
         if self.static:
             self.disabled('test')
 
-        command_args, args = utilities.remove_args(self.command_args,
-                                                   self.args, ['--destroy'])
+        command_args, args = utilities.remove_args(
+            self.command_args, self.args, self.command_args)
```
kireledan: Just to point this out - this performs the same thing as before. This just makes it easier to add in flags that dont match the commands in the test chain.

molecule/commands.py
```diff
@@ -442,26 +446,32 @@ class Test(AbstractCommand):
...
 
         for task in self.molecule._config.config['molecule']['test'][
                 'sequence']:
             command = getattr(sys.modules[__name__], task.capitalize())
             c = command(command_args, args)
+
+            for argument in self.command_args:
+                if argument in c.args:
+                    c.args[argument] = self.args[argument]
+
```
kireledan: This is where the flags get passed in if they apply to a specific command.

retr0h:
Nice cleanup 👍

abrown-sg:
Yeah, this looks really good, thanks for the cleanup. 👍

kireledan:
No twitter unfortunately :P

PR #178, opened 2016-06-11
Add '--no-trail' options to verify command for disabling trailing spaces checks

strangeman:
During active development may be useful launch tests as fast as possible, without frustrating lint checks

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -354,13 +354,14 @@ class Verify(AbstractCommand):
     Performs verification steps on running instances.
 
     Usage:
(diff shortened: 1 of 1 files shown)


conorsch:
If you don't like the linting setup, you can use the `molecule.ignore_paths` list to skip linting ([docs](http://molecule.readthedocs.io/en/master/usage.html#configuration)). 

Discussion of linting strategies and options is happening in #182 if you care to chime in.

retr0h:
This tool for better or worse prescriptive.  Disabling linters and validation is something we want to avoid.  See  #182.

PR #179, opened 2016-06-11
Add tab completion support

rgreinho:
This patch provides a script to add tab completion support to molecule.
It auto-completes the subcommands, options and dynamic arguments such as
platform, providers, and hosts.

To ease tabular display and porcelain mode, a new
``_display_tabulate_data`` function was implemented in the Molecule
object.

The documentation was updated accordingly.

### DIFF
--- a/assets/bash_completion/molecule.bash-completion.sh
+++ b/assets/bash_completion/molecule.bash-completion.sh
@@ -1,0 +1,167 @@
+#!/bin/bash
+
+# Credits:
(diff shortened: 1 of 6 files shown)


abrown-sg:
This looks great! I think you can probably pull PrettyTable from requirements.txt since you removed the only reference to it.

rgreinho:
Reference removed.

retr0h:
> Add tab completion support  …

dang `push -f` :)

retr0h:
👍

Issue #185, opened 2016-06-15
Ability to change containers command


retr0h:
Need the ability to change the command the container is executing.  This is necessary when trying to test services such as [etcd](https://github.com/retr0h/ansible-etcd).  Systemd is necessary and /usr/sbin/init needs to run to properly install and test the role. This should be configurable though molecule.yml.

retr0h:
Actually, more I think about it, we shouldn't specify a command, and allow the command the container was built with to [run](https://github.com/retr0h/docker-images/blob/master/Dockerfile-centos7-systemd#L22).

PR #186, opened 2016-06-15
Do not specify command to run

retr0h:
Require the container to supply the command to execute.
Nearly all Dockerfiles supply a command to execute, we shouldn't
supply our own.  This is particularly useful for images which
need to start an init daemon.

Fixes: #185


kireledan:
looks good to me. We'll need to update the documentation then.

abrown-sg:
👍

Issue #183, opened 2016-06-14
`molecule test` is successful if the last test passes (even though another one failed)


techraf:
With:

```
    sequence:
      - idempotence
      - verify
```

If  Specinfra tests pass, even if idempotence test failed, `molecule` returns exit code 0.

With:

```
    sequence:
      - verify
      - idempotence
```

If idempotence test passes, even if Specinfra tests failed, `molecule` returns exit code 0.

Is it by design?

kireledan:
The return code depends on the success of the last running item. It definitely shouldn't return 0 if one of those fails because idempotence will return 1 if the idempotence test fails. I would say this is not as intended and will fix this so that ```test``` will return the non-zero error code if it encounters one.

retr0h:
This should now be resolved.

Issue #180, opened 2016-06-11
Errors are displayed on stdout


rgreinho:
Errors are displayed on the standard output stream instead of the standard error stream. This causes issues when using redirections. For instance adding ``2>/dev/null`` at the end of a command has no effect.

A use case where this is really annoying is with the bash completion. The bash completion scripts relies on the ``molecule status`` command to provide the dynamic arguments. If an error occurs, it will also be displayed in stdout and therefore pollute the completion.

An idea to solve this would be to use the logging library to display (at least) errors instead of ``print()``, as the default logging output for errors is ``sys.stderr``.

retr0h:
This should have been corrected with #183.  @rgreinho can you confirm?

retr0h:
@rgreinho I believe this is no longer the case.  Do you mind verifying.

rgreinho:
@retr0h Just did a small testing using the tab completion, and no error was displayed in the terminal, so it looks like we're good!

PR #191, opened 2016-06-20
Test sequence will now 'fail fast' (#183 fix) 

kireledan:
This PR addresses #183 . If a step in the ```molecule test``` sequence failed that was not the ending step, the command could still return code ```0```. 

Simply put - this PR causes ```molecule test``` to abort if any sequence fails.  Also, I changed the error printing to go to stderr instead of std out in light of #180.

### DIFF
--- a/molecule/utilities.py
+++ b/molecule/utilities.py
@@ -29,8 +29,30 @@ import logging
 import colorama
 import jinja2
 
-log = logging.getLogger()
-log.addHandler(logging.NullHandler())
+
+class LogFilter(object):
+    def __init__(self, level):
+        self.__level = level
+
+    def filter(self, logRecord):
+        return logRecord.levelno <= self.__level
+
+
+logger = logging.getLogger(__name__)
+warn = logging.StreamHandler()
+warn.setLevel(logging.WARN)
+warn.addFilter(LogFilter(logging.WARN))
+warn.setFormatter(logging.Formatter('{}{} --> {}{} %(message)s'.format(
+    colorama.Style.BRIGHT, colorama.Fore.BLUE, colorama.Fore.RESET,
+    colorama.Style.RESET_ALL)))
+
+error = logging.StreamHandler()
+error.setLevel(logging.ERROR)
+error.setFormatter(logging.Formatter('{}{} --> {}{} %(message)s'.format(
+    colorama.Style.BRIGHT, colorama.Fore.RED, colorama.Fore.RESET,
+    colorama.Style.RESET_ALL)))
+logger.addHandler(error)
+logger.addHandler(warn)
 
 
 def merge_dicts(a, b, raise_conflicts=False, path=None):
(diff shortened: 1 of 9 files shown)


dhutty:
👍

rgreinho:
The patch for failing looks nice, and I do not want to be a blocker, but I'm still convinced than we should use the logging library instead of the ``file=sys.stderr`` option of the print function.

kireledan:
I'd be fine adding that to this PR as well. Shouldn't take too long to switch it over to logging.

retr0h:
👍

retr0h:
@kireledan did we want to switch to logging or just go with the current patch?

kireledan:
I think switching to logging will be the best choice. Should have the PR updated before the end of the day.

rgreinho:
@kireledan awesome! I'll be super happy to have a look at it! thanks for doing it.

kireledan:
@rgreinho I'm actually running into an issue with passing the unit tests though. For some reason the logging is not being picked up on the mock sys.stderr. I've tried adding a handler for sys.stderr to see if that resolved the test, but it doesn't. Is the mock sys.stderr just not able to pick it up?

rgreinho:
@kireledan I'm not sure. Add another commit to this PR, so that I can have a look.

kireledan:
I've pushed what I have so far. I must be missing something with the logging configuration.

molecule/utilities.py
```diff
@@ -24,9 +24,14 @@
 import os
 import sys
 
+import logging
+
 import colorama
 import jinja2
 
+log = logging.getLogger()
+log.addHandler(logging.NullHandler())
```
rgreinho: Try replacing these two lines:
```
logging.basicConfig(level=logging.DEBUG)
log = logging.getLogger(__name__)
```

rgreinho: Maybe also renaming ``log`` with ``logger`` (nitpicking I know, but it sound clearer)

kireledan:
Still doesn't seem to be passing the unit test.

kireledan:
I've updated the tests to use a mock logger instead. Also I added some formatting to distinguish between warnings and errors.

rgreinho:
The formatting should go to the logger configuration itself. for instance:
```
logging.basicConfig(format='[%(levelname).4s] %(asctime)s %(message)s')
```
Also the utilities functions ``print_error`` and ``print_warning`` do not make so much sense if we are using the logger.

kireledan:
I have the formatting done in the separate handlers so that there is a difference in formatting for warnings and errors. The ```print_error``` and ```print_warning``` functions are still useful in that we can control the formatting / method of printing throughout the entire application with ease. Maybe change them to ```log_error``` and ```log_warning```?

rgreinho:
The names are fine, my point was more: the abstraction provided by these 2 functions is pretty much useless. Calling the logger directly would be clearer.

kireledan:
Yeah that's a fair point. I just didn't want to have that giant block of code that sets up the log format in each python file.

rgreinho:
``sys.argv < 3`` does no cover all the cases. For instance, ``molecule status --porcelain`` does not display anything.

Instead of using ``sys.argv``, you may want to use the docopt args for this filtering the output of this command:
```
(Pdb) self.args
{'--debug': False,
 '--hosts': False,
 '--platforms': False,
 '--porcelain': False,
 '--providers': False,
 '<command>': 'status'}
```

Also, you must update the tab completion file if you modify the arguments of a command[0]. 
---
[0]: https://github.com/metacloud/molecule/blob/master/assets/bash_completion/molecule.bash-completion.sh#L37

retr0h:
👍

PR #194, opened 2016-06-21
Update requirements

rgreinho:


retr0h:
Should we pin everything then?

kireledan:
I guess it wouldn't hurt to. Although it seems a bit too much. There isn't anything that relies on a specific version as far as I can tell.

rgreinho:
Well, it's best practice. However, YAPF is changing very frequently, so I am in favor of unpinning that one, as there is no benefit of pinning it if we update it every month.

kireledan:
So maybe pin everything except YAPF ? Would that still be in best practice? I'm not familiar with python packaging best practices.

retr0h:
> Well, it's best practice. However, YAPF is changing very frequently, so I am in favor of unpinning that one, as there is no benefit of pinning it if we update it every month.

YAPF was pinned for a reason that different PRs would have different syntax due to different versions being used to perform the feature.  I prefer to keep it pinned so that we all use the same version.

rgreinho:
@retr0h I'm fine with it, as long as we update it regularly. Their updates always bring more clarity to the code formatting.

retr0h:
Should we remove all the pinning w/ the exception of YAPF then?

rgreinho:
I think we should keep them, and update them once in a while.

PR #196, opened 2016-06-23
Initialize ansible roles in offline mode

sebinjohn:
At the moment, `molecule init` requires internet connection to create the scaffolding of an Ansible role.
I find it inconvenient. `ansible-galaxy` supports `--offline` option so we could make use of it.
This PR addresses this.

There are two changes in tox.ini which fixes some of the build failure issues(https://github.com/metacloud/molecule/issues/192)


kireledan:
:+1:

PR #197, opened 2016-06-23
Changed molecule status to display hosts by default.

kireledan:
Currently, if you try running ```molecule status``` you get nothing unless you pass in any flag. It seems more logical to at least display the hosts by default instead of having to pass in anything for that. 

This PR:
* removes ```--hosts``` flag in ```Status```
* Prints the status of hosts by default

### DIFF
--- a/assets/bash_completion/molecule.bash-completion.sh
+++ b/assets/bash_completion/molecule.bash-completion.sh
@@ -34,7 +34,7 @@ _molecule(){
   INIT_OPTIONS="--docker"
   LIST_OPTIONS="--debug -m"
   LOGIN_OPTIONS=""
-  STATUS_OPTIONS="-a --debug --hosts --platforms --porcelain --providers"
+  STATUS_OPTIONS="--debug --hosts --platforms --porcelain --providers"
   TEST_OPTIONS="--debug --platform --provider --tags --sudo"
   VERIFY_OPTIONS="--debug --platform --provider --tags --sudo"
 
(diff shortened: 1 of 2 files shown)


rgreinho:
👎 We need this for the tab completion. If you want to remove it, please provide an alternative.

kireledan:
What should the molecule status command print by default then? It doesn't really make sense that it just prints nothing unless you pass a flag in.

rgreinho:
I see. It does not really bother me, but I would propose this: 
* no arg means "display all", like the ``-a`` arg.
* if (an) arg(s) is/are specified, you display only its/their results.

What do you think?

kireledan:
Yeah. I like that way more. 👍 . I'll change the PR then.

rgreinho:
Fantastic! Thanks.

rgreinho:
``sys.argv < 3`` does no cover all the cases. For instance, ``molecule status --porcelain`` does not display anything.

Instead of using ``sys.argv``, you may want to use the docopt args for this filtering the output of this command:
```
(Pdb) self.args
{'--debug': False,
 '--hosts': False,
 '--platforms': False,
 '--porcelain': False,
 '--providers': False,
 '<command>': 'status'}
```

Also, you must update the tab completion file if you modify the arguments of a command (https://github.com/metacloud/molecule/blob/master/assets/bash_completion/molecule.bash-completion.sh#L37)

rgreinho:
Thank you 👍!

Issue #193, opened 2016-06-21
Number of processes not configurable for PyTest


kernelpanek-tendril:
The number of processes for PyTest is hardcoded for both [Vagrant](https://github.com/metacloud/molecule/blob/master/molecule/provisioners.py#L342) and [Docker](https://github.com/metacloud/molecule/blob/master/molecule/provisioners.py#L628) provisioners.

I ran into an issue where my test is validating a user's ability to stop/start/restart a service and these tests had unexpected results.  The issue only exists when more than 1 process is used but I could not alter this through the molecule.yml file as testinfra_args and testinfra_kwargs are not exposed to the molecule configuration file (or it appears that they are not, but maybe it is documented somewhere?) Let me know if I'm missing something or if you believe this is a legitimate issue with no workaround.

kireledan:
There is no way to pass testinfra args/kwargs at the moment. I do agree that you should be able to though. I'll propose a PR with a way to pass them in.

kernelpanek-tendril:
@kireledan Thank you! I wanted to get a sense on whether anyone felt this would fit into the roadmap for Molecule before I started taking a crack at a PR.

retr0h:
Implemented in #198.

PR #198, opened 2016-06-24
Adds testinfra parameters through molecule configuration

kernelpanek-tendril:
This PR addresses issue #193 and passes arguments to testinfra from molecule configuration.  Parameters merge with provisioner-specific arguments and override defaults.

Example molecule.yml:

```yaml
...

testinfra:
  n: 1

...
```

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -299,6 +299,27 @@ A specific registry can also be defined with the ``registry`` option in the cont
 When accessing a private registry, ensure your docker client is logged into whichever
 registry you are trying to access.
 
+Testinfra
+---------
+
+In the testinfra section, you can configure flags exactly as they're
+passed to `testinfra`. Some flags, such as ``ansible-inventory``,
+``connection`` and ``hosts``, are already set by Molecule. Any flag
+set in this section will override the defaults. See more details on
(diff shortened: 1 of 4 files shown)


kireledan:
👍

retr0h:
Looks like somehow the sudo option was removed from the testinfra options.  Can we please add sudo as a default option outside of docker.

kernelpanek-tendril:
I updated the usage doc, @kireledan.

kernelpanek-tendril:
@retr0h I think we could add the ability to override Testinfra arguments per Provisioner and have defaults set in molecule/conf/defaults.yml for Vagrant.  Does it make sense to open that in a separate issue/PR?

kireledan:
We have the ```--sudo``` flag in ```verify``` for this. Not sure how logical it would be to have each provisioner have a different value of default. And thanks for updating the docs @kernelpanek-tendril

DavidXArnold:
👍

retr0h:
👍

abrown-sg:
I really like this approach. Things were getting a little out of hand before. We should duplicate this for serverspec args in the future.  👍 Also, good to see another Denver user of molecule!

Issue #110, opened 2016-02-03
Feature idea: Add a new molecule command to perform a syntax check


bborysenko:
**IDEA:**  Add a new molecule command to run `ansible-playbook` locally with `--syntax-check` option. Something like:

```
ANSIBLE_CONFIG=.molecule/ansible.cfg ansible-playbook playbook.yml --inventory-file='localhost,' --connection=local --syntax-check
```

Change sequence of commands to run when performing `molecule test` to 

```
  test:
    # sequence of commands to run when performing `molecule test`
    sequence:
      - destroy
      - syntax-check
      - create
```

The main goal is to fail as soon as possible.

**NOTE:** #105 then should go from `converge` step to `syntax-check`

retr0h:
@bborysenko Sounds reasonable to me.

retr0h:
This work was completed.

PR #201, opened 2016-06-27
[Feature Implementation] Molecule Check

kireledan:
This PR implements feature described in #110. This feature is small yet can save some time because the playbook syntax isn't checked until the molecule test instances have been created. 

The test sequence has been modified to fit the check sequence. Now, the ```check``` action follows directly before ```create```.


retr0h:
Nice 👍

abrown-sg:
💯 Thanks!

PR #202, opened 2016-06-27
Added host_vars and group_vars specification

kireledan:
This PR adds the ability to specify host variables and group variables. Sometimes playbooks will require host variables being set and this allows you test those playbooks without modifying the tasks to accommodate molecule. 

Here is an example of adding specifying some variables

``` yaml
    ansible:
      playbook: playbook.yml
      group_vars:
        foo1:
          - test: key
            var2: value
        foo2:
          - test: key
            var: value
      host_vars:
        foo1-01:
          - set_this_value: True
```

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -138,6 +138,7 @@ is where you give molecule role-specific behavior.
         - -o StrictHostKeyChecking=false
         - -o UserKnownHostsFile=/dev/null
 
+
 Ansible
 -------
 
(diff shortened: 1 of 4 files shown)


rgreinho:
👎 as is. We need to be able to test our roles in the context of our groups without having to copy all the variables over.

A compromise could be for you to find a way to include the current approach in your patch.

retr0h:
> 👎 as is. We need to be able to test our roles in the context of our groups without having to copy all the variables over.

@rgreinho Not sure what the problem with this patch is.  If he wishes to add group/host vars, that seems fine to me.  This is an alternate method to your symlinking approach.  Both have their own use cases.

kireledan:
Well so it wouldn't be too big a deal to maybe just have a flag to use the symlink method of obtaining the group_vars. How does that sound?

conorsch:
This isn't a feature I can see myself using, but the use case is something I can at least understand. If I'm understanding the patch correctly, it seems like it could be merged and I could completely ignore, continuing to test host/group vars as I do now, which is to have a `molecule.yml` file in the root of my project directory, so host/group vars are available in adjacent directories, as Ansible expects. 

In general I think there's a bit of a divide between the appropriate of testing Ansible _roles_ vs. _projects_, but that's neither here nor there.

rgreinho:
@kireledan Yes, a flag would be fine. The bottom line is that we must have a way to provide a symlink to existing group_vars.

freeseacher:
i prefer that way over symlink. 
especially for my task. i want to set ansible_python_interpreter for freebsd. 
that approach wiil give me possibility to control interpreter in multi os playbook.

rgreinho:
Cool, I'll test it with our workflow as soon as I can, and I'll give you some feedback.

rgreinho:
👍

PR #208, opened 2016-06-30
Would never update RHEL based images

retr0h:
Required a `-y` flag to `yum update`.  Also, added `yum makecache` prior
to upgrade.


abrown-sg:
👍

PR #210, opened 2016-06-30
Added slack channel info

retr0h:


dhutty:
how about IRC on freenode? If we want to be open (source), shouldn't we also not default to a walled communications garden.

We could use #molecule-users and/or #molecule-dev

retr0h:
> how about IRC on freenode? If we want to be open (source), shouldn't we also not default to a walled communications garden.

I'll let the community decide this one.

retr0h:
In favor of slack.

retr0h:
In favor of freenode.

conorsch:
At the risk of complicating discussion, there's also Gittr, which uses GitHub for authentication. Might be a fne compromise.

dhutty:
After all, since we are using github, PRs are the definitive "where the action happens". Major changes/roadmap planning *should* happen in a github [Issue](https://github.com/metacloud/molecule/issues) as described in the very first line of our [CONTRIBUTING](https://github.com/metacloud/molecule/blob/master/CONTRIBUTING.rst#contributing). Chat channels, partly because of the plethora of protocols/networks and partly because of the lack of open/public history are not suitable for anything but the most ephemeral communication anyway: https://m.signalvnoise.com/is-group-chat-making-you-sweat-744659addf7d#.6jlw4uc6o

There's nothing to stop us publicizing multiple chat channel/network options.

conorsch:
Sounds like Slack wins. :shipit:

retr0h:
> Sounds like Slack wins.

So far.  However, I'm somewhat hesitant publishing a channel when we only have a handful of voters anyways.  Maybe it's moot. 🍤

Mil0dV:
> because of the lack of open/public history

http://slackarchive.io/

barkerd427:
I'd vote for slack.

kostyrevaa:
Kinda sucks when you have to provide your email just to ask a question or to see what you guys are up to.

retr0h:
I'm inclined to choose freenode, simply b/c more ppl seem to be against slack than against freenode.

retr0h:
There are far more people opposed to slack than IRC.  Freenode doesn't require registration to join.  IRC may not be as "kewl" as Slack, but it allows more people to join to conversation.  I'm going with IRC.

conatus:
IRC please.

PR #215, opened 2016-07-01
defend against "AttributeError: NoneType object has no attribute rstrip"

dhutty:


lyssalivingston:
Should there be a test for this?

abrown-sg:
👍

abrown-sg:
Although isn't this going to fail the same way on True?

dhutty:
I was kinda hoping that if record.msg was not None, then it would be a string and .rstrip would be available

PR #218, opened 2016-07-05
Switched to using colorama autoreset

abrown-sg:


retr0h:
It would be nice if we can move the colorama code into a single module.  Along, with moving the future print as well.  We could then have methods named log_error, log_warn, etc... which handle the colorizing, with a single autoreset.

retr0h:
Superseded by  #226.

Issue #217, opened 2016-07-03
Check mode introduces issue with `molecule test`


retr0h:
Check mode from #201 introduces a bug with `molecule test` in the vagrant provisioner.  Does not appear to impact docker.

```
[jodewey:~/git/ansible-systems/roles/etc_hosts] [venv] master+ 1 ± rm -rf .molecule
[jodewey:~/git/ansible-systems/roles/etc_hosts] [venv] master+ 1 ± molecule test
ERROR: Unable to find an inventory file, specify one with -i ?
None
```

retr0h:
Assigning to @kireledan.

kireledan:
Oh interesting. I didn't notice any bugs with mine. I'll take a look when I get back from vacation this weekend.

retr0h:
> Oh interesting. I didn't notice any bugs with mine. I'll take a look when I get back from vacation this weekend.

Yeah, it looks to occur when the inventory file has been removed.  So removing the `.molecule/` directory triggers the bug.

kireledan:
I can't seem to reproduce this. I've deleted my .molecule directory, but my ```molecule test``` still works.

retr0h:
> I can't seem to reproduce this. I've deleted my .molecule directory, but my molecule test still works.

Have you tested with a vagrant provider?  I can't reproduce with docker.

kireledan:
Yeah. Im testing with a vagrant provider.

retr0h:
Corrected in #222.

PR #222, opened 2016-07-05
[Bug Fix] Added inventory flag to check command. 

kireledan:
Addresses issue #217 

Turns out ansible 1.9 needs the ```--inventory-file``` passed in with the syntax-check flag. In ansible 2.1 it does not need it.


retr0h:
Confirmed corrects issue under ansible 1.9 with existing playbooks.

PR #224, opened 2016-07-06
Make all testinfra args configurable

abrown-sg:
The sudo and debug options for testinfra could optionally be set as CLI args, however the way this was done prevented them from also being set as options in the testinra config block. This change allows support for both methods.


dhutty:
👍

PR #227, opened 2016-07-08
Corrected serverspec args

retr0h:
Debug was getting passed as None to serverspec tests.


abrown-sg:
👍

dhutty:
Remove .DS_Store!
This is why we used to have this in .gitignore, to prevent accidental adding of bogus files.

retr0h:
> Remove .DS_Store!
> This is why we used to have this in .gitignore, to prevent accidental adding of bogus files.

Yeah, but I don't like that ;) 

All good git citizens should have their own ~/.gitconfig with the proper git ignores.  What can I say, I'm an idealist.

Issue #148, opened 2016-04-04
[v1.4.1] Testing multiple roles


skornehl:
Hi,

I'm having a test project with "tests" folders in different roles and one in the root folder. When I run ```molecule verify``` in version 1.4.0 it says:

```
Executing testinfra tests found in tests/.
============================= test session starts ==============================
platform darwin -- Python 2.7.9, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
rootdir: /Users/skornehl/workspace/ansible-intro, inifile: 
plugins: testinfra-1.1.1
collected 21 items 

roles/httpd/tests/httpd_test.py ......
```

in version 1.4.1 i only get:

```
Executing testinfra tests found in tests/.
============================= test session starts ==============================
platform darwin -- Python 2.7.9, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
rootdir: /Users/skornehl/workspace/ansible-intro/tests, inifile: 
plugins: testinfra-1.1.1
collected 0 items 
```

My tree looks like this:

```
├── molecule.yml
├── roles
│   ├── httpd
│   │   ├── tasks
│   │   │   └── main.yml
│   │   └── tests
│   │       └── httpd_test.py
├── site.yml
├── stage
└── tests
```

Also: I need an empty test folder in the root directory to make molecule use the folders in the roles, too.

retr0h:
@skornehl have the repo anywhere I can clone?

skornehl:
Hi @retr0h,

here it is: https://github.com/skornehl/ansible-molecule-test

Thanks!

retr0h:
@skornehl I believe we had a bug some time ago that was corrected.  Are you still having this issue on 1.7 series?

retr0h:
Wow @skornehl my apologies.  Not sure how I missed this issue for sooo long.

I have submitted a PR to correct this issue.  You will need to update your `molecule.yml` to make this work properly, since we no longer span the tree for any py files.  We are being prescriptive with tests matching `test_*.py`

```
molecule:
  testinfra_dir: ./roles/*/tests
```

PR #234, opened 2016-07-16
A custom `testinfra_dir` was not being handled

retr0h:
We were performing validation around the spec/ and tests/ dirs
in two different ways/places.  Also, we were not handling the
passing of the `testinfra_dir` correctly to the validator.

Fixes: #148


abrown-sg:
👍

dhutty:
👍

PR #240, opened 2016-07-16
Added Erik as an author

retr0h:
Erik has contributed quite a few huge additions to molecule.
Felt he should be added as an author.


retr0h:
@kireledan feel free to 👎 if you prefer not to be mentioned.

kireledan:
I'm totally fine with this. Thanks for adding me as an author! :)

Issue #236, opened 2016-07-16
Switch to py.test


retr0h:
Once #226 lands, work on moving to py.test.

kireledan:
Yeah on that note I was going to work on adding some more tests so that we don't have to rely on the demo playbook so much.

PR #242, opened 2016-07-17
Converted tests to py.test

retr0h:
* Ported tests over to py.test.  They still require much love, but
  py.test is a much better environment to work within
* Changed assertions to <expected> == <received>
* Moved demo out of tests dir into root of project

Much more to come...

Fixes: #236


abrown-sg:
Wow! This makes a lot more sense given our familiarity with testinfra. 👍

kireledan:
I'm a big fan of this change. 👍

PR #244, opened 2016-07-18
Native support for triggers (as playbooks)

abrown-sg:
Triggers
------------

Triggers are optional Ansible playbooks that are tied to the commands `molecule create` and `molecule destroy`. This can be useful for performing tasks on an instance outside of what is applied as part of your playbook.yml.

For example, if you wanted to make a RHEL instance unregister itself from subscription manager before being destroyed, you would create a playbook that looks like this:

```yaml
- hosts: all
  tasks:
    - name: Unregister from subscription manager
      command: subscription-manager unregister
```

You can target your instances using the hosts portion of your trigger playbooks just like you would in playbook.yml.

Trigger playbooks should be created under the triggers subdirectory:

```
role_name/
├── ...
└── triggers/
    ├── create.yml
    └── destroy.yml
```


retr0h:
I'm actually not entirely sure we even need to do this.  We can put in our playbook pre_task, post_task, etc.. which get run before and after roles.  Not too sure I like having another set of playbooks/directories to deal with.

abrown-sg:
@retr0h I don't see a way with post_tasks to handle the case of having a system perform an action right before it's destroyed. Like unregistering itself from a service, monitoring, inventory, etc. The trigger for create is probably unneeded though.

If we wanted to keep this all in the single playbook.yml, we could have a special tag `molecule_destroy` that is targeted during a `molecule destroy` and skipped during all other operations.

retr0h:
> Like unregistering itself from a service, monitoring, inventory, etc.

Oh right right.

Could we do something a bit more generic, and simply allow molecule to execute a list of plays one at a time.  One could then add create, converge, destroy, and whatever else.  It's just not tied to triggers.

abrown-sg:
> Could we do something a bit more generic, and simply allow molecule to execute a list of plays one at a time. One could then add create, converge, destroy, and whatever else. It's just not tied to triggers.

Not sure I follow. Where would the plays be defined?

retr0h:
> Not sure I follow. Where would the plays be defined?

Well, you are defining playbooks under a triggers directory.  I'm simply saying we should make it more generic.  I'm leaving that exercise up to you to determine.

kireledan:
Couldn't you also just define these in the playbook file? I feel like it wouldn't need to be apart of molecule itself.

abrown-sg:
The case I can't resolve with the playbook file alone is tying plays to `molecule destroy`. For example, Internally, we use the vagrant-triggers plugin to unregister from subscription manager when an instance is destroyed. But that doesn't work with other providers. Also, being able to define those triggers via plays gives us the flexibility of targeting instances using inventory.

To keep this functionality in playbook.yml rather than specialized playbooks, I think we'd need to internally use special tags, like molecule_destroy.

When `molecule destroy` is called it would internally call converge, but would pass --tags molecule_destroy. All other converge commands would have to call --skip-tags molecule_destroy. I like this approach the best since it should noop if nothing is tagged with molecule_destroy.  And while specialized tags is a little bit magical, it's not unheard of, Ansible does it too.

I could also drop the triggers/ directory entirely and just look for the presence of a configurable destroy playbook in the role's root. Like, `if os.file.exists('destroy.yml'): ...`. The tag thing does appeal to me though since it keeps all our tasks in one place.

retr0h:
> unregister from subscription manager

We no longer internally have this use case.  If others in the community aren't leveraging the triggers functionality, we could probably remove this all together.

abrown-sg:
You're right, waste of time.

retr0h:
> You're right, waste of time.

I like the idea of ripping out the triggers code all together.  Although, I don't know if ppl even use it.

Issue #245, opened 2016-07-18
turn off automatic update during molecule test


sebinjohn:
Docker provisioner always tries to update the image to make it ansible compatible.
Is there a way I can turn this off so that a prebuilt image from a local registry can be used for testing?

This can speed up the testing process.

kireledan:
There is currently no way to turn it off. However, I will make a PR to add a flag to skip-build for the docker driver.

retr0h:
Implemented in #246.

PR #246, opened 2016-07-18
[Docker Driver] Added install_python flag to allow skipping of python installation

kireledan:
This PR addresses the issue #245 to allow the user to skip the installation of python on a docker image. 
By default, the docker driver will build an image on top of the user specified image with python guaranteed to be installed. If the user passes in the flag ```install_python: no``` into either the docker driver config or container, it will be skipped. 

## Example
To skip the installation for all instances...
``` yaml
---
docker:
  install_python: no
  containers:
    - name: test1
      image: ubuntu
      image_version: latest
```

To skip the installation on only one instance
``` yaml
---
docker:
  containers:
  - name: test1-01
    ansible_groups:
      - group1
    image: ubuntu
    image_version: latest
    install_python: no
  - name: test1-02
    ansible_groups:
      - group2
    image: ubuntu
    image_version: latest

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -28,7 +28,7 @@ From the root for the project, run:
 .. code-block:: bash
 
   $ tox -r yapf
-  $ source .tox/py27/bin/activate
+  $ source .tox/yapf/bin/activate
   $ yapf -i -r molecule/ tests/
 
 .. _`YAPF`: https://github.com/google/yapf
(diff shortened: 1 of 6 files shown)


retr0h:
Can we update the docs

kireledan:
updated.

retr0h:
👍

abrown-sg:
👍

kireledan:
some of the tests borked after rebasing with pytest change - fixing

retr0h:
We should probably remove from [test-requirements](https://github.com/metacloud/molecule/blob/master/test-requirements.txt#L7), yeah?

retr0h:
👍

Issue #238, opened 2016-07-16
Implement a state object


retr0h:
Implement a state object that manages serialization/deserialization of the state file.  Currently there is too much involved when working with the state file.  Clean up the API around state, so it's easier to work with.

PR #254, opened 2016-07-19
Implemented/integrated a basic state class

retr0h:
A class which manages the state file.  Intended to be used as a singleton
throughout molecule.  The classes' initial state is serialized to disk if the
file does not exist, otherwise is deserialized from the existing state file.
Changes made to the object are immediately serialized.

Fixes: #238

### DIFF
--- a/molecule/commands.py
+++ b/molecule/commands.py
@@ -144,12 +144,9 @@ class Create(AbstractCommand):
         try:
             utilities.print_info("Creating instances ...")
             self.molecule._provisioner.up(no_provision=True)
-            self.molecule._state['created'] = True
+            self.molecule._state.change_state('created', True)
             if self.args['--platform'] == 'all':
-                self.molecule._state['multiple_platforms'] = True
-            else:
-                self.molecule._state['multiple_platforms'] = False
-            self.molecule._write_state_file()
+                self.molecule._state.change_state('multiple_platforms', True)
         except subprocess.CalledProcessError as e:
             utilities.logger.error('ERROR: {}'.format(e))
             if exit:
@@ -186,12 +183,7 @@ class Destroy(AbstractCommand):
         try:
             utilities.print_info("Destroying instances ...")
             self.molecule._provisioner.destroy()
-            self.molecule._state['default_platform'] = False
-            self.molecule._state['default_provider'] = False
-            self.molecule._state['created'] = False
-            self.molecule._state['converged'] = False
-            self.molecule._state['multiple_platforms'] = False
-            self.molecule._write_state_file()
+            self.molecule._state.reset()
         except subprocess.CalledProcessError as e:
             utilities.logger.error('ERROR: {}'.format(e))
             if exit:
@@ -228,17 +220,17 @@ class Converge(AbstractCommand):
         :param create_instances: Toggle instance creation
         :return: Provisioning output
         """
-        if self.molecule._state.get('created'):
+        if self.molecule._state.created:
             create_instances = False
 
-        if self.molecule._state.get('converged'):
+        if self.molecule._state.converged:
             create_inventory = False
 
-        if self.molecule._state.get('multiple_platforms'):
+        if self.molecule._state.multiple_platforms:
             self.args['--platform'] = 'all'
         else:
-            if self.args['--platform'] == 'all' and self.molecule._state.get(
-                    'created'):
+            if self.args[
+                    '--platform'] == 'all' and self.molecule._state.created:
                 create_instances = True
                 create_inventory = True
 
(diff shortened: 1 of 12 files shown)


molecule/provisioners/vagrantprovisioner.py
```diff
@@ -126,10 +128,10 @@ def default_platform(self):
                 0]['name']
 
         # default to first entry if no entry for platform exists or platform is false
-        if not self.m._state.get('default_platform'):
+        if not self.m._state.default_platform:
```
retr0h: I'm not sure I understand why this is even necessary.  Why cant we simply look at `self.m_args['--platform']`

molecule/provisioners/vagrantprovisioner.py
```diff
@@ -126,10 +128,10 @@ def default_platform(self):
...
             return default_platform
 
-        return self.m._state['default_platform']
+        return self.m._state.default_platform
```
retr0h: Same here. :/

molecule/provisioners/vagrantprovisioner.py
```diff
@@ -103,10 +105,10 @@ def default_provider(self):
                 0]['name']
 
         # default to first entry if no entry for provider exists or provider is false
-        if not self.m._state.get('default_provider'):
+        if not self.m._state.default_provider:
```
retr0h: Why is this needed in a state file?

molecule/provisioners/dockerprovisioner.py
```diff
@@ -161,7 +161,6 @@ def build_image(self):
                         tag_string))
```
retr0h: Why does the dockerprovisioner not use the state file, and use it's own status.

/cc @kireledan

abrown-sg:
I believe the point of these was to maintain state after creating a machine. So that if you do `molecule converge --platform ubuntu` and then later do `molecule converge`, it will continue to use the platform that's currently running. Without it, the second `molecule converge` would stand up new instances with whatever the default platform is.  Trying to remember to add --platform to every converge call when you're developing a role is impossible.

retr0h:
> I believe the point of these was to maintain state after creating an instance, and to persist it while the instances are running.

Ahhh, makes sense now.

kireledan:
Definitely a much cleaner way of maintaining state. 👍

molecule/state.py
```diff
@@ -0,0 +1,117 @@
...
+#  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+#  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+#  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+#  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+#  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+#  THE SOFTWARE.
+"""
+A class which manages the state file.  Intended to be used as a singleton
+throughout molecule.  The classes' initial state is serialized to disk if the
```
dhutty: I think you want "class's", since it's a singleton? Alternatively, rephrase for clarity?

molecule/commands.py
```diff
@@ -186,12 +183,7 @@ def execute(self, exit=True):
         try:
             utilities.print_info("Destroying instances ...")
             self.molecule._provisioner.destroy()
-            self.molecule._state['default_platform'] = False
-            self.molecule._state['default_provider'] = False
-            self.molecule._state['created'] = False
-            self.molecule._state['converged'] = False
-            self.molecule._state['multiple_platforms'] = False
-            self.molecule._write_state_file()
+            self.molecule._state.reset()
```
abrown-sg: ❤️ this!

molecule/state.py
```diff
@@ -0,0 +1,117 @@
...
+#  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+#  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+#  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+#  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+#  THE SOFTWARE.
+"""
+A class which manages the state file.  Intended to be used as a singleton
+throughout molecule.  The initial state is serialized to disk if the file does
+not exist, otherwise is deserialized from the existing state file.  Changesmade
```
abrown-sg: Changesmade should be Changes made

abrown-sg:
One small typo, otherwise this is marvelous! 👍

Issue #256, opened 2016-07-20
move commands into directory


retr0h:
The commands class is huge, lets move it into a directory and improve the test coverage.

PR #257, opened 2016-07-20
Moved commands into their own directory

retr0h:
* Broke command classes out into their own files.  The commands file
 was quite large, and breaking out into individual files felt like the
 right thing to do.
* Added missing tests.

Fixes: #256


dhutty:
👍 for the idea. Untested.

abrown-sg:
Excellent! We were long overdue for this. Haven't tested, but 👍 for implementation.

retr0h:
Will merge once we release 1.8.

Issue #259, opened 2016-07-20
molecule init doesn't create tests directory


alop:
```
molecule init alop.test
Traceback (most recent call last):
  File "/usr/local/bin/molecule", line 11, in <module>
    sys.exit(main())
  File "/Users/abelopez/Library/Python/2.7/lib/python/site-packages/molecule/cli.py", line 69, in main
    CLI().main()
  File "/Users/abelopez/Library/Python/2.7/lib/python/site-packages/molecule/cli.py", line 65, in main
    sys.exit(c.execute()[0])
  File "/Users/abelopez/Library/Python/2.7/lib/python/site-packages/molecule/commands.py", line 792, in execute
    with open(os.path.join(testinfra_path, 'test_default.py'), 'w') as f:
IOError: [Errno 2] No such file or directory: './alop.test/tests/test_default.py'
```

kireledan:
I am unable to re-create this. What version are you running?

alop:
molecule -v
1.7.0

kireledan:
Odd. I've tried creating a new environment and installing from pip and also checked out the 1.7 branch. I'm still unable to recreate this error. I'll go ahead and add a check for the directory before making the test defaults to ensure this error does not pop up again for you in later versions.

alop:
no, looks like it was my env.

retr0h:
Readdressed by #261.

Issue #233, opened 2016-07-14
[Feature Suggestion] Init with existing role/playbook


kireledan:
So I was thinking about adding support for using init with existing roles/playbooks. Essentially this would just place a template molecule file in the current directory. Some of the more interesting things I was thinking about is being able to parse the roles in a playbook directory and place them into the generated playbook too so that they are instantly available to run.

Anyone else feel like this is a needed feature and or have some suggestions?

retr0h:
Sounds useful.

retr0h:
Implemented in #249.

moh-abk:
This doesn't create the `./molecule` directory and files so getting this error when I run `molecule test`



`IOError: [Errno 2] No such file or directory: u'.molecule/state.yml'`

kireledan:
I'll look into this really quickly.

retr0h:
Addressed with #261.

PR #261, opened 2016-07-20
Directory checks

kireledan:
This PR address #233 and #259. Although 259 was resolved due to an environment issue, it still doesn't hurt to check if the directory exists. 

For addressing 233, the merge of the state objects seemed to have cause an issue with the creation of the .molecule folder. This has been fixed with a check for the .molecule folder along with a creation of the folder if it is missing.

### DIFF
--- a/molecule/commands/init.py
+++ b/molecule/commands/init.py
@@ -55,8 +55,6 @@ class Init(base.BaseCommand):
             role_path = os.getcwd()
             utilities.print_info(
                 "Initializing molecule in current directory...")
-            if not os.path.isdir(os.path.join(role_path, "tests")):
-                os.mkdir(os.path.join(role_path, "tests"))
 
         else:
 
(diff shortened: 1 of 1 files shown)


retr0h:
Mind making these changes against #257, since that is about to land soon?

kireledan:
Yeah sure thing.

kireledan:
Updated to be used against #257

molecule/core.py
```diff
@@ -61,6 +61,10 @@ def main(self):
         # merge in molecule.yml
         self._config.merge_molecule_file()
 
+        # ensure the .molecule directory exists
+        if not os.path.isdir(os.path.join(os.curdir, ".molecule")):
```
abrown-sg: This should leverage the config value molecule_dir: https://github.com/metacloud/molecule/blob/master/molecule/conf/defaults.yml#L19

abrown-sg:
Also looks like this needs a rebase.

kireledan:
I'll rebase again once #257 is merged in that way there are less conflicts.

retr0h:
👍

PR #263, opened 2016-07-20
Remove vagrant triggers

abrown-sg:
In an effort to move away from provider specific features, we're removing vagrant triggers support. If this feature is needed or being used by anybody, we can implement it in a better way using playbooks.


retr0h:
Nice

👍

PR #267, opened 2016-07-21
Debug cleanup

abrown-sg:
* Debug of running configs now uses safe_dump so it doesn't print more info than we need
* Made spacing between debug blocks more consistent
* Added \__str\__ to state class for easy printing of object contents
* Added state debugging info to `molecule converge`


abrown-sg:
Gonna rework this.

Issue #275, opened 2016-07-22
Molecule init broken


retr0h:
Molecule init broke with the addition of  #260.  Since this is on master (which is currently a develop branch it is not critical).  We have the `stable/1.x` brach for bug/hotfixes.

PR #276, opened 2016-07-23
Corrected `molecule init`

retr0h:
Corrected bug introduced in #268.  Removed duplicate "are we running
the init command" logic from both the core and command base class.

The `init` sub command is unique to other commands.  It doesn't require
a `molecule.yml`.  Moved this logic into a method, in the parent class,
so the children can change the behavior.

Also disabled static inventory introduced in #40.  We never used this,
and the core needs refactored before we attempt to reintroduce it.

Fixes: #275

### DIFF
--- a/molecule/commands/base.py
+++ b/molecule/commands/base.py
@@ -32,58 +32,28 @@ class InvalidHost(Exception):
 class BaseCommand:
     __metaclass__ = abc.ABCMeta
 
-    def __init__(self, command_args, global_args, molecule=False):
+    def __init__(self, command_args, args, molecule=False):
         """
         Initialize commands
 
         :param command_args: arguments of the command
-        :param global_args: arguments from the CLI
+        :param args: arguments from the CLI
         :param molecule: molecule instance
         """
         self.args = docopt.docopt(self.__doc__, argv=command_args)
         self.args['<command>'] = self.__class__.__name__.lower()
         self.command_args = command_args
 
-        self.static = False
+        self.molecule = core.Molecule(self.args)
+        self.main()
 
-        # give us the option to reuse an existing molecule instance
-        if not molecule:
-            self.molecule = core.Molecule(self.args)
-            self.molecule.main()
-        else:
-            self.molecule = molecule
-
-        # init doesn't need to load molecule files
-        if self.__class__.__name__ == 'Init':
-            return
-
-        # assume static inventory if no vagrant config block is defined
-        if self.molecule._provisioner is None:
-            self.static = True
-
-        # assume static inventory if no instances are defined in vagrant config block
-        if self.molecule._provisioner.instances is None:
-            self.static = True
-
-        # Add or update the group_vars if needed.
-        self.molecule._add_or_update_vars('group_vars')
-
-        # Add or update the host_vars if needed
-        self.molecule._add_or_update_vars('host_vars')
-
-        # Update symlinks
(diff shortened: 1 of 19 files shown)


molecule/core.py
```diff
@@ -85,9 +79,6 @@ def main(self):
             self._print_valid_platforms()
             utilities.sysexit()
 
-        if not os.path.exists(self._config.config['molecule']['molecule_dir']):
-            os.makedirs(self._config.config['molecule']['molecule_dir'])
-
```
retr0h: This logic was duplicated twice.

molecule/commands/base.py
```diff
@@ -32,58 +32,28 @@ class InvalidHost(Exception):
...
-            self.static = True
-
-        # Add or update the group_vars if needed.
-        self.molecule._add_or_update_vars('group_vars')
-
-        # Add or update the host_vars if needed
-        self.molecule._add_or_update_vars('host_vars')
-
-        # Update symlinks
-        self.molecule._symlink_vars()
```
retr0h: symlinks, hostvars, and group_vars were moved into core.  Don't see why commands base would need to initialize it, when base calls `molecule.main()`.

molecule/commands/base.py
```diff
@@ -32,58 +32,28 @@ class InvalidHost(Exception):
 class BaseCommand:
     __metaclass__ = abc.ABCMeta
 
-    def __init__(self, command_args, global_args, molecule=False):
+    def __init__(self, command_args, args, molecule=False):
```
abrown-sg: Can we remove molecule from the arg list since we're not using it when it's passed in now?

abrown-sg:
Another great cleanup. I'm glad to see static go for now, there was a lot of logic around it that was confusing. Other than the one question I had above, 👍 from me.

kireledan:
Looking good. 👍

russ216:
Yay i was just running into some confusion about this static part here.

molecule/commands/base.py
retr0h: It's still used as far as I can tell.

molecule/commands/base.py
```diff
@@ -32,58 +32,28 @@ class InvalidHost(Exception):
...
-        self.static = False
+        self.molecule = core.Molecule(self.args)
+        self.main()
 
-        # give us the option to reuse an existing molecule instance
-        if not molecule:
-            self.molecule = core.Molecule(self.args)
-            self.molecule.main()
-        else:
-            self.molecule = molecule
```
abrown-sg: If other subcommands are still passing in and using the molecule instance, I think that will be ignored now since this change makes self.molecule always point to a new instance.

russ216: I was just attempting to do something like this for Issue #279 - and the first one worked fine, or something like `molecule create --provider=openstack` was fine. but when it would run through here and create new molecules, those args were lost. Looks like the ones i was using weren't attempting to reuse the molecule instance

abrown-sg: Yeah, I think that's why we were allowing the Converge command to pass in an instance of molecule instead of creating a new one for each command. You can see calls to converge still try to pass in self.molecule even though it's no longer being reused. I'm gonna do a following PR to re-enable this until we can examine it more closely.

retr0h: @russ216 I have added this logic back.

PR #277, opened 2016-07-25
Added docker volume mounting

kireledan:
This PR adds the ability to mount volumes to specified containers. This is helpful in running services in container that require specific directories to be mounted. 

Docs have been updated with example usage.

# Using the mount bindings
``` yaml
    ---
    docker:
      containers:
        - name: foo-01
          image: ubuntu
          image_version: latest
          volume_mounts:
            - '/this/volume:/to/this:rw'
          command: '/bin/mycommand'
```

## Edit 
Also added the ability to specify the command. This is useful for testing ansible roles that need systemd such as mysql on centos7.

### DIFF
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -24,7 +24,6 @@ import os.path
 from molecule import utilities
 import anyconfig
 
-
 DEFAULT_CONFIG = os.path.join(os.path.dirname(__file__), 'conf/defaults.yml')
 PROJECT_CONFIG = 'molecule.yml'
 LOCAL_CONFIG = '~/.config/molecule/config.yml'
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -21,8 +21,9 @@
 import os
 import os.path
 
-from molecule import utilities
 import anyconfig
+
+from molecule import utilities
 
 DEFAULT_CONFIG = os.path.join(os.path.dirname(__file__), 'conf/defaults.yml')
 PROJECT_CONFIG = 'molecule.yml'
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -47,10 +47,12 @@ class Config(object):
         :param platform: platform name to pass to ``format_instance_name`` call
         :return: None
         """
-        for instance in self.config['vagrant']['instances']:
-            instance['vm_name'] = utilities.format_instance_name(
-                instance['name'], platform,
-                self.config['vagrant']['instances'])
+
+        if 'vagrant' in self.config:
+            for instance in self.config['vagrant']['instances']:
+                instance['vm_name'] = utilities.format_instance_name(
+                    instance['name'], platform,
+                    self.config['vagrant']['instances'])
 
     def molecule_file_exists(self):
         return os.path.isfile(self.molecule_file)
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -21,7 +21,9 @@
 import os
 import os.path
 
+from molecule import utilities
 import anyconfig
+
 
 DEFAULT_CONFIG = os.path.join(os.path.dirname(__file__), 'conf/defaults.yml')
 PROJECT_CONFIG = 'molecule.yml'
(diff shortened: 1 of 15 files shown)


retr0h:
Could we add a test for this?

kireledan:
Tests added as well as ability to specify command as well.

tests/provisioner/test_dockerprovisioner.py
```diff
@@ -70,9 +72,14 @@ def molecule_instance(temp_files, docker_data):
 
 
 @pytest.fixture()
-def docker_instance(molecule_instance):
+def docker_instance(molecule_instance, request):
```
retr0h: nice, did we want to remove the calls to destroy elsewhere?

molecule/config.py
```diff
@@ -47,10 +47,12 @@ def populate_instance_names(self, platform):
         :param platform: platform name to pass to ``format_instance_name`` call
         :return: None
         """
-        for instance in self.config['vagrant']['instances']:
-            instance['vm_name'] = utilities.format_instance_name(
-                instance['name'], platform,
-                self.config['vagrant']['instances'])
+
+        if 'vagrant' in self.config:
```
retr0h: Did I somehow introduce this bug?

tests/provisioner/test_dockerprovisioner.py
kireledan: Yeah I can go ahead and clean those up a bit.

molecule/config.py
kireledan: Potentially - I didn't seem to have the issue on the master branch though. The method should ideally be inside the vagrant provisioner.

retr0h: Could I convince you to migrate it there as part of this effort?

kireledan: Yes. I was also running into an issue where molecule overwrites the ansible.cfg file after each run. I can no longer specify custom configs for changing role path. Going to try and fix it.

molecule/core.py
```diff
@@ -181,20 +178,25 @@ def _remove_templates(self):
...
+            os.remove(self._config.config['ansible']['config_file'])
 
     def _create_templates(self):
         """
         Creates the templates used by molecule.
 
         :return: None
         """
-        # ansible.cfg
+        # # ansible.cfg
```
retr0h: rogue extra `#`?

molecule/config.py
```diff
@@ -39,19 +37,6 @@ def __init__(self, configs=[DEFAULT_CONFIG, PROJECT_CONFIG, LOCAL_CONFIG]):
     def molecule_file(self):
         return PROJECT_CONFIG
 
-    def populate_instance_names(self, platform):
```
retr0h: Was this relocated?  I don't see it moved elsewhere, however, we still have tests for this functionality.

molecule/state.py
```diff
@@ -71,6 +71,10 @@ def default_provider(self):
     def multiple_platforms(self):
         return self._data.get('multiple_platforms')
 
+    @property
+    def customconf(self):
+        return self._data.get('customconf')
+
```
retr0h: Need getter tests.

molecule/config.py
retr0h: Wouldn't ansible.cfg only be rewritten if we run `molecule init`.  Other than that it should never be re-generated.

kireledan: Well before it was just overwriting the custom one. Instead I can have it such that it will only be re-generated if it is missing.

retr0h: I think I'd like the code to work, where init is the only command that creates it.  I dont believe there is any other use case outside of that where it would be created?

kireledan: Sure. I'll go ahead and move it to init then

molecule/config.py
```diff
@@ -21,9 +21,8 @@
 import os
 import os.path
 
-import anyconfig
-
 from molecule import utilities
+import anyconfig
```
abrown-sg: Not sure why this moved? I think our standard order is:

1. Core modules
2. 3rd party modules
3. Our modules

tests/test_state.py
```diff
@@ -57,6 +57,15 @@ def test_default_provider(state_instance):
 def test_multiple_platforms(state_instance):
     assert not state_instance.multiple_platforms
 
+def test_customconf(state_instance):
```
retr0h: I think YAPF will fail due to missing 2 lines between methods.  Might wanna re-run the formatter.

abrown-sg:
👍

retr0h:
👍

PR #280, opened 2016-07-25
Restore the ability to reuse molecule instance

abrown-sg:


molecule/commands/base.py
```diff
@@ -44,8 +44,12 @@ def __init__(self, command_args, args, molecule=False):
...
         self.command_args = command_args
 
-        self.molecule = core.Molecule(self.args)
-        self.main()
+        # allow us to reuse an existing molecule instance
+        if not molecule:
+            self.molecule = core.Molecule(self.args)
+            self.main()
+        else:
+            self.molecule = molecule
```
retr0h: Would be nice to know the use case for this, and if it is even necessary.  I would prefer to leave it out, unless we can duplicate a bug by it not being in.

retr0h: Ah, documented in #276.

retr0h:
👍

retr0h:
```
  flake8: commands succeeded
  yapf: commands succeeded
  docs: commands succeeded
  py27: commands succeeded
  congratulations :)
```

PR #282, opened 2016-07-25
Pushed logic into config class

retr0h:
* Converted `build_easy_paths` to "private".  That logic should be
  handled inside the config class as part of init.  We aren't going
  to initialize a config object without doing this.
* Removed `update_ansible_defaults` all together.  Not sure why this
  is necessary.  These options are specific to ansible, not molecule,
  so should live under the ansible section.
* Removed env['MOLECULE_PLATFORM'].  No idea why env var was being used
  for.


abrown-sg:
Looks good! 👍

alop:
LGTM 👍

Issue #278, opened 2016-07-25
Trap Ctrl-C and exit cleanly


abrown-sg:
If you ctrl+c out of molecule, you end up with a stack trace. We should follow the lead of others and instead exit clean. Should be able to accomplish this in a straightforward way by registering a SIGINT signal handler. Thinking probably in core, or possibly cli.

retr0h:
Closing until @abrown-sg can pick it up.

PR #285, opened 2016-07-26
Added trapping for ctrl+c

abrown-sg:
* Implements #278

### DIFF
--- a/molecule/__init__.py
+++ b/molecule/__init__.py
@@ -18,9 +18,8 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
-from signal import SIGINT
-from signal import signal
-from sys import exit
+import signal
+import sys
 
 from pbr import version
 
@@ -28,10 +27,10 @@ from pbr import version
 def signal_handler(signal, frame):
     msg = '{red}Exiting molecule on Ctrl+c.{reset}'
     print(msg.format(red='\033[31m', reset='\033[0m'))
-    exit()
+    sys.exit()
 
 
-signal(SIGINT, signal_handler)
+signal.signal(signal.SIGINT, signal_handler)
 
 try:
     version_info = version.VersionInfo('molecule')
--- a/molecule/__init__.py
+++ b/molecule/__init__.py
@@ -18,8 +18,9 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
-import signal
-import sys
+from signal import SIGINT
+from signal import signal
+from sys import exit
 
 from pbr import version
 
@@ -27,10 +28,10 @@ from pbr import version
 def signal_handler(signal, frame):
     msg = '{red}Exiting molecule on Ctrl+c.{reset}'
     print(msg.format(red='\033[31m', reset='\033[0m'))
-    sys.exit()
+    exit()
 
 
-signal.signal(signal.SIGINT, signal_handler)
+signal(SIGINT, signal_handler)
 
 try:
     version_info = version.VersionInfo('molecule')
--- a/molecule/__init__.py
+++ b/molecule/__init__.py
@@ -18,7 +18,19 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
+import signal
+import sys
+
 from pbr import version
+
+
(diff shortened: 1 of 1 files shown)


molecule/__init__.py
```diff
@@ -18,8 +18,19 @@
...
 #  THE SOFTWARE.
 
+import signal
+import sys
+
 from pbr import version
 
+
+def signal_handler(signal, frame):
+    print('\033[31mExiting molecule on Ctrl+c.\033[0m')
```
abrown-sg: Prints the message to the console using ANSI red without having outside dependencies on anything else (like colorama).

molecule/__init__.py
```diff
@@ -18,8 +18,20 @@
...
 
+import signal
+import sys
+
 from pbr import version
 
+
+def signal_handler(signal, frame):
+    msg = '{red}Exiting molecule on Ctrl+c.{reset}'
+    print(msg.format(red='\033[31m', reset='\033[0m'))
```
retr0h: We have methods which return red error text.

molecule/__init__.py
```diff
@@ -18,8 +18,20 @@
...
+    sys.exit()
```
retr0h: We have a sysexit method now.

retr0h: Also, not sure if this belongs in init.  Why not install in cli?

abrown-sg: So when testing this, I started out using the molecule.utilities helper functions you mentioned, and had this code in cli.py. I noticed there was a window after executing molecule where ^C would still cause a strack trace. Generally this was occurring during the module import process, and (subjectively) was a second or more after executing the program.

So, the motivation for making the change in this way was to place the signal handler as early in the code path as possible, which makes that window smaller. It's still possible to catch it during the pbr import, but it requires a nearly instant ^C to do that. Registering the handler before that import wouldn't pass pep8.

So, ultimately, I don't have strong feelings about where this goes, but the whole point is to catch ^C, and it doesn't do that as effectively if we let it load up all the imports in cli.py and utilities.py before registering the signal handler.

molecule/__init__.py
```diff
@@ -18,19 +18,20 @@
...
+from sys import exit
 
 from pbr import version
 
 
 def signal_handler(signal, frame):
     msg = '{red}Exiting molecule on Ctrl+c.{reset}'
     print(msg.format(red='\033[31m', reset='\033[0m'))
-    sys.exit()
+    exit()
```
retr0h: Sorry, I meant we have a utility for this.

https://github.com/metacloud/molecule/blob/master/molecule/utilities.py#L218-L219

abrown-sg: Right, but bringing in utilities loads a ton of code before the handler is defined. See my earlier comment for more explanation.

molecule/__init__.py
retr0h: We should use `utilities.logger.error`.

abrown-sg: Even in light of my comment about it making trapping ^C significantly less effective?

molecule/__init__.py
retr0h: Well won't we still get stack traces depending on when ^C is pressed?  If we don't press it when imports are occurring in __init__, whats to say we wont get them when imports are loaded elsewhere?

molecule/__init__.py
retr0h: Well won't we still get stack traces depending on when ^C is pressed? If we don't press it when imports are occurring in init, whats to say we wont get them when imports are loaded elsewhere?

molecule/__init__.py
abrown-sg: As far as I can tell, once the handler is registered, it will be caught even if we're loading other modules after it. It seems to benefit from being as early in the code path as possible, which is why I was trying not to bury it under a bunch of includes.

molecule/__init__.py
abrown-sg: As far as I can tell, once the handler is registered, it will be caught even if we're loading other modules after it. It seems to benefit from being as early in the code path as possible, which is why I was trying not to bury it under a bunch of includes.

retr0h: Ah, I have no idea.  Maybe someone else can give a 👍 👎 .

molecule/__init__.py
```diff
@@ -18,8 +18,21 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
+from signal import SIGINT
+from signal import signal
+from sys import exit
```
retr0h: One small nit.  Third party code in molecule was refactored to `import library`.  Can you follow that?

```
import signal
import sys
```

PR #286, opened 2016-07-26
Expand logging levels for cleaner output

abrown-sg:
* Adds info logging so all output isn't colored yellow
* Adds debug logging to stdout for future cleanup


abrown-sg:
**BEFORE**
screen shot 2016-07-26 at 11 57 09 am

**AFTER**
screen shot 2016-07-26 at 1 14 26 pm

kireledan:
👍  love this.

Issue #181, opened 2016-06-11
Molecule status is too slow (vagrant provider)


rgreinho:
The ``molecule status`` command, used with the vagrant provider is too slow. It takes an average of 3s to execute. This is particularly annoying as this command is the one providing the dynamic arguments during the bash completion.

Would it be possible to extract the information directly from the molecule configuration file rather than querying vagrant for this?

conorsch:
Is running `vagrant ssh-config` any faster? The slowness you're describing is likely vagrant itself. Reading machine information from the YAML files on disk isn't as accurate, since the status info displays dynamic information such as machine state (e.g. "running"). 

Great point about the bash completion though. Can you provide example commands for what you're trying to tab-complete? If it's just machine names, then it would make sense to read them from the local YAML files.

rgreinho:
Yes you're right, the slowness totally comes from vagrant itself.

The tab-completion for any command that can get the dynamic arguments through ``molecule status`` [0] is too slow (~3s to complete). This does not provide a good user experience.

As we are just looking for platforms, providers, or hosts, I think I could just use the molecule.yml instead of using the provider to retrieve this info.

---
[0] | https://github.com/metacloud/molecule/blob/master/assets/bash_completion/molecule.bash-completion.sh#L9-L19

retr0h:
@rgreinho Is this still an issue?

PR #287, opened 2016-07-27
Update and fix serverspec for all platforms

### LINKED ISSUE #199: Multiple platforms support (--platform option not working)
```
$ molecule --version
1.5.1
```

I would assume that Molecule would run `molecule test` for each platform listed in platforms but that doesn't seem to be the case. Below is my `molecule.yml` file.

```yaml
vagrant:
  raw_config_args:
    - "ssh.insert_key = false"

  platforms:
    - name: trusty64
      box: ubuntu/trusty64
    - name: precise64
      box: ubuntu/precise64

  providers:
    - name: virtualbox
      type: virtualbox
      options:
        memory: 1024
        cpus: 2

  instances:
    - name: s1
      options:
        append_platform_to_hostname: no
```

When running `molecule test`: 

Only `trusty64` platform is tested in Vagrant. I would expect Molecule to run tests for each platform listed in `platforms`.

When running `molecule test --platform=precise64`:

Still trusty64 is being used.... `--platform` option doesn't change behaviour.

```
Bringing machine 's1' up with 'virtualbox' provider...
==> s1: Cloning VM...
==> s1: Matching MAC address for NAT networking...
==> s1: Checking if box 'ubuntu/trusty64' is up to date...
```

abrown-sg:
* Updated demo.molecule to have serverspec and testinfra tests that target by ansible group.
* Added method to core to save information about running instances to state. This is used now to pass info to rakefile, but will be useful in future cleanups that try to limit calls to vagrant for info.
* Updated state class to handle new information about running instances (hosts).
* Greatly simplified rakefile, and fixed serverspec so that it works with `molecule test --platform all`.
* Updated rake validator to output at INFO log level to remain consistent.
* Added tests for new core method.
* Fixes #199

### DIFF
--- a/molecule/core.py
+++ b/molecule/core.py
@@ -224,10 +224,10 @@ class Molecule(object):
             instance_name = utilities.format_instance_name(
                 instance['name'], self._provisioner._platform,
                 self._provisioner.instances)
-            instances[instance_name] = []
+            instances[instance_name]['ansible_groups'] = []
             if 'ansible_groups' in instance:
                 for group in instance['ansible_groups']:
-                    instances[instance_name].append(group)
+                    instances[instance_name]['ansible_groups'].append(group)
 
         self._state.change_state('instances', instances)
 
--- a/molecule/core.py
+++ b/molecule/core.py
@@ -18,6 +18,7 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
+import collections
 import fcntl
 import os
 import re
@@ -211,25 +212,28 @@ class Molecule(object):
             self._config.config['molecule']['rakefile_template'],
             self._config.config['molecule']['rakefile_file'], kwargs=kwargs)
 
-    def _create_instances_state(self):
+    @property
+    def _instances_state(self):
         """
         Creates a dict of formatted instances names and the group(s) they're
         part of to be added to state.
 
-        :return: None
+        :return: Dict containing state information about current instances
         """
 
-        instances = {}
+        instances = collections.defaultdict(dict)
         for instance in self._provisioner.instances:
             instance_name = utilities.format_instance_name(
                 instance['name'], self._provisioner._platform,
                 self._provisioner.instances)
-            instances[instance_name]['ansible_groups'] = []
-            if 'ansible_groups' in instance:
-                for group in instance['ansible_groups']:
-                    instances[instance_name]['ansible_groups'].append(group)
 
-        self._state.change_state('instances', instances)
+            if 'ansible_groups' in instance:
+                instances[instance_name][
+                    'groups'] = [x for x in instance['ansible_groups']]
+            else:
+                instances[instance_name]['groups'] = []
+
+        return dict(instances)
 
     def _create_inventory_file(self):
         """
@@ -263,7 +267,7 @@ class Molecule(object):
                     self._provisioner.instances))
 
         inventory_file = self._config.config['ansible']['inventory_file']
-        self._create_instances_state()
+        self._state.change_state('hosts', self._instances_state)
         try:
             utilities.write_file(inventory_file, inventory)
         except IOError:
--- a/molecule/core.py
+++ b/molecule/core.py
@@ -204,13 +204,32 @@ class Molecule(object):
 
         # rakefile
         kwargs = {
-            'molecule_file': self._config.molecule_file,
-            'current_platform': self._provisioner._platform,
+            'state_file': self._config.config['molecule']['state_file'],
             'serverspec_dir': self._config.config['molecule']['serverspec_dir']
         }
         utilities.write_template(
             self._config.config['molecule']['rakefile_template'],
             self._config.config['molecule']['rakefile_file'], kwargs=kwargs)
+
+    def _create_instances_state(self):
+        """
+        Creates a dict of formatted instances names and the group(s) they're
+        part of to be added to state.
+
+        :return: None
+        """
+
+        instances = {}
+        for instance in self._provisioner.instances:
+            instance_name = utilities.format_instance_name(
+                instance['name'], self._provisioner._platform,
+                self._provisioner.instances)
+            instances[instance_name] = []
+            if 'ansible_groups' in instance:
+                for group in instance['ansible_groups']:
+                    instances[instance_name].append(group)
+
+        self._state.change_state('instances', instances)
 
     def _create_inventory_file(self):
         """
@@ -244,6 +263,7 @@ class Molecule(object):
                     self._provisioner.instances))
 
         inventory_file = self._config.config['ansible']['inventory_file']
+        self._create_instances_state()
         try:
             utilities.write_file(inventory_file, inventory)
         except IOError:
--- a/molecule/core.py
+++ b/molecule/core.py
@@ -234,6 +234,9 @@ class Molecule(object):
 
         return dict(instances)
 
+    def _write_instances_state(self):
+        self._state.change_state('hosts', self._instances_state())
+
     def _create_inventory_file(self):
         """
         Creates the inventory file used by molecule and later passed to ansible-playbook.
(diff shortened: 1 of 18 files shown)


abrown-sg:
Hold off for a minute, gonna refactor this to avoid writing a new file.

abrown-sg:
The reason for this change is that serverspec needs to know the names of our instances, which is hard to infer since we can append the platform to the hostname either via CLI (by passing --platform all) or via molecule.yml.  Previously, rakefile was reading in molecule.yml, but this didn't provide enough information, especially with `--platform all`.

This change adds information about running instances and the groups they're part of to the state file, which allows other programs to read it in and act on it. This is how the new rakefile works.

I investigated reading the inventory file directly, but none of the Ruby ini parsers I tried were able to deal with its syntax.

molecule/core.py
```diff
@@ -198,14 +198,33 @@ def _create_templates(self):
...
+        for instance in self._provisioner.instances:
+            instance_name = utilities.format_instance_name(
+                instance['name'], self._provisioner._platform,
+                self._provisioner.instances)
+            instances[instance_name] = []
+            if 'ansible_groups' in instance:
+                for group in instance['ansible_groups']:
+                    instances[instance_name].append(group)
+
+        self._state.change_state('instances', instances)
```
retr0h: I really would like to avoid using the state file as a container for random things.  It should honestly be used sparingly.  It contains state about molecule, this is pushing it IMO.  Whats wrong with leaving the logic in the rakefile like before?

abrown-sg: The issue is that the rakefile needs to know the instance names. We used to be able to infer that from the molecule.yml file, because the names were either as named in the file, or with the platform appended, which is also an option in that file.  That's no longer the case since `--platform all` automatically appends the platform name to a hostname. That's why this was broken. The rakefile was trying to hit demo-01-all.

So we need a way to pass that information to the Rakefile now. Ruby can't easily parse the actual inventory file, so the state file seemed like the next best choice since it's YML, and the actual instance names being used are now part of the state of molecule while they're running.

molecule/state.py
```diff
@@ -68,6 +68,10 @@ def default_provider(self):
         return self._data.get('default_provider')
 
     @property
+    def instances(self):
+        return self._data.get('instances')
```
retr0h: We need tests for this.

molecule/core.py
retr0h: I'm not entirely sold on this, but there isn't really a more elegant solution.  We could also benefit from this data living in the state file per #181.

molecule/core.py
```diff
@@ -198,14 +198,33 @@ def _create_templates(self):
...
-            'molecule_file': self._config.molecule_file,
-            'current_platform': self._provisioner._platform,
+            'state_file': self._config.config['molecule']['state_file'],
             'serverspec_dir': self._config.config['molecule']['serverspec_dir']
         }
         utilities.write_template(
             self._config.config['molecule']['rakefile_template'],
             self._config.config['molecule']['rakefile_file'], kwargs=kwargs)
 
+    def _create_instances_state(self):
```
retr0h: Lets add tests for this.

retr0h: We need unit tests for this.

molecule/state.py
retr0h: We need unit tests for this.

molecule/core.py
retr0h: unit tests that is

abrown-sg: Yup, I gotcha. Just haven't got there yet. This is still a WIP now. Needed better test coverage of group targeting with serverspec and testinfra for this particular change.

molecule/core.py
```diff
@@ -198,14 +198,33 @@ def _create_templates(self):
...
+        instances = {}
+        for instance in self._provisioner.instances:
+            instance_name = utilities.format_instance_name(
+                instance['name'], self._provisioner._platform,
+                self._provisioner.instances)
+            instances[instance_name]['ansible_groups'] = []
+            if 'ansible_groups' in instance:
+                for group in instance['ansible_groups']:
+                    instances[instance_name]['ansible_groups'].append(group)
+
```
retr0h: Lets change the syntax to.

```
hosts:
  foo-centos7:
    groups:
      - one
      - two
```

abrown-sg:
@retr0h Okay, I think this is ready to be reviewed again. I updated the first comment to explain the changes.

abrown-sg:
Sorry for the force push, had to rebase with master to deconflict.

tests/test_core.py
```diff
@@ -59,3 +69,8 @@ def test_parse_provisioning_output_success_00(molecule):
 
     assert res
     assert [] == changed_tasks
+
+
+def test_instances_state_00(molecule):
```
retr0h: this belongs in test_state.py

tests/test_core.py
```diff
@@ -23,9 +23,19 @@
 from molecule import core
 
 
+class TestProvisioner():
+    def __init__(self):
+        self.instances = [{'name': 'demo-01',
+                           'ansible_groups': ['demo', 'demo1'],
+                           'options': {'append_platform_to_hostname': True}}]
+        self._platform = 'centos7'
```
retr0h: The py.test way would be to make this a fixture.

molecule/state.py
```diff
@@ -68,6 +68,10 @@ def default_provider(self):
         return self._data.get('default_provider')
 
     @property
+    def hosts(self):
```
retr0h: Should add a test for the getter in test_state.py

tests/test_core.py
abrown-sg: Can pytest make a class a fixture? Core expects self._provisioner to be a class.

molecule/provisioners/dockerprovisioner.py
```diff
@@ -113,7 +113,7 @@ def build_image(self):
             dockerfile = '''
             FROM {}:{}
             RUN bash -c 'if [ -x "$(command -v apt-get)" ]; then apt-get update && apt-get install -y python sudo; fi'
-            RUN bash -c 'if [ -x "$(command -v yum)" ]; then yum makecache fast && yum update -y && yum install -y python sudo; fi'  # NOQA
+            RUN bash -c 'if [ -x "$(command -v yum)" ]; then yum makecache fast && yum update -y && yum install -y python sudo; fi'
```
retr0h: This prevents flake 8 errors for the long line.

abrown-sg: That's a NOQA that I added and then removed because it was in the context of the line. There's already a noqa in there, though it doesn't work for me.

tests/test_core.py
retr0h: Oh, yeah good point.  There are probably ways to do it, but I'm not going to worry about it.

molecule/provisioners/dockerprovisioner.py
abrown-sg: Looks like my flake8 was old, the existing noqa in the code works for me now. This change was simply removing one that I had added. There's zero net change to that file.

retr0h:
Sweet

molecule/templates/rakefile.j2
```diff
@@ -8,44 +8,22 @@ require 'fileutils'
...
-  if instance.key?('options')
-    if instance['options'].key?('append_platform_to_hostname')
-      if instance['options']['append_platform_to_hostname']
-        instance_name = instance['name'] + '-{{ current_platform }}'
-      end
-    end
-  end
-  instances_to_test[instance_name] = instance
-end
+hosts = YAML.load_file('{{ state_file }}')['hosts']
```
retr0h: For a fute followup,  does this file even need to get rendered through jinja anymore?  Could just be copied to disk, ya?

abrown-sg: We still template the location of the state file and the directory under which to look for serverspec tests.

retr0h:
👍 Looks legit

kireledan:
👍

PR #293, opened 2016-07-28
Removed the link to travis

retr0h:
Our org is not linked to travis, so the badge pointing to a personal
travis URL didn't make sense.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -15,10 +15,6 @@ Molecule
    :target: https://molecule.readthedocs.org/en/latest/
    :alt: Documentation Status
 
-.. image:: https://travis-ci.org/rgreinho/molecule.svg?branch=master
-   :target: https://travis-ci.org/rgreinho/molecule
-   :alt: Build Status
-
 .. image:: https://requires.io/github/rgreinho/molecule/requirements.svg?branch=master
    :target: https://requires.io/github/rgreinho/molecule/requirements/?branch=master
    :alt: Requirements Status


abrown-sg:
👍

rgreinho:
I disagree 👎 . The badge has proven useful several times, and it should be useful now as it shows the build is broken.

I sync my clone about once a week, which I agree may not be enough, but if this frequency is not at par with our development pace, we should have another member, who updates more frequently, placing his badge in the README file.

dhutty:
then make an org account?

retr0h:
> then make an org account?

I'm not willing to grant travis access to our org, and have the ability to clone our private repos.

retr0h:
Will close this out for now.  We will likely just point to our own CI infra vs travis.

retr0h:
Lets revisit this again.

Since the badge is on a personal travis account.  It does not reflect the current state of the project.  Since we will move CI in house, I would like to close this out.

rgreinho:
I see that you're itching to remove the link, but I still think it is super useful. People should totally pay attention to it and master should never be broken (like it is now).

So let's compromise. Let's make the build succeed again, and then we could remove the link. Deal?

retr0h:
Ok, will just get the tests to pass in travis, and hope it stays green for good.  Currently the tests are passing locally, but travis needs some work-a-rounds.  Will submit PRs.

PR #295, opened 2016-07-29
Fixed parse error in openstack driver

kireledan:
This PR fixes some parsing issues the openstack driver had when trying to grab the proper SSH information. Also, it no longer requires a security_group for specifying an openstack instance.


abrown-sg:
👍

retr0h:
👍

PR #301, opened 2016-07-31
Core test cleanup using pytest fixture

abrown-sg:
* Quick fix to use a fixture instead of a bare class.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -23,18 +23,23 @@ import pytest
 from molecule import core
 
 
-class TestProvisioner():
-    def __init__(self):
-        self.instances = [{'name': 'demo-01',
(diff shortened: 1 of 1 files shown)


abrown-sg:
I wanted to use a namedtuple here since it mimics a class somewhat and lets you access its values using dotted notation, but it won't work with field names that start with an underscore. _platform messed it up for us.

retr0h:
I don't think what you did was bad.  We can just leave it at that.  Ultimately we are going to need to instantiate an instance of core, so that we can unit test it.  Currently it's missing tons of coverage.  So this test is going to change regardless.

abrown-sg:
Sounds good.

Issue #297, opened 2016-07-29
Molecule broken - ansible v2.1.1.0


moh-abk:
Molecule version: `1.8.3`
Command: `molecule test`

The latest version of molecule is broken. I first tried with my roles and then I cloned the molecule repo and tried with the demo and go the same error.

```
ERROR! the role 'molecule' was not found in /Users/mohammed/Downloads/molecule-master/demo/molecule/roles:/Users/macuser/Downloads/molecule-master/demo/molecule:/Users/macuser/Downloads/molecule-master/demo/molecule/.molecule/.molecule/roles:/Users/macuser/Downloads/molecule-master/demo/molecule

The error appears to have been in '/Users/macuser/Downloads/molecule-master/demo/molecule/playbook.yml': line 4, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  roles:
    - role: molecule
      ^ here

None
```

retr0h:
Works here.

```
[jodewey:~/git/molecule/demo/molecule] [venv] master+ ± molecule --version
1.8.3
[jodewey:~/git/molecule/demo/molecule] [venv] master+ ± pwd
/Users/jodewey/git/molecule/demo/molecule
[jodewey:~/git/molecule/demo/molecule] [venv] master+ ± molecule test
--> Destroying instances ...
--> Checking playbooks syntax ...

playbook: playbook.yml

--> Creating instances ...
Bringing machine 'demo-01' up with 'virtualbox' provider...
Bringing machine 'demo-02' up with 'virtualbox' provider...
...

**Comment by moh-abk** [comment]
works with `ansible-2.0.2.0` not `ansible-2.1.1.0` just tested now :(

**Comment by retr0h** [comment]
@moh-abk ok confirmed, thanks!

**Comment by retr0h** [comment]
Looks like the problem is with the syntax command.  We need to pass the env to syntax so it can find the role.

**Comment by kireledan** [comment]
Ahh cool. I was wondering why this was happening. I ran into this yesterday too

**Comment by kostyrevaa** [comment]
@moh-abk Actually, it **does not** work with ansible **2.1.0.0**
```
$ ansible --version
ansible 2.1.0.0
$ molecule --version
1.8.3
```

retr0h:
Ansible must have changed ever so slightly.  However, 2.1.0.0 does work when ran manually.

[...]

PR #305, opened 2016-08-01
Added ../../ to rolepath to fix ansible 2.1.1 default role search

kireledan:
This addresses #297. 

For some reason in Ansible 2.1.1, the default search for the role path is relative to the task folder or at least one directory down from where it used to be. Adding this quick bit to the template resolves the error and still makes it compatible with the other ansible versions.


retr0h:
Would be good to find out if this is a bug or desired change.

retr0h:
Opened upstream bug https://github.com/ansible/ansible/issues/16911

However, this PR seems logical.  The default .ansible.cfg lives in `.molecule`, which means it needs to step up two directories.

retr0h:
Merging so we can cherry-pick into stable/1.x and cut a 1.8.4 release of molecule.

PR #306, opened 2016-08-01
Added driver specific functional tests

retr0h:
* Moved functional tests into their own directories.
* Renamed molecule role to full and moved it under the vagrant
  functional tests.
* Removed tasks from the full role which do not have verification
  tests.


abrown-sg:
💯

kireledan:
🎱  👍

PR #307, opened 2016-08-01
Started to implement scenarios

retr0h:
* Added test command scenario to docker.
* Reworked the vagrant driver to support scenarios as well.
* Created a shared role directory for scenarios.


abrown-sg:
👍

PR #309, opened 2016-08-02
Added color to functional tests

retr0h:
It's a bit tricky to identify where the functional tests begin.
This should help with that.


retr0h:
screen shot 2016-08-01 at 6 13 03 pm

abrown-sg:
👍

Issue #314, opened 2016-08-02
Skip tests if docker executable isn't installed


retr0h:
We already have the hooks in place.  Just need to wire them in for docker executable missing.

PR #316, opened 2016-08-02
Skip unit tests if docker executable not found

retr0h:
This matches the work we did in the vagrant provisioner.

Fixes: #314

### DIFF
--- a/tests/unit/provisioner/test_dockerprovisioner.py
+++ b/tests/unit/provisioner/test_dockerprovisioner.py
@@ -18,6 +18,7 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
+import distutils.spawn
 import distutils.version
 import logging
 import pytest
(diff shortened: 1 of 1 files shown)


tests/unit/provisioner/test_dockerprovisioner.py
```diff
@@ -31,10 +32,22 @@
...
-    distutils.version.LooseVersion('2.0'),
-    reason='No Ansible v2 executable found - skipping docker tests')
+
+def ansible_v1():
+    d = distutils.version
+    return (d.LooseVersion(ansible.__version__) < d.LooseVersion('2.0'))
+
+
+def get_docker_executable():
+    return distutils.spawn.find_executable('docker') is None
```
rgreinho: Maybe nitpicking, but I would have done the opposite: if there is no docker executable returns False, otherwise True.

tests/unit/provisioner/test_dockerprovisioner.py
```diff
@@ -31,10 +32,22 @@
...
+
+
+if get_docker_executable():
```
rgreinho: And then reverse or negate the if/else.

rgreinho:
👍

Issue #300, opened 2016-07-31
Ability to target docker hosts


retr0h:
We cannot [target](https://github.com/metacloud/molecule/blob/master/demo/molecule/tests/test_demo1.py#L1-Lundefined) docker hosts in testinfra like we can when using the ansible backend.
This was discovered while creating #298.

The problem is due to our docker driver passing all hosts to testinfra via the `--hosts` flag.  This forces all test files to execute across all hosts.

retr0h:
Raised upstream https://github.com/philpep/testinfra/issues/120

rcuza:
It looks like philpep/testinfra#120 was closed but the `--host` option in molecule does not appear to be a feature for `verify`. Is there a reason something like `molecule verify --host bionic` cannot work? If there isn't, I'll open an issue and investigate making a pull request.

This would be helpful for me to debug issues with roles for a particular OS version. During 
a debugging session, I could use this flag to just check the OS that had a problem instead of testing everything.

lwm:
> If there isn't, I'll open an issue and investigate making a pull request.

go go go go go go go :rocket: :rocket: :rocket: :rocket:

rcuza:
#1731 

@retr0h, @lwm - There seems to be a number of different ways that this was asked for, but I couldn't tell if there was a technical reason for not doing it or a project resource reason. I'll see what I can find out.

If you have suggestions for how to improve the wording of #1731, please let me know in that ticket.

PR #318, opened 2016-08-02
Ability to target ansible groups

retr0h:
This gives the docker container similar functionality to our
vagrant workflow.

Fixes: #300


abrown-sg:
💯

PR #319, opened 2016-08-02
Prep work before a potential 1.9 release

retr0h:
* Change the config merging order to the existing strategy
   used in molecule stable/1.x series.  The merge order is
   default -> local -> project.
* Moved unit test logging into conftest.


abrown-sg:
👍

PR #323, opened 2016-08-03
Pinned runtime requirements

retr0h:
Pulling in the current pip dependencies can cause stability
issues.  We want to pin to known working/tested versions
of pypi packages.  We cannot guarantee against breakage
between major/minor/patch.


retr0h:
```
  syntax: commands succeeded
  docs: commands succeeded
```

tox.ini
```diff
@@ -24,6 +24,8 @@ format = pylint
 [testenv:syntax]
 deps = 
     flake8
+    # NOTE(retr0h): We specifically pin to a known version so formatting
+    # is consistent across developer/PR.
     yapf==0.10.0
```
rgreinho: I absolutely agree with pinning the runtime requirements, but we should not put packages likes that in tox. If we do that, people cannot know that they need to install yapf 0.10.0.

tox.ini
```diff
@@ -32,7 +34,7 @@ commands =
 [testenv:docs]
 passenv = *
 deps=
-    Sphinx==1.4.4
+    Sphinx
     sphinx_bootstrap_theme
```
rgreinho: Same thing for the docs. If I want to build the documentation I have to read the tox file instead of just installing requirements-*.txt files.

tox.ini
retr0h: People don't know they need to install yapf?  It installs yapf for them.  No action on users part.

tox.ini
retr0h: Tox is the single entry point to developing molecule.  If the user wants docs they run `$tox -e docs`.

They don't need to worry about installing requirements or anything.  Tox handles it for them.

tox.ini
rgreinho: That's only one use case. For instance my editor needs know where yapf is, because it reformats my the files every time I save them.

The other concern is totally valid. Here is a pattern I use often to solve that:
```
├── requirements
│   ├── _base.txt
│   ├── _docs.txt
│   ├── local.txt
│   ├── production.txt
│   └── _tests.txt
├── requirements.txt
```
Where:

1. local.txt installs the base, docs and tests packages 
2. production installs base only
3. The main requirements.txt always installs production

tox.ini
rgreinho: I agree with the simplicity of using tox for this use case, but there are way more efficient ways to build the docs:

- Building docs using tox: 1m52s (1st time) / 9s (w/ existing venv)
- Building docs using the Makefile: 2s
- Building docs using setup.py: 3s

retr0h: Sure, but again, tox is the entry point.  I am trying to promote a single way to do things within the community.

tox.ini
retr0h: > That's only one use case. For instance my editor needs know where yapf is, because it reformats my the files every time I save them.

But that happens to be your use case.  Most ppl don't have this wired into their editor.  To be honest, I do the same thing with my editor, and since I am very savy with the project, I have the proper yapf installed on my system.  However, I am trying to keep things simple for the community at large.

rgreinho:
Understanding your concerns, I give a 👍

PR #337, opened 2016-08-07
Added provider core tests

retr0h:
Core has a unique set of tests, some change based on the provider.
Started implementing some of the tests for better coverage.


alop:
Looks like test additions, and some safety checks 👍

Issue #332, opened 2016-08-04
Molecule checks playbook syntax too early (before requirement.yml is resolved)


tknerr:
Using a `requirements.yml` like that:

```yaml
---
- name: timezone
  src: yatesr.timezone
```

With this `molecule.yml`:

```yaml
---
#
# see here for the possible configuration options:
# https://molecule.readthedocs.io/en/master/usage.html#configuration
#

# configure serverspec and testinfra directories
molecule:
  serverspec_dir: disabled
  testinfra_dir: spec

# force the ssh connection
testinfra:
  connection: ssh
  hosts: helloworld
  ssh-config: .vagrant/ssh-config
  spec: yes

# ansible specific configuration, disabling SSH connection sharing
ansible:
  playbook: site.yml
  verbose: vvvv
  requirements_file: requirements.yml
  raw_ssh_args:
    - -o UserKnownHostsFile=/dev/null
    - -o IdentitiesOnly=yes
    - -o ControlMaster=no

# vagrant and instances configuration
vagrant:
  raw_config_args:
    - "ssh.insert_key = false"
  platforms:
    - name: ubuntu1604
      box: bento/ubuntu-16.04
  providers:
    - name: virtualbox
      type: virtualbox
      options:
        name: '"molecule-helloworld"'
        memory: 512
        cpus: 2
  instances:
    - name: helloworld
      options:
        append_platform_to_hostname: no

```

When I run `molecule test`, it fails because it can not find the "timezone" dependency role:

```
Y:\repo\devops\devops-acceptance-tests\vagrant-ansible>molecule test
--> Destroying instances ...
--> Checking playbooks syntax ...
Using .molecule/ansible.cfg as config file
ERROR! the role 'timezone' was not found in /cygdrive/y/repo/devops/devops-acceptance-tests/vagrant-ansible/roles:/cygdrive/y/repo/devops/devops-acceptance-tests/vagrant-ansible:./roles

The error appears to have been in '/cygdrive/y/repo/devops/devops-acceptance-tests/vagrant-ansible/site.yml': line 4, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  roles:
    - timezone
      ^ here

None
```

Imho it should first resolve the dependecies via `ansible-galaxy`, then check the playbook syntax.

retr0h:
Duplicate of #288.

tknerr:
Workaround is to manually put the `syntax` check after the `converge` step:

```yaml
[...]

PR #340, opened 2016-08-09
Added dependency resolution to Syntax check

kireledan:
This PR addresses #332. 

It adds the state ```installed_deps``` to prevent molecule from repeatedly downloading dependencies when they have already been installed. 

This has not been run with the unit tests. I'm currently only able to test it with the docker driver. It seemed to pass on my travis instance. But please feel free to double check this PR with a full tox suite test.

### DIFF
--- a/molecule/ansible_galaxy_install.py
+++ b/molecule/ansible_galaxy_install.py
@@ -91,3 +91,9 @@ class AnsibleGalaxyInstall:
         except sh.ErrorReturnCode as e:
             LOG.error('ERROR: {}'.format(e))
             utilities.sysexit(e.exit_code)
+
+    def download(self, config_file):
+        utilities.print_info('Installing role dependencies ...')
+        self.add_env_arg('ANSIBLE_CONFIG', config_file)
+        self.bake()
+        self.execute()
--- a/molecule/commands/converge.py
+++ b/molecule/commands/converge.py
@@ -23,7 +23,6 @@ import sys
 
 import yaml
 
-from molecule import ansible_galaxy_install
 from molecule import ansible_playbook
 from molecule import utilities
 from molecule.commands import base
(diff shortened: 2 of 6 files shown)


molecule/core.py
```diff
@@ -33,6 +33,7 @@
 from molecule import config
 from molecule import state
 from molecule import utilities
+from molecule import ansible_galaxy_install
```
retr0h: This should be imported alphabetically.

molecule/state.py
```diff
@@ -79,6 +80,10 @@ def multiple_platforms(self):
     def customconf(self):
         return self._data.get('customconf')
 
+    @property
+    def installed_deps(self):
+        return self._data.get('installed_deps')
```
retr0h: Can we get a simple test for this in the state file.  Can just copy an [existing](https://github.com/metacloud/molecule/blob/master/tests/unit/test_state.py#L51-Lundefined) one.

We don't need to really assert it changes, just making sure it doesn't throw an exception due to the property not being there.

molecule/core.py
```diff
@@ -235,6 +236,18 @@ def _instances_state(self):
     def _write_instances_state(self):
         self._state.change_state('hosts', self._instances_state())
 
+    def download_dependencies(self, requirements_file):
+
```
retr0h: Could we remove the extra space between the method and the code?

molecule/commands/converge.py
```diff
@@ -82,14 +81,8 @@ def execute(self,
...
                 'ansible']:
-            utilities.print_info('Installing role dependencies ...')
-            galaxy_install = ansible_galaxy_install.AnsibleGalaxyInstall(
-                self.molecule.config.config['ansible']['requirements_file'])
-            galaxy_install.add_env_arg(
-                'ANSIBLE_CONFIG',
-                self.molecule.config.config['ansible']['config_file'])
-            galaxy_install.bake()
-            output = galaxy_install.execute()
+            self.molecule.download_dependencies(self.molecule.config.config[
```
retr0h: This feels like it belongs in the ansible_galaxy install class vs core.

We should simply call `ansible_galaxy_install.install()` or `ansible_galaxy_install.download()`.  Core IMO is too large and largely untested (#337).

molecule/commands/converge.py
```diff
@@ -25,6 +25,7 @@
 
 from molecule import ansible_playbook
 from molecule import utilities
+from molecule.ansible_galaxy_install import AnsibleGalaxyInstall
```
retr0h: Can we follow our current convention?

`from molecule import ansible_galaxy_install` in alphabetical order.

molecule/ansible_galaxy_install.py
```diff
@@ -91,3 +91,9 @@ def execute(self):
         except sh.ErrorReturnCode as e:
             LOG.error('ERROR: {}'.format(e))
             utilities.sysexit(e.exit_code)
+
+    def download(self, config_file):
```
retr0h: Can we write a unit test for this.

tests/unit/test_state.py
```diff
@@ -118,6 +118,16 @@ def test_change_state_persists(state_instance):
 
         assert d.get('created')
 
+def test_change_state_deps_installed(state_instance):
+    assert not state_instance.created
+
```
retr0h: I think we are okay with just the above.  We don't need the rest of the test.  We already have tests that we can set/get through `change_state`.

retr0h:
I'm in the process of testing this.  I'll update the PR with the peer review.  It's just a couple of quick fixes.

Issue #204, opened 2016-06-28
Update molecule's config syntax to be more like test kitchen


retr0h:
This is a fairly tough job to merge into molecule, however, the config would benefit from following the test kitchen "syntax".  Especially b/c testing various scenarios would be beneficial.  I have started to [prototype](https://github.com/retr0h/maquina/blob/master/tests/conftest.py#L68-L95) this.

This would bump molecule to 2.x as part of the change.

kireledan:
Im a huge fan of this. 👍 👍 👍

dhutty:
I'm very much in favour of this, it was the original intent to be as close to test-kitchen's UX as made sense, to default to the kitchen UX and only deviate when necessary or for good reason. In reality, molecule 1.x UX is quite far from kitchen and this would be a breaking change for (almost?) every molecule.yml, so I agree that this is worth a major version bump.

And to go along with that we should be more careful than usual about documenting the upgrade/testing/releasing, but with that caveat: 👍

tknerr:
+1, would love to see suites support (as long as there can be still multiple instances per suite, which is atm not supported by TK afaik)

retr0h:
Molecule 2.x config syntax proposal.  We would like to use a similar config to test-kitchen.  However, we likely cannot be 100% compatible _nor_ do we need to be.

Below is more vagrant specific, but we should try and normalize the config across drivers. 

```
---
driver:
  name: vagrant
  network:
    - ip: 192.168.50.4
    - type: dhcp
       auto_config: false
  customize:
    memory: 1024
    cpu: 1

provisioner:
  name: ansible
  flag1: foo
  flag2: bar
  env:
    FOO: bar

platforms:
  - name: ubuntu-14.04
     box: ubuntu/trusty64
  - name: ubuntu-12.04
     box: hashicorp/precise64

suites:
  - name: scenario-1
    run_list:
      playbook.yml
```

How do we represent multiple-instances?  Test kitchen combines the suite and platform to create a set of tests it loops through.

I was thinking we treat platforms as instances.  The number of systems listed in platforms == the number of instances we converge.  We then combine that with the suites, so we can create scenario basted testing.

So this would effectively become.

```
platforms:
  - name: ubuntu-14.04
     box: ubuntu/trusty64
  - name: ubuntu-12.04
     box: hashicorp/precise64

suites:
  - name: scenario-1
    run_list:
      playbook1.yml
  - name: scenario-2
    run_list:
      playbook2.yml
```

[...]

PR #360, opened 2016-08-15
Renamed provisioner to driver

retr0h:
So we can start making changes to the core config syntax (#204), we should
start calling the provisioner as driver.  Renamed all mentions of
provisioner to driver.

This is also the beginning of renaming our directories to non-plural.
Will submit a follow-up pull request once the outstanding PRs land.

This is a non-breaking config change.


retr0h:
```
  syntax: commands succeeded
  docs: commands succeeded
  congratulations :)
```

abrown-sg:
This looks good. We're still inconsistent in places where we used the term provisioner and provider interchangeably. Would be nice to get everything renamed "driver," though that's a breaking change for the CLI option. It might make sense to tackle that in another PR or even as part of 2.0 with other breaking changes.

👍

kireledan:
:+1: And I agree with @abrown-sg. It will definitely be better to do that in another PR.

PR #372, opened 2016-08-16
Added auto docker api version recognition to prevent api mismatch errors

kireledan:
This PR alleviates errors that occur when the server and client of a docker daemon/client do not match. It ignores the environment variable DOCKER_API_VERSION so the client needs to be passed the 'auto' version flag to detect the api difference.

### DIFF
--- a/molecule/driver/dockerdriver.py
+++ b/molecule/driver/dockerdriver.py
@@ -23,7 +23,6 @@ import io
 import json
 
 import docker
-from docker.utils import kwargs_from_env
 
 from molecule import util
 from molecule.driver import basedriver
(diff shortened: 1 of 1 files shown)


molecule/driver/dockerdriver.py
```diff
@@ -23,6 +23,7 @@
 import json
 
 import docker
+from docker.utils import kwargs_from_env
```
retr0h: This is already imported as docker.  You're safe to simply use `docker.utils.kwargs_from_env()` below.

kireledan:
```
  syntax: commands succeeded
  doc: commands succeeded
  congratulations :)

retr0h:
👍

PR #373, opened 2016-08-16
Implemented alabaster

retr0h:
Using the alabaster theme so we can get nice logo placement, and
a similar clean theme to testinfra.


kireledan:
Good theme choice 👍

PR #377, opened 2016-08-17
Reworking doc requirements

retr0h:
Also, no longer display the requirements button, which will never
be up to date, since we pin our runtime deps.


abrown-sg:
👍

kireledan:
:+1:

Issue #188, opened 2016-06-16
Does test always need to destroy machine after a full run?


techraf:
If I define my test sequence in `Molecule.yml` like this:

```
molecule:
  test:
    sequence:
      - destroy
      - create
      - converge
      - idempotence
      - verify
#      - destroy
```

my expectation is that the machine will remain up. It doesn't.

Shouldn't the logic be: if user specified `test:` they should be considered in full control (and thus Molecule should expect explicit `destroy` at the end instead of inferring one)?

kireledan:
The behavior of automatically destroying the instances after testing is similar to test-kitchen behavior. To me, this was completely expected, but I see what you mean about being more explicit. 

The ```--destroy``` flag in test is what controls this behavior currently. The choice here would be to either remove the hardcoded destroy sequence and get rid of the ```--destroy``` flag. Or maybe just make it clearer in the docs.

PR #392, opened 2016-08-22
Clarified usage verbiage

retr0h:
We had outdated CLI usage, destroy verbiage, which caused
confusion.

Fixes: #188


alop:
Much clearer 👍

chrisplo:
👍

PR #415, opened 2016-08-26
Added ability to generate and destroy temporary openstack keypair and ssh key file if they are not specified in the molecule.yml

rjfellman:
This change is so that molecule can generate an ssh keyfile and openstack keypair when not specified in the molecule.yml

The reason for this change, is we plan to use this mostly in continuous integration systems. We don't really care about persisting a key or keypair since the instance running molecule will be destroyed once molecule is finished.

Molecule only deletes the keypair and keyfiles that it creates. `/tmp/molecule_rsa` and `molecule_<abcdef0123>`

### DIFF
--- a/molecule/driver/openstackdriver.py
+++ b/molecule/driver/openstackdriver.py
@@ -35,6 +35,26 @@ class OpenstackDriver(basedriver.BaseDriver):
         self._provider = self._get_provider()
         self._platform = self._get_platform()
         self._openstack = shade.openstack_cloud()
+        self._keypair_name = None
+        self._molecule_generated_ssh_key = False
+
+    def get_keyfile(self):
+        if ('keyfile' in self.molecule.config.config['openstack']):
+            return self.molecule.config.config['openstack']['keyfile']
+        elif self._molecule_generated_ssh_key:
+            return util.generated_ssh_key_file_location()
+        else:
+            LOG.info(
+                'Keyfile not specified. molecule will generate a temporary one.')
+            self._molecule_generated_ssh_key = True
+            return util.generate_temp_ssh_key()
+
+    def get_keypair_name(self):
+        if ('keypair' in self.molecule.config.config['openstack']):
+            return self.molecule.config.config['openstack']['keypair']
+        else:
+            LOG.info('Keypair not specified. molecule will generate one.')
+            return util.generate_random_keypair_name('molecule', 10)
 
     @property
     def name(self):
@@ -64,6 +84,13 @@ class OpenstackDriver(basedriver.BaseDriver):
     def platform(self, val):
         self._platform = val
 
+    def molecule_generated_keypair(self):
+        return 'keypair' not in self.molecule.config.config['openstack']
+
+    @property
+    def molecule_generated_ssh_key(self):
+        return self._molecule_generated_ssh_key
+
     @property
     def valid_providers(self):
         return [{'name': self.provider}]
@@ -79,6 +106,9 @@ class OpenstackDriver(basedriver.BaseDriver):
     @property
     def ansible_connection_params(self):
         return {'connection': 'ssh'}
+
+    def keypair_name(self):
+        return self._keypair_name
 
     @property
     def testinfra_args(self):
@@ -109,8 +139,8 @@ class OpenstackDriver(basedriver.BaseDriver):
                     flavor=self._openstack.get_flavor(instance['flavor']),
                     auto_ip=True,
                     wait=True,
-                    key_name=self.molecule.config.config['openstack'][
-                        'keypair'], security_groups=instance['security_groups']
+                    key_name=self._keypair_name,
+                    security_groups=instance['security_groups']
                     if 'security_groups' in instance else None)
                 util.reset_known_host_key(server['interface_ip'])
                 instance['created'] = True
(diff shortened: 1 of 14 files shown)


kireledan:
:+1: Nice addition

molecule/driver/openstackdriver.py
```diff
@@ -64,6 +84,13 @@ def platform(self):
     def platform(self, val):
         self._platform = val
 
+    def molecule_generated_keypair(self):
+        return 'keypair' not in self.molecule.config.config['openstack']
+
+    @property
+    def molecule_generated_ssh_key(self):
+        return self._molecule_generated_ssh_key
+
```
retr0h: I don't think a property is really needed.  We generally use properties in our drivers for abstract properties that are common across all drivers.

molecule/util.py
```diff
@@ -32,6 +32,8 @@
 
 colorama.init(autoreset=True)
 
+GENERATED_SSH_KEY_LOCATION = '/tmp/molecule_rsa'
```
retr0h: Would be nice to use something that chooses the OS tempdir.

molecule/util.py
```diff
@@ -189,20 +191,51 @@ def reset_known_host_key(hostname):
...
             paramiko.SSHException, socket.error):
         time.sleep(timeout)
         return False
 
 
+def generated_ssh_key_file_location():
+    return GENERATED_SSH_KEY_LOCATION
+
+
+def generate_temp_ssh_key():
```
retr0h: Does this really belong in util?  Since it is specific to openstack ,and only used int he openstack driver, prob should be relocated to openstack driver class.

kireledan: It has the potential to be used for the AWS driver too. So it might be a useful thing to keep in util.

retr0h: Sure, I understand, but current it isn't.  If and when that time comes we can evaluate.

rjfellman:
Thanks to both of you for checking it out. I think those are all great suggestions, I'll make the changes. I put it in util because `check_ssh_availability` was also only used in openstack, but seemed to be put there as a place where it could grow. 

```
[~/Development/comcast/sed]$ grep -R "check_ssh_availability" *
molecule/molecule/driver/openstackdriver.py:                while not util.check_ssh_availability(
molecule/molecule/driver/openstackdriver.py:            return util.check_ssh_availability(instance_ip,
molecule/molecule/util.py:def check_ssh_availability(hostip, user, timeout):
molecule/test/unit/test_util.py:def test_check_ssh_availability():
```

I can pull mine out into the driver and that one out too if you'd like.

retr0h:
> I can pull mine out into the driver and that one out too if you'd like.

Yeah, `check_ssh_availability` should be in openstack driver as well.

rjfellman:
Okay this is ready for review again - thanks again for the suggestions! Let me know if there's anything else.

molecule/driver/openstackdriver.py
```diff
@@ -253,14 +258,61 @@ def _set_keypair(self):
             self._openstack.create_keypair(kpn, open(pub_key_file,
                                                      'r').read().strip())
 
+    def reset_known_host_key(self, hostname):
+        return os.system('ssh-keygen -R {}'.format(hostname))
+
+    def check_ssh_availability(hostip, user, timeout, sshkey_filename):
+        import socket
```
retr0h: Lets move this import to the top of the file in alphabetical order, like we do with the rest of our imports.

retr0h:
Hi @rjfellman looking better.  Just about there.  Could we be sure to make these methods "private"?  We tend to follow the pattern of _method_name for things that are not used outside the module.  Other than that, I think your feature is looking good.

molecule/driver/openstackdriver.py
```diff
@@ -22,6 +22,9 @@
 import os
 
 import shade
+import paramiko
+import time
+import tempfile
```
retr0h: Since time and tempfile are part of the standard library, they should be imorted with OS, in alphabetical order.  Paramiki should stay with shade, however, should go before it.

molecule/driver/openstackdriver.py
```diff
@@ -109,16 +141,18 @@ def up(self, no_provision=True):
                     flavor=self._openstack.get_flavor(instance['flavor']),
                     auto_ip=True,
                     wait=True,
-                    key_name=self.molecule.config.config['openstack'][
-                        'keypair'], security_groups=instance['security_groups']
+                    key_name=self._keypair_name,
```
retr0h: This is referenced as _keypair_name, but we have a method named `keypair_name` above?

molecule/driver/openstackdriver.py
```diff
@@ -80,6 +109,9 @@ def ssh_config_file(self):
     def ansible_connection_params(self):
         return {'connection': 'ssh'}
 
+    def keypair_name(self):
+        return self._keypair_name
+
```
retr0h: Looks like this isn't used?

molecule/driver/openstackdriver.py
```diff
@@ -35,6 +38,26 @@ def __init__(self, molecule):
         self._provider = self._get_provider()
         self._platform = self._get_platform()
         self._openstack = shade.openstack_cloud()
+        self._keypair_name = None
+        self._molecule_generated_ssh_key = False
```
bborysenko: 3уж яся о

rjfellman: thanks, for checking it out. Can you please give me an english translation?

rjfellman:
So i think i've addressed everything from your comments @retr0h . @bborysenko - if you can please give me an english translation i'd appreciate it.

retr0h:
Thanks @rjfellman I'll look it over this week.

retr0h:
@rjfellman I made a few updates to your PR.  We have limited openstack tests ATM.  Please verify all is well.

PR #474, opened 2016-09-14
Corrected molecule login usage

retr0h:
When cutting over to click, `molecule login` usage was not preserved.


alop:
Looks fine 👍

Issue #496, opened 2016-09-21
Testinfra doesn't pass the sudo option


retr0h:
The sudo option does not get passed to testinfra?

```
verifier:
  name: testinfra
  options:
    sudo: True
    v: True
```

retr0h:
Pushed molecule 1.11.2 to pypi.

PR #497, opened 2016-09-21
Preserve sudo passed in config section

retr0h:
The sudo option was getting reset, and would not respect option set
in molecule.yml.

Fixes: #496

### DIFF
--- a/molecule/verifier/testinfra.py
+++ b/molecule/verifier/testinfra.py
@@ -55,7 +55,8 @@ class Testinfra(base.Base):
             self._molecule.verifier_options)
         testinfra_options['env'] = ansible.env
         testinfra_options['debug'] = self._molecule.args.get('debug', False)
-        testinfra_options['sudo'] = self._molecule.args.get('sudo', False)
+        if self._molecule.args.get('sudo'):
+            testinfra_options['sudo'] = True
 
         tests_glob = self._get_tests()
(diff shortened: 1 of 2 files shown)


molecule/verifier/testinfra.py
```diff
@@ -55,7 +55,8 @@ def execute(self):
             self._molecule.verifier_options)
         testinfra_options['env'] = ansible.env
         testinfra_options['debug'] = self._molecule.args.get('debug', False)
-        testinfra_options['sudo'] = self._molecule.args.get('sudo', False)
+        if self._molecule.args.get('sudo'):
+            testinfra_options['sudo'] = True
```
signed8bit: Would this affect the debug flag in the same way?

retr0h: I don't believe so, b/c the debug flag is never passed in as an option to testinfra.

signed8bit: So this one is really just not a True/False? Meaning that 'sudo' will never be present in self._molecule.args with a value of False?

retr0h: sudo in the context of `self._molecule.args` would apply here.

One could override testinfra to always set sudo, or simply supply on the command line.

retr0h:
```
  syntax: commands succeeded
  doc: commands succeeded
  congratulations :)
```

Issue #289, opened 2016-07-27
molecule has lots of dependencies


dhutty:
Molecule should try: import foo the modules that are needed for each provisioner and only error if you attempt to use a provisioner that's not available.

retr0h:
What do you mean @dhutty?  The requirements file brings in the necessary dependencies.  The tests  will be testing all the dependencies.  So unfortunately if you don't have docker installed on your system, you wont be able to run unittests.  Just the nature of the beast.

Can you give more details?

dhutty:
Sure, if you want to test molecule itself. But if I want merely to *use* molecule, I don't necessarily want to bother installing, for example, all of the openstack deps if I'm never going to use it because I'm, say, a virtualbox user.

kireledan:
You do have a point - it is a bit much to have all of those dependencies if the user only wants to use a specific driver. This is definitely something we should look into.

retr0h:
> Sure, if you want to test molecule itself. But if I want merely to use molecule, I don't necessarily want to bother installing, for example, all of the openstack deps if I'm never going to use it because I'm, say, a virtualbox user.

But what does it hurt?  They are python dependencies as the projects dependencies.  How does this matter?  The python dependencies shouldn't be anywhere else but in the requirements file for setup.py.

tknerr:
test-kitchen does this via additional drivers that can be installed as plugins. Something like this would be nice. Eg I don't need all the openstack dependencies while I'm not going to use them. They should be optional.

retr0h:
@tkimball83 @kireledan @dhutty at some point we will break things out into plugins.  The project is in it's early stages.  It's too premature ATM.

retr0h:
Closing this in favor of #312, which will address the dependencies.

PR #499, opened 2016-09-22
Removed the driver requirements

retr0h:
Now must install the appropriate python package based on the
driver used.  Opted not to add try catch around the imports,
since this is called out in the docs, and fairly obvious.
The requirements are also pinned in test-requirements so we
can test against a specific known version.

Fixes: #289

### DIFF
--- a/molecule/driver/dockerdriver.py
+++ b/molecule/driver/dockerdriver.py
@@ -21,11 +21,12 @@
 import collections
 import io
 import json
+import sys
 
 try:
     import docker
 except ImportError:
-    LOG.error('Driver missing, install docker-py!')
+    sys.exit('ERROR: Driver missing, install docker-py!')
 
 from molecule import util
 from molecule.driver import basedriver
--- a/molecule/driver/dockerdriver.py
+++ b/molecule/driver/dockerdriver.py
@@ -22,7 +22,10 @@ import collections
 import io
 import json
 
-import docker
+try:
+    import docker
+except ImportError:
+    LOG.error('Driver missing, install docker-py!')
 
 from molecule import util
 from molecule.driver import basedriver
--- a/molecule/driver/openstackdriver.py
+++ b/molecule/driver/openstackdriver.py
@@ -22,6 +22,7 @@ import collections
 import os
 import random
 import socket
+import sys
 import tempfile
 import time
 
@@ -29,7 +30,7 @@ import paramiko
 try:
     import shade
 except ImportError:
-    LOG.error('Driver missing, install shade!')
+    sys.exit('ERROR: Driver missing, install shade!')
 
 from molecule import util
 from molecule.driver import basedriver
--- a/molecule/driver/openstackdriver.py
+++ b/molecule/driver/openstackdriver.py
@@ -26,7 +26,10 @@ import tempfile
 import time
 
 import paramiko
-import shade
+try:
+    import shade
+except ImportError:
+    LOG.error('Driver missing, install shade!')
 
 from molecule import util
 from molecule.driver import basedriver
--- a/molecule/driver/vagrantdriver.py
+++ b/molecule/driver/vagrantdriver.py
@@ -22,7 +22,10 @@ import collections
 import copy
 import os
 
-import vagrant
+try:
+    import vagrant
+except ImportError:
+    LOG.error('Driver missing, install python-vagrant!')
 
 from molecule import util
 from molecule.driver import basedriver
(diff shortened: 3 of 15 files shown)


abrown-sg:
Looks good. Would be nice if molecule threw some sort of error about reading the docs if no drivers are installed. That probably gets into the try/catch scenario you were trying to avoid though.

retr0h:
> Looks good. Would be nice if molecule threw some sort of error about reading the docs if no drivers are installed since that's the most likely failure scenario. That probably gets into the try/catch situation you were trying to avoid though.

Yeah, if people can't figure out why docker-py isn't found, they probably shouldn't be using python or developing roles.  It's well documented, and really hate catering to newbz.

retr0h:
```
  syntax: commands succeeded
  doc: commands succeeded
  congratulations :)
```

signed8bit:
How do we feel about the case where someone wants to try out molecule does the following.
        
`pip install molecule `

Reads enough of the docs to do molecule init and gets a stack. It might be cool to add some exception handling around the specific driver import and add some hinting as to what the end user needs to do.

```
try:
    import vagrant
except ImportError:
    LOG.error('Driver missing, install python-vagrant')
```

retr0h:
> How do we feel about the case where someone wants to try out molecule does the following.

I think the docs and README adequately document what is necessary.  README entails everything necessary to use molecule for the first time.

retr0h:
> I think the docs and README adequately document what is necessary. README entails everything necessary to use molecule for the first time.

Since @signed8bit and @abrown-sg like this, I'll add it :)

signed8bit:
Thanks @retr0h! I can certainly see someone trying to use molecule that isn't a very/all fluent user of python from a development standpoint and we want to be as welcoming as possible if we are requiring some extra steps to make it usable.

molecule/driver/vagrantdriver.py
```diff
@@ -22,7 +22,10 @@
 import copy
 import os
 
-import vagrant
+try:
+    import vagrant
+except ImportError:
+    LOG.error('Driver missing, install python-vagrant!')
```
signed8bit: LOG hasn't been imported at this point so this fails. In hindsight I think what we are really going to want is something like this.

```
try:
    import vagrant
except ImportError:
    sys.exit('ERROR: Driver missing, install python-vagrant!')
```

This way we don't require the LOG import and the program exits with a non-zero status.

molecule/driver/openstackdriver.py
```diff
@@ -26,7 +26,10 @@
 import time
 
 import paramiko
-import shade
+try:
+    import shade
+except ImportError:
+    LOG.error('Driver missing, install shade!')
```
signed8bit: LOG hasn't been imported at this point so this fails. In hindsight I think what we are really going to want is something like this.

```
try:
    import shade
except ImportError:
    sys.exit('ERROR: Driver missing, install shade!')
```

This way we don't require the LOG import and the program exits with a non-zero status.

molecule/driver/dockerdriver.py
```diff
@@ -22,7 +22,10 @@
 import io
 import json
 
-import docker
+try:
+    import docker
+except ImportError:
+    LOG.error('Driver missing, install docker-py!')
```
signed8bit: LOG hasn't been imported at this point so this fails. In hindsight I think what we are really going to want is something like this.

```
try:
    import docker
except ImportError:
    sys.exit('ERROR: Driver missing, install python-docker!')
```

This way we don't require the LOG import and the program exits with a non-zero status.

molecule/driver/vagrantdriver.py
signed8bit: Needs import sys too

molecule/driver/openstackdriver.py
signed8bit: import sys

molecule/driver/dockerdriver.py
signed8bit: import sys

dhutty:
I think this is good, with one caveat: if we take out the (pinned) requirements, how is the user supposed to know which versions of python-vagrant, docker-py, shade are tested/known to work with their version of molecule?

retr0h:
> I think this is good, with one caveat: if we take out the (pinned) requirements, how is the user supposed to know which versions of python-vagrant, docker-py, shade are tested/known to work with their version of molecule?

Yeah, we have the test-requirements file which pins to tested versions.  We are allowing the user to install off main "and hope for the best".  It's not ideal, but it's generally what most projects do.

retr0h:
```
  syntax: commands succeeded
  doc: commands succeeded
  congratulations :)
```

abrown-sg:
👍

PR #511, opened 2016-09-30
Set ssh key if specified.

lewg:
SSH wasn’t working for non-auto-generated keys.


retr0h:
@lewg would you like this backported, and released to pypi, so you do not need to run molecule off master?  Master has [breaking changes](https://github.com/metacloud/molecule/blob/master/CHANGELOG.rst#breaking-changes) around the CLI.

lewg:
I'm fine running from master if a release is in the near future.

retr0h:
@lewg I released 1.11.5 to pypi with this functionality.

PR #520, opened 2016-10-03
added Dockerfile

robtec:
Hi,

As I don't have a python development environment, I made a Dockerfile to build the latest.
Figured might be worth contributing it back to you guys

```
$ docker build -t robtec/molecule .
...
$ docker run -it robtec/molecule molecule --version
molecule, version 1.10.1.dev82
```

Initial thoughts?

Signed-off-by: Rob Powell <rob@robpowell.engineering>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM python:3.3
+
+RUN mkdir /home/molecule
+WORKDIR /home/molecule
+
+COPY requirements.txt /home/molecule/requirements.txt
+RUN pip install -r requirements.txt
+
+COPY test-requirements.txt /home/molecule/test-requirements.txt
+RUN pip install -r test-requirements.txt
+
+COPY . /home/molecule
+RUN pip install .


Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.3
```
retr0h: FYI - Molecule is currently only python 2.7 compatible.

retr0h:
Hi @robtec, thanks for thinking of us and sending us a PR.  I'm probably going to hold off on this, and get a feel for what other developers are doing.  We haven't had any issues around this, and might cause confusion if we cannot support a molecule-docker workflow.

Also, to run a molecule development environment, it requires vagrant/virtualbox and docker as well.

Again, appreciate the PR, I'd just like to hold off for now.

robtec:
Hey, yes this wouldn't play well with virtual box, would require a bind mount to the host docker sock for docker to work also.

Appreciate your feedback, thank you

PR #526, opened 2016-10-04
Remove travis integration

retr0h:
We do not use travis the way it is intended to be used.  Having
broken badges on the page, gives the illusion of a broken project.


retr0h:
Alpha dogging this one.

Issue #527, opened 2016-10-04
Idempotence command does not display the non-idempotent tasks


rgreinho:
The `molecule idempotence` command does not display the non idempotent tasks anymore. This forces the user to run `molecule converge` an extra time and manually look for the guilty tasks.

```sh
$ molecule --version
1.11.4
```

### Repro steps

```sh
mkdir /tmp/molecule-idempotence
cd /tmp/molecule-idempotence
molecule init
```
Use the following playbook:
```yaml
---
- hosts: all
  pre_tasks:
    - name: Idempotence test
      command: echo "Molecule idempotence check"

```
Then run:
```sh
molecule idempotence
```
Output:
```sh
--> Idempotence test in progress (can take a few minutes) ...
--> Starting Ansible Run ...
Idempotence test failed.
```

PR #528, opened 2016-10-04
Display the list of non idempotence tasks

rgreinho:
When running the `molecule idempotence` command, the output does not
show any information regarding the non idempotence tasks found if any.

This patch gives the ability to the command to log the list of guilty
task, and prevents the user to have to perform manual verifications to
fix the problem.

Fixes: #527

### DIFF
--- a/molecule/command/idempotence.py
+++ b/molecule/command/idempotence.py
@@ -56,7 +56,9 @@ class Idempotence(base.Base):
             util.print_success('Idempotence test passed.')
             return None, None
         else:
-            LOG.error('Idempotence test failed.')
+            LOG.error(
+                'Idempotence test failed because of the following tasks:')
+            LOG.error('\n'.join(self._non_idempotent_tasks(output)))
             if exit:
                 util.sysexit()
 
@@ -81,6 +83,27 @@ class Idempotence(base.Base):
             return False
 
         return True
+
+    def _non_idempotent_tasks(self, output):
+        """
+        Parses the output to identify the non idempotent tasks.
+
+        :param (str) output: A string containing the output of the ansible run.
+        :return: A list containing the names of the non idempotent tasks.
+        """
+        # Remove blank lines to make regex matches easier.
+        output = re.sub("\n\s*\n*", "\n", output)
+
+        # Split the output into a list and go through it.
+        output_lines = output.split('\n')
+        res = []
+        for idx, line in enumerate(output_lines):
+            if line.startswith('changed'):
+                guilty_line = output_lines[idx - 1]
+                task_name = re.search(r'\[(.*)\]', guilty_line)
+                res.append('* {}'.format(task_name.groups()[0]))
+
+        return res
 
 
 @click.command()
--- a/test/unit/command/test_idempotence.py
+++ b/test/unit/command/test_idempotence.py
@@ -18,6 +18,8 @@
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
 
+import textwrap
+
 import pytest
 
 from molecule.command import idempotence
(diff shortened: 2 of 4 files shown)


tox.ini
```diff
@@ -11,7 +11,7 @@ deps =
     ansible20: ansible==2.0.2.0
     ansible21: ansible==2.1.1.0
 commands =
-    unit: py.test --cov={toxinidir}/molecule/ test/unit/ {posargs}
+    unit: py.test -x --cov={toxinidir}/molecule/ test/unit/ {posargs}
```
retr0h: Good idea to exit quickly.

test/unit/command/test_idempotence.py
```diff
@@ -50,23 +50,25 @@ def test_execute_does_not_raise_on_converge_error(
...
         patched_converge, patched_logger_error, molecule_instance):
-    patched_converge.return_value = None, 'changed=2'
+    output = 'check-command-01: ok=2    changed=1    unreachable=0    failed=0'
+    patched_converge.return_value = None, output
 
     i = idempotence.Idempotence({}, {}, molecule_instance)
     result = i.execute(exit=False)
 
-    msg = 'Idempotence test failed.'
+    msg = ''
```
retr0h: Not sure I understand why msg is being set to empty string and asserting it was called empty?

molecule/command/idempotence.py
```diff
@@ -82,6 +84,27 @@ def _is_idempotent(self, output):
...
+        output_lines = output.split('\n')
+        res = []
+        for idx, line in enumerate(output_lines):
+            if line.startswith('changed'):
+                guilty_line = output_lines[idx - 1]
+                task_name = re.search(r'\[(.*)\]', guilty_line)
+                res.append('* {}'.format(task_name.groups()[0]))
+
+        return res
+
```
retr0h: I don't see this function being tested anywhere.

test/unit/command/test_idempotence.py
rgreinho: Because the output message in the test does not contain any information about the tasks, and the assert checks only for the last logged line.

retr0h: Oh, I see..... So really the problem is [trying to assert 2 calls to the same mock], and we see the last one (https://github.com/rgreinho/molecule/blob/8e801284089d4545147cac66296e05b04b3c5add/molecule/command/idempotence.py#L59-L61).  

I think you could do something like [this](https://github.com/metacloud/molecule/blob/ab46e9e7fc5808ebce57a9f1d5db4271e12f981d/test/unit/command/test_converge.py#L159-L165).

rgreinho: I found a "better"/different way to do it, still preserving the order or the calls.

signed8bit:
There are some minor flake8 validations in test_idempotence.py.

rgreinho:
Weird, tox did not find them. Are you sure you got the latest patch? If so could you post the errors please?

signed8bit:
Oh, let me make sure it isn't something environmental.

rgreinho:
I do get the congratulations message, so the syntax check should be good.

```sh
syntax: commands succeeded
congratulations :)
```

signed8bit:
Here is what I am getting.

        tox -e syntax
       ...
        syntax runtests: commands[1] | flake8
        ./test/unit/command/test_idempotence.py:88: [E501] line too long (87 > 79 characters)
        ./test/unit/command/test_idempotence.py:89: [E501] line too long (87 > 79 characters)
        ./test/unit/command/test_idempotence.py:91: [E501] line too long (87 > 79 characters)
        ./test/unit/command/test_idempotence.py:93: [E501] line too long (87 > 79 characters)
        ./test/unit/command/test_idempotence.py:94: [E501] line too long (83 > 79 characters)
        ./test/unit/command/test_idempotence.py:102: [F811] redefinition of unused 'test_no_non_idempotent_tasks' from line 85
        ./test/unit/command/test_idempotence.py:105: [E501] line too long (87 > 79 characters)
        ./test/unit/command/test_idempotence.py:106: [E501] line too long (87 > 79 characters)
        ./test/unit/command/test_idempotence.py:108: [E501] line too long (87 > 79 characters)
        ./test/unit/command/test_idempotence.py:110: [E501] line too long (87 > 79 characters)
        ./test/unit/command/test_idempotence.py:111: [E501] line too long (83 > 79 characters)

signed8bit:
There was a recent change to conform to the default 79 chars. But that was a couple weeks ago so should have been in the latest tox.ini.

rgreinho:
I did update the yapf file to reflect that change, but for some reason it does not appear in the patchset...

retr0h:
> I did update the yapf file to reflect that change, but for some reason it does not appear in the patchset...

Update YAPF for that?  We shouldn't need to update any YAPF file, we should simply fix the errors reported.  @signed8bit went through and corrected everything to 79 or less.

signed8bit:
@rgreinho's passes locally and mine doesn't @retr0h. Trying to figure out the difference in the environment. Can you run in yours and see what you get?

rgreinho:
@retr0h incorrect! We want to force yapf to have a line length of 79 chars, otherwise it takes the system config, which in my case is 120 chars.

```
column_limit = 79
```

But I don't understand why my change does not appear in the patchset.

rgreinho:
image

`gitk` does see my change in `.style.yapf`.

rgreinho:
OK everything is there now .

retr0h:
> OK everything is there now .

Syntax looks good now.

retr0h:
```
  syntax: commands succeeded
  doc: commands succeeded
  congratulations :)
```

PR #529, opened 2016-10-04
Added Gemfile to serverspec verifier

retr0h:
When running `init` with the serverspec verifier, added a Gemfile
with necessary serverspec dependencies.


rgreinho:
👍

PR #533, opened 2016-10-05
Removed customconf from state file

retr0h:
The state file was used to determine if ansible.cfg was custom or
provided by the user.  However, the logic was flawed.  We execute
`destroy` which first executes `remove_templates`.  Unfortunately,
`remove_templates` relies on the state file having `customconf` set,
but the setting occurs in another method, which is called after
`destroy`.

We now look for a "Molecule managed" string, to know if safe to delete
or not.

### DIFF
--- a/molecule/core.py
+++ b/molecule/core.py
@@ -164,9 +164,13 @@ class Molecule(object):
         """
         if os.path.exists(self.config.config['molecule']['rakefile_file']):
             os.remove(self.config.config['molecule']['rakefile_file'])
-        if not self.state.customconf:
-            if os.path.exists(self.config.config['ansible']['config_file']):
-                os.remove(self.config.config['ansible']['config_file'])
+
+        config = self.config.config['ansible']['config_file']
+        if os.path.exists(config):
+            with open(config, 'r') as stream:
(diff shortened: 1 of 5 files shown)


retr0h:
```
  syntax: commands succeeded
  doc: commands succeeded
  congratulations :)
```

molecule/core.py
```diff
@@ -164,9 +164,13 @@ def remove_templates(self):
...
             os.remove(self.config.config['molecule']['rakefile_file'])
-        if not self.state.customconf:
-            if os.path.exists(self.config.config['ansible']['config_file']):
-                os.remove(self.config.config['ansible']['config_file'])
+
+        config = self.config.config['ansible']['config_file']
+        if os.path.exists(config):
+            with open(config, 'r') as stream:
+                data = stream.read().splitlines()
+                if '# Molecule managed' in data:
```
rgreinho: Maybe nitpicking a tiny bit, but a regex would probably be more efficient.
Something like:
```python
data = stream.read()
if re.search('# Molecule managed', data, re.IGNORECASE):
    os.remove(config)
```

abrown-sg:
👍

PR #540, opened 2016-10-06
Write templates even when a custom ansible.cfg is specified

abrown-sg:
This guard was removed because it isn't needed, and was preventing all templates from being written when specifying a custom ansible.cfg file. This change will cause molecule to create .molecule/ansible.cfg to be created even a user specifies their own, but it won't be used. The path to ansible.cfg specified in molecule.yml will get priority.


retr0h:
Merging, we tested this locally.  Will submit a follow up pr with functional test.

PR #541, opened 2016-10-06
Added custom ansible functional scenario

retr0h:
Verify we still write out ansible.cfg even when passing a custom
one.  Also, verify we do not delete the custom one on destroy.
This will go away once we have core coverage.


abrown-sg:
👍

PR #542, opened 2016-10-06
Added a private disabled top level key

retr0h:
This is temporary and intended to ease our migration to a modern
molecule.  It will be removed and should not be relied upon by
anyone.


abrown-sg:
👍

PR #550, opened 2016-10-10
Expose driver to login

retr0h:
If an instance was created using a different driver than the one
specified in the config, allow login to function.


retr0h:
> This looks fine, but question - will driver become a required option if the config is correct?

If config is correct it simply sets the driver to it's value internally.

alop:
ok

PR #553, opened 2016-10-13
Make sure ansible-lint respects the molecule ignore_paths

lewg:
Based on the documentation, I think this is the desired behavior. You don't want tests to fail because of ansible galaxy roles you brought in via requirements.yml.


retr0h:
Released fix to pypi as 1.12.6.

Issue #554, opened 2016-10-13
Idempotence edge case


retr0h:
```
+ molecule idempotence
--> Idempotence test in progress (can take a few minutes) ...
--> Starting Ansible Run ...
Idempotence test failed because of the following tasks:
Traceback (most recent call last):
  File "/home/leeroy-jenkins/jenkins/workspace/ci-molecule-ansible-systems-cisco.metapod_common/venv/bin/molecule", line 11, in <module>
    sys.exit(main())
  File "/home/leeroy-jenkins/jenkins/workspace/ci-molecule-ansible-systems-cisco.metapod_common/venv/lib/python2.7/site-packages/molecule/cli.py", line 41, in main
    cli(obj={})
  File "/home/leeroy-jenkins/jenkins/workspace/ci-molecule-ansible-systems-cisco.metapod_common/venv/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/home/leeroy-jenkins/jenkins/workspace/ci-molecule-ansible-systems-cisco.metapod_common/venv/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/home/leeroy-jenkins/jenkins/workspace/ci-molecule-ansible-systems-cisco.metapod_common/venv/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/leeroy-jenkins/jenkins/workspace/ci-molecule-ansible-systems-cisco.metapod_common/venv/lib/python2.7/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/leeroy-jenkins/jenkins/workspace/ci-molecule-ansible-systems-cisco.metapod_common/venv/lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/home/leeroy-jenkins/jenkins/workspace/ci-molecule-ansible-systems-cisco.metapod_common/venv/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/leeroy-jenkins/jenkins/workspace/ci-molecule-ansible-systems-cisco.metapod_common/venv/lib/python2.7/site-packages/molecule/command/idempotence.py", line 119, in idempotence
[...]

PR #556, opened 2016-10-14
Disable diff when executing idempotent check

retr0h:
When performing the idempotence check, the regexp fails when output
contains `--diff` output.  Run the reconverge without `--diff`.

Fixes: #554


kevinbringard:
👍

PR #563, opened 2016-10-19
Pass config object to molecule

retr0h:
Pass the config object to molecule.  This will allow us to treat the
config more like a standlone object, and add properties to it.  This
is also the first step in implementing v2 of the config file syntax.
We will be able to initialize a v2 core object based on the version
of the config object.


retr0h:
```
  syntax: commands succeeded
  doc: commands succeeded
  congratulations :)
```

PR #566, opened 2016-10-22
Implemented basic v2 hooks

retr0h:
* CLI now has a v2 hook.  This allows us to instantiate any v2 module
  we like, when the CLI is called.
* ConfigV1 class is "proxying" to the Config class.  Will break this
  out further once V2 config is created.


retr0h:
```
  congratulations :)
```

PR #569, opened 2016-10-23
Created a molecule2 command

retr0h:
We will likely be dealing with subcommands differently in v2.
Due to clicks `add_command` usage, we cannot side load different
command usage based on a CLI flag (inception).  Instead we created
a `molecule2` command to experiment with.


retr0h:
```
  congratulations :)
```

Issue #571, opened 2016-10-25
Support --skip-tags in addition to --tags


tknerr:
Sometimes it's handy to run `molecule converge --tags "foo"` to converge only the roles / tasks with the given tag.

But now lets assume that a role tagged with "foo" depends on another role "heavystuff". In this case, running  `molecule converge --tags "foo"` also implies running "heavystuff" due to the role dependencies (which is good actually :)).

So if you don't want to run "heavystuff" but only "foo", you would have to pass `--tags "foo" --skip-tags "heavystuff"` to ansible.

Molecule currently only provides passing `--tags`, so additionally passing `--skip-tags` as well would be nice.

tknerr:
Btw I aslo tried `--tags "foo,!heavystuff"` but this seems to be not supported by ansible

retr0h:
@tknerr is this workflow something you want to occur every time you run molecule?  Each time you converge you wish to skip tags for a particular role your developing?

tknerr:
@retr0h nope this was more like a one-off use case where I was making changes in "foo" and didn't want to run all the "heavystuff" too... just to iterate a bit more quickly during development

tknerr:
Wow, that was fast. Thanks @retr0h !

retr0h:
This feature was released with molecule 1.13 which is live on pypi.

tdschmidl:
Dear all,

just for understanding - this patch only applies to `molecule converge` and not `molecule test` - would it be possible to add for latter too?

FrankApiyo:
I'm not sure if I misunderstood but this does not seem to work for me:
```
(ona) user@user-ThinkPad-T580:~/work/playbooks/roles/zebra$ molecule converge --tags "test-logrotate"
Usage: molecule converge [OPTIONS] [ANSIBLE_ARGS]...
Try 'molecule converge --help' for help.

Error: no such option: --tags
(ona) user@user-ThinkPad-T580:~/work/playbooks/roles/zebra$ molecule --version
molecule 3.0.3
   ansible==2.8.0 python==3.6

```

FrankApiyo:
huh! ok... seems what I needed to do was `molecule converge -- --tags "test-logrotate"`

PR #572, opened 2016-10-25
Ability to pass arbitrary ansible cli flags

retr0h:
We supported `--tags` when converging, and had a request to also
support `--skip-tags`.  Rather than supporting ansible options,
simply pass everything after -- to ansible.  Flags are added to
the end of the `ansible-playbook` command and take precedence over
anything before.

  molecule converge -- -v --tags foo,bar --skip-tags baz

Fixes: #571


retr0h:
```
  syntax: commands succeeded
  doc: commands succeeded
```

Issue #578, opened 2016-10-26
Debug should print the commands executed


retr0h:
We should print the different commands executed when running with the `--debug` flag.  Currently we only display the converge ansible-playbook command.

retr0h:
This feature was released with molecule 1.13 which is live on pypi.

PR #589, opened 2016-11-05
Implemented logging of sh commands when debug flag

retr0h:
* Implemented a run_command function to execute, and log executed commands.
* Corrected arg handling, where args and command_args were switched.
* Removed unnecessary ansible-galaxy install function, and simply use
  execute().
* Updated tests.

Fixes: #578


retr0h:
Tests succeeded.

PR #599, opened 2016-11-18
Update dynamic ssh key generation and fix molecule login ...

nmische:
... for openstack driver.

This PR seeks to resolve a few issues around the dynamic provisioning of local ssh keys and openstack keypairs. First, `molecule login` did not work when using a dynamically provisioned ssh key, or any key other than that configured in the user's local ssh config. Second, the dynamically provisioned openstack keypair was not correctly deleted if destroy was not called in the same run as up.


retr0h:
Thanks for the contribution @nmische.  The unit tests are not passing.  Could you look into this?

```
_________________________________________________________________ test_generate_temp_ssh_key _________________________________________________________________

openstack_instance = <molecule.driver.openstackdriver.OpenstackDriver object at 0x10b563550>

    def test_generate_temp_ssh_key(openstack_instance):
>       fileloc = openstack_instance._generated_ssh_key_location()
E       AttributeError: 'OpenstackDriver' object has no attribute '_generated_ssh_key_location'
```

nmische:
I've fixed the unit tests and increased the coverage slightly.

retr0h:
Released to pypi as 1.14.1.

nmische:
Thanks for the quick turnaround!

PR #606, opened 2016-11-22
Added a shell dependency manager

retr0h:
Instead of using ansible-galaxy one can use the command specified.
Useful, since ansible-galaxy blows.

### DIFF
--- a/test/unit/driver/test_dockerdriver.py
+++ b/test/unit/driver/test_dockerdriver.py
@@ -20,7 +20,6 @@
 
 import pytest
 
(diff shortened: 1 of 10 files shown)


kireledan:
Awesome. :+1:

test/unit/driver/test_dockerdriver.py
```diff
@@ -20,7 +20,6 @@
 
 import pytest
 
-from molecule import ansible_playbook
```
dhutty: should this be part of this PR?

PR #610, opened 2016-11-22
Rename dependencies key to dependency

retr0h:
Should follow suit with singular form of molecule config sections.
This functionality was added in 1.14, and this follow-up corrects
the usage, before 1.14 was utilized.

Breaking Change: The galaxy override options have been moved to the
`dependency` section of molecule's config.  No longer support a top
level `dependencies` config key.


kevinbringard:
👍

PR #616, opened 2016-11-23
Removed logger usage from base

retr0h:
Using the logger to print doesn't make sense.  We are configuring it
simply to write to stdout/error.

Loggers are only used when passed to the `sh` module as callbacks.


retr0h:
closing while wip.

PR #617, opened 2016-11-23
Fix task determination on idempotence failure

janderson2:
Previously when the idempotence test failed when run the failure message
was returning the hostname rather than the task name.
This fix returns both the hostname and task name to aid in isolating the
failure.


retr0h:
@janderson2 thanks for the PR, much appreciated!

PR #624, opened 2016-11-29
Added information on Gilt usage

retr0h:


doc/source/configuration.rst
```diff
@@ -228,6 +228,22 @@ to execute the command provided.
     name: shell
     command: script --flag1 subcommand --flag2
 
+`Gilt`_ can be used to manage dependencies, by creating a ``gilt.yml`` in the
+root of the role, and configuring molecule as follows.
```
abrown-sg: Probably should be `as follows:`

PR #633, opened 2016-12-06
Added missing completion targets.

pporada-gl:
Missing targets will now appear as follows

```
$ molecule 
check        converge     create       dependency   destroy      idempotence  init         list         login        status       syntax       test         verify ```

### DIFF
--- a/asset/bash_completion/molecule.bash-completion.sh
+++ b/asset/bash_completion/molecule.bash-completion.sh
@@ -24,7 +24,7 @@ _molecule(){
 	prev=${COMP_WORDS[COMP_CWORD-1]}
 	firstword=$(_get_firstword)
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: asset/bash_completion/molecule.bash-completion.sh [hunk 1]
```diff
@@ -68,6 +73,9 @@ _molecule(){
           ;;
       esac
       ;;
+    dependency)
+      complete_options="${CHECK_OPTIONS}"
```
rgreinho: Should be `DEPENDENCY_OPTIONS`.

rgreinho:
A small copy/paste mistake, but good job otherwise 👍

pporada-gl:
Fixed, good spot.

PR #643, opened 2016-12-11
Make ansible.cfg settings configurable

alvaroaleman:
This PR allows to set arbitrary settings for the `defaults` and `ssh_opts` section of `ansible.cfg`.

IMHO this way is much more intuitive than environment vars, especially since Ansible environment variables are mostly documented in source code only.

Also, retry_files get disabled by default - If you think that is out of scope for this PR, I'll remove that commit.

Any feedback on the general approach is welcome.


.gitignore
```diff
@@ -11,3 +11,4 @@
 .vagrant/
 doc/build/
 htmlcov/
+*.sw[po]
```
retr0h: If the developer uses vim, then they should be ignoring these in their own global gitignore file.  The project shouldn't be aware of this.

Your `~/.gitconfig` can have a section pointing to a global .gitignore. 

```
[core]
	excludesfile = /Users/jodewey/.gitignore
	
```

retr0h:
@alvaroaleman Thanks for the patch.  Will need to review it. :guitar:

retr0h:
FYI - Implemented similar logic [into molecule v2](https://github.com/metacloud/molecule/commit/7143b30dc3de423e40b91ef672ee631b4de72b1f).

alvaroaleman:
Your implementation appears way cleaner, but the name `config_options` doesn't make quite clear its about ansible.cfg options IMHO.

retr0h:
Thanks @alvaroaleman.  Will get released with v1.18.

PR #644, opened 2016-12-11
Openstack: Fix security_groups default for newer shade versions

alvaroaleman:
The openstack driver currently pases a `None` object to shade in case the `security_groups` paramter is undefined. In the current devel version of shade however, the `security_groups` parameter is put into a list in case it is not a list itself: https://github.com/openstack-infra/shade/blob/master/shade/openstackcloud.py#L4995

This results in a `[None]` list for the `security_groups` paramter which then makes the `create_server` call an invalid request.

### DIFF
--- a/molecule/driver/openstackdriver.py
+++ b/molecule/driver/openstackdriver.py
@@ -120,8 +120,7 @@ class OpenstackDriver(basedriver.BaseDriver):
                     auto_ip=True,
                     wait=True,
                     key_name=kpn,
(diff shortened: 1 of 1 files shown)


retr0h:
Unfortunately we do not maintain the OpenStack driver, and rely upon community contributions.  I'm going to merge this so we can release it.  The patch is fairly straightforward.  Hopefully I can get a set of functional tests around this driver soon.

PR #645, opened 2016-12-11
Make Openstack ip pool configurable

### LINKED ISSUE #639: No openstack floating ip support
# Issue Type

- Bug report

```
molecule, version 1.16.1
```

- Molecule installation method: pip

# Desired Behaviour

The openstack provider creates an openstack instance, associates a floating ip to it and connects to the instance using the floating ip.

# Actual Behaviour (Bug report only)

The openstack provider creates an openstack instance and tries to connects to the private ip which obviously doesn't work.

alvaroaleman:
Fixes #639


retr0h:
Unfortunately we do not maintain the OpenStack driver, and rely upon community contributions.  I'm going to merge this so we can release it.  The patch is fairly straightforward.  Hopefully I can get a set of functional tests around this driver soon.

Issue #650, opened 2016-12-14
Add networks parameter to docker driver


abtreece:
# Feature request
Add the networks parameter to the docker driver.

The default bridge network does not support automatic service discovery. I created another bridge network to enable this feature, but am unable to launch my molecule controlled container on this network.

# Molecule and Ansible details

```
ansible 2.2.0.0
  config file =
  configured module search path = Default w/o overrides
neutron:kong abtreece$ molecule --version
molecule, version 1.16.1
```

- Molecule installation method: pip
- Ansible installation method: pip

# Desired Behaviour

Would like to be able to specify a container network other than the default bridge. Adding the links parameter would be an additional bonus although it is a deprecated feature.

# Actual Behaviour (Bug report only)

N/A

kireledan:
I'll be able to get to this next week when I'm on my winter break if no one else has done it by then.

retr0h:
@kireledan would be great if you could tackle it.

kireledan:
The docker driver rework I have will cover this. I'll make a PR for it in a couple days.

retr0h:
@kireledan Any chance you were able to tackle this?

kireledan:
Yeah - currently just trying to get the tests to pass because it throws an error code 1 even though it doesn't error out.

retr0h:
FYI - I just fixed functional testing on master earlier today.  For some reason I ran into a bug I never ran into before.

kireledan:
Roger. Ill merge that into my branch. Currently my error is just that in using ansible to run the driver, the warning that ansible pops up with about localhost causes the test to fail immediately.

lhoss:
we also desperately require this feature 👍 
as it (using a custom network/bridge) seems to be the cleanest way to deploy a cluster of docker containers with *working* hostname resolution (for ex. required to test a multi-host hadoop/hdfs deployment in our case) without any custom /etc/hosts hacks.
ps: I couldn't find any branch/commit.

As I'm totally new to molecule, I then wonder where's the best place in a molecule build to create that network ( `docker network create --driver bridge isolated_nw` ) ?https://docs.docker.com/engine/userguide/networking/

kireledan:
I currently have it pushed to my playbook_proto branch. It is currently functional and does support networks, but it does not pass the unit tests because of some strange issues I'm having at the moment. https://github.com/kireledan/molecule/tree/playbook_proto

abtreece:
[...]

PR #651, opened 2016-12-15
Feature/add links param to docker

abtreece:
Adds the links parameter which ended up being an easy add. The links functionality is deprecated in Docker though, so the networks parameter is still desireable. This was the "bonus" request in Issue #650


retr0h:
Hi @abtreece thanks for the PR.  Can you add some tests for this?  See the driver/docker unit tests for reference.

abtreece:
ah, yes... i'll try.

abtreece:
added unit test for links.

retr0h:
Awesome @abtreece.  Thanks a ton!  Will get released to pypi as 1.18 in the next week or so.

PR #649, opened 2016-12-14
Fix environment passed to ansible-lint (closes #642)

techraf:
Pass `PYTHONPATH` environment variable to `ansible-lint`, which when stripped, renders Molecule incompatible with any Ansible version running from source (i.e. failing on each `Executing ansible-lint...` task).

PR #653, opened 2016-12-15
Revert "Fix environment passed to ansible-lint (#649)"

retr0h:
This reverts commit c7ffdf877daea2785cd409706201d25b4770d591.

This commit broke functional tests.  Reverting until submitter corrects it to work with tox.

### DIFF
--- a/molecule/verifier/ansible_lint.py
+++ b/molecule/verifier/ansible_lint.py
@@ -49,7 +49,6 @@ class AnsibleLint(base.Base):
         env = {
             'ANSIBLE_CONFIG':
             self._molecule.config.config['ansible']['config_file'],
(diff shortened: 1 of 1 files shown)


techraf:
I'm not sure how passing a standard variable [`PYTHONPATH`](https://docs.python.org/2/using/cmdline.html#envvar-PYTHONPATH) to a program which requires this very variable to run (`ansible-lint`) to access Ansible modules breaks something in an unrelated part.

It rather looks like someone has that variable set to an improper value and should correct their real problem.

It sounds surreal to me that code must take into account incorrect environment settings, not the other way.

retr0h:
Running the tests broke.  I simply reverted it so I could put out a new release.

Issue #642, opened 2016-12-11
ansible-lint called from Molecule fails when no Ansible-installed-with-pip is present


techraf:
# Issue Type

- Bug report

# Molecule and Ansible details

```
# ansible --version
ansible 2.2.1.0 (stable-2.2 acad2ba246) last updated 2016/12/11 20:27:02 (GMT +900)
  lib/ansible/modules/core: (detached HEAD 8139278530) last updated 2016/12/11 20:30:10 (GMT +900)
  lib/ansible/modules/extras: (detached HEAD f5f1fc934a) last updated 2016/12/11 20:30:10 (GMT +900)
  config file =
  configured module search path = Default w/o overrides
# molecule --version
molecule, version 1.16.1
```

- Molecule installation method: pip
- Ansible installation method: source

# Desired Behaviour

```
# molecule verify
--> Executing ansible-lint...
[ANSIBLE0002] Trailing whitespace
playbook.yml:7
    - ansible-unix-python-environment
```

# Actual Behaviour (Bug report only)

```
# pip uninstall ansible
((( cut )))
  Successfully uninstalled ansible-2.2.0.0

# . /usr/local/src/ansible/hacking/env-setup
((( cut )))
PYTHONPATH=/usr/local/src/ansible/lib:
((( cut )))

# ansible --version
ansible 2.2.1.0 (stable-2.2 acad2ba246) last updated 2016/12/11 20:27:02 (GMT +900)
  lib/ansible/modules/core: (detached HEAD 8139278530) last updated 2016/12/11 20:30:10 (GMT +900)
  lib/ansible/modules/extras: (detached HEAD f5f1fc934a) last updated 2016/12/11 20:30:10 (GMT +900)
  config file =
  configured module search path = Default w/o overrides

# molecule --debug verify
--> Executing ansible-lint...
DEBUG: COMMAND
/usr/local/bin/ansible-lint playbook.yml --exclude .git --exclude .vagrant --exclude .molecule
Traceback (most recent call last):
  File "/usr/local/bin/ansible-lint", line 30, in <module>
    import ansiblelint
  File "/usr/local/lib/python2.7/site-packages/ansiblelint/__init__.py", line 26, in <module>
    import ansiblelint.utils
  File "/usr/local/lib/python2.7/site-packages/ansiblelint/utils.py", line 25, in <module>
    import ansible.constants as C
ImportError: No module named ansible.constants

[...]

PR #657, opened 2016-12-16
Pass env to `ansible-lint`

retr0h:
Reimplements the revert from #649.
Reimplements the feature from #642.


retr0h:
@techraf Can you verify this patch works for you?  This is passing unit and functional tests, and should pass the entire env to `ansible-lint`.

PR #659, opened 2016-12-17
Ensure raw_env_vars contains string values only

### LINKED ISSUE #658: If raw_env_vars contains a var whose value is not a string, molecule silently fails to execute Ansible
# Issue Type

- Bug report

```
ansible --version
2.2.0.0

molecule --version

1.14.1.dev37
```

Sha256 of commit is 8eed3f539f0b30b04d98f8ab418a2a5d55cce01f, molecule --version for source checkout seems to be stuck on 1.14

- Molecule installation method: source

# Desired Behaviour

There is no possible key value pair in the ``ansible: {raw_env_vars: {}}`` section that prevents Ansible from getting executed

# Actual Behaviour (Bug report only)

Except from molecule.yml:

```
  raw_env_vars:
    RETRY_FILES_ENABLED: 0
```

```
molecule --debug syntax; echo $?
--> Checking playbook's syntax...
DEBUG: COMMAND
/usr/bin/ansible-playbook tests/test.yml -vvvv --inventory-file=localhost, --syntax-check --limit=all --timeout=30 --diff
255
```

```
$ /usr/bin/ansible-playbook tests/test.yml -vvvv --inventory-file=localhost, --syntax-check --limit=all --timeout=30 --diff; echo $?
Using $ROLE_DIR/ansible.cfg as config file
Set default localhost to localhost
3 plays in tests/test.yml

playbook: tests/test.yml
0
```
The same happens with ``molecule converge``.

This is especially bad since that one is really hard to debug, because there is no error message at all.

I am not even sure if this is the fault of molecule, because [the doc to sh.pys `_env` arg only mentions it must be a dict, not that all values must be string types](https://amoffat.github.io/sh/sections/special_arguments.html?highlight=_env#env).

alvaroaleman:
Required to not silently break Ansible runs.
Fixes #658

### DIFF
--- a/molecule/ansible_playbook.py
+++ b/molecule/ansible_playbook.py
@@ -95,6 +95,8 @@ class AnsiblePlaybook(object):
 
         if name == 'raw_env_vars':
             for k, v in value.iteritems():
+                if not isinstance(v, basestring):
+                    v = unicode(v)
                 self.add_env_arg(k, v)
             return
 
--- a/test/functional/test_docker_scenarios.py
+++ b/test/functional/test_docker_scenarios.py
@@ -145,6 +145,14 @@ def test_custom_ansible_cfg(scenario_setup):
 
 
 @pytest.mark.parametrize(
+    'scenario_setup', ['non_string_env'], indirect=['scenario_setup'])
+def test_non_string_env(scenario_setup):
+    sh.molecule('create')
+    sh.molecule('syntax')
+    sh.molecule('destroy')
+
+
+@pytest.mark.parametrize(
     'scenario_setup', ['dockerfile'], indirect=['scenario_setup'])
 def test_dockerfile(scenario_setup):
     sh.molecule('test')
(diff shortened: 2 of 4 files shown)


retr0h:
Can we add an ansible platbook unit test for this?

alvaroaleman:
Added that as a docker scenario.

As a sidenote, the unittests don't work for me because my `ls` is at `/usr/bin/ls` and not at `/bin/ls` as expected: https://github.com/metacloud/molecule/blob/master/test/unit/test_util.py#L213

alvaroaleman:
Please do not merge this right now, because I am currently doing another PR that as a side effect also fixes this issue, making a3b5c78 obsolete.

It will probably do not harm to still keep the testcase in cf36101 thought.

alvaroaleman:
Scratch that, I originally thought #667 would be fixing this issue.

retr0h:
@alvaroaleman Thanks for the contribution.  Will be available in 1.17.2.

PR #662, opened 2016-12-17
Fix unittests to allow ls to be in both /usr/bin and /bin

alvaroaleman:
As mentioned in #659, the unittests didn't work for me because my $PATH does not contain `/bin` but `/usr/bin` instead. This is default for RHEL/CentOS 7 and Fedora.

### DIFF
--- a/test/unit/command/test_verify.py
+++ b/test/unit/command/test_verify.py
@@ -20,6 +20,7 @@
 
 import pytest
 import sh
+import os
 
 from molecule.command import verify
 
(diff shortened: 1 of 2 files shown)


retr0h:
@alvaroaleman I simply used ls as a way to throw a side effect.  It may make sense to replace `ls` with something else or perform a regexp so we are not having to worry about pathing.

I think I would prefer a regexp.

P.S. Thanks for all of your issues and PRs.  Appreciate the help.

alvaroaleman:
Regexp definitely sounds like a good idea. However I did not find a way make `assert_called_with()` test against one. Did I miss something there or did you have another approach in mind?

retr0h:
@alvaroaleman I went ahead and fixed the test.  Thanks for your contribution and my apologies for taking so long with this.  This will be released under 1.17.2.

PR #672, opened 2016-12-31
Use 'become' options instead of 'sudo' options

elliotweiser:
As of Ansible 1.9, `sudo` options are deprecated for achieving
privelege escalation in favor of the `become` options.
This change will cause the generated `ansible-playbook` command to
use the following options in favor of their deprecated counterparts.

* `become` instead of `sudo`
* `become_user` instead of `sudo_user`
* `ask_become_pass` instead of `ask_sudo_pass`

### DIFF
--- a/doc/source/configuration.rst
+++ b/doc/source/configuration.rst
@@ -46,7 +46,7 @@ may look something like this:
   ansible:
     inventory_file: ../../inventory/
     diff: False
-    sudo: True
+    become: True
     vault_password_file: ~/.vault
 
 As you can see, the names of these values correspond to what the underlying
--- a/doc/source/provider/virtualbox.rst
+++ b/doc/source/provider/virtualbox.rst
@@ -26,8 +26,8 @@ of options that you probably don't want to set.
   ---
   ansible:
     playbook: playbook.yml
-    sudo: True
-    sudo_user: False
+    become: True
+    become_user: False
     verbose: vvvv
 
   vagrant:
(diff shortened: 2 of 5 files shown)


elliotweiser:
@retr0h Do you happen to recall the reasoning behind making the default for `sudo: True`? I was doing some debugging and I realized that `ansible_user_id` was consistently picked up as `root` after gathering facts. It made me wonder if privilege escalation shouldn't be turned off by default.

retr0h:
@elliotweiser thanks for the PR.  Looks good 👍.

> @retr0h Do you happen to recall the reasoning behind making the default for sudo: True? I was doing some debugging and I realized that ansible_user_id was consistently picked up as root after gathering facts. It made me wonder if privilege escalation shouldn't be turned off by default.

I believe we originally enabled ansible to run with sudo (become), since it was making changes to the system.  We really didn't think anyone would be invoking ansible, through molecule, to not make a change, so having sudo (become) False, didn't make a lot of sense.

I do understand there are work-flows, where people write ansible the opposite.  I'm open to a PR so we can discuss this, but again, people are generally running molecule, to test the role.  We just picked a direction and went with it, with not major discussion around it.

elliotweiser:
Thanks for the explanation. Fortunately, it turned out to be pretty easy to turn off the behavior by just setting `sudo: False`.

It might be worth adding some "usage" documentation on the value of running `molecule <command> --debug` to see exactly what commands are being executed. That can be a separate PR.

retr0h:
`molecule --debug <subcommand>` will print out the exact commands being executed.

elliotweiser:
Oops, had the order wrong 😜

PR #675, opened 2017-01-04
[v2] syntax and test tweaks

detiber:
I hit some test issues on the v2 branch, one related to the absolute path of a particular command, and the others were caused by the ordering of arguments by `sh` not being the same between runs.  Other diffs are related to yapf formatting.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,4 +1,5 @@
 docker-py
+gilt-python
 mock
 pytest
 pytest-cov
--- a/molecule/command/init.py
+++ b/molecule/command/init.py
@@ -27,9 +27,8 @@ import cookiecutter.main
 from molecule import config
 from molecule import util
 
-def _process_templates(template_dir,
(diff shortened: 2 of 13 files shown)


retr0h:
> diffs are related to yapf formatting

Are you using the YAPF version supplied in tox.ini?  We should both be on the same version.  Did you run `tox syntax`?

detiber:
@retr0h yes, the diff is the result of running tox syntax

test-requirements.txt
```diff
@@ -5,3 +5,4 @@ pytest-cov
 pytest-helpers-namespace
 pytest-mock
 wheel
+gilt-python
```
retr0h: Great cleanups.  Could you alphabetize the new dep?  Other than that looks good!

detiber: Not a problem, re-ordered and pushed.

retr0h:
Thanks @detiber

Issue #688, opened 2017-01-05
Add openstack multiple network support


cseke:
# Issue Type

- Feature request

# Molecule and Ansible details

```
molecule, version 1.14.1.dev53
ansible 2.2.0.0
```

- Molecule installation method: pip
- Ansible installation method: pip

# Desired Behaviour

The openstack provider creates an openstack instance and attach multiple network interfaces. Networks could be defined in molecule.yml explicitly.

# Actual Behaviour (Bug report only)

Frankly speaking I could not realize how it should work currently. I have always error message:
```
shade.exc.OpenStackCloudException: Error in creating instance (Inner Exception: Multiple possible networks found, use a Network ID to be more specific. (HTTP 409) (Request-ID: req-547b661b-cd0f-43aa-a087-053c34aa0188))
```

retr0h:
Addressed by #689.

PR #689, opened 2017-01-05
Add openstack multiple network support

cseke:
#688


retr0h:
Thanks @cseke 🎸 Will get released to pypi as 1.18 in the next week or so.

PR #715, opened 2017-01-14
Fix availability timeout in Openstack driver

alvaroaleman:


molecule/driver/openstackdriver.py
```diff
@@ -134,15 +134,23 @@ def up(self, no_provision=True):
...
                     util.print_info('\t Waiting for ssh availability...')
                     num_retries += 1
+                    if self._check_ssh_availability(
+                            server['interface_ip'],
+                            instance['sshuser'],
+                            timeout=6,
+                            sshkey_filename=self._get_keyfile()):
+                        instance['reachable'] = True
+                        break
+                if not instance['reachable']:
```
kireledan: It seems a bit redundant to me to have the reachable flag. The num retries reaching a max threshold would already be a clear indicator of not being reachable, no?

I do like the add in of the print_error and sys.exit because that behavior needed to be added.

alvaroaleman: Just to give some context, the original while loop never matches, because `not ssh_availability` evaluates to `True` when the instance is not reachable, thus the `num_retries == 5` never gets checked.

Since it is possible the instance gets reached on the last iteration, a check on `num_retries` is not enough here.

alvaroaleman: Looking at it, `num_retries` is completely obsolete now, I'll replace it with a `range(5)` if you have no further objections.

kireledan: Sounds good to me. 👍

alvaroaleman: Done @retr0h

PR #716, opened 2017-01-14
Do not alter users known_hosts file in Openstack driver

alvaroaleman:
Currently, the Openstack driver uses ssh-keygen to remove any entrys
for the address of the given instance from the users known_hosts file.
Since an altered known_hosts file is not something the user expects,
this patch changes that behaviour to directly try to connect with
hostkey checking disabled.

### DIFF
--- a/molecule/driver/openstackdriver.py
+++ b/molecule/driver/openstackdriver.py
@@ -21,9 +21,10 @@
 import collections
 import hashlib
 import os
-import socket
 import sys
 import time
+
+from subprocess import check_output, CalledProcessError, STDOUT
 
 import paramiko
 try:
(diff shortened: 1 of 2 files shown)


kireledan:
👍 I definitely approve of this because at the time I had forgotten about the hostkey checking flag.

alvaroaleman:
Hey @retr0h,

IMHO there are a couple of good reasons for this:

* The whole point of a `known_hosts` file is to make ssh error if a host key changed. This also means, that if for whatever reason a hostkey ends up there, no tool should ever remove it implicitly without asking for permission

* The host key of a test instance is obviously ephemeral, thus should not be added to the `known_hosts` file

* In contrast to what you said, the current implementation does leave junk behind, because the host key is only deleted on instance creation, not on tear down

* `ssh-keygen` is not available everywhere, e.G. the `ubuntu:16:04` docker image does not have it. Also, the case of `ssh-keygen` not being available isn't handled at all, the user just gets a `ssh-keygen: no such file or directory` followed by a stracktrace.

retr0h:
good points @alvaroaleman 👍

PR #726, opened 2017-01-24
Openstack: Performance improvements for multiinstance setups

### LINKED ISSUE #713: Openstack instances get created sequentially
# Issue Type

- Bug report

Right now molecule creates  Openstack instances sequentially, waits until the instance was fully created and can be reached via SSH, then proceeds to the next instance. This is very time consuming.

Instead, the Openstack driver should just create all instances without waiting and afterwards wait until it can reach them all.

alvaroaleman:
This PR improves the instance creation performance for multiinstance setups, by dividing it into three distinct steps:

1. Create all instances, don't wait for them to be in state active
2. Wait for all instances to get to state active
3. Wait for all instances to be available via SSH

Fixes #713 

@kireledan: Could you verify this works as intended?

### DIFF
--- a/molecule/driver/openstackdriver.py
+++ b/molecule/driver/openstackdriver.py
@@ -117,7 +117,7 @@ class OpenstackDriver(basedriver.BaseDriver):
 
         active_instances = self._openstack.list_servers()
         active_instance_names = {
-            instance['name']: instance['status']
+            instance['name']: instance['interface_ip']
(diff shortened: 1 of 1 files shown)


kireledan:
I can try and test it, but I am not in the same environment I was in when I developed this. I might not be able to get a testing environment up because the openstack instance was already all configured when working on this over the summer.

PR #730, opened 2017-01-27
Do not require a project_config when a local_config is present

alvaroaleman:


alvaroaleman:
It doesn't clearly state that a `molecule.yml` is required either. IMHO putting that info into the ``Config File`` section may be more irritating than helping, since that passage is supposed to be an introduction for people that are new to Molecule.

When I first tried to use a `~/.config/molecule/config.yml` only, I was surprised that threw an error. So IMHO it is more intuitive to not require a project `molecule.yml` if all required settings are already present in the global one.

retr0h:
> It doesn't clearly state that a molecule.yml is required either. IMHO putting that info into the Config File section may be more irritating than helping, since that passage is supposed to be an introduction for people that are new to Molecule.

You're adding new functionality.  I'd like it documented that you can leave it out and it comes from local.

PR #737, opened 2017-01-31
[v2] allow tests to run on python35

detiber:
- add python35 targets to travis and tox
- update tox to run linter for all python versions
- update tests that use asssert_called_once to test against call_count
  instead (assert_called_once is only supported for python < 3 and >= 3.6)
- update tox config to not fail if a target interpreter is missing


retr0h:
@detiber Looks like I screwed you up with my current force push on v2.  Mind rebasing?  My apologies.

PR #738, opened 2017-02-01
[v2]Fix lint environments in tox

detiber:
With the change to prefixing the lint env, the lint commands needed to be moved to the [testenv] stanza.

### DIFF
--- a/molecule/provisioner/ansible.py
+++ b/molecule/provisioner/ansible.py
@@ -305,10 +305,13 @@ class Ansible(object):
             vars_target = copy.deepcopy(self.host_vars)
             # Append the scenario-name
             for instance_name, _ in self.host_vars.items():
-                instance_with_scenario_name = util.instance_with_scenario_name(
-                    instance_name, self._config.scenario.name)
-                vars_target[instance_with_scenario_name] = vars_target.pop(
-                    instance_name)
+                if instance_name == 'localhost':
+                    instance_key = instance_name
+                else:
+                    instance_key = util.instance_with_scenario_name(
+                        instance_name, self._config.scenario.name)
+
+                vars_target[instance_key] = vars_target.pop(instance_name)
 
         elif target == 'group_vars':
             vars_target = self.group_vars
--- a/test/unit/provisioner/test_ansible.py
+++ b/test/unit/provisioner/test_ansible.py
@@ -47,6 +47,9 @@ def molecule_provisioner_section_data():
                 'instance-1': [{
                     'foo': 'bar'
                 }],
+                'localhost': [{
+                    'foo': 'baz'
+                }]
             },
             'group_vars': {
                 'example_group1': [{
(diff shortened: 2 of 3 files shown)


detiber:
This PR kind of scope creeped as I ran into issues with the docs and trying to set host_vars for setup/teardown playbooks.  I need to go back and fix the coverage regression for adding the localhost exception still.

molecule/provisioner/ansible.py
```diff
@@ -33,42 +33,6 @@ class Ansible(object):
...
-              - destroy
-              - dependency
-              - syntax
-              - create
-              - converge
-              - idempotence
-              - lint
-              - verify
-              - destroy
-
```
retr0h: Why was the documentation removed?

detiber: It was removed because they are not properties of the ansible config, but rather the scenario config (where they were already documented).

detiber:
@retr0h once I fix the coverage regression, I will break these changes out.

molecule/provisioner/ansible.py
retr0h: Ah, thanks, bad cut and paste on my part :)

test/unit/provisioner/test_ansible.py
```diff
@@ -28,37 +28,6 @@
 
 
 @pytest.fixture
-def ansible_data():
```
detiber: This fixture was unused.

test/unit/provisioner/test_ansible.py
```diff
@@ -78,6 +47,9 @@ def molecule_provisioner_section_data():
                 'instance-1': [{
                     'foo': 'bar'
                 }],
+                'localhost': [{
+                    'foo': 'baz'
+                }]
```
detiber: This and line 156 below adds coverage for the special case of localhost host_vars

tox.ini
```diff
@@ -17,21 +17,18 @@ deps =
     -rtest-requirements.txt
     ansible22: ansible==2.2.1.0
 commands =
-    unit: py.test -vv -x --cov={toxinidir}/molecule/ {posargs}
+    unit: py.test -vv --cov-report=term-missing --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
```
detiber: - Remove -x so that other unit tests can run even if one fails.
  - I wasn't sure if there was a specific reason for using -x here
- use term-missing for the coverage report, to get a list of uncovered lines
- add --no-cov-on-fail to prevent an additional failure for coverage when there is a test failure.

tox.ini
```diff
@@ -17,21 +17,18 @@ deps =
...
     functional: py.test -vv -x test/functional/ {posargs}
+    lint: flake8
```
detiber: moves the lint test to the main [testenv] stanza so that it can be invoked by tox-travis

detiber:
@retr0h there is definitely scope creep here, I'm happy to break out one or more changes into a separate pr if needed.

retr0h:
@dedihub Can you rebase?

detiber:
@retr0h done.

It looks like the current v2 branch is failing unit tests for coverage.

retr0h:
> It looks like the current v2 branch is failing unit tests for coverage.

Corrected

Issue #749, opened 2017-02-20
`molecule login` - Too many authentication failures error


gtirloni:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible 2.2.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides

molecule, version 1.20.0

vagrant version 1.9.1
```

- Molecule installation method: pip
- Ansible installation method: OS package

# Actual Behaviour

By default, SSH will try to offer all keys found under ~/.ssh. If it reaches the limit of authentication failures (default=6), the connection will be closed. This is a problem for people with many SSH keys. To work around this issue and only use the specified IdentifyFile (-i), the option IdentitiesOnly should be specified.

```
     IdentitiesOnly
             Specifies that ssh(1) should only use the authentication identity and certificate files
             explicitly configured in the ssh_config files or passed on the ssh(1) command-line, even if
             ssh-agent(1) or a PKCS11Provider offers more identities.  The argument to this keyword must be
             yes or no (the default).  This option is intended for situations where ssh-agent offers many
             different identities.
```

It seems Molecule retrieves the SSH client configuration from Vagrant (`vagrant ssh-config`) but fails to activate this option. This causes the issue to happen with molecule but it cannot be reproduced using Vagrant directly (with molecule's vagrantfile).

This is the error from molecule:
```
$ molecule login
Warning: Permanently added '[127.0.0.1]:2222' (ECDSA) to the list of known hosts.
Received disconnect from 127.0.0.1 port 2222:2: Too many authentication failures for vagrant
Authentication failed.

$ cp .molecule/vagrantfile Vagrantfile
$ vagrant ssh
Last login: Mon Feb 20 01:25:07 2017 from 10.0.2.2
[vagrant@rolename ~]$ 
```

I was tempted to edit [driver/vagrantdriver#L190](https://github.com/metacloud/molecule/blob/master/molecule/driver/vagrantdriver.py#L190) and add "-o IdentitiesOnly=yes" there but I'm not sure if that's appropriate (it works though).

I understand I can add this option to .molecule/config but I stil think it's a bug that molecule and Vagrant differ when using the vagrant driver.

# Additional Information

```
$ vagrant ssh-config
Host rolename
  HostName 127.0.0.1
  User vagrant
  Port 2222
  UserKnownHostsFile /dev/null
  StrictHostKeyChecking no
  PasswordAuthentication no
[...]

PR #750, opened 2017-02-21
Correct too many authentication failures error

retr0h:
By default, SSH will try to offer all keys found under ~/.ssh. If it
reaches the limit of authentication failures (default=6), the connection
will be closed. This is a problem for people with many SSH keys. To work
around this issue and only use the specified IdentifyFile (-i), the
option IdentitiesOnly should be specified.

Fixes: #749

### DIFF
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -174,8 +174,11 @@ class ConfigV1(Config):
                 'limit': 'all',
                 'playbook': 'playbook.yml',
                 'raw_ssh_args': [
-                    '-o UserKnownHostsFile=/dev/null', '-o IdentitiesOnly=yes',
-                    '-o ControlMaster=auto', '-o ControlPersist=60s'
+                    '-o UserKnownHostsFile=/dev/null',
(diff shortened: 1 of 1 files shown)


molecule/config.py
```diff
@@ -174,8 +174,11 @@ def _get_defaults(self):
                 'limit': 'all',
                 'playbook': 'playbook.yml',
                 'raw_ssh_args': [
-                    '-o UserKnownHostsFile=/dev/null', '-o IdentitiesOnly=yes',
-                    '-o ControlMaster=auto', '-o ControlPersist=60s'
+                    '-o UserKnownHostsFile=/dev/null',
+                    '-o IdentitiesOnly=yes',
+                    '-o ControlMaster=auto',
+                    '-o ControlPersist=60s',
+                    '-o IdentitiesOnly=yes',
```
alop: Aren't you duplicating this option?

retr0h: It's used in two diff sections.

PR #756, opened 2017-03-04
docker driver 'build_image' added instance specific overrides applciable

j0lly:
Found myself in need of using other containers as dependencies for a given scenarios and I would not create a molecule_local tagged image for a container I just need as dependency, so i gave build_image config option to be overridden by container specific configuration (with priority over driver generic config)


j0lly:
second push is a bugfix:
did not tested from clean situation (missing molecule_local images).
I think this also deserves a test case and some setup during tests (clean docker host images or such)

j0lly:
Hey @retr0h did u got the chance to look at this PR?

retr0h:
@j0lly needs a rebase.

retr0h:
@kireledan how do you feel about this change?

Issue #743, opened 2017-02-06
docker-py package breaks using docker-compose


timja-kainos:
# Issue Type

- Bug report

# Molecule and Ansible details
- Molecule installation method: One of source, pip
pip
- Ansible installation method: One of source, pip, OS package
os package
# Desired Behaviour
I can use docker-compose and molecule on the same machine

# Actual Behaviour (Bug report only)
Installing `docker-py` conflicts with the new package `docker`

```
Traceback (most recent call last):
  File "/bin/docker-compose", line 7, in <module>
    from compose.cli.main import main
  File "/usr/lib/python2.7/site-packages/compose/cli/main.py", line 20, in <module>
    from ..bundle import get_image_digests
  File "/usr/lib/python2.7/site-packages/compose/bundle.py", line 13, in <module>
    from .network import get_network_defs_for_service
  File "/usr/lib/python2.7/site-packages/compose/network.py", line 7, in <module>
    from docker.types import IPAMConfig
ImportError: cannot import name IPAMConfig
```

See:
https://github.com/docker/compose/issues/4401
https://github.com/docker/docker-py/issues/1431

`docker-py` appears to be obsoleted and `docker` is the new pip package

abtreece:
Docker changed the name of the Docker Python SDK to just `docker` when they bumped to version 2. The last docker-py version is 1.10.6.  Molecule is only capable of using docker-py due to changes that have been to the API in v2. AFAIK we won't be supporting the v2 docker b/c molecule is going to transition to using the Ansible implementation of Docker.

As you've discovered you can't have both `docker-py` and `docker` installed which is by design by Docker.

abtreece:
Thinking about this more... in order to use docker-compose and molecule together you will likely need to install docker-compose via a method other than `pip`. On a Mac or Windows machine docker-compose comes with the installation of Docker Engine. You can also pull down the docker-compose binary by itself via a curl command documented here... https://docs.docker.com/compose/install/

retr0h:
> AFAIK we won't be supporting the v2 docker b/c molecule is going to transition to using the Ansible implementation of Docker.

Also, these `docker_` modules require `docker-py`, so I probably wont be spending effort porting Molecule to the `docker` pip package.

retr0h:
[...]

PR #764, opened 2017-03-11
Support for new docker sdk (quick fix)

kireledan:
In regards to #743 

This is a quick fix to alleviate the issue with the SDK changes.  I can spend more time to utilize the abstraction layer built on top of the 'low-level API' if desired.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -104,7 +104,7 @@ Install Molecule using pip:
 .. code-block:: bash
 
   $ pip install ansible
-  $ pip install docker-py
+  $ pip install docker
   $ pip install molecule
 
 Create a new role with the docker driver:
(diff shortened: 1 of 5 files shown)


retr0h:
Thanks @kireledan.

Can we update `test-requirements.txt` with the new pinning of `docker-2.1.0`.   We also need to update `doc/source/driver/docker/index.rst`, `doc/source/driver/docker/usage.rst`, qmolecule/driver/dockerdriver.py`, and `README.rst` with `docker` instead of `docker-py`.

retr0h:
Once the above is addressed, this commit looks good.

kireledan:
I tried to make the error message as clear as I could. Just saying to install docker could be a bit confusing.

PR #766, opened 2017-03-11
Default to driver from state if possible

alvaroaleman:


retr0h:
We should prob add a unit test to `test/unit/core/test_core.py`.

alvaroaleman:
@retr0h: Did that.

retr0h:
Thanks @alvaroaleman 🎸

PR #768, opened 2017-03-11
Get rid of terminal warnings issue while running apt

SergK:
Let's get rid of some warnings during packages installation on
ubuntu, e.g:

   debconf: unable to initialize frontend: Dialog
   debconf: (TERM is not set, so the dialog frontend is not usable.)
   debconf: falling back to frontend: Readline


retr0h:
Thanks @SergK 🏎

PR #769, opened 2017-03-13
Remove exception for setting the hostname on libvirt

bol:
This PR reverts the exception for setting the hostname when using libvirt in vagrant.

It was introduced back in [655d20b0bd6ebf995915f33b06d16bd338e0bc52](https://github.com/metacloud/molecule/commit/655d20b0bd6ebf995915f33b06d16bd338e0bc52) which claims that the vagrant-libvirt backend doesn't support it.
I don't know which version @conorsch tested against but vagrant-libvirt has supported setting the hostname since the [0.0.5 release](https://github.com/vagrant-libvirt/vagrant-libvirt/blob/master/CHANGELOG.md#005-may-10-2013) which came out in 2013.
Can we have it removed? It messes up integration with the landrush dns for vagrant.


retr0h:
@bol sounds reasonable.

Issue #771, opened 2017-03-14
AttributeError: 'module' object has no attribute 'APIClient'


popstas:
# Issue Type
- Bug report

# Molecule and Ansible details

```
molecule 1.21.0
docker-py 1.10.6
```

Appears today, after molecule update.

```
$ molecule test
Traceback (most recent call last):
  File "/usr/local/bin/molecule", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/site-packages/molecule/cli.py", line 41, in main
    cli(obj={})
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/molecule/command/test.py", line 94, in test
    t = Test(ctx.obj.get('args'), command_args)
  File "/usr/local/lib/python2.7/site-packages/molecule/command/base.py", line 60, in __init__
    self.main()
  File "/usr/local/lib/python2.7/site-packages/molecule/command/base.py", line 83, in main
    self.molecule.main()
  File "/usr/local/lib/python2.7/site-packages/molecule/core.py", line 56, in main
    self.driver = self._get_driver()
  File "/usr/local/lib/python2.7/site-packages/molecule/core.py", line 298, in _get_driver
    return dockerdriver.DockerDriver(self)
  File "/usr/local/lib/python2.7/site-packages/molecule/driver/dockerdriver.py", line 39, in __init__
    self._docker = docker.APIClient(
AttributeError: 'module' object has no attribute 'APIClient'
```

alvaroaleman:
The underlying issue is the docker-py python module, which is now called `docker` and has an incompatible api.

[...]

PR #772, opened 2017-03-14
Docs: install docker package in travis usage

popstas:
After 1.21 update travis builds fails with `docker-py` driver.
Related: #764, #771

### DIFF
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -60,6 +60,7 @@ A ``.travis.yml`` testing a role named foo1 with the Docker driver.
     - sudo apt-get install -o Dpkg::Options::="--force-confold" --force-yes -y docker-engine
   install:
     - pip install molecule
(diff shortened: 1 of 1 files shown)


doc/source/usage.rst
```diff
@@ -59,7 +59,7 @@ A ``.travis.yml`` testing a role named foo1 with the Docker driver.
     - sudo apt-get -qq update
     - sudo apt-get install -o Dpkg::Options::="--force-confold" --force-yes -y docker-engine
   install:
-    - pip install molecule
+    - pip install molecule docker
```
retr0h: I think we want to make this a comment.  This assumes everyone will be using docker, when we support a few drivers.  Could we change this to a comment?

```
 - pip install molecule
# - pip install required driver (e.g. docker, python-vagrant, shade)
```

popstas: Ok, I changed commit. Are you think that anyone will use travis with vagrant or openstack?

Issue #748, opened 2017-02-16
Feature/support docker '--network' parameter providing userdefined networks (thus dns/autodiscovery for its containers)


lhoss:
# Issue Type
- Feature request
Feature already asked in issue https://github.com/metacloud/molecule/issues/650#issuecomment-270351777 but  PR (https://github.com/metacloud/molecule/pull/725) was merged for for v1.20, without addressing the '--network' requirement

# Molecule and Ansible details
```
$ ansible --version
ansible 2.2.0.0
$ molecule --version
molecule, version 1.20.0
```

# Desired Behaviour
In many (testing) cases (at least in bigdata), you want to deploy a sample cluster of containers, providing:
* full connectivity between containers, by re-using a customizable docker network (name)
* automatic DNS resolution (so that each container in the same network can connect to another container, through it's container/host-name)
The good news, this functionality is provided out-of-the-box by using docker user-defined networks, see: https://docs.docker.com/engine/userguide/networking/#/user-defined-networks
So we just need to expose additional parameters to molecule.

I propose to fully expose the original docker params, so not to limit the configurability in any way.
[ An alternative (not recommended!), possibly simpler impl. could be adding a new (molecule custom) 'network mode' (like `udf_network`) that would ensure that a hardcoded custom network gets created and used by all containers that have the same setting (`udf_network`) ]

# Actual Behaviour (Bug report only)
Since v1.20 one can change the docker network *mode* ('bridge' being the default), but still not set a custom network, working in that default `bridge` mode.

abtreece:
@lhoss DNS resolution does not work for the default bridge network. If you create a new (non-default) bridge network it will work. The current docker network capabilities in Molecule basically only allow you to attach them to an existing network. Thus in order to use a non-default bridge you would need to manually create a new bridge network and then set `network_mode` to the name of that network.

I thought (briefly) about adding the ability to manage docker networks via molecule, but that significantly increased the scope of the changes I was making.

lhoss:
OK, now got it (working) 👍    with manually creating a docker network, and then setting network_mode to that network.

The 'network_mode' doc  IMO is confusing, if not wrong on following:
> bridge creates a new network stack for the container on the Docker bridge
This suggests that when using 'bridge' already a *new* (non default) docker network is created. Instead this is simply re-using the default one.

Anyway, my next question  (and another small feature request) is, where/how best to add the network creation ?

[...]

PR #780, opened 2017-03-20
Basic docker network handling for cluster setups

storpipfugl:
Allow a "network" section in the docker driver for basic handling of docker networks which enables testing of cluster setups with docker (#748). Requires docker >= 1.10
Also allow setting "hostname" for docker containers

Tests pass except py27-ansible19-unit which fails due to low unit test coverage but I'm not sure what to do about that when it skips large parts of the test suites

### DIFF
--- a/doc/source/driver/docker/index.rst
+++ b/doc/source/driver/docker/index.rst
@@ -48,12 +48,19 @@ Options
   existing Docker network.
 * ``build_image`` - **(default=None)** override global driver value
   to build (or not) an image for use with Molecule.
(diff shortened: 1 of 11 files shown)


retr0h:
@storpipfugl wonderful patch.  Can you please update the documentation for the new keys you're adding.

storpipfugl:
documentation updated with key description and usage example

retr0h:
Thanks @storpipfugl, very nice work! 🚀

PR #784, opened 2017-03-23
Ansible v2 has deprecated ansible_ssh_{host,port,user}

gzm55:
from [Ansible Document](http://docs.ansible.com/ansible/intro_inventory.html):

> Note
Ansible 2.0 has deprecated the “ssh” from ansible_ssh_user, ansible_ssh_host, and ansible_ssh_port to become ansible_user, ansible_host, and ansible_port. If you are using a version of Ansible prior to 2.0, you should continue using the older style variables (ansible_ssh_*). These shorter variables are ignored, without warning, in older versions of Ansible.

This will drop the support ansible<2.0 (we already done now?), but i think ansible>=2.0 is very popular now.

P.S. If really not able to avoid supporting old deprecated behaviors, we still could define both ansible_* and ansible_ssh_*, but the generated code is not clean.

### DIFF
--- a/molecule/driver/openstackdriver.py
+++ b/molecule/driver/openstackdriver.py
@@ -384,8 +384,8 @@ class OpenstackDriver(basedriver.BaseDriver):
                 os.remove(publoc)
 
     def _host_template(self):
-        return ('{hostname} ansible_ssh_host={interface_ip_address} '
-                'ansible_ssh_user={ssh_username} '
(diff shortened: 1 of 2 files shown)


retr0h:
My only concern is supporting Ansible 1.9.6.  I don't know how many users of Molecule use <2.0.

retr0h:
@gzm55 can you rebase your PR?

gzm55:
rebase done

PR #785, opened 2017-03-23
docker driver: use POSIX shell and support more linux package systems

gzm55:
1. 'bash -c' ==> '/bin/sh -c', when build image, we may not have bash installed
2. default Dockerfile will install bash, cause login_cmd() will always use bash (why not only use /bin/sh)
3. add support more pkg system: apk, pacman, dnf, emerge


retr0h:
Nice work @gzm55 🚀

Issue #783, opened 2017-03-22
BUG: Space in directory name causes problems


hkroger:
# Issue Type

- Bug report

# Molecule and Ansible details

```
$ ansible --version
molecule --version
ansible 2.2.1.0
  config file =
  configured module search path = Default w/o overrides

$ molecule --version
molecule, version 1.20.3
```

- Molecule installation method: brew install molecule
- Ansible installation method: brew install ansible

# Desired Behaviour

I ran `molecule converge` to provision machines. I expected the process to go smoothly and machines to be running after command.

# Actual Behaviour (Bug report only)

Ran `molecule converge` but command failed with following error:

    --> Starting Ansible Run...
    ERROR! Attempted to read ".molecule/ansible_inventory" as YAML: 'AnsibleUnicode' object has no attribute 'keys'
    Attempted to read ".molecule/ansible_inventory" as ini file: .molecule/ansible_inventory:1: Expected key=value host variable assignment, got: Documents/com~apple~CloudDocs/work/ansible/molecule/.vagrant/machines/db1/virtualbox/private_key

I saw that `.molecule/ansible_inventory` had some problematic rows because of spaces in directory names.

One of the problematic rows:
`db1 ansible_ssh_host=127.0.0.1 ansible_ssh_port=2222 ansible_ssh_private_key_file=/Users/hkroger/Library/Mobile Documents/com~apple~CloudDocs/work/ansible/molecule/.vagrant/machines/db1/virtualbox/private_key ansible_ssh_user=vagrant`

hkroger:
At least the issue is different now.

```
$ molecule --debug converge
DEBUG: ANSIBLE ENVIRONMENT
ANSIBLE_CONFIG: .molecule/ansible.cfg
ANSIBLE_FORCE_COLOR: 'true'
ANSIBLE_HOST_KEY_CHECKING: 'false'
ANSIBLE_SSH_ARGS: -o UserKnownHostsFile=/dev/null -o ControlMaster=auto -o ControlPersist=60s
  -o IdentitiesOnly=yes

--> Starting Ansible Run...
DEBUG: COMMAND
/usr/local/bin/ansible-playbook playbook.yml --inventory-file=.molecule/ansible_inventory --connection=ssh --limit=all --user=vagrant --timeout=30 --become --diff
ERROR! Unexpected Exception: 'Host' object has no attribute 'remove_group'
to see the full traceback, use -vvv
```

retr0h:
[...]

PR #787, opened 2017-03-24
Add quotes around ansible_ssh_private_key_file format

vegai:
Proposed simple fix for #783


retr0h:
@vegai Simple solution, you rock 🎸

PR #788, opened 2017-03-25
Clean up {group,host}_vars

alvaroaleman:


molecule/core.py
```diff
@@ -28,6 +28,8 @@
 from molecule import util
 from molecule.driver import basedriver
 
+from shutil import rmtree
```
retr0h: We do imports in a particular order.  Since this is a module bundled with python it should be imported under os on line `22`.  We also have a convention to import the base module, so code should be `import shutil` and reference `shutil.rmtree`.

alvaroaleman:
Thanks for the hint @retr0h. Are those molecule-specific conventions or general Python best practices?

retr0h:
> Thanks for the hint @retr0h. Are those molecule-specific conventions or general Python best practices?

pep8 and some [openstack coding standards](https://docs.openstack.org/developer/hacking/#imports).

molecule/command/destroy.py
```diff
@@ -48,6 +48,7 @@ def execute(self, exit=True):
             return e.returncode, e.message
         self.molecule.remove_templates()
         self.molecule.remove_inventory_file()
+        self.molecule.remove_vars_files()
```
retr0h: We should assert this is called in [test_destroy](https://github.com/metacloud/molecule/blob/master/test/unit/command/test_destroy.py).

molecule/core.py
```diff
@@ -234,6 +235,18 @@ def remove_inventory_file(self):
...
+            os.path.abspath(self.config.config['molecule']['molecule_dir']),
+            'group_vars')
+        host_vars_dir = os.path.join(
+            os.path.abspath(self.config.config['molecule']['molecule_dir']),
+            'host_vars')
+        if os.path.exists(group_vars_dir):
+            shutil.rmtree(group_vars_dir)
+        if os.path.exists(host_vars_dir):
+            shutil.rmtree(host_vars_dir)
+
```
retr0h: We should write a unit test or this.

alvaroaleman: Tbh I am not extremely comfortable with pytest and all the tests in `test/unit/core/test_core.py` that would be testing something alike have a `@pytest.mark.skip(reason='TODO(retr0h): Determine best way to test this')` decorator, thus there is no sample to adapt from.

alvaroaleman:
Thinking about it, `remove_vars_files()` shouldn't be called in any command but in `Molecule.main() right before `_add_or_update_vars()`

alvaroaleman:
Hey @retr0h, from mein POV this is ready. Does this look acceptable to you?

retr0h:
Hi @alvaroaleman will check it out this week.  Thanks as always 🚀

retr0h:
@retr0h remember to port to v2.

retr0h:
This does not need back ported to v2.  Molecule v2 already prunes the .moleule directory.

```
[jodewey:~/git/molecule_2/test/scenarios/host_group_vars] [molecule2-env-2.7.13] v2(+3/-0)+ ± touch molecule/default/.molecule/foo
[jodewey:~/git/molecule_2/test/scenarios/host_group_vars] [molecule2-env-2.7.13] v2(+3/-0)+ ± ls !$
ls molecule/default/.molecule/foo
molecule/default/.molecule/foo
[jodewey:~/git/molecule_2/test/scenarios/host_group_vars] [molecule2-env-2.7.13] v2(+3/-0)+ ± molecule create
--> Scenario: [default]
--> Provisioner: [ansible]
--> Driver: [docker]
--> Playbook: [create.yml]
Skipping, instances already created.
[jodewey:~/git/molecule_2/test/scenarios/host_group_vars] [molecule2-env-2.7.13] v2(+3/-0)+ ± ls molecule/default/.molecule/foo
ls: molecule/default/.molecule/foo: No such file or directory
```

PR #837, opened 2017-05-23
Quick one liner to enable ssh key forwarding

jamesla:


molecule/cookiecutter/driver/vagrant-runtime/{{cookiecutter.repo_name}}/vagrantfile
```diff
@@ -4,6 +4,8 @@ $state = YAML::load_file('{{ cookiecutter.vagrantfile_state_file }}')
 
 Vagrant.configure('2') do |config|
 
+    config.ssh.forward_agent = true
```
retr0h: Can you make this configurable with it set to False by default?

PR #842, opened 2017-05-29
Allow disable_cache parameter for Docker containers

james-portman:


retr0h:
@james-portman awesome.  Can you make sure [docs](https://github.com/metacloud/molecule/blob/master/doc/source/driver/docker/index.rst) are updated as well.

molecule/driver/dockerdriver.py
```diff
@@ -303,12 +303,17 @@ def _build_ansible_compatible_image(self, container):
         tag_string = container['image_tag'].format(container['image'],
                                                    container['image_version'])
 
+        if 'disable_cache' in container and container['disable_cache'] is True:
+            nocache = True
+        else:
+            nocache = False
+
```
retr0h: This would prob be simpler syntax.

```
nocache = container.get('disable_cache', False)
```

retr0h:
Thanks @james-portman 👍

Issue #845, opened 2017-06-01
if ANSIBLE_FILTER_PLUGINS set, provisioner create.yml can't find molecule_instance_with_scenario_name


ingvagabund:
# Issue Type

- Bug report

# Molecule and Ansible details

```
$ ansible --version
ansible 2.2.3.0
$ molecule --version
molecule, version 2.0.0.0rc6
```

- Molecule installation method: pip
- Ansible installation method: OS package

# Desired Behaviour

If `ANSIBLE_FILTER_PLUGINS` is set in molecule.yml file (`provisioner.env`), molecule provisioning is still working and docker container are destroyed/created.

# Actual Behaviour (Bug report only)

If `ANSIBLE_FILTER_PLUGINS` is set in molecule.yml file (`provisioner.env`), molecule provisioning of docker container fails with:
```
template error while templating string: no filter named 'molecule_instance_with_scenario_name'. String: {{ item.name | molecule_instance_with_scenario_name(molecule_yml.scenario.name) }}
```

# Reproducer

Take one of the examples in `https://github.com/metacloud/molecule/tree/v2/test/scenarios/driver/docker` directory and set the env. E.g.:
```yaml
...
provisioner:
  name: ansible
  env:
    ANSIBLE_FILTER_PLUGINS: ../../../../filter_plugins
...
```

When the `ansible-playbook` command for provisioning is created [1], all available env variables inside the provisioner are handed as well. So even if the generated `ansible.cfg` file is correct and `ANSIBLE_FILTER_PLUGINS` is set to `/usr/lib/python2.7/site-packages/molecule/provisioner/../../molecule/provisioner/ansible/plugins/filters:$ANSIBLE_FILTER_PLUGINS`, the command is run as:

```sh
ANSIBLE_FILTER_PLUGINS=../../../../filter_plugins /usr/bin/ansible-playbook ...
```

Given the `ANSIBLE_FILTER_PLUGINS` has higher precedence than the same env set in the `ansible.cfg` file, the `molecule/provisioner/ansible/plugins/filters` is always skipped.

I would suggest to check if the env variable is set inside `molecule.yml` file and if so, just filter it out from the constructed command.

[1] https://github.com/metacloud/molecule/blob/v2/molecule/provisioner/ansible_playbook.py#L71

ingvagabund:
Re-examing the generated `ansible.cfg` file again:
```yaml
# Molecule managed
[ssh_connection]
scp_if_ssh = True
[defaults]
[...]

PR #847, opened 2017-06-02
replace ANSIBLE_ like envs with their value when generating ansible.cfg

ingvagabund:
ANSIBLE_ like environment variables have higher precedence over their equivalents inside ansible.cfg. Thus, if the envs are set on CLI, the roles_path, library and filter_plugins inside ansible.cfg are completely ignored.
    
Given the ansible.cfg is generated each time `molecule test` command is run, it is ok to replace the envs with they values and filtering all three ANSIBLE_ like envs when generating ansible-playbook CLI command.

Fixes: #845

### DIFF
--- a/molecule/provisioner/ansible.py
+++ b/molecule/provisioner/ansible.py
@@ -206,6 +206,30 @@ class Ansible(base.Base):
 
         :return: dict
         """
+        envs = self._config.config['provisioner']['env']
+
+        roles_path = '../../../../'
+        filter_plugins = self._get_filter_plugin_directory()
+        library = self._get_libraries_directory()
+
+        # ANSIBLE_ like environments set on command line have higher precedence
+        # than their equivalents in the ansible.cfg file.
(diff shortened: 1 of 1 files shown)


ingvagabund:
@retr0h

molecule/provisioner/ansible.py
```diff
@@ -206,18 +206,40 @@ def default_config_options(self):
 
         :return: dict
         """
+        envs = self._config.config['provisioner']['env']
+
+        roles_path = '../../../../'
```
ashcrow: Will this always be accurate? (4 dirs back)

retr0h:
I highly suggest looking at molecule v2 branch.  This is the future of the project as v1 is on the way out.

retr0h:
Think we will need some unit tests reproducing these overriden scenarios.

retr0h:
Implemented at 38e8acb.

PR #849, opened 2017-06-02
Update for python 2 and 3

ashcrow:
This change moves metaclasses, string checks, and unicode casting to six
and also moves from iteritems to items.

**Note**: I found unittests fail with and without this patch.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ PyYAML==3.12
 sh==1.12.13
 tabulate==0.7.7
 testinfra==1.5.5
+six
--- a/molecule/ansible_playbook.py
+++ b/molecule/ansible_playbook.py
@@ -19,6 +19,8 @@
 #  DEALINGS IN THE SOFTWARE.
 
 import os
+
+import six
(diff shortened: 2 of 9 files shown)


requirements.txt
```diff
@@ -12,3 +12,4 @@ PyYAML==3.12
 sh==1.12.13
 tabulate==0.7.7
 testinfra==1.5.5
+six
```
retr0h: Can you keep the list alphabetical and pin to a release of six.

retr0h:
Also, as an FYI the v1 version of molecule is on it's way out in favor of v2.  Might be of interest to look at v2's py3 compatibility.  I believe I did a better job with that one.

ashcrow:
Understood. Closing this PR.

retr0h:
I'm good with merging this, but I wanted to point out v2.

ashcrow:
@retr0h thanks. Unless v1 will live on for quite a while I think taking a gander (and possibly fixing if need be) v2 is probably of more benefit.

PR #861, opened 2017-06-11
docs: fixed initializing new scenario in v2

butla:
I think there's something missing there. I hope I've fixed it properly.


retr0h:
@butla Thanks for cleaning this up!

PR #882, opened 2017-06-30
Make Dockerfile templated, reducing the default configuration

saez0pub:
Using default(omit) in docker create.yml, we have less configuration in
mlecule.yml.
The Dockerfile was not explicit when we want to build image because of
the "FROM centos:7".

### DIFF
--- a/molecule/cookiecutter/scenario/driver/docker/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
+++ b/molecule/cookiecutter/scenario/driver/docker/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
@@ -8,21 +8,30 @@
     molecule_scenario_directory: "{{ lookup('env','MOLECULE_SCENARIO_DIRECTORY') }}"
     molecule_yml: "{{ lookup('file', molecule_file) | from_yaml }}"
   tasks:
+    - name: Create Dockerfiles from image names
+      template:
+        src: "{{ molecule_scenario_directory }}/Dockerfile-template"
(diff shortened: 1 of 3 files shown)


molecule/cookiecutter/scenario/driver/docker/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
```diff
@@ -8,21 +8,30 @@
     molecule_scenario_directory: "{{ lookup('env','MOLECULE_SCENARIO_DIRECTORY') }}"
     molecule_yml: "{{ lookup('file', molecule_file) | from_yaml }}"
   tasks:
+    - name: Create Dockerfiles from image names
+      template:
+        src: "{{ molecule_scenario_directory }}/Dockerfile-template"
```
retr0h: I don't see this template existing in this PR.

molecule/cookiecutter/scenario/driver/docker/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/molecule.yml
```diff
@@ -7,9 +7,9 @@ lint:
   name: {{ cookiecutter.lint_name }}
 platforms:
   - name: instance-1
-    image: molecule_local/centos:7
-    command: sleep infinity
-    dockerfile: Dockerfile
+    image: centos:7
+    #command: sleep infinity
+    #dockerfile: Dockerfile
```
retr0h: We can remove the commented lines.

saez0pub:
Done, I had fogotten the git add :)
I also removed the comments.
Maybe  it would be better if the generated template is placed under the .molecule/ directory.

retr0h:
Thanks @saez0pub 🎸

Issue #884, opened 2017-07-04
Feature request: Add the possibility to avoid the scenario to be added to the hostname


Dorn-:
# Issue Type
- Feature request

# Molecule and Ansible details

```
ansible 2.3.1.0
molecule, version 2.0.0.0rc11
```

- Molecule installation method: pip
- Ansible installation method: pip

# Desired Behaviour

As your documentation says: 

> Molecule automatically appends the scenario name to the instances it is testing. It doesn’t seem useful to converge each scenario against the same static host.

But in special scenario we could need to avoid this functionnality. Let me take two examples happening in my case:
- I want to use the static drivers, I have in my possession differents VM on my enterprise ESX, each VM corresponding to a different OS version or to a different configuration. I just wanna set to my platforms configuration the IP to the VM.

- I would test my ansible role with different scenario on the same machine (VM, Vagrant, docker ...). I would love to test that my role is still working with differents vars. I have a role which install php, I want to install with default value, let's say PHP5.6, then relaunch the role with PHP7.1 on the same machine. It should find that there is already a PHP installed, remove it, install the new one etc ...

As you could see, english is not my first language, I hope I was clear enough.

I'll send you a PR to add this feature if you think it could be a good idea.

retr0h:
@Dorn- I wasn't initially a fan of this.  However, your PR is very clean.  I'll look this over more.

There are some use cases I don't think you are handling.   If a user does not use the static driver and still relies on molecule to create the instances.

The [molecule_instance_with_scenario_name](https://github.com/metacloud/molecule/blob/master/molecule/provisioner/ansible/plugins/filters/molecule_core.py#L25) will still try to append the scenario name via [instance_with_scenario_name](https://github.com/metacloud/molecule/blob/master/molecule/util.py#L194).

If the user provides `

```
platforms:
  - name: static-instance-vagrant
    append_scenario: False
```

Then that should short circuit all scenario name appending everywhere.

retr0h:
Also, the reason I like having the scenario name appended, it is very clear what is going on. 

> I want to use the static drivers, I have in my possession differents VM on my enterprise ESX, each VM corresponding to a different OS version or to a different configuration. I just wanna set to my platforms configuration the IP to the VM.
> 

I would simply argue you should spin up static instances with the scenario name appended.

[...]

PR #885, opened 2017-07-04
Add feature #884, add an option to avoid adding scenario name to the hostname

Dorn-:
- Add a boolean into the platform configuration `append_scenario` to let the user choose if he wants a different host for each of his scenario
- Update the documentation
- Update unit test for this new functionality


retr0h:
@Dorn- Could I get a little background on the use cases for this?

retr0h:
Supplemented by #895.

PR #888, opened 2017-07-05
force stop lxd containers on destruction

MarkusTeufelberger:
No need to wait for timeouts or whatnot since we want to delete everything anyways.

This might also be a useful patch for upstream (set force if removing containers), but I guess in some cases people might want their systems to gracefully shut down - even if they are deleted afterwards. I doubt that this is the case for the destroy phase of molecule though.


retr0h:
@MarkusTeufelberger thanks!

Can you also update our [functional scenario](https://github.com/metacloud/molecule/blob/master/test/resources/playbooks/lxd/create.yml) tests as well?

MarkusTeufelberger:
/done, this could also be done for Docker containers (using `force_kill`) by the way.

retr0h:
Thanks @MarkusTeufelberger 🎸

Issue #889, opened 2017-07-05
Update RUN statements in Dockerfile.j2


aliusmiles:
# Issue Type

- Feature request

# Molecule and Ansible details

```
ansible 2.3.1.0
molecule, version 2.0.0.0rc11
```

- Molecule installation method: pip
- Ansible installation method: brew

# Desired Behaviour

Multiple RUN statements result in multiple layers created, even though they do not increase image size. I propose to join them into one single RUN as follows:

```
RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get upgrade -y && apt-get install -y python sudo bash; \
    elif [ $(command -v yum) ]; then yum makecache fast && yum update -y && yum install -y python sudo yum-plugin-ovl bash && sed -i 's/plugins=0/plugins=1/g' /etc/yum.conf && yum clean all; \
    elif [ $(command -v zypper) ]; then zypper refresh && zypper update -y && zypper install -y python sudo bash; \
    elif [ $(command -v apk) ]; then apk update && apk add python sudo bash; \
    elif [ $(command -v pacman) ]; then pacman --sync --noconfirm --refresh python2 bash; \
    elif [ $(command -v dnf) ]; then dnf makecache fast && dnf --assumeyes install python python-devel python2-dnf bash; \
    elif [ $(command -v emerge) ]; then emerge --ask n =dev-lang/python-2\* gentoolkit; fi
```
Other changes (most are rather cosmetic):
* added `yum clean all` to reduce image size (from 331MB to 244MB for centos 7)
* removed `touch /var/lib/rpm/*` -- wasn't able to found it's purpose
* replaced `-x "$(command -v yum)"` with simple `$(command -v yum)` -- exit code 0 equals true anyway
* removed `/bin/sh -c` -- seems to me it does not have any visible effect, but increase clutter instead.
* replaced `;` with `&&` for dnf

aliusmiles:
I could create PR if needed.

retr0h:
> I could create PR if needed.

Please do

aliusmiles:
PR created: #891

retr0h:
Implemented.

PR #891, opened 2017-07-06
Updated Dockerfile template as per #889

aliusmiles:
* Joined multiple RUN statements into one
* Simplified if statements syntax
* Added cache cleaning commands

### DIFF
--- a/molecule/cookiecutter/scenario/driver/docker/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/Dockerfile.j2
+++ b/molecule/cookiecutter/scenario/driver/docker/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/Dockerfile.j2
@@ -1,11 +1,9 @@
 {% raw -%}
 FROM {{ item.image }}
 
-RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get upgrade -y && apt-get install -y python sudo bash && apt-get clean; \
+RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get upgrade -y && apt-get install -y python sudo bash ca-certificates && apt-get clean; \
     elif [ $(command -v yum) ]; then yum makecache fast && yum update -y && yum install -y python sudo yum-plugin-ovl bash && sed -i 's/plugins=0/plugins=1/g' /etc/yum.conf && yum clean all; \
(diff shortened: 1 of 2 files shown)


MarkusTeufelberger:
I ran into issues (not 100% related to this PR) when using Ubuntu:16.04 containers and Goss. The "Download and install Goss" stage (a https download) fails because ca-certificates is not installed by default in the container.

Maybe consider adding the respective ca-cert package to the ones installed by default (python, sudo and bash) for each distro.

aliusmiles:
Will look into this.

retr0h:
Also, be sure to update the test docker scenario [Dockerfile](https://github.com/metacloud/molecule/blob/master/test/resources/playbooks/docker/Dockerfile.j2).

aliusmiles:
It appears that ca-certificates are required for ubuntu and apline only.
I've also removed commands for Arch and Gentoo, for the following reasons:
* Building image from _archlinux/base_ fails due to:
```
warning: database file for 'core' does not exist
warning: database file for 'extra' does not exist
warning: database file for 'community' does not exist
error: no operation specified (use -h for help)
```
* Image builds successfully from _base/archlinux_, but converge fails with:
```
 fatal: [arch]: FAILED! => {"changed": false, "failed": true, "module_stderr": "/bin/sh: /usr/bin/python: No such file or directory\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 0}
```
`pacman python2` does not create /usr/bin/python symlink, and there seems to be neither systemd neither init-scripts in this image, so it's basically unusable with ansible

* _gentoo/stage3-amd64_ requires volumes_from _gentoo/portage_ or running `emerge-webrsync`, which pulls ~1GB portage tree, and compiling python2 from sources takes quite some time.

I'm frankly not sure how common is testing arch/gentoo in docker, and current RUN commands for those were not working with ansible anyway. Whole idea of Dockerfile in scenario folder is to allow user's customization, right?

MarkusTeufelberger:
For Arch Linux I'd recommend https://hub.docker.com/r/archimg/base/ - seems regularly updated, not extremely large and works with these commands for me. Unfortunately there doesn't seem to be an actual "official" image on dockerhub.

It would still be nice to have the commands for Arch in there by default, other distributions also have /usr/bin/python pointing to Python 3. Which module fails exactly? This seems more like a problem with the failing module than anything else.

retr0h:
@MarkusTeufelberger, yah the Arch commands are in there via subsequent PRs.

PR #940, opened 2017-09-04
Fix dockerfile for fedora

Anthony25:
Move dnf test before yum, as fedora 26 has symlinked yum to dnf. Avoid `dnf makecache` to complain because of the `fast` argument used by yum.


retr0h:
Thanks @Anthony25 can you also update https://github.com/metacloud/molecule/blob/master/test/resources/playbooks/docker/Dockerfile.j2?

Anthony25:
Sure, sorry.

Anthony25:
Thanks!

Issue #936, opened 2017-08-31
Allow Action: 'destroy' and Action: 'create' when driver.name: 'delegated'


qrrock:
# Issue Type

- Feature request

# Desired Behaviour

It would be nice to use already designed hooks   

```
driver:
  name: delegated
  options:
    followScenarioStrict: true        <----------- OR if == true ( SUGGESTION 1)


provisioner:
  name: ansible
  playbooks:
    create: create.yml                    <----------- OR  if != 'null'  (SUGGESTION 3)
    converge: playbook.yml
    destroy: destroy.yml                <----------- OR  if != 'null'  (SUGGESTION 3)
scenario:
  name: someCustomScenario   <----------- OR  if  != 'default'  (SUGGESTION 2)
  check_sequence:
    - destroy
    - create
    - converge
    - check
    - destroy
  converge_sequence:
    - create
    - converge
  destroy_sequence:
    - destroy
  test_sequence:
    - destroy
    - create
    - converge
    - lint
    - verify
    - destroy
```

# Actual Behaviour 
...
--> Action: 'destroy'
Skipping, instances are delegated.
--> Scenario: 'someCustomScenario'
--> Action: 'create'
Skipping, instances are delegated.
...

retr0h:
The whole point of delegated is not to destroy instances.  The delegated instances are managed externally to molecule.

kireledan:
This would be kinda cool though. Similar to how the openstack one is, as long as their create.yml populates the instance config correctly, molecule can use the instances. Then anyone could tinker with their own create/destroy yamls. Instead of 'delegated', we could let someone say their driver is 'custom'. The docs could describe general guidelines for creating the proper instance config.

retr0h:
This feature feels more like an edge case, and if we find people wanting support for a particular driver, its fairly simple enough to request the driver be added.

I'm not against what you describe @kireledan, I just want to be sure we aren't adding features for one off use-cases, or something that really wont be used much at all.

kireledan:
Yeah - I definitely understand what you mean. I'll be submitting a PR soon for a KVM driver. I'll include another doc with guidelines on creating proper drivers with it as well. It should help with anyone wanting to submit a PR.

retr0h:
I suppose the more I think about it the delegated driver really means we are delegating to you the developer vs my original plan of delegating to a server.  I'll think about this a bit more.

retr0h:
[...]

PR #941, opened 2017-09-05
Delegated driver may or may not manage instances

retr0h:
Reworked the delegated to fully delegate to the developer.

1. Developer may delegate the instance create/destroy to the delegate
   driver.
2. Developer may not use create/destroy and delegate to a manual
   process.

Fixes: #936


kireledan:
👍 awesome!

PR #947, opened 2017-09-12
update to ansible-lint 3.4.15

MarkusTeufelberger:
This should fix some errors with Ansible 2.4.

I haven't checked the remaining versions of requirements, maybe there are more that could/should be updated.


retr0h:
Thanks @MarkusTeufelberger 🎸

Issue #963, opened 2017-09-20
feature: Add a 'prepare' step as part of creation of VM


mzsilak:
# Issue Type
- Feature request

# Molecule and Ansible details

```
ansible 2.3.2.0
molecule master
```

- Molecule installation method: pip
- Ansible installation method: OS package

# Desired Behaviour

The _create_ step of a scenario shall be able to execute tasks on a VM after its creation through a new template file _prepare.yml_. The file shall tasks to prepare the VM for the intended scenario and role.

Minimum content (moved from playbook.yml):
```
- name: Prepare
  hosts: all
  gather_facts: False
  tasks:
    - name: Install python for Ansible
      raw: test -e /usr/bin/python || (yum clean all -y && yum install python-minimal -y)
      become: True
```

The _prepare.yml_ shall only be executed as part of VM creation. If the VM exists already it must not be executed to keep the idempotence of the _converge_ step

# Motivation

The current execution steps always target a fresh installation of a role. The addition of a _prepare_ step allows testing of roles in existing or (manually) modified environments and fits the possibility to create multiple scenarios with molecule.

retr0h:
I'm planning to implement anything like this at moment.  The playbook can be used to run any prepare statements necessary.  If you are concerned about idempotence mark the tasks or roles running as `changed_when: False` in playbook.yml.  If the prepare work is done correctly it can be made idempotent so it does not impact the converge steps.

mzsilak:
I guess you meant '... **not** planning to implement...'.

I have the issue that the my prepare tasks basically contradict the tasks in the role. So whenever I run the prepare tasks, the role has to change it back. (i.e. testing a package upgrade).

I created a fork and have some success already. I am not struggling with the tests and tox-tags. The changes I did so far would create a prepare.yml on new scenarios, but will also work with existing scenarios. If its fine for you, I would submit a pull request as soon as the tests are green.

retr0h:
Implemented in #971.

PR #968, opened 2017-09-21
Added a 'prepare' step as part of creation of VM (see also #963)

mzsilak:
This is an implementation of feature request #963

### DIFF
--- a/molecule/cookiecutter/scenario/driver/delegated/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/prepare.yml
+++ b/molecule/cookiecutter/scenario/driver/delegated/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/prepare.yml
@@ -1,0 +1,11 @@
+---
+{% raw -%}
+- name: Prepare
+  hosts: all
+  gather_facts: False
+  tasks:
+    - name: Install python for Ansible
+      raw: test -e /usr/bin/python || (apt -y update && apt install -y python-minimal)
+      become: True
+      changed_when: False
+{%- endraw %}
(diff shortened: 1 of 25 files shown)


retr0h:
Hi @mzsilak I'll check this out.

One thing I was thinking.  Maybe use the terminology customize vs prepare, but we can talk about that later.  I'll look over the code tho.

kireledan:
This looks pretty interesting. I'm definitely a fan of this.

molecule/cookiecutter/scenario/driver/delegated/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/prepare.yml
```diff
@@ -0,0 +1,11 @@
+---
+{% raw -%}
+- name: Prepare
+  hosts: all
+  gather_facts: False
+  tasks:
+    - name: Install python for Ansible
+      raw: test -e /usr/bin/python || (apt -y update && apt install -y python-minimal)
```
kireledan: Would it be better to use the 'package' module for the 'default' prepare step? That way non apt package managers don't get errors on first run.

mzsilak:
Hi @retr0h, apart from the terminology I also thought about doing the _prepare_ similarly to _side_effects_. Meaning only execute it when it is set. The term _customize_ would fit to this.
I personally opted for the current way, because the _install python_ task is mandatory anyway.

Hi @kireledan, unless I did a copy and paste mistake, the task was already as it is and I just moved it from playbook.yml to prepare.yml. But I agree, the package module would be better.
I could do it once this merge gets accepted or declined.

retr0h:
I personally like the terminology customize.  I'll review this soon enough.

retr0h:
Closing this one, reimplemented in #971.

PR #971, opened 2017-09-25
Execute a prepare playbook post create pre converge

retr0h:
This is a rework of #968.


retr0h:
@mzsilak You good with this?

@kireledan any thoughts?

retr0h:
So question are we for or against the naming customize vs prepare vs something else?

retr0h:
Customize

retr0h:
Prepare

retr0h:
Other

kireledan:
I like this rework, but I would prefer "Prepare" over customize. Prepare just seems to make more sense in regards to "Prepping" an instance for testing.

mzsilak:
Looks good, my tests were successful. I am still for _prepare_.

mzsilak:
I just noticed two things:
1) In my local setup, I am unable to call **molecule customize**. It does not do anything and it seems the test matrix is empty.
2) Having a dedicated customize command (instead of implicitly on create) might make troubleshooting for idempotence more difficult. 
It is not really possible to execute _converge_ twice to see what changes, because the customize might always change something.

mzsilak:
Another possible issue:

> We ship an empty task list in molecule init for drivers which don't need customization by default.
That does not work: ERROR! playbooks must be a list of plays

retr0h:
> In my local setup, I am unable to call molecule customize. It does not do anything and it seems the test matrix is empty.

Corrected

> Having a dedicated customize command (instead of implicitly on create) might make troubleshooting for idempotence more difficult.

Idempotence only runs converge, twice.  So it shouldn't be an issue.

> It is not really possible to execute converge twice to see what changes, because the customize might always change something.

Customize doesn't run on converge, and it only ever runs once per create/destroy.

retr0h:
I will rework this a bit tho.  It probably shouldn't have it's own sub command.  We should just execute the playbook in the create action.

retr0h:
Reworked the PR to use prepare and get called in create.

PR #975, opened 2017-09-29
use become=True when using the vagrant driver

pescobar:
I have started to test molecule and I found that when using the Vagrant driver my roles were not working unless I use `become: True` with ansible because the ssh connection is done as the vagrant user. 

This patch updates the cookiecutter template so when the Vagrant driver is used the `become: True` option is added by default to `molecule.yml`


retr0h:
Just add become to your playbook.  Molecule doesn't make any assumptions based on sudo access, so it is up to the user to add become to their [playbook](https://github.com/metacloud/molecule/blob/master/test/scenarios/driver/vagrant/molecule/default/playbook.yml#L4).  This is just standard ansible.

retr0h:
I'll provide more details here.

Originally, in Molecule v1 we set the ansible connection to use sudo.  However, many people did not like this b/c it didn't represent how they worked in production or allow them to test roles utilize `become`.

Molecule v2 decided to not pass sudo to testinfra or ansible, which would allow users the flexibility to test their roles in the way they desire.  True, the user does have some additional configuration, but the `molecule init --url` cookiecutter templates helps to solve that.

PR #977, opened 2017-09-29
Revert "Add support import data from original ansible.cfg"

retr0h:
This reverts commit 555dc2b571b5c78fac5b177496ca8b5d53763656.


retr0h:
This feature breaks ansible 2.4 compatibility, and goes against the desired way of modifying Molecle's internal ansible.cfg.

Issue #992, opened 2017-10-07
Unable to query for become sudo passphrase


msheiny:
# Issue Type

- Feature request - Add an option to molecule to pass a sudo password to ansible

I'm unable to prompt for a become/sudo password because of how molecule is calling ansible. I've been trying to work around it by using an askpass program but that doesn't cache and will prompt me at every task :( 

I gather the next question will be why do I want this feature? Well the more I read about docker security, it seems problematic to have a user configured to be able to run docker commands without credentials -- its essentially the same as giving your user password-less sudo access (http://www.projectatomic.io/blog/2015/08/why-we-dont-let-non-root-users-run-docker-in-centos-fedora-or-rhel/).


# Molecule and Ansible details

```
$ molecule --version    
molecule, version 2.2.1   

$ ansible --version     
ansible 2.4.0.0
```

- Molecule installation method: pip
- Ansible installation method: pip

# Desired Behaviour

```
$ molecule --ask-sudo-pass converge -s dev
# insert password prompt here
```

# Current Behaviour

```
╰─$ molecule create                                                                      
--> Test matrix                                     
└── default               
    ├── create            
    └── prepare           
--> Scenario: 'default'   
--> Action: 'create'      
/usr/lib64/python2.7/getpass.py:83: GetPassWarning: Can not control echo on the terminal.
  passwd = fallback_getpass(prompt, stream)         
Warning: Password input may be echoed.              

```

## Askpass notes

I'd actually be happy if someone could point me to a GUI askpass solution that would work in this case, the few I can find in my repository designed for openssh to not cache and are prompting for each become task :( I realize that question is totally out of scope of molecule and this ticket but I still feel like it would be great to document in your repo if future folks have this same question.

 Was calling that like so:

molecule.yml:
```yml
provisioner:                                                                                                                                                                                                   
  name: ansible                                                                                                                                                                                                
    config_options:                                                                                                                                                                                              
      defaults:                                                                                                                                                                                                  
        sudo_flags: -A   
```

```bash
$ SUDO_ASKPASS=/usr/libexec/openssh/x11-ssh-askpass molecule converge -s dev
```

retr0h:
Molecule assumes Docker is running on a desktop or a controlled system.  I suggest simply adding the user running Molecule to the docker group on the system.  This will prevent just any user from running Docker on the system.

msheiny:
yeah that's what I'm trying to avoid. I concede that's a problem I can
[...]

PR #1000, opened 2017-10-13
Implement sudo password via CLI

retr0h:
Ability to securely collect the sudo password and use it in various
playbooks.

Fixes: #992

### DIFF
--- a/doc/source/examples.rst
+++ b/doc/source/examples.rst
@@ -64,3 +64,31 @@ project.
     name: ansible
     env:
       ANSIBLE_$VAR: $VALUE
+
+Sudo
+====
+
+Some developers may wish to execute the playbooks with sudo.  As an example
+they may not want to add their user to the docker group.  Molecule provides
+a way to securely use the password as an environment variable in such
+playbooks.
+
+Adding the `ansible_sudo_pass` and necessary `become` options to the play, will
+enable such functionality.
+
+.. note::
+
+    Molecule accepts the `sudo` flag globally.  However, this does not mean
+    every subcommand will be executed with sudo.  The `sudo` flag is intended
+    as a means to pass the sudo password safely to the provisioner.
+
+.. code-block:: yaml
+
+    - name: Name of playbook
+      hosts: localhost
+      connection: local
+      gather_facts: False
+      become: True
+      no_log: "{{ not lookup('env', 'MOLECULE_DEBUG') | bool }}"
+      vars:
+        ansible_sudo_pass: "{{ lookup('env', 'MOLECULE_SUDO_PASSWORD') }}"
--- a/molecule/command/check.py
+++ b/molecule/command/check.py
@@ -40,6 +40,10 @@ class Check(base.Base):
     Executing with `debug`:
 
     >>> molecule --debug check
+
+    Executing with `sudo`:
+
+    >>> molecule --sudo check
     """
 
     def execute(self):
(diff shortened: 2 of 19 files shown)


retr0h:
I've added a `--sudo` flag to the base Molecule command (`molecule --sudo`), this will prompt for your sudo password, and make a `MOLECULE_SUDO_PASSWORD` environment available to Molecule.

Now for you to use this, you will need to configure your playbooks which use docker, lxc, whichever as follows.

```
- name: Destroy
  hosts: localhost
  connection: local
  gather_facts: False
  become: True
  no_log: "{{ not lookup('env', 'MOLECULE_DEBUG') | bool }}"
  vars:
    ansible_sudo_pass: "{{ lookup('env', 'MOLECULE_SUDO_PASSWORD') }}"
    molecule_file: "{{ lookup('env', 'MOLECULE_FILE') }}"
    molecule_yml: "{{ lookup('file', molecule_file) | molecule_from_yaml }}"
  tasks:
```

retr0h:
PR 1000 LOL

~[1000](https://media.giphy.com/media/pBevuxpD4Tq4U/giphy.gif)

fujin:
It's over :nine: :nine: :nine:! Congratulations.

msheiny:
hey @retr0h - this is working great on the `destroy` and `create` but is bombing out on the actual `converge` playbook. Whats the best way I can pass this environment variable to the runner that is calling this playbook? 

```
                       
    TASK [Gathering Facts] *********************************************************                     
    fatal: [gnomeshell_role_f26]: FAILED! => {"failed": true, "msg": "Docker version check (['/usr/bin/docker', 'version', '--format', \"'{{.Server.Version}}'\"]) failed: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.26/version: dial unix /var/run/docker.sock: connect: permission denied\n"}
```

retr0h:
> hey @retr0h - this is working great on the destroy and create but is bombing out on the actual converge playbook. Whats the best way I can pass this environment variable to the runner that is calling this playbook?

You can't just add the inventory var to your playbook.yml as well?

Maybe this is an incomplete solution to the actual problem.

msheiny:
> You can't just add the inventory var to your playbook.yml as well?

No :( Honestly, I think I have to do more research. This is a great start though! I suspect I could be hitting issues with upstream ansible and how they make `docker` connection calls. Maybe I'll need to run ansible as sudo during the converge playbook call ...

retr0h:
Yah, I can just seeing there being all sorts of weird edge cases.  For example if testinfra needs to talk to docker and requires sudo as well, then this solution really wont work.

Can backout to a Molecule on pypi and simply run `sudo molecule test` instead and let the OS prompt you for your sudo password and run molecule itself in privileged mode?

msheiny:
Uggghh totally I can see that cropping up with testinfra and friends. Dang it. Thanks for trying to tackle this. I do really appreciate it and think this is an important issue that needs to get solved but definitely outside the scope of molecule.

I think I'll just use docker-machine to run docker under a VM to solve my woes. However,  I still think this is a valuable PR to leave in. I can't think of any use-case off-hand but I can image a scenario spin-up/spin-down task requires sudo but the rest can be run as regular user. Would just have to add a big disclaimer to the docs.

retr0h:
For shits 'n giggles does `sudo molecule test` work?

msheiny:
errrrr @retr0h - i'll try running that in a disposable VM later, makes me nervous to run the entire molecule environment as root 😬

retr0h:
Closing

PR #1003, opened 2017-10-15
fix pragma: no cover comment typo

brad:
Hey folks,
I noticed this issue with the `# pragma: no cover` comment. Looks like a copy/paste accident.


retr0h:
Thanks @brad

PR #1004, opened 2017-10-15
fix python 3 compatibility checker

brad:
Hi folks,
Looks like there is a little bug in the Python 3 check. Every time I try to run with Python 3 it says my Python is incompatibile. It seems like it is though, and Python 3.6 is mentioned in the setup.cfg. I made a couple tiny changes that I think makes it work more as it was intended. The pseudocode:
```
if using a supported python version:
    if not using a supported ansible version:
        show the appropriate error and exit
else:
    # not using a supported python version
    show the appropriate error and exit
```

### DIFF
--- a/molecule/shell.py
+++ b/molecule/shell.py
@@ -51,16 +51,18 @@ def _supported_python3_version():  # pragma: no cover
 
 
 def _supported_ansible_version():  # pragma: no cover
-    if _supported_python2_version():
+    if _supported_python2_version() or _supported_python3_version():
         if (distutils.version.LooseVersion(_get_ansible_version()) <=
                 distutils.version.LooseVersion('2.2')):
             msg = ("Ansible version '{}' not supported.  "
                    'Molecule only supports Ansible versions '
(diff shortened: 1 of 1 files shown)


retr0h:
Hey @brad.  Molecule does not support Python.  We don't ship py3 files to pypi since Ansible does not seem to have a working project under py3.  Once Ansible actually works under py3 we will remove this check.

brad:
@retr0h Oh! Ansible [does have some support](http://docs.ansible.com/ansible/latest/python_3_support.html) for Python 3 even if it's not official yet. My role is working completely with Molecule 2.3.1, Pythen 3.6.3, and Ansible 2.3

retr0h:
> @retr0h Oh! Ansible does have some support for Python 3 even if it's not official yet. My role is working completely with Molecule 2.3.1, Pythen 3.6.3, and Ansible 2.3.

Once Molecule functional test pass under 3.6 I'll feel comfortable removing the check.  Unfortunately, I have run into issues various issues with every release since 2.2.  Because of this, it's hard for me to support it.

PR #1009, opened 2017-10-22
Add sudo package by default on Fedora

elliotweiser:
While testing an Ansible role against Fedora-flavored Docker
images, package installation would fail because `become: yes`
was set and `sudo` is missing.

There exists a workaround by installing `sudo` during the
preparation phase (i.e. the prepare.yml playbook). However, that
places the burden on the user to address it. Furthermore, other
package managers in the Dockerfile.j2 template installs `sudo`
by default, so this would make that pattern more consistent.


retr0h:
Thanks @elliotweiser

PR #1016, opened 2017-10-26
Fix/lint errors

jamatute:
If you run `molecule test` without an `.yamllint` in your role you get the following errors with the docker driver:

```
    /XXX/molecule/default/create.yml
      5:17      warning  truthy value is not quoted  (truthy)
      9:81      error    line too long (87 > 80 characters)  (line-length)
      10:81     error    line too long (85 > 80 characters)  (line-length)
      16:81     error    line too long (116 > 80 characters)  (line-length)
      30:81     error    line too long (92 > 80 characters)  (line-length)
      33:81     error    line too long (124 > 80 characters)  (line-length)
      41:19     warning  truthy value is not quoted  (truthy)
    
    /XXX/molecule/default/prepare.yml
      4:17      warning  truthy value is not quoted  (truthy)
    
    /XXX/molecule/default/destroy.yml
      5:17      warning  truthy value is not quoted  (truthy)
```
I've fixed this warnings and errors for `docker` and `vagrant`

Cheers

PR #1018, opened 2017-10-27
Added yamlint to init scenario

retr0h:
Molecule ships template files which pass internal yamllint validation.
Molecule also ships a `.yamllint` on `init role`.  However, molecule
did not ship a `.yamllint` file  with `init scenario`.  This patch
corrects the issue, and creates a generic cookiecutter template for
these types of future files.


retr0h:
Fixes an issue brought up in #1016.

PR #1024, opened 2017-11-01
support expose_ports option in docker driver

rwaweber:
While this is more or less on the end user to change their `create.yml` I think that these changes make it a little bit more ergonomic, so that those using molecule don't have to fiddle too much with files outside of the `molecule.yml`.

After reading through [this](https://github.com/metacloud/molecule/issues/953) I'm okay with changing the key name to `published_ports` if that's preferred.


retr0h:
Thanks @rwaweber can you also update the functional test's [create.yml](https://github.com/metacloud/molecule/blob/master/test/resources/playbooks/docker/create.yml).  Thanks!

rwaweber:
Sorry about that, made the suggestions!

retr0h:
> Sorry about that, made the suggestions!

no problem, I didn't expect you to know about that one :)

retr0h:
Thanks @rwaweber

PR #1032, opened 2017-11-14
Docs: usage of dependency role-file instead of requirements_file

popstas:
For avoid exploring molecule sources.


retr0h:
Thanks @popstas

Issue #1043, opened 2017-11-18
When using docker (image alpine:3.6): Authentication or permission failure


robertdebock:
# Bug report

When trying to test a role against alpine:3.6 (and alpine:3.5) I see an unexected error. Other platforms, such as Fedora, Ubuntu, Debian and CentOS seem to work fine.

# Molecule and Ansible details
```
ansible --version
ansible 2.4.1.0
  config file = None
  configured module search path = [u'/Users/username/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python2.7/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 2.7.14 (default, Sep 25 2017, 09:53:22) [GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.37)]

molecule --version
molecule, version 2.4.0
```

- Molecule installation method: pip
- Ansible installation method: pip

# Desired Behaviour

Create a role (works)
```
molecule init role --driver-name docker --role-name ansible-role-molecule
```

Modify molecule/default/molecule.yml to run on alpine:3.6 (works)
```
---
dependency:
  name: galaxy
driver:
  name: docker
lint:
  name: yamllint
platforms:
  - name: instance
    image: alpine:3.6
provisioner:
  name: ansible
  lint:
    name: ansible-lint
scenario:
  name: default
verifier:
  name: testinfra
  lint:
    name: flake8
molecule test
```
Normally I would expect a successful test:

```
molecule test
```

# Actual Behaviour

Run the tests (molecule test) fails:
```
--> Action: 'converge'
    
    PLAY [Converge] ****************************************************************
    
    TASK [Gathering Facts] *********************************************************
    fatal: [instance]: UNREACHABLE! => {"changed": false, "msg": "Authentication or permission failure. In some cases, you may have been able to authenticate and did not have permissions on the target directory. Consider changing the remote temp path in ansible.cfg to a path rooted in \"/tmp\". Failed command was: ( umask 77 && mkdir -p \"` echo ~/.ansible/tmp/ansible-tmp-1511002266.48-250329601373921 `\" && echo ansible-tmp-1511002266.48-250329601373921=\"` echo ~/.ansible/tmp/ansible-tmp-1511002266.48-250329601373921 `\" ), exited with result 1", "unreachable": true}
    
    PLAY RECAP *********************************************************************
[...]

PR #1044, opened 2017-11-20
A portable default command

retr0h:
Molecule attempts to boot the container with the `sleep infinity` command,
when command is absent.  However, this is not portable across supported
distros.  Updated command with a fairly generic while loop.

Fixes: #1043


robertdebock:
You're using bash in your PR, but bash is not installed on every image.

retr0h:
> You're using bash in your PR, but bash is not installed on every image.

It is for the images we support out of the box.  Plus the Dockerfile template and create/destroy files are free for the developer to change if they wish to do things differently.

https://github.com/metacloud/molecule/blob/master/molecule/cookiecutter/scenario/driver/docker/%7B%7Bcookiecutter.molecule_directory%7D%7D/%7B%7Bcookiecutter.scenario_name%7D%7D/Dockerfile.j2

PR #1064, opened 2017-12-15
Adds an option to allow vagrant to run the provisioning step

RuriRyan:
There's a new module argument called "no_provision" which, if set
to False, causes Vagrant to run the provisioning step.

I ran into a problem where a Vagrant Box wouldn't work because it has a Vagrantfile included with some shell script, which is needed for the Box to work correctly.

### DIFF
--- a/test/scenarios/driver/vagrant/molecule/default/tests/test_default.py
+++ b/test/scenarios/driver/vagrant/molecule/default/tests/test_default.py
@@ -7,7 +7,7 @@ testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
 
 
 def test_hostname(host):
-    assert 'instance' == host.check_output('hostname -s')
+    assert host.check_output('hostname -s') == 'instance'
 
 
 def test_etc_molecule_directory(host):
--- a/test/scenarios/driver/vagrant/molecule/default/tests/test_default.py
+++ b/test/scenarios/driver/vagrant/molecule/default/tests/test_default.py
@@ -7,7 +7,7 @@ testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
 
 
 def test_hostname(host):
-    assert host.check_output('hostname -s') == 'instance'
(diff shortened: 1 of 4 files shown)


test/scenarios/driver/vagrant/molecule/default/tests/test_default.py
```diff
@@ -7,7 +7,7 @@
 
 
 def test_hostname(host):
-    assert 'instance' == host.check_output('hostname -s')
+    assert host.check_output('hostname -s') == 'instance'
```
retr0h: Why is this flipped?  Looks unrelated and our pattern is `expected == returned`.

RuriRyan: This was actually the only test case which was written as `expected == returned` and my linter complained that test cases should actually be `returned == expected`. But I can revert it back, if it bothers you.

madrover:
Looking forward to this one!

retr0h:
We need to upgrade the cookiecutter templates.

* https://github.com/metacloud/molecule/blob/master/molecule/cookiecutter/scenario/driver/vagrant/%7B%7Bcookiecutter.molecule_directory%7D%7D/%7B%7Bcookiecutter.scenario_name%7D%7D/create.yml
* https://github.com/metacloud/molecule/blob/master/molecule/cookiecutter/molecule/%7B%7Bcookiecutter.role_name%7D%7D/%7B%7Bcookiecutter.molecule_directory%7D%7D/%7B%7Bcookiecutter.scenario_name%7D%7D/molecule.yml#L36-L39

nisimond:
I am, as a matter of principle, opposed to a value of `False` being used to mean "do something".

If you flip the semantics on this option, it will work the same, and IMO be clearer.

retr0h:
@RuriRyan please update with feedback provided.  Thanks.

retr0h:
Closing have not heard back from submitter.  Please reopen when you are able and willing.  Thanks again.

Issue #1065, opened 2017-12-16
goss can't be found in some cases


robertdebock:
# Issue Type
- Bug report

# Molecule and Ansible details

```
ansible --version
ansible 2.4.1.0
  config file = None
  configured module search path = [u'/Users/yf30lg/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python2.7/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 2.7.14 (default, Sep 25 2017, 09:53:22) [GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.37)]
```

```
molecule --version
molecule, version 2.4.0
```

- Molecule installation method: pip
- Ansible installation method: pip

# Desired Behaviour
Goss can be found on every type of system.

# Actual Behaviour (Bug report only)
When using goss on CentOS virtual machines, goss (located in /usr/local/bin) can't be found.

If goss is selected as the verifier:
```
verifier:
  name: goss
  lint:
    name: 'None'
```

On a vagrant CentOS (6/7) machine:
```
platforms:
  - name: centos-6-default
    box: centos/6
```

Goss can't be found:
```
    failed: [centos-6-default] (item=/tmp/test_default.yml) => {"changed": false, "cmd": "goss -g /tmp/test_default.yml validate --format documentation", "failed": true, "item": "/tmp/test_default.yml", "msg": "[Errno 2] No such file or directory", "rc": 2}
```

I'll make a merge requests to:
- use `{{ goss_dst }}` instead of `goss`. (That basically makes this execution independent of the set PATH.
- Remove `ignore_errors: True`. (IMHO a play should fail if there is an error. I was a little confused by the message where it actually failed now: `exception: 'dict object' has no attribute 'stdout_lines'`
- Use `become: yes`, because these tasks require elevated privileges.

retr0h:
These issues were merged.

PR #1066, opened 2017-12-16
Update verifier.yml

robertdebock:
instead of "goss", us the (variable) "{{ goss_dst }}" which can always be found. Solves #1065.


retr0h:
Thanks @robertdebock

PR #1068, opened 2017-12-16
Update verifier.yml

robertdebock:
Removing ignore_errors, this makes it easier to see where the issue is. Described in #1065.


MarkusTeufelberger:
Maybe also update the goss version, 0.3.5 is out for a while now.

retr0h:
Can you also [update our functional tests](https://github.com/metacloud/molecule/blob/master/test/scenarios/verifier/molecule/goss/verifier.yml).

PR #1079, opened 2017-12-28
Path searching to check ephemeral dir first

retr0h:
A slight rework of agreed upon changes from #973.  Searching the parent
directory first, may have unintended consequences.  Switched the order
to ephemeral directory first, so roles can be galaxy/gilted into the
project prior to tests.


champain:
👍

Issue #1084, opened 2018-01-09
Add vagrant 2.X


saez0pub:
# Issue Type

- Feature request

# Molecule and Ansible details

```
ansible --version
ansible 2.4.2.0
molecule --version
molecule, version 2.6.0
```

- Molecule installation method: pip
- Ansible installation method: pip

# Desired Behaviour

Activate linked_clone by default for vagrant virtualbox 2.X
Desired Vagrantfile : 
```
      config.vm.provider provider['name'] do |virtualbox|
        virtualbox.memory = provider['options']['memory']
        virtualbox.cpus = provider['options']['cpus']

        if provider['options'] and provider['options']['linked_clone']
          if Vagrant::VERSION =~ /^1.[89]/
            virtualbox.linked_clone = provider['options']['linked_clone']
          end
        else
          if Vagrant::VERSION =~ /^1.[89]/ or Vagrant::VERSION =~ /^2/
            virtualbox.linked_clone = true
          end
        end
```

# Actual Behaviour (Bug report only)

linked clone is activated only for vagran 1.8x or 1.9X 

Actual Vagrant file : 
```
      config.vm.provider provider['name'] do |virtualbox|
        virtualbox.memory = provider['options']['memory']
        virtualbox.cpus = provider['options']['cpus']

        if provider['options'] and provider['options']['linked_clone']
          if Vagrant::VERSION =~ /^1.[89]/
            virtualbox.linked_clone = provider['options']['linked_clone']
          end
        else
          if Vagrant::VERSION =~ /^1.[89]/
            virtualbox.linked_clone = true
          end
        end
```

saez0pub:
another solution : 
```
      config.vm.provider provider['name'] do |virtualbox|
        virtualbox.memory = provider['options']['memory']
        virtualbox.cpus = provider['options']['cpus']

        if provider['options'] and provider['options']['linked_clone']
          if Vagrant::VERSION =~ /^1.[89]/
            virtualbox.linked_clone = provider['options']['linked_clone']
          end
        else
          virtualbox.linked_clone = true if Gem::Version.new(Vagrant::VERSION) >= Gem::Version.new('1.8.0')
        end
```

retr0h:
Merged

PR #1085, opened 2018-01-09
Support linked_clone for Vagrant 2.X 

saez0pub:
Add support for vagrant 2.X resolves the #1084

### DIFF
--- a/molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
+++ b/molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
@@ -162,13 +162,9 @@ Vagrant.configure('2') do |config|
         virtualbox.cpus = provider['options']['cpus']
 
         if provider['options'] and provider['options']['linked_clone']
-          if Vagrant::VERSION =~ /^1.[89]/
-            virtualbox.linked_clone = provider['options']['linked_clone']
-          end
+          virtualbox.linked_clone = provider['options']['linked_clone'] if Gem::Version.new(Vagrant::VERSION) >= Gem::Version.new('1.8.0')
         else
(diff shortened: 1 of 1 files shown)


retr0h:
Please correct the commit message so it has an informative message.

retr0h:
Please correct the commit message so it has a descriptive commit message in the subject and body.

molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
```diff
@@ -162,13 +162,9 @@
         virtualbox.cpus = provider['options']['cpus']
 
         if provider['options'] and provider['options']['linked_clone']
-          if Vagrant::VERSION =~ /^1.[89]/
-            virtualbox.linked_clone = provider['options']['linked_clone']
-          end
+          virtualbox.linked_clone = provider['options']['linked_clone'] if Gem::Version.new(Vagrant::VERSION) >= Gem::Version.new('1.8.0')
```
retr0h: While I understand this is a more ruby way of doing things with the one line setting of a conditional var, it will exceed the project limit line length.  Please use the Gem version comparison, but keep the conditional on a separate line.

molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
```diff
@@ -162,13 +162,9 @@
...
         else
-          if Vagrant::VERSION =~ /^1.[89]/
-            virtualbox.linked_clone = true
-          end
+          virtualbox.linked_clone = true if Gem::Version.new(Vagrant::VERSION) >= Gem::Version.new('1.8.0')
```
retr0h: ditto

Issue #1091, opened 2018-01-12
Driver Docker missing ulimits


robertdebock:
# Feature request

Allow molecule to set ulimit values for the Docker Driver.

# Molecule and Ansible details

```
ansible --version
ansible 2.4.2.0
  config file = None
  configured module search path = [u'/Users/yf30lg/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python2.7/site-packages/ansible
  executable location = /Users/yf30lg/Library/Python/2.7/bin/ansible
  python version = 2.7.14 (default, Sep 25 2017, 09:53:22) [GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.37)]
```

```
molecule --version
molecule, version 2.6.0
```

- Molecule installation method: pip
- Ansible installation method: pip

# Desired Behaviour
In some cases it may be required to set a different ulimit for a Docker instance. For example "nofile". The Ansible module [Docker Container](http://docs.ansible.com/ansible/latest/docker_container_module.html#docker-container) allows this using the list "ulimits".

I'll prepare a merge request and link it to this feature request.

A hint of what work on my laptop now:

molecule.yml:
```
...
  - name: release-centos-6
    image: centos:6
    ulimits:
      - nofile:262144:262144
...
```

create.yml:
```
    - name: Create molecule instance(s)
      docker_container:
...
        ulimits: "{{ item.ulimits | default(omit) }}"
      with_items: "{{ molecule_yml.platforms }}"
```

robertdebock:
I've tests (part of) the code on [Travis CI](https://travis-ci.org/robertdebock/ansible-role-release/builds/328160911).

retr0h:
Merged.  Thnks @robertdebock !

PR #1092, opened 2018-01-12
Add the ability to set a ulimit for the Docker driver.

robertdebock:
Solves #1091 by passing ulimits (list) to the ansible docker container module.


retr0h:
Don't forget to update the other create.yml files.  Similar to what you did in #1090.

PR #1094, opened 2018-01-18
EC2: support ansible 2.4

siavashs:
This PR fixes some issues in ec2 driver when using ansible 2.4, `region` or `ec2_url` should be provided for some modules.
Since `include` is now deprecated in 2.4, we skip those in molecule generated playbooks.
Also `user` and `vpc_id` are added to platforms, since they are needed for some AMIs and subnets.


siavashs:
It seems you completely ignored my comments regarding **AMI usernames** and **VPC IDs**!

retr0h:
> It seems you completely ignored my comments regarding AMI usernames and VPC IDs!

Feel free to submit a PR with those changes.  My problem was with the other changes.  Sorry about that.

PR #1097, opened 2018-01-19
Add docker network

dj-wasabi:
This PR contains the following change:

* Creation/Deletion of Docker networks via the `create.yml` and `delete.yml` playbooks;

A filter `molecule_get_docker_networks` will be used that checks if a `networks` key is configured in the platforms. If so, it then checks if it has a `name` key and if that exists to, this name will be used with creating or deleting the docker network. If no `networks` key is configured in the platform (Or if so, but has no `name` key), then an empty list is returned and task is skipped.

Example usage for `molecule.yml`:

```
platforms:
  - name: ansible-docker
    image: milcom/centos7-systemd
    groups:
      - group1
    privileged: True
    networks:
      - name: my_fancy_network
```

If a better approach is needed, please let me know and I'll try again.
(I hope I have updated all correct test files :-) )

### DIFF
--- a/molecule/cookiecutter/scenario/driver/docker/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
+++ b/molecule/cookiecutter/scenario/driver/docker/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
@@ -33,6 +33,12 @@
       with_items: "{{ platforms.results }}"
       when: platforms.changed or docker_images.results | map(attribute='images') | select('equalto', []) | list | count >= 0
 
+    - name: Create docker network(s)
+      docker_network:
+        name: "{{ item }}"
+        state: present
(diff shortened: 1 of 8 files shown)


jdvr:
Amazing work. Thank you.

retr0h:
Hi @dj-wasabi Can you update the cookiecutter molecule.yml with the whole default omit pattern.

Also, can you update the test files and docker driver (for documentation) as well.  See #1092 for reference.

dj-wasabi:
@retr0h 

Thank you for the suggestion. I believe I have updated all files which were also in the PR you mentioned, if not please let me know and will update them to with this PR.

retr0h:
Hi @dj-wasabi can you be sure to update the docker driver class comments so the documentation gets updated.    I believe #1092 does this for reference.

dj-wasabi:
Hi @retr0h Sorry! I thought I had updated that file to/

retr0h:
@dj-wasabi 🎸

Issue #1060, opened 2017-12-04
regex support in additional_files_or_dirs:


NasrHassanein:
# Issue Type

- Feature request

# Molecule and Ansible details

```
ansible, version 2.4.2.0
molecule, version 2.5.0
```

- Molecule installation method:  pip
- Ansible installation method:  pip

# Desired Behaviour

[additional_files_or_dirs](https://molecule.readthedocs.io/en/latest/configuration.html#testinfra) is really good. I think it would be better if you added wildcard and regex support. That will enable us to include the tests inside the ansible role itself.

so it could be something like the following:
```
verifier:
  name: testinfra
  additional_files_or_dirs:
    - ../../ansible-roles/test_*.py
    - ../../ansible-roles/*/tests/
```

retr0h:
@dj-wasabi  interested in implementing this one?

dj-wasabi:
@retr0h I was already thinking about it. I'm not able to do anything after the weekend though. If issue is still open after this weekend, I'll give it a try.

hmarlo:
Nice feature! I definitely will use this.  Now we will be able to have the tests grouped in one folder:

```
tests/
    test_role_X.py
    test_role_Y.py
roles/
    role_X/
    role_Y/
```

I commented the WIP merge request, thanks @dj-wasabi

retr0h:
Implemented in 2.8.

PR #1105, opened 2018-01-26
regex support in additional_files_or_dirs

dj-wasabi:
Currently a WIP for issue #1060 , need to figure out how to properly test this. I created the PR already, so you can already comment on changes. 😄 

When a `molecule.yml` is configured like this:
```
verifier:
  name: testinfra
  additional_files_or_dirs:
    - ../../tests/test_*.py
    - ../../tests/*/tests/
  lint:
    name: flake8
```
The `molecule verify` will be using the additional file and directories:
```
    platform darwin -- Python 2.7.10, pytest-3.3.2, py-1.5.2, pluggy-0.6.0
    rootdir: /Users/wdijkerman/git/ansible/access, inifile:
    plugins: testinfra-1.7.1
collected 4 items                                                              
    
    tests/test_default.py .                                                  [ 25%]
    ../../tests/test_first.py .                                              [ 50%]
    ../../tests/dir2/tests/test_dir2.py .                                    [ 75%]
    ../../tests/dir1/tests/test_dir1.py .                                    [100%]
    
    =============================== warnings summary ===============================
```

I'm not sure if is needed so I will just mention it here, but a check is added to see if the additional file or directory exists. If this doesn't exist, the entry will be skipped and the `verify` step will not fail on this.

### DIFF
--- a/molecule/verifier/testinfra.py
+++ b/molecule/verifier/testinfra.py
@@ -19,7 +19,7 @@
 #  DEALINGS IN THE SOFTWARE.
 
 import os
-
+import glob
 import sh
 
 from molecule import logger
@@ -126,9 +126,18 @@ class Testinfra(base.Base):
 
         :return: None
         """
+        list_of_files_or_dirs = []
         options = self.options
         verbose_flag = util.verbose_flag(options)
-        args = verbose_flag + self.additional_files_or_dirs
+        additional_files_or_dirs = self.additional_files_or_dirs
+        for additional_path in additional_files_or_dirs:
+            glob_add_path_dir = glob.glob(self._config.scenario.directory + '/' + additional_path)
+            # If file/directory not exists, length will be 0 and continue with next path.
+            if len(glob_add_path_dir) == 0:
+                continue
+            list_of_files_or_dirs.append(glob_add_path_dir)
+
+        args = verbose_flag + list_of_files_or_dirs
 
         self._testinfra_command = sh.Command('py.test').bake(
             options,
--- a/molecule/verifier/testinfra.py
+++ b/molecule/verifier/testinfra.py
@@ -131,11 +131,11 @@ class Testinfra(base.Base):
         verbose_flag = util.verbose_flag(options)
         additional_files_or_dirs = self.additional_files_or_dirs
         for additional_path in additional_files_or_dirs:
-            glob_add_path_dir = glob.glob(self._config.scenario.directory + '/' + additional_path)
-            # If file/directory not exists, length will be 0 and continue with next path.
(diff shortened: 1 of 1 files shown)


molecule/verifier/testinfra.py
```diff
@@ -126,9 +126,18 @@ def bake(self):
...
         """
+        list_of_files_or_dirs = []
         options = self.options
         verbose_flag = util.verbose_flag(options)
-        args = verbose_flag + self.additional_files_or_dirs
+        additional_files_or_dirs = self.additional_files_or_dirs
+        for additional_path in additional_files_or_dirs:
+            glob_add_path_dir = glob.glob(self._config.scenario.directory + '/' + additional_path)
+            # If file/directory not exists, length will be 0 and continue with next path.
+            if len(glob_add_path_dir) == 0:
```
hmarlo: You can also write this as
```
if not glob_add_path_dir:
    continue
```

molecule/verifier/testinfra.py
```diff
@@ -126,9 +126,18 @@ def bake(self):
 
         :return: None
         """
+        list_of_files_or_dirs = []
         options = self.options
         verbose_flag = util.verbose_flag(options)
-        args = verbose_flag + self.additional_files_or_dirs
+        additional_files_or_dirs = self.additional_files_or_dirs
+        for additional_path in additional_files_or_dirs:
+            glob_add_path_dir = glob.glob(self._config.scenario.directory + '/' + additional_path)
```
hmarlo: In order to concatenate paths, the recommended way is to use `os.path.join` (or methods from Pathlib), since `/` is the POSIX separator, and if maybe will work on Windows, it's not the main separator there. [Reference](https://docs.python.org/3.6/library/os.html#os.sep).

molecule/verifier/testinfra.py
dj-wasabi: Ah, yes. Thank you.

molecule/verifier/testinfra.py
dj-wasabi: Indeed, I forgot about Windows. Will update it. Thanks!

dj-wasabi:
Is there someone that is able to help with the tests?

retr0h:
I'll have a look this week.

hmarlo:
Brainstorming:

* (subprocess approach) Maybe running testinfra with dry-run mode, capture the output and check which files are listed?
* (inside python approach) Running testinfra from a test and using a pytest feature to get the list of files before running them, and then check there and exit?

dj-wasabi:
Thanks @retr0h !

Issue #1069, opened 2017-12-17
Remove psutil dependency


pierrecarre:
# Issue Type

- Feature request

# Molecule and Ansible details

```
ansible --version
ansible 2.4.2.0
  config file = /cygdrive/c/home/user/.cygwin/.ansible.cfg
  configured module search path = [u'/cygdrive/c/home/user/.cygwin/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /cygdrive/c/home/user/.cygwin/.virtualenvs/test/lib/python2.7/site-packages/ansible
  executable location = /cygdrive/c/home/user/.cygwin/.virtualenvs/test/bin/ansible
  python version = 2.7.13 (default, Mar 13 2017, 20:56:15) [GCC 5.4.0]

molecule version: 2.5.0
```

- Molecule installation method: pip
- Ansible installation method: pip

# Desired Behaviour

I cannot install Molecule using pip on a Cygwin environment as psutil is not available on this kind of environment. I would like to know if you could remove the dependency to psutil, as it seems not used in the sources of Molecule. Is there a reason to have it in the requirements ? 
When I remove the psutil from requirements and build molecule from sources, it seems that it is working.

```
molecule --version
molecule, version 2.5.1.dev21
```

retr0h:
`psutil` is needed for click_completion, which is a feature the community enjoys.

retr0h:
I can probably leave psutil out of the requirements file, and users can install it if they wish to have completion work.  I'll need to think about this. 

It's kinda making one use case worse, for a very slim user base.

pierrecarre:
Thanks for the answer. I understand your point of view.

For me, as `psutil` is needed for click_completion, it should not be managed on your side, but on click_completion side (and this is done [here](https://github.com/click-contrib/click-completion#dependencies)).

retr0h:
@pierrecarre Can you update `requirements.txt` and verify this works `psutil==5.2.2; sys.platform != 'win32'`.

retr0h:
Implemented in #1080

pierrecarre:
@retr0h Hi, I tried to install the latest version, and it still did not work on Cygwin. 

I managed it in the `requirements.txt` using `psutil==5.2.2; sys_platform != 'win32' and sys_platform != 'cygwin'`. I'm going to create a pull request if you agree with this solution.

retr0h:
@pierrecarre sounds good

[...]

PR #1131, opened 2018-02-15
Ignore psutil on cygwin platform

pierrecarre:
Omit installation of psutil on cygwin, where it does not exist

Fixes Issue #1069


retr0h:
Thanks @pierrecarre 🎸

PR #1134, opened 2018-02-16
Vagrant provider - Add basic support for libvirt

msheiny:
This was pretty much a copy/paste from the parallels block above, I was considering appending to the
parallels section and adding an `if provider == parallels or provider == libvirt` but seems like the
original design wanted them intentionally separated for future provider specific functionality calls.

Let me know if you want me to back-pedal and do that instead.


retr0h:
@msheiny Thanks looks good.

PR #1135, opened 2018-02-18
Fix confusion between exposed and published ports in docker create playbook

informancer:


retr0h:
Please update the following files also.

* test/resources/playbooks/docker/create.yml

Update docs in:

* molecule/driver/docker.py

informancer:
Here you are :D

retr0h:
Thanks @informancer 🎸

Issue #1182, opened 2018-03-13
Dependency Provider: shell 


lxhunter:
Hi, 

its me again :) , i was looking through v1 and saw there was a shell dependency provider. I wrote a tool ([gooh](https://github.com/lxhunter/gooh)), close to gilt, which uses [tilde and caret](https://blog.madewithlove.be/post/tilde-and-caret-constraints/) like bundler, npm or composer does, so that i get updates on patch and minor changes. 

like so:
```yml
- src: https://github.com/lxhunter/versions.git
  version: '~1.0.0'
```
Is there any plan to implement this, i could give it a try, as i see it is really close to the Tilt Provider.

What do you think?

Cheers,
Alex

retr0h:
Shell is probably one of the dependency managers I should forward port from v1.  Also, would you be interested in adding gooh semver support to Gilt?

lxhunter:
Aside from a lot of excuses which directly spring into mind, like i am a python beginner and don't know a lot about the whole build stack. I can give it a try. I really like to contribute... but you have to be honest if i made a mess in my code. What do you think, still try?

retr0h:
> What do you think, still try?

Sure!  Looking at `gooh` it is solving a similar problem to `Gilt`.  If `Gilt` could do semver tag checkout, that would give you what you wanted, yeah?

PR #1185, opened 2018-03-14
Forward port Molecule v1 shell dependency manager

retr0h:
Forward ported Molecule's v1 shell dependency manager to v1,
which mostly turned into a rewrite due to the differences between
v1 and v2, but :shussh: who is really looking at this message
anyway.

Fixes: #1182


lxhunter:
I do 😬

dj-wasabi:
LOL. 😄 👍

Issue #1171, opened 2018-03-07
Rework schema validation


retr0h:
# Issue Type

- Feature request

# Desired Behavior

We currently use [Marshmallow](https://github.com/marshmallow-code/marshmallow) for schema validation.  It would be great to validate the schema and introduce a key to identify which version of Molecule the key was introduced in.

* Identify if Marshmallow should continue to be used.
* Ability to unmarshal the config into a different version (e.g v1 to v2+microversion).  Looks like this will need to be built.
* Extensible.  Able to provide custom rules and validators.
* Implement a `version_added` tag to the schema.

retr0h:
Looking into the following. 

* [JSON schema](http://json-schema.org/)
* [PyKwalify](https://github.com/Grokzen/pykwalify)
* [Cerberus](http://docs.python-cerberus.org/en/stable/)
* [Voluptuous](https://pypi.python.org/pypi/voluptuous)
* [Rx](http://rx.codesimply.com/)

PR #1194, opened 2018-03-20
Reimplemented schema validation with Cerberus

retr0h:
The Molecule file is thoroughly validated.  This ended up
being a much larger change than I wanted.

Found a major issue with unit tests.  Running all tests
would end up modifying subsequent test data.  Tests were
merging the same underlying dict.  This went unnoticed
because the config was not thoroughly validated.

* Docker driver credential user/password must be interpolated
  to ensure users do not check-in a user/password.
* The Docker and Vagrant driver Molecule files have enhanced
  validation.
* Ported disallowed from Marshmallow into Cerberus validation.
* Reworked unit tests to deal with config pollution.
* Provisioner sanitize env moved into schema validation.
* Removed dependency name validation from config and moved to
  schema validation.
* Ported existing v1 migration to new schema validator.

Fixes: #1171


retr0h:
tenor-37482529

PR #1211, opened 2018-03-27
AnsibleLint: fix docstring

maxdanilov:
Fix an example for how to pass -x option to ansible-lint in case multiple rules need to be ignored.
The current solution does not work for both rules (only one of them is excluded from checks).

### DIFF
--- a/molecule/provisioner/lint/ansible_lint.py
+++ b/molecule/provisioner/lint/ansible_lint.py
@@ -49,10 +49,10 @@ class AnsibleLint(base.Base):
               exclude:
                 - path/exclude1
                 - path/exclude2
-              x:
-                - ANSIBLE0011
-                - ANSIBLE0012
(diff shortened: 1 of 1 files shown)


maxdanilov:
The current implementation calls `ansible-lint` like this:
`ansible-lint -x ANSIBLE0012 -x ANSIBLE0013 ...`
It needs to be this instead:
`ansible-lint -x ANSIBLE0012,ANSIBLE0013 ...`

retr0h:
This is an [ansible-lint bug](https://github.com/willthames/ansible-lint/issues/279).

retr0h:
@maxdanilov Can you put a comment pointing to the ansible-lint bug as well, so that we can uncomment the other -x option passing when ansible-lint fixes their bug?

maxdanilov:
@retr0h I've added a comment (not sure though if this is want you've meant).

retr0h:
thanks @maxdanilov 🎸

PR #1213, opened 2018-03-28
openstack: set delete_fip = yes for os_server resources

berendt:
Floating IP associated with created instances will be deleted
along with the instance.


berendt:
As reference: http://docs.ansible.com/ansible/latest/modules/os_server_module.html

retr0h:
Thanks @berendt

PR #1215, opened 2018-03-29
Remove include tasks from driver playbooks

berendt:
* the include task itself is outdated
* the outsourced tasks are only used in a single playbook (create)
* it's easier to keep the playbooks in sync with a single playbook
  without additional files


berendt:
Does that make sense to you? then I would clean up the test directory.

Then I would move the variables set directly in the playbook into a group_vars file (in it's own PR). This makes the playbook completely static.

retr0h:
Ya, makes sense.  Good idea.  Could you also cleanup the includes in the `https://github.com/metacloud/molecule/tree/master/test/resources/playbooks` directory?

berendt:
@retr0h Did a full sync of the playbooks in test. Please double check them.

retr0h:
Thanks @berendt you rock 🎸

PR #1230, opened 2018-04-03
Moved functional test provisioning vars

retr0h:
Moved vars into `molecule.yml` instead of the playbook.  This is
now the recommended way to manage data between `molecule.yml`
and the provisioning playbooks.


berendt:
@retr0h This one is closed but not yet merged..?

retr0h:
> @retr0h This one is closed but not yet merged..?

@berendt I opted against it at the moment.  Doesn't really make anything clearer.

berendt:
@retr0h It is fine for me.

PR #1233, opened 2018-04-06
Add 'halt' option to vagrant module

msheiny:
Added a `halt` option to `state` to shut-down a VM without deleting it.

### DIFF
--- a/molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
+++ b/molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
@@ -354,6 +354,15 @@ class VagrantClient(object):
 
         self._module.exit_json(changed=changed)
 
+    def halt(self):
+        changed = False
+        cd = self._created()
+        if cd:
+            changed = True
+            self._vagrant.halt(force=self._module.params['force_stop'])
+
+        self._module.exit_json(changed=changed)
+
     def _conf(self):
         instance_name = self._module.params['instance_name']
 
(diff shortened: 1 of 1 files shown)


retr0h:
Doesn't this functionality already [exist](https://github.com/msheiny/molecule/blob/master/molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py#L309-L310)?

msheiny:
hey @retr0h !! technically the section you highlighted already performs a halt but then immediately afterwards it performs a destroy

```python
            if self._module.params['force_stop']:
                self._vagrant.halt(force=True)
            self._vagrant.destroy()
```

I just want the halt functionality without destroying the machine. Hope that makes more sense.

retr0h:
@msheiny oh yeah, I see.  Can you update the [module docs](https://github.com/msheiny/molecule/blob/master/molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py#L112) and add an [example](https://github.com/metacloud/molecule/blob/master/doc/source/configuration.rst#molecule-vagrant-module).

molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
```diff
@@ -354,6 +354,15 @@ def destroy(self):
 
         self._module.exit_json(changed=changed)
 
+    def halt(self):
+        changed = False
+        cd = self._created()
+        if cd:
```
berendt: Why not "if self._created()"?

msheiny: Thats a good observation @berendt ... I'd like to keep my PR pretty lean though and since that paradigm is already used throughout the file. I think filing a different PR to cleanup that specific logic would be a better course of action.

msheiny:
updated @retr0h ! :+1:

molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
berendt: @msheiny Have not checked the rest of the file. if it is used everywhere else it fits.

retr0h: @msheiny good catch, cleaned those up (https://github.com/metacloud/molecule/commit/507893079802ca594a671bafb90c21f9255f1387)

PR #1235, opened 2018-04-06
Add parameter for vagrant provider override

msheiny:
See the docs for additional info on this option:
https://www.vagrantup.com/docs/providers/configuration.html#overriding-configuration

While trying to disable nfs explicitly on a box that had it enabled... doing an over-ride on the provider was the only way I could get it working :man_shrugging: This might be a libvirt specific problem but adding this option fixed it for me and its optional.

### DIFF
--- a/molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
+++ b/molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
@@ -100,6 +100,11 @@ options:
       - Additional provider options not explcitly exposed by this module.
     required: False
     default: {}
+  provider_override_args:
+    description:
+      - Additional override options not explcitly exposed by this module.
+    required: False
+    default: None
   provider_raw_config_args:
     description:
       - Additional Vagrant options not explcitly exposed by this module.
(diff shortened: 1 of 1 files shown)


retr0h:
@msheiny can you provide more information, specifically usage of your new option.

msheiny:
sureeeeeeeee @retr0h i'm specifically using it in this scenario:

`molecule.yml`:
```
driver:
  name: vagrant
  provider:
    name: libvirt
platforms:
  - name: app-staging
    box: fpf/securedrop-app
    box_url: "../vagrant_packager/box_files/app_metadata.json"
    box_version: "0.6"
    instance_raw_config_args:
      - "ssh.insert_key = false"
    provider_override_args:
      - "vm.synced_folder './', '/vagrant', disabled: true, type: 'nfs'"
    memory: 1024
    private_ip: 10.0.1.2
    groups:
      - securedrop_application_server
      - securedrop
      - staging
```

I'm trying to disable synced folder mapping and NFS for the underlying box. No matter where else I try to put the NFS value I was unable to stop vagrant from bombing out without sticking it under the `override` key.

retr0h:
Thanks @msheiny can you add this example to https://github.com/metacloud/molecule/blob/master/doc/source/configuration.rst#molecule-vagrant-module ?

PR #1249, opened 2018-04-13
Retry downloading goss 3 times

robertdebock:
Retry downloading goss 3 times as it fails sometimes. Fixes https://github.com/metacloud/molecule/issues/1248


retr0h:
Great fix @robertdebock you rock 🎸

Issue #1245, opened 2018-04-12
Handle Vagrant provider errors


baccenfutter:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version
2.5.0
```

```
molecule --version
molecule, version 2.12.1
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

# Desired Behavior

When Vagrant detects and reports a provider error, Molecule should handle that error and throw an error message at the user with a clear, concise and human-readable hint pointing to the actual cause of the issue.

# Actual Behaviour (Bug report only)

When issuing a `vagrant up --provider=virtualbox` and Vagrant detects the the desired provider (VirtualBox) is available in only unsupported versions, it throws a clear, concise and human-readable error message pointing to the actual cause of the problem. Molecule just throws a `python-vagrant` stack-trace that contains no indication of the actual cause what so ever and renders the user completely clueless about what the heck is going on.

retr0h:
I'm going to need more info here.  Need to know how to reproduce the error you are seeing so that a nice message can be reported.   Unfortunately, I cannot capture any possible error vagrant may report, and provide a nice message as part of this bug report.

baccenfutter:
Example using `vagrant-1.8.1` in combination with `virtualbox-5.1`:
```
# molecule create
    [...]
    The full traceback is:
    Traceback (most recent call last):
      File "/tmp/ansible_AyuYHB/ansible_module_molecule_vagrant.py", line 476, in <module>
        main()
      File "/tmp/ansible_AyuYHB/ansible_module_molecule_vagrant.py", line 469, in main
        v.up()
      File "/tmp/ansible_AyuYHB/ansible_module_molecule_vagrant.py", line 334, in up
        cd = self._created()
      File "/tmp/ansible_AyuYHB/ansible_module_molecule_vagrant.py", line 369, in _created
        status = self._status()
      File "/tmp/ansible_AyuYHB/ansible_module_molecule_vagrant.py", line 364, in _status
        s = self._vagrant.status(vm_name=instance_name)[0]
      File "/home/brian/.virtualenvs/playbook-programmfabrik/local/lib/python2.7/site-packages/vagrant/__init__.py", line 493, in status
        output = self._run_vagrant_command(['status', '--machine-readable', vm_name])
      File "/home/brian/.virtualenvs/playbook-programmfabrik/local/lib/python2.7/site-packages/vagrant/__init__.py", line 976, in _run_vagrant_command
[...]

PR #1251, opened 2018-04-16
Log vagrant errors

retr0h:
Molecule may run into issues with Vagrant for various reasons.  Log
these errors to a log file, and notify the developer of the path to
the log on module failure.

Fixes: #1245


berendt:
@retr0h Why is this a breaking change? maybe it makes sense to document the break, I am not familiar with Vagrant.

retr0h:
> @retr0h Why is this a breaking change? maybe it makes sense to document the break, I am not familiar with Vagrant.

@berendt wrong label.  Not a breaking change.  Sorry about that.

PR #1258, opened 2018-04-18
Support Void Linux when using Docker driver.

rosstimson:


retr0h:
Thanks as always @rosstimson

PR #1320, opened 2018-05-25
Docker create playbooks: add tmpfs & security_opts docker_container parameters

pilou-:
Docker create playbooks: add `tmpfs` & `security_opts` [docker_container module](http://docs.ansible.com/ansible/2.5/modules/docker_container_module.html) parameters.


retr0h:
Can you also update the [docs]( https://github.com/metacloud/molecule/blob/master/molecule/driver/docker.py#L41-L74) with these new options?

pilou-:
Doc and schema updated.

PR #1334, opened 2018-06-13
I update the dockerfile for work with az client, rubocop

luupux:
The actual container isn't  able to execute test on windows vm with inspec and rubocop I thought it's a good idea to update the Dockerfile and share with you

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -18,13 +18,21 @@ ENV PACKAGES="\
     linux-headers \
     sudo \
     rsync \
+    curl \ 
+    ruby \
+    ruby-dev \ 
 "
 
 ENV PIP_PACKAGES="\
+    ansible[azure] \
     virtualenv \
     molecule \
     ansible \
     docker-py \
+    azure-cli \
+    packaging \
+    msrestazure \
+    pywinrm \ 
 "
 
 RUN \


luupux:
Hi, it's possible to know the timeline for my request or if exist one way to contribute to the project to merge my request  fast ?

retr0h:
Hi @luupux I'm not sure I want to bake everything into the Molecule default image.  I think in this case you could simply use the Molecule image as a base for your Dockerfile and add your additions ontop.  I appreciate the PR, but for now I think it would be best to build ontop of Molecule's image.

luupux:
Ok , but if you want to use default molecule with inspec test and azure resource you need extra
package and I think to the it a good Idea to have a standard  image  for every driver/provisioner/lint/ecc 
supported in molecule .
I spent some time to find the good way to install extra package  using alpine image  for use molecule with inspec and rubocop(the only  tools good anough for windows and azure) .
I think it's good idea for the project to support as soon as possible the windows enviroment and
I give you this opportunity and sincerly I don't understand the problem to update Docker image for support in the best way the windows enviroment  and Inspec verify  in azure.
I appreciate so much you effort and I hope you give me an answer

retr0h:
@luupux good points.  I was trying to keep the size of the image small but it probably makes sense to accept PR of this nature that is helpful to others using Molecule.

retr0h:
Thanks @luupux

PR #1342, opened 2018-06-22
Add support for multiple distributions

dhoppeIT:
This will close #1341.


dhoppeIT:
@retr0h Do you have any doubts about this change? I think it would be useful to support more than one distribution / release by default.

PR #1348, opened 2018-06-26
Add support for Systemd

dhoppeIT:


dhoppeIT:
@retr0h I think it would be useful to support Systemd by default because we would be able to test services.

retr0h:
If users wish to use systemd they are free to configure it.  Not everyone who uses Molecule uses systemd containers.  Molecule has a nice doc on [enabling](http://molecule.readthedocs.io/en/latest/examples.html#systemd-container) this if one desires.  However, for now it shouldn't be a default.

Issue #1382, opened 2018-07-11
Support Python 3.7


webknjaz:
# Issue Type

- Bug report

# Desired Behavior

I want to be able to run it under Python 3.7, there's no fatal difference, which would prevent it from running under forward-compatible pythons, so it should be possible to painlessly change the check to be >= 3.6

# Actual Behaviour (Bug report only)

```
$ molecule check
ERROR: Python version '(3, 7)' not supported.  Molecule only supports Python versions '2.7' and '3.6'.
```

retr0h:
At some point.

PR #1388, opened 2018-07-15
Added python 3.7 support

retr0h:
Updated Molecule to support python versions >= 3.6
starting with Ansible 2.4.

Fixes: #1382


lwm:
:fire: :fire: :fire:

PR #1390, opened 2018-07-16
add docker python and rubocop dependencies

jamesla:
Replaces the discontinued docker-py library with the modern docker library which allows new versions of ansible to work properly. Also adds Rubocop and its dependencies so that the inspec linter will run.


retr0h:
Thanks @jamesla \m/

Issue #1346, opened 2018-06-26
Feature request, setting no_log from playbook.yml


rgarrigue:
# Issue Type

- Feature request

# Molecule and Ansible details

```
ansible 2.5.5
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/gitlab-runner/builds/54a89bc8/0/r.garrigue.ext/ansible-role-webenv/virtenv/lib/python3.6/site-packages/ansible
  executable location = /home/gitlab-runner/builds/54a89bc8/0/r.garrigue.ext/ansible-role-webenv/virtenv/bin/ansible
  python version = 3.6.5 (default, Apr 10 2018, 17:08:37) [GCC 4.8.5 20150623 (Red Hat 4.8.5-16)]

```

```
molecule, version 2.15.0

```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

# Desired Behavior

I'ld like to be able to set `no_log: false` from `molecule/default/playbook.yml`. Debug is way too verbose for me, I just wan't to see the censored message

![image](https://user-images.githubusercontent.com/1872031/41896483-fd86198a-7924-11e8-8c87-219b403e198f.png)

retr0h:
You are free to modify your create/destroy playbooks as you choose.  However, this isn't a default we will be setting.  Again, you are free to change your own setup how you see fit.

rgarrigue:
@retr0h I don't have the above mentionned playbooks in the 2.15.0 generated default scenario

![image](https://user-images.githubusercontent.com/1872031/41957098-1ce0a9c8-79e6-11e8-980c-ec81eaa2d9f2.png)

Your soft, your call, though if there's no way to adjust verbosity I'm curious to know how you spot the issues in a debug output. As for me I can't run molecule on my windows workstation, I only have the log shown by Gitlab CI without coloration, I'm looking for "error" with firefox's Ctrl-F... not exactly efficient :-/

retr0h:
Let me think about this more.

retr0h:
I don't have the cycles to implement this.  Curious if anyone else does.

retr0h:
@rgarrigue when 2.17 lands you can use this feature.  Update your `molecule.yml`'s provisioner section with:

```
provisioner:
  name: ansible
  log: true
```

rgarrigue:
Great, I'll test this asap, thanks !--
[...]

PR #1406, opened 2018-07-24
Knob to change Ansible `no_log`

retr0h:
Provisioner will take a log option to enable/disable Ansible
log output.

Fixes: #1346

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -3,7 +3,7 @@
   hosts: localhost
   connection: local
   gather_facts: false
-  no_log: "{{ not lookup('env', 'MOLECULE_DEBUG') | bool }}"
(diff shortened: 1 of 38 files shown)


molecule/provisioner/ansible/playbooks/docker/create.yml
```diff
@@ -3,7 +3,7 @@
   hosts: localhost
   connection: local
   gather_facts: false
-  no_log: "{{ not lookup('env', 'MOLECULE_DEBUG') | bool }}"
+  no_log: "{{ not (lookup('env', 'MOLECULE_DEBUG') | bool or molecule_yml.provisioner.log|default(false) | bool) }}"
```
Rickkwa: If `provisioner.log` is set to `true` in `molecule.yml`, then I should expect `no_log` to be `false`. However, `no_log` is being evaluated to `true`  here.

PR #1417, opened 2018-08-03
Added inspec download for Ubuntu 14.04

loadstar81:
Simple addition to the inspec verifier for Ubuntu 14.04 Trusty


retr0h:
Thanks @loadstar81 \m/

espoelstra:
@loadstar81 would it make sense to (ab)use the Omnitruck.chef.io installer URL and allow passing versions via additional facts like "-v 2.2.61" or leaving it off for the `stable` or `current` channels?

The [documentation](https://docs.chef.io/install_omnibus.html) explicitly mentions the `chefdk`, but it works for `chef` and most other products they have, and it has platform auto-detection built into their `install.sh` script already.

`curl https://omnitruck.chef.io/install.sh | sudo bash -s -- -c stable -P inspec`

PR #1418, opened 2018-08-04
Add git to docker DIND container

jamesla:
Ansible-galaxy can require git to be installed to pull down certain dependencies


retr0h:
Thanks @jamesla

Issue #1419, opened 2018-08-04
Problems with testing on Arch Linux


skhoroshavin:
# Issue Type
- Bug report

# Details

```
$ ansible --version
ansible 2.6.2
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/xop/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/xop/.venv/molecule/lib/python3.6/site-packages/ansible
  executable location = /home/xop/.venv/molecule/bin/ansible
  python version = 3.6.6 (default, Jun 27 2018, 13:11:40) [GCC 8.1.1 20180531]
```

```
$ tox --version
3.1.3 imported from /home/xop/.venv/molecule/lib/python3.6/site-packages/tox/__init__.py
```

Ansible installation method:
- pip

# Desired Behavior

Cloned molecule repository and followed [instructions](http://molecule.readthedocs.io/en/latest/testing.html) to test it using tox:
- tried to install requirements using `pip install -r test-requirements.txt -t requirements.txt`, expected no errors
- tried to run `tox` in order to see if everything is good, expected no errors
- tried to run `tox -t unit`, expected no errors

# Actual Behaviour

- when tried to install requirements there was an error
```
azure-cli-core 2.0.35 has requirement wheel==0.30.0, but you'll have wheel 0.31.1 which is incompatible.
```
I was able to fix it by pinning wheel version in `test-requirements.txt`

- when tried to run `tox` there were lots of errors in functional tests, most are related to Azure and some to rubocop (which isn't installed and there was no mention of it in instructions), although unit tests passed just fine

- when tried to run `tox -t unit` it said there's no such option, and when running `tox --help` there's really no such version

# Questions
- is it expected that Azure functional tests fail?
- should I provide some more info on Azure test failures or it is already known problem?
- is some older version of tox is expected for tests?
- are there any plans to update documentation?

retr0h:
> is it expected that Azure functional tests fail?

Yes, b/c you need Azure credentials.  So you would to source your Azure creds for it to work.

> is some older version of tox is expected for tests?

No

> are there any plans to update documentation?

With what?

> when tried to run tox -t unit it said there's no such option, and when running tox --help there's really no such version

You should prob read the [testing docs](http://molecule.readthedocs.io/en/latest/testing.html).  You need to install tox tags to make it work.

retr0h:
[...]

PR #1420, opened 2018-08-04
Pin wheel version to 0.30.0

skhoroshavin:
This PR is related to #1419

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -15,5 +15,5 @@ python-vagrant
 # The error was: KeyError: 'created_at'
 shade==1.22.2
 twine
-wheel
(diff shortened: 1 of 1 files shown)


skhoroshavin:
Tried `pip install -r test-requirements.txt -r requirements.txt` with Ubuntu 18.04/Python3, Ubuntu 16.04/Python3, Ubuntu 16.04/Python2 and it failed with `azure-cli-core 2.0.35 has requirement wheel==0.30.0, but you'll have wheel 0.31.1 which is incompatible.` in all three cases. So it's definitely not an Arch issue, and can cause problems for anyone starting from clean environment.

PR #1422, opened 2018-08-05
Support passing arbitrary keys to vm.network 

skhoroshavin:
I wanted to be sure that `virtualbox__intnet` is really passed to Vagrant, so I added additional network to vagrant driver scenario and updated tests (and checked that they actually pass after implementing feature). Also I think that having an example of how to add VirtualBox internal network might be useful. Still, if you think these changes are too VirtualBox-specific and not appropriate I can just remove related commit.

Another way to test this might be just removing `auto_config: true` from test scenario since it's default in Vagrant and with generic key passing it's no longer necessary to pass it explicitly. I can add these modifications if you think it's appropriate.

Also when I started working with clean codebase I still got some functional tests failing, so I fixed them in my first commit. The problem was with output lines ordering, sometimes it was different from test expectations.

### DIFF
--- a/test/scenarios/driver/vagrant/molecule/default/molecule.yml
+++ b/test/scenarios/driver/vagrant/molecule/default/molecule.yml
@@ -16,6 +16,9 @@ platforms:
       - auto_config: true
         network_name: private_network
         type: dhcp
+      - virtualbox__intnet: test_network
+        network_name: private_network
+        ip: 192.168.0.1
     provision: false
 provisioner:
   name: ansible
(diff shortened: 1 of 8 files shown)


test/scenarios/driver/vagrant/molecule/default/molecule.yml
```diff
@@ -16,6 +16,9 @@ platforms:
       - auto_config: true
         network_name: private_network
         type: dhcp
+      - virtualbox__intnet: test_network
```
retr0h: Why is it called `virtualbox__intnet`?

skhoroshavin: Because it is called so in vagrant: https://www.vagrantup.com/docs/virtualbox/networking.html

PR #1435, opened 2018-08-13
Adds tox-tags to test-requirements.txt

wilmardo:
As the title states, adds tox-tags to test-requirements.txt. Uses the "editable" form ([pip docs](https://pip.readthedocs.io/en/1.1/requirements.html#requirements-file-format)) in the test-requirements.txt and my source on [SO](https://stackoverflow.com/questions/16584552/how-to-state-in-requirements-txt-a-direct-github-source).

I could pin the latest commit of tox-tags in the requirements to prevent breakage if tox-tags gets updated?

### DIFF
--- a/doc/source/testing.rst
+++ b/doc/source/testing.rst
@@ -18,10 +18,6 @@ Install the test framework `Tox`_.
 
     $ pip install tox
 
-Install `tox-tags`_.
-
-.. _`tox-tags`: https://github.com/AndreLouisCaron/tox-tags
-
 Install the remaining requirements in a venv (optional).
 
 .. code-block:: bash
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -17,3 +17,4 @@ shade==1.22.2
 twine
 wheel==0.30.0
 yapf==0.21.0
+-e git://github.com/AndreLouisCaron/tox-tags@master#egg=tox-tags


retr0h:
rad \m/

retr0h:
@wilmardo actually, I need to revert this.  Adding tox-tags to the test environment doesn't accomplish what we want b/c tox is installed outside of the tox virtual environments.  Tox tags needs installed in the py env running tox not the inner one.

wilmardo:
@retr0h Are you sure about this? I just tested it again on a fresh LXC Xenial container:

```
apt-get update && apt-get install -y python-pip virtualenv git ruby ruby-dev
git clone https://github.com/metacloud/molecule && cd molecule
pip install tox
virtualenv venv && source venv/bin/activate
pip install -r test-requirements.txt -r requirements.txt
```
After this I ran `tox -t unit`and it seems to work as expected.

Besides this, are you open for a PR implementing Travis CI testing for Molecule? It integrates really well into PR's like this to see if things are still working as expected. I did not start yet but it does not seem that hard since Tox is used.

retr0h:
> pip install -r test-requirements.txt -r requirements.txt

Ok, ya I think I am good with this.  That makes sense.

lwm:
> Besides this, are you open for a PR implementing Travis CI testing for Molecule?

Please see https://github.com/metacloud/molecule/issues/1350.

PR #1447, opened 2018-08-17
Adds pre_build_image option to Docker create playbook

wilmardo:
As the title states adds a `pre_build_image`option to the Docker API. 
When user want to user their own prebuild images instead of running the default Molecule Dockerfile.j2 build. When set to true the `image:` get passed to docker_container without the `molecule_local` prefix.

In my case I use my own images which use a script to simulate systemd in the container ([repo](https://github.com/LANsible/docker-base-ansible)). This allows me to skip all the build steps.


retr0h:
@wilmardo thanks!

PR #1461, opened 2018-08-30
1415 - Handle long interpreter directives in external programs

### LINKED ISSUE #1415: Molecule shouldn't call external programs directly to avoid shebang limitations
# Issue Type

- Bug report

```
$ ansible --version
ansible 2.6.2
```

```
$ molecule --version
molecule, version 2.16.0
```

Molecule installation method:

- pip

Ansible installation method:

- pip

# Desired Behavior

Molecule should work as expected even with virtual envs nested deeply in directory tree.

# Actual Behaviour

The length of shebangs in Linux is limited by `BINPRM_BUF_SIZE` set in `include/linux/binfmts.h`. By default `BINPRM_BUF_SIZE=128`. This value can't be changed without recompiling the kernel.

We currently run molecule with [Tox](https://github.com/tox-dev/tox) and Jenkins and sometimes, when path to some executable (i.e. ansible-playbook) becomes too long, Molecule fails. For example, here is an excerpt from Molecule's log running in our typical environment: 

```
--> Scenario: 'default'
--> Action: 'syntax'
<REDACTED>
DEBUG: COMMAND

/opt/jenkins/workspace/Molecule/test_role/roles/some_longer_than_usual_role_name/.tox/some_longer_than_usual_role_name-current/bin/ansible-playbook --skip-tags=molecule-notest,notest --diff --inventory=/tmp/longer_than_usual_role_name-current/ansible_inventory.yml --syntax-check /opt/jenkins/workspace/Molecule/test_role/roles/some_longer_than_usual_role_name/molecule/default/playbook.yml -vvv

ERROR: 

An error occurred during the test sequence action: 'syntax'. Cleaning up.
```

Executable `/opt/jenkins/workspace/Molecule/test_role/roles/some_longer_than_usual_role_name/.tox/some_longer_than_usual_role_name-current/bin/ansible-playbook` has rather long shebang:
```
#!/opt/jenkins/workspace/Molecule/test_role/roles/some_longer_than_usual_role_name/.tox/some_longer_than_usual_role_name-current/bin/python
```

Its length is 137 which obviously exceeds BINPRM_BUF_SIZE limitation mentioned above.

I see several possible **workarounds** for this problem:

* **Do not use tox and venvs.** This kills one of the most cool things about Molecule: the ability to test roles in different environments against multiple versions of Ansible.
* **Do not use long roles names and/or long tox env names.** We've been trying to do this for a while but it's not always possible and it's incredibly inconvenient.
* **Do not call external executables directly**, instead parse shebang (if said executable has one) and call the interpreter explicitly. Actually Tox took this exact approach some time ago:
    * https://github.com/tox-dev/tox/issues/66
    * https://github.com/tox-dev/tox/blob/46fca4c3c817f0a4728552e99e89ace9b2cc6234/tox/venv.py#L420

I've [tried implementing](https://github.com/metacloud/molecule/compare/master...kdask:shebang-limitations-workaround) this last workaround and it does seem to solve the problem. I don't know how I feel about this implementation though: looks a bit hacky to me, but, well :) Tell me what you think!

kdask:
Fixes #1415 

I'm not sure about how to test it, though. Any suggestions?

### DIFF
--- a/molecule/util.py
+++ b/molecule/util.py
@@ -87,6 +87,46 @@ def sysexit_with_message(msg, code=1):
     sysexit(code)
 
 
+def prepend_shebang_interpreter(cmd):
+    """
+    Inspired by tox:
+    https://github.com/tox-dev/tox/blob/46fca4c3c817f0a4728552e99e89ace9b2cc6234/tox/venv.py#L420
+
+    prepend interpreter directive (if any) to the provided `sh.Command` object
+
+    When preparing virtual environments in a file container which has large
+    length, the system might not be able to invoke shebang scripts which
+    define interpreters beyond system limits (e.x. Linux as a limit of 128;
+    BINPRM_BUF_SIZE).
+    This method can be used to check if the executable is
+    a script containing a shebang line. If so, extract the interpreter (and
+    possible optional argument) and prepend the values to the provided ``sh``
+    command. Molecule will only attempt to read an interpreter directive of
+    a maximum size of 2048 bytes to limit excessive reading and support UNIX
+    systems which may support a longer interpret length.
+
+    :param cmd: a `sh.Command` object
+    :return: a `sh.Command` object
+
+    """
+    cmd_path = cmd._path
+    try:
+        with open(cmd_path, 'rb') as f:
+            if f.read(1) == b'#' and f.read(1) == b'!':
+                cmd._partial_baked_args.insert(0, cmd_path)
+                MAXINTERP = 2048
+                interp = f.readline(MAXINTERP).rstrip()
+                interp_args = interp.split(None, 1)[:2]
+                cmd._path = interp_args[0]
+                baked_args = cmd._partial_baked_args
+                cmd._partial_baked_args = interp_args[1:] + baked_args
+                return cmd
+    except IOError:
+        pass
+
+    return cmd
+
+
 def run_command(cmd, debug=False):
     """
     Execute the given command and returns None.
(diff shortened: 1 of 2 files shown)


retr0h:
fyi - this isn't something I have a good way of testing.  I'm not really inclined to merge this without a way to reproduce it through unit tests to verify it actually worked.  Also, the code submitted lowers the over all code coverage, since it also does not contain unit tests.

retr0h:
I'm going to hold off from merging this.  As you know [Ansible is adopting Molecule](https://groups.google.com/forum/#!topic/ansible-project/ehrb6AEptzA), therefore I do not want to merge anything that may impact future architectural decisions, as I will no longer be leading the development of this project.

gundalow:
Once the repo has been moved under Ansible we can look at getting this merged.

In the short term we welcome testing and feedback on this PR.

Thanks for your understanding.

gundalow:
@kdask thank you for the PR. Looks like for this to progress we need to work out a way to test it, any ideas?

molecule/util.py
```diff
@@ -95,6 +135,8 @@ def run_command(cmd, debug=False):
     :param debug: An optional bool to toggle debug output.
     :return: ``sh`` object
     """
+    if 'MOLECULE_LIMITED_SHEBANG' in os.environ:
```
webknjaz: maybe booleanize this?

`to_bool(os.getenv('MOLECULE_LIMITED_SHEBANG', False))`

themr0c:
close/repopren to trigger travis build.

ssbarnea:
Now that I look better a the code I realise that it is totally outside molecule scope to fix broken shebangs lines in scripts not produced by it.

Molecule is not the place to address this issue.

Issue #1456, opened 2018-08-27
Support restart_policy option for docker driver


borisnaydis:
# Issue Type

- Feature request

# Molecule and Ansible details

```
ansible --version
```

```
molecule --version
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

# Desired Behavior

When `restart_policy` is set for a platform instance, it should be propagated to a corresponding container. The reason for adding this functionality is to test reboot action with subsequent reconnect in Ansible playbooks.

# Actual Behaviour (Bug report only)

Now this setting is not defined in result containers because it is not defined in [create.yml](https://github.com/metacloud/molecule/blob/master/molecule/provisioner/ansible/playbooks/docker/create.yml#L57).

retr0h:
Please submit a PR for this feature.  You can reference #1413.

borisnaydis:
Ok, I'll try to find time for this during this week.

PR #1465, opened 2018-09-03
Add support of restart_policy and restart_retries to docker driver

borisnaydis:
Fixes: #1456

### DIFF
--- a/test/unit/model/v2/test_platforms_section.py
+++ b/test/unit/model/v2/test_platforms_section.py
@@ -82,6 +82,10 @@ def _model_platforms_docker_section_data():
             'env': {
                 'FOO': 'bar',
             },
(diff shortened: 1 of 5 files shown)


test/unit/model/v2/test_platforms_section.py
```diff
@@ -149,6 +153,8 @@ def _model_platforms_docker_errors_section_data():
                 int(),
             ],
             'env': str(),
+            'restart_policy': str(),
+            'restart_retries': int(),
```
retr0h: These should be invalid options so the test fails and we capture it failed as you did below.

borisnaydis: Done.

PR #1467, opened 2018-09-04
Updates tox-tags url in test-requirements.txt

wilmardo:
Worked behind a strict firewall today and the previous approach wasn't working. So dug around and found out a https:// approach is also an option. I think a better approach since this is never blocked on a firewall and even faster since the repo does not get cloned ([source](https://stackoverflow.com/a/24811490))

Another option which does clone the repository:
`git+https://github.com/AndreLouisCaron/tox-tags.git@master`


retr0h:
Seems reasonable.

PR #1468, opened 2018-09-04
Adds install instructions for RuboCop

wilmardo:
Adds install instructions for RuboCop to documentation as mentioned in #1419.

Might be better to rephrase `For some tests` to a more specific context?


retr0h:
Thanks as always @wilmardo \m/

PR #1473, opened 2018-09-13
WIP: Add VMware driver

bandit145:
I've been working on a VMware driver and it is functional and works right now. I'm just working on adding it to the unit/functional tests and updating docs.

One thing I'm not sure of is if you guys are interested is how to add functional tests unless you have some VMware infra sitting around.

The reason I'm even attempting to add an on-prem provider is that I work in a secure environment where we can't utilize Azure/AWS instances, even for testing. This would allow people in these sorts of environments to still leverage molecule.


retr0h:
Hi @bandit145, I am hesitant to commit drivers which rely on community members maintaining drivers which most people do not use.  To be honest, I think we already have too many drivers.  Most people simply use Vagrant and Docker.  Very few people use the others, and I don't actively test the others.

 I do appreciate you're taking the time to try and implement this, but at this time it's more maintenance than it's prob worth.

bandit145:
That's understandable but thanks for taking a look anyway.

lwm:
Regarding community drivers, I opened https://github.com/metacloud/molecule/issues/1479!

gundalow:
Hi,
As detailed in https://groups.google.com/forum/#!topic/molecule-users/O1a_YNAjy-Q `Ansible to adopt molecule and ansible-lint projects`.

I'm in the process of doing the technical move. Feel free to reach out in `#molecule-users` for more info

PR #1474, opened 2018-09-13
Docs: Recommend prepare playbook for node setup

### LINKED ISSUE #1472: Docs: Confusing advice on create vs converge playbooks
In the [docs for the Ansible provisioner](https://molecule.readthedocs.io/en/latest/configuration.html#id12) it says (emphasis mine):
> Reserve the create and destroy playbooks for provisioning. Do not attempt to gather facts or perform operations on the provisioned nodes inside these playbooks. Due to the gymnastics necessary to sync state between Ansible and Molecule, it is best to perform these tasks in the **converge** playbook.

Shouldn't it say
>... it is best to perform these tasks in the **prepare** playbook.

here instead? Or at least "prepare **or** converge playbooks"? See also #1351.

From what I understand, *converge* is meant for the actual execution of the role, isn't it? Of course, the cited paragraph makes sense if that's what "perform operations on the provisioned nodes" refers to, but then I don't think anyone would think to put this into *create*... confusion between *create* and *prepare*, on the other hand, seems more likely.

smheidrich:
Fixes #1472 (rationale for this PR is described in the issue so I won't repeat it here).

### DIFF
--- a/molecule/provisioner/ansible.py
+++ b/molecule/provisioner/ansible.py
@@ -50,7 +50,7 @@ class Ansible(base.Base):
         attempt to gather facts or perform operations on the provisioned nodes
         inside these playbooks.  Due to the gymnastics necessary to sync state
         between Ansible and Molecule, it is best to perform these tasks in the
-        converge playbook.
+        prepare or converge playbooks.
 
         It is the developers responsiblity to properly map the modules's fact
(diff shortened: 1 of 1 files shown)


retr0h:
Thanks @smheidrich \m/

smheidrich:
Thanks for merging :+1:

PR #1487, opened 2018-09-20
cleanup of pip packages list and proposal for additional packages

themr0c:
Hello, this is a proposal for some changes in the list of pip packages. I believe it makes sense to integrate them to the official molecule image. 

First some cleanup:

Removal of ansible from the list: 

calling `pip install ansible[azure] ansible` makes no sense, only first occurence `ansible[azure]` will be installed, but as it implies `ansible` that's fine. For the record `pip install ansible ansible[azure]` does not install `ansible[azure]`, but only `ansible`.

Other removals: 

ansible[azure] implies packaging, msrestazure, azure-cli.

Additions:

ansible-lint-junit
detox
junit-xml
tox

As a maintainer of a repository of playbooks containing multiple ansible roles, I am using detox and tox to automate and parallelize the run of molecule test. 
I am also using ansible-lint-junit and junit-xml to let the build system generate nice reports. 

I really think that these packages are a nice fit for the official molecule docker image, as they are expanding the usecases. 

Here is a sample tox.ini to illustrate the usecase:

```
# Invocation:
# Run all tests in parallel (where possible):
#   detox
# Run all tests sequentially:
#   tox
# Run tests sequentially for specified roles:
#   tox -e some_role,additional_role
[tox]
# Each role is defined as an environment, to enable parallel runs of command 'molecule test'
envlist =
  some_role
  another_role
  playbooks
minversion = 3.2.1
skipsdist = true
[testenv]
basepython = python2
commands = bash -c "(cd {toxinidir}/roles/{envname} && molecule --debug test --all)"
description = molecule test on {envname}
passenv = *
setenv =
  # necessary for parallelization with detox
  # see https://molecule.readthedocs.io/en/latest/testing.html
  MOLECULE_EPHEMERAL_DIRECTORY={envname}
sitepackages = true
whitelist_externals =
  /bin/bash
  /usr/bin/rubocop
[testenv:playbooks]
deps = ansible-lint-junit
description = ansible-lint on all playbooks
commands = bash -c "find {toxinidir} -maxdepth 1 -name '*.yml' | xargs -n1 ansible-lint -v"
whitelist_externals =
  /bin/bash
  /usr/bin/find
  /usr/bin/xargs
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -28,14 +28,13 @@ ENV PACKAGES="\
 
 ENV PIP_PACKAGES="\
     ansible[azure] \
-    virtualenv \
-    molecule \
-    ansible \
+    ansible-lint-junit \
+    detox \
     docker \
-    azure-cli \
-    packaging \
-    msrestazure \
+    junit-xml \
+    molecule \
     pywinrm \
+    virtualenv \
 "
 
 ENV GEM_PACKAGES="\


themr0c:
Indeed, but if I don't ask, I will not know if it makes sense for other people :)

retr0h:
> And yet others will have no use for these packages, right? If you've seen how many issues get closed around here (😅)

awwww :sad

lwm:
lol, apologies if I am being a dick :boom: . You're right @themr0c!

retr0h:
I'm not so sure what I think about this.  On one hand I think it's fine, on the other hand it seems a bit unnecessary, when this can be solved with your own Dockerfile.  I'll look this over more this week.  My apologies for the delay.

themr0c:
I am not completely sure either :).
Key question is: does the use case make sense to other than ours ?

Maybe the tox and the junit cases should not be discussed in the same PR?

For tox: is the user story convincing and in the scope of molecule?

Given a repository containing multiple roles, performing CI tests on the repository implies the need to orchestrate multiple molecule tests.
Should this orchestration task stay out of the scope of molecule ? 
Or is this idea to use (de)tox generic (and simple) enough to become the orchestrator of multiple molecule tests ?

For junit, I have more afterthoughts. 

It is less necessary: once you have tox installed, it is very simple to add dependencies in the tox.ini configuration file.

It is maybe not the right place to implement it. If we want to make the molecule image feature complete for all possible ansible scenarios, we are missing also other packages, and the list can grow infinitely. For instance python-consul would be necessary to use consul_kv module and lookup plugin. Maybe the management of additional dependencies should be implemented at ansible level, not in molecule docker image dependencies: we should be able to ask for ansible[azure,consul,docker], etc.

retr0h:
I'm going to hold off from merging this.  As you know [Ansible is adopting Molecule](https://groups.google.com/forum/#!topic/ansible-project/ehrb6AEptzA), therefore I do not want to merge anything that may impact future architectural decisions, as I will no longer be leading the development of this project.

themr0c:
Fine for me :).

Versions on pip packages is a great idea. 

BTW in our (private) pipeline using molecule image, we have oped in for versioned pip packages in an external `requirement.txt` that is suitable for all scenarios (automation pipelines, docker build, detox scenarios, user installation...). It could be a better option than this actual proposal.

webknjaz:
@themr0c in general, we discuss high-level proposals under ansible/proposals repo, but for more specific or straightforward things sometimes issues in projects work.

We're thinking already about having extras, I've got a WIP PR and it's been mentioned across quite a few issues/PRs recently.

As for pinning I've explained my position @ https://github.com/ansible/molecule/pull/1530#issuecomment-436219312

Also, please be familiar with https://caremad.io/posts/2013/07/setup-vs-requirement/, which describes difference between distribution dependencies and env requirements. (TL;DR pinning env deps is up to user who installs dists and distributions themselves should only specify lower bounds if necessary)

themr0c:
I am closing this PR as with the current changes it is becoming irrelevant: detox has become obsolete, the proposed implementation is not going in the right direction.

I'll maybe come back later with a more appropriate request, which will be more in the direction of  ongoing work on extras.

As tox is given as an example for ci, it will make sense to have tox as an extra  https://molecule.readthedocs.io/en/latest/testing.html#tox

For junit output, we should maybe have a discussion on that topic later, once current PR and issues ile has decreased.

PR #1489, opened 2018-09-26
Bumped testinfra to 1.16.0 due to testinfra bug.

camillehuot:
This is a request to bump testinfra to 1.16.0 because they've fixed PATH issue that prevent using some modules (like sysctl) on Red Hat-type systems.

This change has passed all Tox tests (except some rubocop because I didn't installed it)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,11 @@
 History
 *******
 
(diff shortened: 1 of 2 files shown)


retr0h:
Thanks @camillehuot

PR #1500, opened 2018-10-08
Fix delegate

rahul0705:
The delegate role had a bug in the template. Corrected said bug.

Also updated the delegate docs to show how to pass connection information to ansible.

Added new blurb about how to use localhost as a connection.


lwm:
LGTM :+1: :+1: :+1:

gundalow:
@rahul0705 Thank you for this PR and docs update.

Does the current code function at all. I wonder if this is something we need tests adding for?

rahul0705:
Curious which release this will be added in? or do I need to add additional tests?

lwm:
Think you're good to go. AFAIK, check https://github.com/ansible/molecule/pull/1594 for a temporary stop gap for releases right now. Not sure when next full point release will be out.

gundalow:
@rahul0705 thank you for the fix.
@lwm @themr0c Thank you for the reviews.

PR #1501, opened 2018-10-08
Fix init

rahul0705:
the documentation for template does not make it clear that you can provide a path to a local cookiecutter template directory

also fixed a minor typo.

### DIFF
--- a/molecule/command/init/scenario.py
+++ b/molecule/command/init/scenario.py
@@ -43,7 +43,7 @@ class Scenario(base.Base):
 
     .. option:: cd foo; molecule init scenario --scenario-name bar --role-name foo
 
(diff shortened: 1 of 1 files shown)


lwm:
This is a breaking change on the CLI front-end, right? We probably want to avoid that. Personally, I agree that the `--template` is a better name for this option since we can also pass local paths but could we support both and output a deprecation warning for the `--url` option name is used? This would be better for the users. Your commits are just not for docs, you change functionality, so the commit message should also probably change.

retr0h:
Yeah, unfortunately this is a breaking change.  I'm not ready to accept such a PR.  A local URL is still a valid URL.

PR #1507, opened 2018-10-10
Add provisioner.dockerfile configuration

### LINKED ISSUE #1502: Configuration of Dockerfile.j2 location
# Issue Type

- Feature request

```
ansible 2.7.0
```

```
molecule, version 2.18.1
```

- pip

- pip

# Desired Behavior

I'd like to share the `Dockerfile.j2` file across scenarios / roles, like [defined in the docs here](https://molecule.readthedocs.io/en/latest/examples.html#sharing-across-scenarios)

I'm new to molecule, so help me if I misunderstand. Currently the default `docker/create.yml` does not allow you to change the location of the `Dockerfile.j2` (it is hard coded to the scenario directory, [here](https://github.com/metacloud/molecule/blob/25a9ee3ee0cd2eb880c94c9400b926686e75cab6/molecule/provisioner/ansible/playbooks/docker/create.yml#L23)), so in order to support this I'd need to use symlinks or my own custom `create.yml`, both I'd like to avoid.

I'm not sure the best place to define where to load the Dockerfile.j2 from, but with input on the design I can submit a pull request. Two options would be under `platforms`, or under `provisioner`.

```yaml
platforms:
  - name: instance
    image: debian:stable
    dockerfile: ../resources/Dockerfile.j2
```

```yaml
provisioner:
  name: ansible
  dockerfile: ../resources/Dockerfile.j2
```

Please let me know if I'm missing anything, or if what I'm trying to accomplish can be done another way. If wanted, I can submit a pull request. Thank you!

plasticrake:
Fixes #1502 

I wanted to add a test, but I wasn't sure exactly where and how to test this. I did run tox and also tested my own roles against this version with no issues.

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -20,7 +20,7 @@
 
     - name: Create Dockerfiles from image names
       template:
-        src: "{{ molecule_scenario_directory }}/Dockerfile.j2"
+        src: "{{ molecule_scenario_directory }}/{{ molecule_yml.provisioner.dockerfile | default('Dockerfile.j2') }}"
         dest: "{{ molecule_ephemeral_directory }}/Dockerfile_{{ item.image | regex_replace('[^a-zA-Z0-9_]', '_') }}"
       with_items: "{{ molecule_yml.platforms }}"
       when: not item.pre_build_image | default(false)


molecule/provisioner/ansible/playbooks/docker/create.yml
```diff
@@ -20,7 +20,7 @@
 
     - name: Create Dockerfiles from image names
       template:
-        src: "{{ molecule_scenario_directory }}/Dockerfile.j2"
+        src: "{{ molecule_scenario_directory }}/{{ molecule_yml.provisioner.dockerfile | default('Dockerfile.j2') }}"
```
retr0h: We should retain the default behavior here.

```
src: molecule_yml.provisioner.dockerfile | default("{{ molecule_scenario_directory }}/Dockerfile.j2")
```

retr0h:
Also, can you update `test/resources/playbooks/docker/create.yml`.

molecule/provisioner/ansible/playbooks/docker/create.yml
plasticrake: I actually had it that way first. The issue with that is the  search path is under `.../molecule/provisioner/ansible/playbooks/docker/` (relative to the default create.yml) so if you just specify `Dockerfile2.j2` for example it wouldn't work. The way I have it, it is relative to the scenario path so you can have `Dockerfile2.j2` in your scenario directory and just refer to it that way.

The default is already there to preserve the existing default behavior.

plasticrake:
`test/resources/playbooks/docker/create.yml` was set to:
```yaml
Dockerfile.j2
```

`provisioner/ansible/playbooks/docker/create.yml` was set to:
```yaml
{{ molecule_scenario_directory }}/Dockerfile.j2
```

In order to preserve the differing default behavior the updates in each file will be different. My intention is to **preserve the existing default behavior** in both files and allow a custom Dockerfile to be specified **relative to the scenario directory**.

For `test/resources/playbooks/docker/create.yml` it would look something like this, do you have any preference between them or did you have something else in mind?
```yaml
src: "{{ molecule_scenario_directory ~ '/' ~ molecule_yml.provisioner.dockerfile if molecule_yml.provisioner.dockerfile else 'Dockerfile.j2' }}"
```
```yaml
src: "{{ molecule_yml.provisioner.dockerfile | ternary(molecule_scenario_directory ~ '/' ~ molecule_yml.provisioner.dockerfile, 'Dockerfile.j2') }}"
```


Alternatively the `Dockerfile.j2` that is now generated by `molecule init role` could be placed here: `provisioner/ansible/playbooks/docker/` and used as the default `Dockerfile.j2` without having to place default `Dockerfile.j2`'s  in every role. Just a thought.

themr0c:
close/repopren to trigger travis build.

decentral1se:
We've got something cooking for this on https://github.com/ansible/molecule/pull/1769.

PR #1510, opened 2018-10-15
only colorize output to tty

pahoughton:
The escape sequences make the output hard to read when output to a file and when shown on a white background.

### DIFF
--- a/molecule/logger.py
+++ b/molecule/logger.py
@@ -171,4 +171,7 @@ def cyan_text(msg):
 
 
 def color_text(color, msg):
-    return '{}{}{}'.format(color, msg, colorama.Style.RESET_ALL)
+    if sys.stdout.isatty():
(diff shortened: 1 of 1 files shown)


gundalow:
Is there a standard ENV var that Travis/Shippable/etc use for this

ssbarnea:
Not as far as I know but I would recommend re-using Ansible logic from https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/color.py

I see no harm in loading the same environment variable.

For Travis just defined it using the env section.

gundalow:
OK, so we will need the docs updating as well.

themr0c:
close/reopen to trigger travis build.

webknjaz:
I think we should close this in favor of #1533 which goes after a more resilient approach.

decentral1se:
Closing based on https://github.com/ansible/molecule/pull/1510#issuecomment-468672712.

Feel free to re-open / further discuss!

PR #1530, opened 2018-10-30
Bump ansible-lint to 3.5.1

### LINKED ISSUE #1525: ansible-lint version needs bumped to 3.5.1
# Issue Type
- Feature request

In the currently referenced version of `ansible-lint`, it will silently not parse tasks files that are included via `import_tasks`/`import_roles`. This bug has since been fixed -- when I manually specify `ansible-lint==3.5.1` in my requirements.txt, the behavior is resolved. It would be nice to get rid of that version pinning in my requirements.txt, and I'm sure users of molecule would appreciate their linter not silently ignoring some of their files :)

I haven't run into any incompatibilities so far.

Here's the PR with the fix https://github.com/willthames/ansible-lint/pull/365

```
ansible --version
```
2.6.1

```
molecule --version
```
2.19.0

- pip

- pip

wilmardo:
Fixes #1525

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-ansible-lint==3.4.23
+ansible-lint==3.5.1
 anyconfig==0.9.7
 cerberus==1.2
 click==6.7


gundalow:
Setting as v2.20 milestone (next release)

webknjaz:
@gundalow automatic periodic version bump PRs can be done by pyup (demo: https://github.com/aio-libs/multidict/pull/300) or dependabot (demo: https://github.com/github/opensource.guide/pull/742)

ssbarnea:
This clearly needs to go in as it fixed a serious bug. I am not sure if pinning of ansible-lint is the best idea but this would deserve a new bug and consultation.

There are pros and cons for both approaches but maybe in the future we could go for a mixed approach: having unpinned versions in requirements.txt and and adding an extra named "pinned" that has the pinned versions. So users that want extra satbility can install `molecule[pinned]` and those that are more flexible can install `molecule`. Obviously that even the non pinned version would need ranged, at least >= if not even capped ones.

The way it was proposed in the tox.ini PR looks like a good traddeoff between safety and maintenance.

webknjaz:
Hey @ssbarnea,

Let me add a few points to the discussion.

1) Let's not confuse environment dependencies and distribution package requirements. It's encouraged to pin deps for env, but not library-like things. The reason is that it breaks user's ability to install things with conflicting deps alongside or to pin them to versions they want within their (virtual)envs, for example.
OTOH setting a lower bound is a good idea and it's fine with me. Another thing we can do is to explicitly forbid certain faulty dependency versions, like `ansible-lint>=x.y.z,!=x.y.a`.
2) I like the idea of having "pinned" extra. Nice thing is that we could do post-releases only updating that pinned dependency (I mean like `.post0`, see PEP 440)

webknjaz:
Hey @wilmardo, please rebase this PR.

wilmardo:
@webknjaz Done :smile:

webknjaz:
It looks like you hit an unstable job. I've restarted it to see what happens :)

PR #1458, opened 2018-08-29
Add Linode driver (v3 API)

lwm:
This is still a work in progress. I've manually tested parts of it but not a full end to end use.

Just wanted to get this up here to see that it is all going in the right direction in case I miss something.

Will close https://github.com/metacloud/molecule/issues/1450.

PR #1533, opened 2018-10-31
Respect PY_COLORS variable

ssbarnea:
Allow PY_COLORS to control coloring support.

Allow user to define PY_COLORS=1 or 0 to force use or not use of
ANSI coloring sequences.

Use of this option overrided default behavior of detecting TTY status
of stdout and allows users to control its use CI systems which
may support ANSI but where there is no TTY.

Implementation based on https://github.com/pytest-dev/py/blob/master/py/_io/terminalwriter.py#L131
which is also used by tox and many other similar tools.

Fixes: #1510

### DIFF
--- a/molecule/logger.py
+++ b/molecule/logger.py
@@ -19,11 +19,22 @@
 #  DEALINGS IN THE SOFTWARE.
 
 import logging
+import os
 import sys
 
 import colorama
+from ansible.module_utils.parsing.convert_bool import boolean as to_bool
 
-colorama.init(autoreset=True)
+
+def should_do_markup():
+    py_colors = os.environ.get('PY_COLORS', None)
+    if py_colors is not None:
+        return to_bool(py_colors, strict=False)
+
+    return sys.stdout.isatty() and os.environ.get('TERM') != 'dumb'
+
+
+colorama.init(autoreset=True, strip=not should_do_markup())
 
 SUCCESS = 100
 OUT = 101
--- a/test/unit/test_logger.py
+++ b/test/unit/test_logger.py
@@ -126,3 +126,29 @@ def test_cyan_text():
     x = '{}{}{}'.format(colorama.Fore.CYAN, 'foo', colorama.Style.RESET_ALL)
 
     assert x == logger.cyan_text('foo')
+
+
+def test_markup_detection_pycolors0(monkeypatch):
+    monkeypatch.setenv('PY_COLORS', '0')
+    assert not logger.should_do_markup()
+
+
+def test_markup_detection_pycolors1(monkeypatch):
+    monkeypatch.setenv('PY_COLORS', '1')
+    assert logger.should_do_markup()
+
+
+def test_markup_detection_tty_yes(mocker):
+    mocker.patch('sys.stdout.isatty', return_value=True)
+    mocker.patch('os.environ', {'TERM': 'xterm'})
+    assert logger.should_do_markup()
+    mocker.resetall()
+    mocker.stopall()
+
+
+def test_markup_detection_tty_no(mocker):
+    mocker.patch('os.environ', {})
+    mocker.patch('sys.stdout.isatty', return_value=False)
+    assert not logger.should_do_markup()
+    mocker.resetall()
+    mocker.stopall()
--- a/doc/source/ci.rst
+++ b/doc/source/ci.rst
@@ -1,6 +1,10 @@
 Continuous integration
 ----------------------
 
+Molecule output will use ``ANSI`` colors if stdout is an interactive TTY and
+``TERM`` value seems to support it. You can define ``PY_COLORS=1`` to force
+use of ``ANSI`` colors, which can be handly for some CI systems.
+
 Travis CI
 ^^^^^^^^^
 
--- a/molecule/util.py
+++ b/molecule/util.py
@@ -30,12 +30,10 @@ import anyconfig
 import colorama
 import yaml
 
-from molecule import logger
+from molecule.logger import get_logger
 
-LOG = logger.get_logger(__name__)
+LOG = get_logger(__name__)
 MERGE_STRATEGY = anyconfig.MS_DICTS
-
-colorama.init(autoreset=True)
 
 
 class SafeDumper(yaml.SafeDumper):


ssbarnea:
I don't know an easy way to test it for CI but you can easily run `PY_COLOR=0 molecule test`, also try with `PY_COLOR=1 molecule test &>mo.log` and it should be enough.

Because ``--help`` does not have any coloring is not a good way of testing it. I will try to rebase my change to get rid of the unrelated change and bring it closer to mergeable.

ssbarnea:
BTW, do you know a way to make the signing automated on git, the DCO stuff is a real PITA.

gundalow:
@ssbarnea https://stackoverflow.com/questions/41881722/how-can-i-force-git-commit-s-using-git-commit-command

molecule/__main__.py
```diff
@@ -0,0 +1,24 @@
+#  Copyright (c) 2015-2018 Cisco Systems, Inc.
```
gundalow: Not sure if this change should be in here, I think this is from your other PR

molecule/logger.py
```diff
@@ -19,11 +19,27 @@
...
+    if os.environ.get('PY_COLORS') == '1':
+        return True
+    if os.environ.get('PY_COLORS') == '0':
+        return False
+    return hasattr(file, 'isatty') and file.isatty() and \
+        os.environ.get('TERM') != 'dumb' \
+        and not (sys.platform.startswith('java') and os._name == 'nt')
+
+
+def colorama_init():
```
lwm: Can we write a unit test directly against `colorama_init` and use https://docs.pytest.org/en/latest/reference.html#_pytest.monkeypatch.MonkeyPatch to mock the necessary environment values? You can also mock `file.isatty` to be sure also.

themr0c:
I believe we can move forward now, what you thing @lwm and @gundalow ?

webknjaz:
@decentral1se did you register @lwm as a backup account? if yes, you might want to add a status/bio there pointing to your new handle...

molecule/logger.py
```diff
@@ -19,11 +19,27 @@
...
-colorama.init(autoreset=True)
+
+def should_do_markup(file=sys.stdout):
+    if os.environ.get('PY_COLORS') == '1':
+        return True
+    if os.environ.get('PY_COLORS') == '0':
+        return False
+    return hasattr(file, 'isatty') and file.isatty() and \
+        os.environ.get('TERM') != 'dumb' \
+        and not (sys.platform.startswith('java') and os._name == 'nt')
```
webknjaz: Wow! Anyone runs in under jython?

molecule/logger.py
```diff
@@ -19,11 +19,27 @@
...
 import logging
+import os
 import sys
 
 import colorama
 
-colorama.init(autoreset=True)
+
+def should_do_markup(file=sys.stdout):
+    if os.environ.get('PY_COLORS') == '1':
```
webknjaz: There's a prior art on boolean detection you may want to reuse:
```python
from ansible.module_utils.parsing.convert_bool import boolean as to_bool
...
is_py_colors_enabled = to_bool(os.environ.get('PY_COLORS'), strict=False)
```

* https://github.com/ansible/ansible/blob/c37f6bed1c44f7d6ef854baec4c5e04a037c048f/lib/ansible/module_utils/parsing/convert_bool.py
* https://github.com/ansible/ansible/blob/c37f6bed1c44f7d6ef854baec4c5e04a037c048f/lib/ansible/plugins/filter/core.py#L89

molecule/logger.py
```diff
@@ -19,11 +19,27 @@
...
+        return True
+    if os.environ.get('PY_COLORS') == '0':
+        return False
+    return hasattr(file, 'isatty') and file.isatty() and \
```
webknjaz: It's more pythonic to use brackets for multiline expressions, using EOL escaping is discouraged in typical pythonic style guides.

Also, why do you need to check whether `file.isatty()` method is present? Isn't that guaranteed?

molecule/logger.py
webknjaz: > os._name

I'd recommend avoiding accessing private attributes. Also, there's a `platform` stdlib module for this kind of env detection.


It's better to go for
```suggestion
        and platform.system() not in ('Java', 'Windows')
```

(of course `import platform` somewhere in the module header)

molecule/logger.py
webknjaz: After looking closer I realized that you use `file()` builtin which is a **deprecated alias** for `open()`. So you basically ignore this check under Python 3 and always disable colors under it, which is probably not what you want.

Moreover, `file` itself doesn't really point to any io stream: you should be checking stdout stream.

Here's what you should be doing instead:
```suggestion
    return sys.stdout.isatty() and \
```

test/unit/test_logger.py
```diff
@@ -126,3 +126,28 @@ def test_cyan_text():
     x = '{}{}{}'.format(colorama.Fore.CYAN, 'foo', colorama.Style.RESET_ALL)
 
     assert x == logger.cyan_text('foo')
+
+
+def test_markup_detection_pycolors_0(monkeypatch):
+    monkeypatch.setenv('PY_COLORS', '0')
+    assert logger.should_do_markup() is False
```
webknjaz: Comparing expressions which are already boolean with boolean literals is unpythonic. Here's how to do it:
```suggestion
    assert not logger.should_do_markup()
```

test/unit/test_logger.py
```diff
@@ -126,3 +126,28 @@ def test_cyan_text():
...
+    stdout_mock = mocker.patch("sys.stdout")
+    stdout_mock.isatty.return_value = True
+    assert logger.should_do_markup(sys.stdout) is True
+
+
+def test_markup_detection_is_not_tty(monkeypatch, mocker):
+    monkeypatch.delenv('PY_COLORS', raising=False)
+    stdout_mock = mocker.patch("sys.stdout")
+    stdout_mock.isatty.return_value = False
+    assert logger.should_do_markup() is False
```
webknjaz:
```suggestion
    assert not logger.should_do_markup()
```

test/unit/test_logger.py
```diff
@@ -126,3 +126,28 @@ def test_cyan_text():
...
+    monkeypatch.setenv('PY_COLORS', '1')
+    assert logger.should_do_markup() is True
+
+
+def test_markup_detection_is_tty(monkeypatch, mocker):
+    monkeypatch.delenv('PY_COLORS', raising=False)
+    monkeypatch.setenv('TERM', 'xterm')
+    stdout_mock = mocker.patch("sys.stdout")
+    stdout_mock.isatty.return_value = True
+    assert logger.should_do_markup(sys.stdout) is True
```
webknjaz:
```suggestion
    assert logger.should_do_markup(sys.stdout)
```

test/unit/test_logger.py
```diff
@@ -126,3 +126,28 @@ def test_cyan_text():
...
+
+
+def test_markup_detection_pycolors_0(monkeypatch):
+    monkeypatch.setenv('PY_COLORS', '0')
+    assert logger.should_do_markup() is False
+
+
+def test_markup_detection_pycolors_1(monkeypatch):
+    monkeypatch.setenv('PY_COLORS', '1')
+    assert logger.should_do_markup() is True
```
webknjaz:
```suggestion
    assert logger.should_do_markup()
```

molecule/logger.py
ssbarnea: Sure, I will use to_bool, funny I was considering doing it.

molecule/logger.py
ssbarnea: I will replace use of reserved file with "fh" as a file-handle is what it can receive. I wonder why our linting does not do anything about that. This is something we should fix inside "lint".

webknjaz: I think you cannot lint this because linters except for pylint do only static analysis + in some version `file` builtin exists. What do you mean by `fh`?

test/unit/test_logger.py
```diff
@@ -126,3 +126,28 @@ def test_cyan_text():
...
+
+def test_markup_detection_pycolors_1(monkeypatch):
+    monkeypatch.setenv('PY_COLORS', '1')
+    assert logger.should_do_markup()
+
+
+def test_markup_detection_is_tty(monkeypatch, mocker):
+    monkeypatch.delenv('PY_COLORS', raising=False)
+    monkeypatch.setenv('TERM', 'xterm')
+    stdout_mock = mocker.patch("sys.stdout")
```
webknjaz: There's no need in patching stdlib objects if you decide what to pass to the tested function anyway?

test/unit/test_logger.py
```diff
@@ -126,3 +126,28 @@ def test_cyan_text():
     x = '{}{}{}'.format(colorama.Fore.CYAN, 'foo', colorama.Style.RESET_ALL)
 
     assert x == logger.cyan_text('foo')
+
+
+def test_markup_detection_pycolors_0(monkeypatch):
```
webknjaz: I think this and the next test could be collapsed into one parametrized test case

test/unit/test_logger.py
```diff
@@ -126,3 +126,28 @@ def test_cyan_text():
...
+    monkeypatch.setenv('PY_COLORS', '0')
+    assert not logger.should_do_markup()
+
+
+def test_markup_detection_pycolors_1(monkeypatch):
+    monkeypatch.setenv('PY_COLORS', '1')
+    assert logger.should_do_markup()
+
+
+def test_markup_detection_is_tty(monkeypatch, mocker):
```
webknjaz: mocker and monkeypatch fixtures have the same purpose. I suggest you go for built-in monkeypatch.

test/unit/test_logger.py
webknjaz:
```suggestion
    monkeypatch.setattr('sys.stdout.isatty', lambda: True)
```

webknjaz: Oh, even better. This suggests using a CM: https://docs.pytest.org/en/latest/monkeypatch.html

test/unit/test_logger.py
```diff
@@ -126,3 +126,28 @@ def test_cyan_text():
...
+    monkeypatch.delenv('PY_COLORS', raising=False)
+    monkeypatch.setenv('TERM', 'xterm')
+    stdout_mock = mocker.patch("sys.stdout")
+    stdout_mock.isatty.return_value = True
+    assert logger.should_do_markup(sys.stdout)
```
webknjaz:
```suggestion
    with monkeypatch.ctx() as mp_ctx:
        mp_ctx.setattr('sys.stdout.isatty', lambda: True)
        assert logger.should_do_markup(sys.stdout)
```

molecule/logger.py
```diff
@@ -19,11 +19,27 @@
...
 import logging
+import os
 import sys
 
 import colorama
+from ansible.module_utils.parsing.convert_bool import boolean as to_bool
 
-colorama.init(autoreset=True)
+
+def should_do_markup(fh=sys.stdout):
```
webknjaz: Why do you need this to be an argument? I see that it's good for unit-testing but you don't really test the integration then.

molecule/logger.py
```diff
@@ -19,11 +19,27 @@
...
+    py_colors = os.environ.get('PY_COLORS', None)
```
webknjaz: Why not fall-back to a boolean value instead? You'd not need to do None-check at all, then.
```suggestion
    is_tty = fh.isatty() and os.environ.get('TERM') != 'dumb'
    py_colors = os.environ.get('PY_COLORS', is_tty)
    return to_bool(py_colors, strict=False)
```

ssbarnea: Not really: we cannot have a default boolean value here because the default behavior is to use detection. Ansible to_bool does return only booleans, it does not have any `undefined` value.

I used None as undefined.

test/unit/test_logger.py
ssbarnea: @webknjaz I had a rough experience with pytest CM so I would prefer if introducing its use in molecule test would be done by someone else, outside the scope of this patch. So far there is no existing code in molecule that uses pytest context manager.

PS. I am not againt CM, just that I already wasted too much time trying to make it work (and ended up with "good" old style).

molecule/logger.py
webknjaz: oh, makes sense

test/unit/test_logger.py
webknjaz: Just a few weeks ago I faced pytest-mock breaking test suite in an exceptionally weird manner in Ansible Core tests. Those patched objects were leaking into pytest internals badly.

And it's still weird to combine several patching tools in one place.

If you want to keep mocker, please ensure to always unpatch everything upon leaving the test function:
```python
    mocker.resetall()
    mocker.stopall()
```
Ref: https://github.com/ansible/ansible/pull/52372/files#diff-7d19c1624d0977d32233184491b27e1bR156

test/unit/test_logger.py
webknjaz: @ssbarnea this is the correct usage of context. just apply it and it should be good to go like this.

test/unit/test_logger.py
ssbarnea: Thanks for the hint! I am sure this may prove handly in the future. Meanwhile I already removed one of the mocks so we should be safe.

molecule/logger.py
ssbarnea: That is unit testing, I cannot really control how the user is calling the unittests and I need to isolate that from the test (it could be inside/outside tox, in CI, command line, with output redirected to file,... all of these could influence what stdout.istty() returns).

test/unit/test_logger.py
```diff
@@ -126,3 +126,33 @@ def test_cyan_text():
...
+    monkeypatch.setenv('PY_COLORS', '0')
+    assert not logger.should_do_markup()
+
+
+def test_markup_detection_pycolors1(monkeypatch):
+    monkeypatch.setenv('PY_COLORS', '1')
+    assert logger.should_do_markup()
+
+
+class FileHandleTTY(object):
```
webknjaz: put
```python
__metaclass__ = type
```
in the beginning of the module and you won't have to inherit from object explicitly.
Oh, and since you don't use features of new-style classes you don't need this anyway...

test/unit/test_logger.py
```diff
@@ -126,3 +126,33 @@ def test_cyan_text():
...
+
+def test_markup_detection_tty_yes(monkeypatch):
+    monkeypatch.delenv('PY_COLORS', raising=False)
+    monkeypatch.setenv('TERM', 'xterm')
+    assert logger.should_do_markup(FileHandleTTY(True))
+
+
+def test_markup_detection_tty_no(monkeypatch):
+    monkeypatch.delenv('PY_COLORS', raising=False)
+    assert not logger.should_do_markup(FileHandleTTY(False))
```
webknjaz: According to `pytest` practices, it is typical to use fixtures for such things.
You could do
```python
@pytest.fixture
def not_tty():
    return FileHandleTTY(False)

def test_markup_detection_tty_no(monkeypatch, not_tty):
```

molecule/logger.py
```diff
@@ -19,11 +19,28 @@
...
+    if py_colors is not None:
+        return to_bool(py_colors, strict=False)
+
+    return (callable(getattr(fh, 'isatty', None)) and
+            fh.isatty() and
+            os.environ.get('TERM') != 'dumb')
+
+
+def colorama_init():
```
webknjaz: Can we test this method in conjunction with the other one, please?
Because currently it looks like

molecule/logger.py
```diff
@@ -19,11 +19,28 @@
...
+from ansible.module_utils.parsing.convert_bool import boolean as to_bool
 
-colorama.init(autoreset=True)
+
+def should_do_markup(fh=sys.stdout):
+    py_colors = os.environ.get('PY_COLORS', None)
+    if py_colors is not None:
+        return to_bool(py_colors, strict=False)
+
+    return (callable(getattr(fh, 'isatty', None)) and
```
webknjaz: This check is completely unnecessary.

xoxys:
This pull request was moved and moved since nearly five months. Anything we can to bring it forward?

webknjaz:
@xoxys I'm happy to accept it once the harmful code is eliminated. I'd prefer to see some regression/integration test included as well, though.

decentral1se:
> This pull request was moved and moved since nearly five months.
> Anything we can to bring it forward?

Yeah, apologies about that @xoxys. If you makes you feel any better, I also had to wait 5 months (and endure 99 comments!) for #1458 ... :) but it's just due transition and https://molecule.readthedocs.io/en/latest/contributing.html#move-to-red-hat etc. We're much improved lately, I think.

Just needs concerns in https://github.com/ansible/molecule/pull/1533#issuecomment-468672031 addressed!

xoxys:
@decentral1se no need to apologize for anything. You all do a great job for the open source community! I am currently moving to a new apartment so there is not much time to do something by my own. But i would love to see this feature in the next time :)

molecule/logger.py
```diff
@@ -19,11 +19,24 @@
 #  DEALINGS IN THE SOFTWARE.
 
 import logging
+import os
 import sys
 
 import colorama
+from ansible.module_utils.parsing.convert_bool import boolean as to_bool
```
webknjaz: Do you mind putting this import one line above to maintain alphabetic order?

test/unit/test_logger.py
```diff
@@ -126,3 +126,25 @@ def test_cyan_text():
...
+
+def test_markup_detection_pycolors1(monkeypatch):
+    monkeypatch.setenv('PY_COLORS', '1')
+    assert logger.should_do_markup()
+
+
+def test_markup_detection_tty_yes(mocker):
+    mocker.patch('sys.stdout.isatty', return_value=True)
+    mocker.patch('os.environ', {'TERM': 'xterm'})
+    assert logger.should_do_markup()
```
webknjaz: As per my previous explanation @ https://github.com/ansible/molecule/pull/1533#discussion_r260881603, if you use `pytest-mock`, you have to manually unpatch everything. Otherwise, it causes external side-effects which break at least some of pytest versions.

webknjaz:
```suggestion
    assert logger.should_do_markup()
    mocker.resetall()
    mocker.stopall()
```

test/unit/test_logger.py
```diff
@@ -126,3 +126,25 @@ def test_cyan_text():
...
+
+
+def test_markup_detection_tty_no(mocker):
+    mocker.patch('os.environ', {})
+    mocker.patch('sys.stdout.isatty', return_value=False)
+    assert not logger.should_do_markup()
```
webknjaz: Same as https://github.com/ansible/molecule/pull/1533/files#r261974746:
```suggestion
    assert not logger.should_do_markup()
    mocker.resetall()
    mocker.stopall()
```

decentral1se:
Inclined to push this on the v2.20 release! The original report talked about improving the readability of Molecule CI logs and I think that is definitely something we want sooner than later. I'll merge later on today unless I hear otherwise.

decentral1se:
This was much needed in the end! See https://github.com/ansible/ansible-runner/issues/200 :+1:

Issue #1479, opened 2018-09-17
molecule-driver-contrib: community managed drivers


lwm:
Regarding https://github.com/metacloud/molecule/pull/1473#issuecomment-421832966.

> Hi @bandit145, I am hesitant to commit drivers which rely on community members maintaining drivers which most people do not use. To be honest, I think we already have too many drivers. Most people simply use Vagrant and Docker. Very few people use the others, and I don't actively test the others.

And since I am working on https://github.com/metacloud/molecule/pull/1458, I am of course interested in this.

So, have you given any thought to a `molecule-driver-contrib` which is pluggable from the end-user side and introduces no maintenance burden on you all here? We could say that those drivers are maintained by the community and those community members are responsible for them. No guarantees are given back in molecule core. That would steer all the maintenance tickets away from here.

I would guess core devs here would know the best way that is possible, if this is feasible.

I'd be happy to QA test any effort here in plugging in my Linode driver functionality.

lwm:
> If just the tickets get issued to the correct maintainer that should fix most of the problems (I think open for discussion of course). When a maintainer seems inactive over a longer period of time the driver should be removed from Molecule core.

I am not sure if you are aware, but this is basically what is already happening. This ticket is about talking about other potential solutions because it seems this is not working for core devs (this may not be the case entirely but from the original motivating comment, I want to start this discussion in case that is so).

> Another thing to consider is a contrib is a real stretch without a publicly visible CI. The community maintainers then must test all the changes locally, this takes a real time effort.

Each community driver maintains its own CI. This CI consists of: `molecule init role -r foobar -d <community-driver> && cd foobar && molecule test`. That's it, right? If the thing passes, the community driver is still working.

bandit145:
Maybe an option would be to handle drivers like Ansible handles module discovery. As an example you could have a custom driver in a "drivers" folder in the cwd and molecule could pick that up and could use the driver(s) in that directory.

retr0h:
So you can more or less do this today with the 'delegated' driver.  The delegated driver defaults to `managed: true`, which means as the developer you bring your own create/destroy playbooks and conform to the `instance_config` API in create.yml.

We have had users already use this for a vmware driver.  Community members simply drop in 3rd party drivers, and set the driver to delegated, and you should be good to go.

wilmardo:
Is it a possibility to do something like Ansible does with for example Azure (ansible[azure)]? 
`pip install molecule` for Molecule with Docker and Vagrant
[...]

PR #1539, opened 2018-11-01
Adds docker python module as requirement

wilmardo:
Adds the docker python module as requirement for molecule.
Can't find the issue or PR at the moment but if I remember correctly @ssbarnea mentioned that the user experience was somewhat lacking since out of the box Molecule doesn't work.

Since Docker is the default driver I don't see a problem with installing it as a dependency for Molecule.

The [docker_container module](https://docs.ansible.com/ansible/latest/modules/docker_container_module.html) requires >1.7.0 but the docker module only goes back to 2.0.0 on [PyPi ](https://pypi.org/project/docker/#history)(then the renaming happened from docker-py > docker). So everything above 2.0.0 should be alright.

The only point for discussion is Python 2.6 which requires the older docker-py module.
But since Ansible dropped some of the 2.6 support in [v2.7](https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_2.7.html) I do not see an issue in this.
Should I add the requirement for Python 2.7 somewhere in the docs?

#### PR Type

- Feature Pull Request

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ click==6.7
 click-completion==0.3.1
 colorama==0.3.9
 cookiecutter==1.6.0
+docker>=2.0.0
 flake8==3.5.0
 python-gilt==1.2.1
 Jinja2==2.10
--- a/molecule/driver/docker.py
+++ b/molecule/driver/docker.py
@@ -85,10 +85,6 @@ class Docker(base.Base):
             restart_policy: on-failure
             restart_retries: 1
 
-    .. code-block:: bash
-
-        $ sudo pip install docker-py
-
     Provide the files Molecule will preserve upon each subcommand execution.
 
     .. code-block:: yaml


ssbarnea:
If there are any concerns regarding not wanting to add uneccesary optional requirementes to molecule, we should at least have an extras options for docker which with do the trick and to mention it on default install documentatiom, like `pip install molecule[docker]`.

My personal preference is to have docker as default, but extras would also be considered acceptable.

I already seen azure driver listed as extras so you already know how how extras do work.

ssbarnea:
@webknjaz ^^ this should be easy to review :D

wilmardo:
> like pip install molecule[docker]

There has been a discussion about this before it moved to Ansible #1479 

I think it is sensible to have a working default with Docker and have a choice like `molecule[lxd]`,  `molecule[vagrant]` and `molecule[azure]` which omits this Docker driver and installs the required modules. This also omits the azure driver from the default install.

webknjaz:
I'd like to have it as extra. Having it required by default might cause side-effects (like deps collision): imagine user wants to avoid having docker-py in their virtualenv, but wants to have something conflicting. by requiring this dist by default you'll break such workflow.

I'm trying to bring extras within my current PRs as well:
* https://github.com/ansible/molecule/pull/1540/files#diff-380c6a8ebbbce17d55d50ef17d3cf906R28
* https://github.com/ansible/molecule/pull/1541/files#diff-380c6a8ebbbce17d55d50ef17d3cf906R35

requirements.txt
```diff
@@ -5,6 +5,7 @@ click==6.7
 click-completion==0.3.1
 colorama==0.3.9
 cookiecutter==1.6.0
+docker>=2.0.0
```
webknjaz: I'd accept it as extra, but not by default. This would have to be a separate discussion.

Also, there's two different dists: `docker` and `docker-py` implementing different API version. It may make sense to allow user to choose those.

P.S. Have you checked that they are well-compatible on the publish API level?

webknjaz:
I will not approve this until #1541 is merged (coming soon!) for stability reasons. I've spent way too much time making tests green.
Thanks for understanding :)

webknjaz:
@wilmardo 

> [...] and have a choice like `molecule[lxd]`,  `molecule[vagrant]` and `molecule[azure]` which omits this Docker [...]

Extras don't work like this. There's no way to set up metadata extra disabling other deps :)
That's Python packaging issue, we cannot work around it, sorry.

wilmardo:
> Extras don't work like this. There's no way to set up metadata extra disabling other deps :)

Oh sad I had no idea :neckbeard: 
Then it might be a good idea to close this and let `pip install molecule` be a base without a driver (so only delegated is possible). Then document clearly that a user needs to choose for a driver with `pip install [<drivername>]` which installs some extra dependencies for it to work.

Would something like that work?

webknjaz:
I think so. This will need additional approvals, not just mine, though.

ssbarnea:
Can we merge it in current form?

webknjaz:
@ssbarnea definitely not. It must go to extras.

webknjaz:
Ref #1561

webknjaz:
@wilmardo hey I've absorbed your commits into #1561.

wilmardo:
Closed by #1561

Issue #1522, opened 2018-10-25
molecule cannot be called as a module


ssbarnea:
- Bug report

# Desired Behavior

`python -m molecule test` sould have the same result as calling `molecule` script directly.

This is an important aspect because module calling can be used in places where scripts are not installable or not in PATH. Also allows you to control which python version to use to call molecule.

Most python based tools are able to be called like this, including `tox`, `flake8`, ....

The good news is that this feature is very easy to implement, not requiring any special things as python has native support for calling modules as scripts. 

# Actual Behaviour (Bug report only)
```
python -m molecule test
/home/travis/virtualenv/python2.7.14/bin/python: No module named molecule.__main__; 'molecule' is a package and cannot be directly executed
```

webknjaz:
Ref #1545

Issue #1521, opened 2018-10-25
molecule fails during linting destroy when run from inside a virtualenv


ssbarnea:
# Issue Type

- Bug report

# Molecule and Ansible details

```
molecule, version 2.19.0 (pip installed)
```

# Desired Behavior

Molecule shout run the same way if run from inside a virtualenv or from outside.

# Actual Behaviour

At this moment running molecule from outside virtualenv works perfectly. Still, when run from insidea a virtualenv created by tox, it fails with an unclear failure to cleanup after linting.

```
--> Executing Ansible Lint on /Users/ssbarnea/rdo/x/ssbarnea.six/ansible-six/molecule/default/playbook.yml...
Lint completed successfully.
--> Scenario: 'default'
--> Action: 'destroy'

    PLAY [Destroy] *****************************************************************

    TASK [Destroy molecule instance(s)] ********************************************
    changed: [localhost] => (item=None)
    changed: [localhost] => (item=None)
    changed: [localhost]

    TASK [Wait for instance(s) deletion to complete] *******************************
    failed: [localhost] (item=None) => {"attempts": 1, "censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": false}
    failed: [localhost] (item=None) => {"attempts": 1, "censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": false}
    fatal: [localhost]: FAILED! => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": false}

    PLAY RECAP *********************************************************************
    localhost                  : ok=1    changed=1    unreachable=0    failed=1


ERROR:
An error occurred during the test sequence action: 'destroy'. Cleaning up.
```

ssbarnea:
It seems that this is also easy reproduceable on Travis without any tox involvment. See https://travis-ci.com/ssbarnea/ansible-six

Travis creates a virtualenv for python projects by default, so this bug also affects it.

Any idea why this happens and how to avoid it? ... other than not using virtualenvs.

webknjaz:
Curious, I'll try to take a look at this next week. If not — ping/remind me.

wilmardo:
Could you add the `--debug` flag to molecule and post the output?

wilmardo:
Ran the build myself with `--debug` and the issue is with FreeBSD:
https://travis-ci.org/wilmardo/ansible-six/builds/448855694#L3125 

[...]

PR #1545, opened 2018-11-03
Add ability to call molecule as a python module

ssbarnea:
Also documents module calling:
    python -m molecule ...

Fixes: #1522 (#1521)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ Important Changes
 
 .. _`Move to Red Hat`: https://molecule.readthedocs.io/en/latest/contributing.html#move-to-red-hat
 .. _`quay.io`: https://quay.io/repository/ansible/molecule
+.. Molecule can now be called as a python module (python -m molecule)
 
 2.19
 ====
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,10 @@ History
 Unreleased
 ==========
 
+* Molecule can now be called as a python module (python -m molecule). Patch by `@ssbarnea`_.
+
+.. _`@ssbarnea`: https://github.com/ssbarnea
+
 Important Changes
 -----------------
 
@@ -13,7 +17,6 @@ Important Changes
 
 .. _`Move to Red Hat`: https://molecule.readthedocs.io/en/latest/contributing.html#move-to-red-hat
 .. _`quay.io`: https://quay.io/repository/ansible/molecule
-.. Molecule can now be called as a python module (python -m molecule)
 
 2.19
 ====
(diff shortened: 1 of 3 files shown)


doc/source/installation.rst
```diff
@@ -53,7 +53,16 @@ Install Molecule:
...
 
-    $ sudo pip install molecule
+    $ pip install --user molecule
+
+Installing molecule package also installed its main script ``molecule``,
+usually in ``PATH``. Users should know that molecule can also be called as a
+python module, using ``python -m molecule ...``. This alternative method has
+some benefits:
+
+* allows to control which python interpreter is used by molecule
```
webknjaz:
```suggestion
* allows to explicitly control which python interpreter is used by molecule
```

doc/source/installation.rst
```diff
@@ -53,7 +53,16 @@ Install Molecule:
...
+* allows molecule installation at user level without even needing to have
```
webknjaz: I'm not sure how this is related to the change.

webknjaz:
@ssbarnea could you please also add an entry to https://github.com/ansible/molecule/blob/master/CHANGELOG.rst in this PR?

doc/source/installation.rst
ssbarnea: This clearly related to this patch because before it, if you would have installed molecule using --user it was likely that you would fail to execute "moleculte" because ~/.local/bin was not in you PATH. Addition of ~/.local/bin to user PATH is something relatively new.

In fact this is one of the main selling points for using module calling instead of script callling, as it does not depend of having the script in PATH.

webknjaz: Oh.. I see. I'd say it's not _directly_ connected, but okay. It probably makes sense to leave this note here as well :)

CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ Important Changes
 
 .. _`Move to Red Hat`: https://molecule.readthedocs.io/en/latest/contributing.html#move-to-red-hat
 .. _`quay.io`: https://quay.io/repository/ansible/molecule
+.. Molecule can now be called as a python module (python -m molecule)
```
webknjaz: Hey, it's a syntax for external hyperlinks in RST, meaning that you put a label + link like this and then use it to highlight some link in the text.

Ref: http://docutils.sourceforge.net/docs/user/rst/quickref.html#external-hyperlink-targets

webknjaz: @ssbarnea I've fixed this myself

webknjaz:
@gundalow I think we can merge this without waiting for the Travis CI PR, because it doesn't change existing code and just adds a new module. What do you think?

Issue #1546, opened 2018-11-03
minimize number of virtualenv used by tox


ssbarnea:
# Issue Type

- Feature request

# Molecule and Ansible details

Currently running tox with default envlist creates 23 virtualenvs and takes about 6GB disk space due to theve virtualenvs. 

Many of these virtualenvs do have the same package lists installed inside them so we are better off reusing virtualenvs between different test commands. 

A very good article explaining how to do this can be found at https://blog.ionelmc.ro/2015/04/14/tox-tricks-and-patterns/ -- this not being a new tox feature, in fact is very old but is less known.

We should be safe to have one virtualenv for each python version, the footprint would be cut to about 1/3 of its current size and also the test duration would be cut considerably as a side effect.

PR #1547, opened 2018-11-03
Reconfigure tox to reuse same envs between unit and functional

ssbarnea:
Removes 9 environments by reusing them across unit and functional
testing commands as they have identical dependencies.

Fixes: #1546


tox.ini
```diff
@@ -1,15 +1,16 @@
...
+    py{27,36,37}-ansible{24,25,26}-functional
+    py{27,36,27}-ansible{24,25,26}-unit
     py{27,36,37}-lint
     format-check
     doc
 skipdist = True
 skip_missing_interpreters=True
+# sitepackages improve speed and footprint at small risk of missing to spot
+# undeclared dependencies which are already installed on the system.
+sitepackages=True
```
webknjaz: I don't think we should do this. Tests are already unstable. No more cumulative risks, please.

tox.ini
```diff
@@ -1,15 +1,16 @@
-
 [tox]
 minversion = 1.8
 envlist =
-    py{27}-ansible{24,25,26}-functional
-    py{27,36,37}-ansible{24,25,26}-unit
-    py{36,37}-ansible{24,25,26}-functional
+    py{27,36,37}-ansible{24,25,26}-functional
```
webknjaz: Please wait until I extract this from my PR. I've got it extended with a few more.

webknjaz:
@ssbarnea please rebase this after #1557 merged

webknjaz:
@ssbarnea you can probably start rebasing this

tox.ini
```diff
@@ -1,15 +1,16 @@
...
+    py{27,36,27}-ansible{24,25,26}-unit
```
webknjaz: you can also collapse the last factor as `{functional,unit}`

tox.ini
```suggestion
    py{27,36,37}-ansible{24,25,26}-{functional,unit}
```

tox.ini
```suggestion
```

themr0c:
@ssbarnea Can you please rebase, so we can have a view if the proposal is still compatible with current tox.ini?

tox.ini
ssbarnea: sure, better to do this isolated if we do it.

PR #1561, opened 2018-11-06
Add optional deps as extras

webknjaz:
#### PR Type

- Feature Pull Request


lwm:
Yes, this seems a nice way to optionally include driver dependencies.

webknjaz:
I skipped `ec2` because I'm not sure about boto vs boto3: are they both needed or conditionally?

Issue #1567, opened 2018-11-07
Allow changes in the base directory name for the role


amarao:
# Issue Type
- Feature request

# Molecule version
2.19.0, via pip

# Desired Behavior
Right now molecule search for role to test in the parent directory. If molecule tests for role `foo`, it expects to have it to be in a directory `foo`. If this directory was renamed into something different ('f.e. `ansible-foo`, or `username.foo`), it fails with following message:

```
ERROR! the role 'foo' was not found in /path/ansible-foo/molecule/default/roles:/tmp/molecule/ansible-foo/default/roles:/path:/path/ansible-foo/molecule/default

The error appears to have been in '/path/ansible-foo/molecule/default/playbook.yml': line 5, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  roles:
    - role: foo
      ^ here
```

The key issue is that most ansible roles are stored not under the role name, but with some prefixes - user (galaxy roles), or, in case of github, ansbile-* is a common prefix. It's really strange to have configuration role to be kept in the same directory. `github.com/example/docker` is expected to my docker, and `github.com/example/ansible-docker` is expected to be ansible role for docker.

I found a workaround for affected people, but it's dirty.

```
mkdir molecule/default/roles
ln -s '../../..' molecule/default/roles/${REAL_ROLE_NAME}
```
f.e. for role 'foo' it should be `molecule/default/roles/foo -> '../../..'`

lwm:
Ah, I might have re-read this issue in the first place when I prompted you to create the issue but this is nonetheless worth having a discussion about. There is also the option to do https://github.com/ansible/molecule/blob/master/test/scenarios/driver/docker/molecule/default/molecule.yml#L22 to specify the roles path. This could potentially need better documenting.


I'll create another issue for the thing I was talking about on IRC now ...

ssbarnea:
As someone that was also confused about role location and how to deal with them, I support fixing this even if this may only need changing documentation or the way messages are printed.

icollar:
Specifying the `ANSIBLE_ROLE_PATH` won't resolve this as Ansible looks for roles by searching all directories on the `ANSIBLE_ROLE_PATH` for directories with the same name as the role. Very quick example of this:

```
icollar@dev:~$ molecule init role --role-name rolename
--> Initializing new role rolename...
Initialized role in /home/icollar/rolename successfully.
icollar@dev:~$ cd rolename 
[...]

PR #1581, opened 2018-11-10
added an alternate Jenkinsfile and comments on Multibranch pipelines.

alainchiasson-guavus:

#### PR Type

- Docs Pull Request

Wit the current docs : 

- The Documented Jenkinsfile did not work on our central servers. The Jenkins sandbox environment limits some changes. I have not been able to determine if it is my specific jenkins or if it is general - so I have not removed the pipeline example.

For Jenkins :
- Jenkins renames the role root directory to the workspace name in multibranch projects. So the role is not found . After finding the comments I tested on internal roles. (  https://github.com/ansible/molecule/issues/1567#issuecomment-436876722 ) - this comment is is for general Jenkins - not just in the docker image.

For Jenkinsfile within a docker environment: 

- Added usage of the official docker image from the top of the examples section (https://molecule.readthedocs.io/en/latest/examples.html). Jenkins takes care of mounting the workspace in the container, so the only volume required to mount is the docker socket.

### DIFF
--- a/doc/source/ci.rst
+++ b/doc/source/ci.rst
@@ -144,7 +144,54 @@ Jenkins Pipeline
 
     }
 
+The following `Jenkinsfile` uses the official 'quay.io/ansible/molecule' image.
 
+.. code-block:: groovy
+
+    pipeline {
+      agent {
+        docker {
+          image 'quay.io/ansible/molecule'
+          args '-v /var/run/docker.sock:/var/run/docker.sock'
+        }
+      }
+
+      stages {
+
+        stage ('Display versions') {
+          steps {
+            sh '''
+              docker -v
+              python -V
+              ansible --version
+              molecule --version
+            '''
+          }
+        }
+
+        stage ('Molecule test') {
+          steps {
+            sh 'sudo molecule test --all'
+          }
+        }
+
+      } // close stages
+    }   // close pipeline
+
+.. note::
+
+    For Jenkins to work properly using a `Multibranch Pipeline` or a `GitHub Organisation` - as used by Blue Ocean, the role name in the scenario/playbook.yml should be changed to perform a lookup of the role root directory. For example :
+
+.. code-block:: yaml
+
+    ---
+    - name: Converge
+      hosts: all
+      roles:
+        - role: "{{ lookup('env', 'MOLECULE_PROJECT_DIRECTORY') | basename }}"
+
+
+This is the cleaner of the current choices. See `issue1567_comment`_ for additional detail.
 
 Tox
 ^^^
@@ -245,3 +292,4 @@ conflict.
 .. _`Gitlab`: https://gitlab.com
 .. _`Tox`: https://tox.readthedocs.io/en/latest
 .. _`--parallel functionality`: https://tox.readthedocs.io/en/latest/config.html#cmdoption-tox-p
+.. _`issue1567_comment`: https://github.com/ansible/molecule/issues/1567#issuecomment-436876722


doc/source/ci.rst
```diff
@@ -144,7 +144,54 @@ Jenkins Pipeline
...
+              python -V
+              ansible --version
+              molecule --version
+            '''
+          }
+        }
+
+        stage ('Molecule test') {
+          steps {
+            sh 'sudo molecule test --all'
```
ssbarnea: TBH, my hopes were to see an example of spreding molecule execution in multiple CI stages, as this would be the only way to make it clear when it fails on certain platform.
 
Stages can run serially or in parallel, this being based on the job defintion but I would really like to see how to split a single execution in multiple ones.

icollar: Something to be aware of is that the general convention as show in the docs, with Docker, is using instance names such as instance-1 and instance-2. If you were to run scenarios in parallel you must be careful of the following:

- The instances are named uniquely if your Jenkins executors are running on the same server. Otherwise the destroy phase would be run end a test run early, or a an instance could fail to startup due to a naming conflict.
- If you happened to run two branches of the same role/molecule repository then they could also conflict, so I'd recommend naming the instance after the current branch as well.

Using the following Jenkinsfile works for the later issue:

```
pipeline {
	agent any
	environment {
		BRANCH_SAFE = sh (script: "echo $BRANCH_NAME | sed 's@/@-@'", returnStdout: true).trim()
	}
	stages {
		stage ('test') {
			steps {
				sh 'molecule test --all'
			}
		}
	}
	post {
		always {
			deleteDir()
		}
	}
}
```

With molecule.yml looking like:
```
platforms:
  - name: ${BRANCH_SAFE:-default}-instance-1
    image: centos:latest
```

Obviously it could be expanded out a little to account for any scenario, e.g. by adding the a variable containing the current molecule scenario, but I didn't need that for my means.

Worth noting that this wouldn't be needed for certain setups of Jenkins where each executor has a completely isolated Docker server, but alas this isn't an option for me.

alainchiasson-guavus: So I agree about additional examples - and this can probably be worked out - but they become more advanced. My goal in publishing the simple example was to get it working for the multi-branch pipeline from the pre-built molecule image. 

The example above it, for me, did not work - it failed with errors on "pipeline durability". To get it to work, we would have had to remove alot of security from Jenkins (do you really want someone else's build script to install software on your slaves ?)

Based on the bug / bug fixes - I don't think I was the only one struggling with this.

alainchiasson-guavus: @ssbarnea 

I did the following and it works (did not place it in the PR yet. See bellow for other ideas I'm looking at.

```
pipeline {
  agent {
    docker {
      image 'quay.io/ansible/molecule'
      args '-v /var/run/docker.sock:/var/run/docker.sock'
    }
  }

  stages {

    stage ('Display versions') {
      steps {
        sh '''
          docker -v
          python -V
          ansible --version
          molecule --version
        '''
      }
    }

    stage ('Molecule pre-clean') {
      steps {
        sh 'sudo molecule destroy'
      }
    }

    stage ('Molecule converge') {
      steps {
        sh 'sudo molecule converge'
      }
    }

    stage ('Molecule idempotence') {
      steps {
        sh 'sudo molecule idempotence'
      }
    }

    stage ('Molecule apply side effect') {
      steps {
        sh 'sudo molecule side-effect'
      }
    }

    stage ('Molecule verify') {
      steps {
        sh 'sudo molecule verify'
      }
    }

    stage ('Molecule post-clean') {
      steps {
        sh 'sudo molecule destroy'
      }
    }

  } // close stages
}   // close pipeline
```

I would like to see If I can run the scenarios in parallel and via "dynamic discovery".

themr0c:
Hi @alainchiasson, can you please hit the Resolve conflicts button or rebase your branch?

alainchiasson:
I'm having a hard time figuring out where the failure is. I'm only changing documentation.

decentral1se:
Looks unrelated @alainchiasson-guavus! I'm re-running that failed build ... thanks for patience :+1:

PR #1591, opened 2018-11-15
Enable LXC and LXD functional tests under Travis CI

webknjaz:
#### PR Type

- Bugfix Pull Request

PR #1594, opened 2018-11-15
Enable auto release and prerelease on pypi

ssbarnea:
Pushed tags will release on pypi if all jobs are passing.

Cron and manual builds will pre-release on pypi if all jobs are passing.

No risks as pypi nevers allows release override. Also all non tagged
versions do count as prereleases.

You will have to update this pull request to add TRAVIS_USERNAME and TRAVIS_PASSWORD into .travis.yml file, last one needs to be encrypted using `travis encryp` cli to secure it.

#### PR Type

- Feature Pull Request

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,10 @@ cache:
 services:
 - docker
 
+stages:
+- test
+- release
+
 language: python
 
 .mixtures:  # is not used by Travis CI, but helps avoid duplication
@@ -287,7 +291,10 @@ jobs:
     env:
       <<: *env-functional-shard_3
       ANSIBLE_VERSION: devel
-
+  - release
+      <<: *if-cron-or-manual-run-or-tagged
+      stage: release
+      script: tox -e upload
 env:
   global:
     TOXENV_TMPL: "'ansible${ANSIBLE_VERSION}-${TESTS_TYPE}'"
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@
 .tox/
 .vagrant/
 .pytest_cache/
+build/
+dist/
 doc/build/
 htmlcov/
 pytestdebug.log
--- a/tox.ini
+++ b/tox.ini
@@ -51,6 +51,16 @@ commands =
 extras =
     docs
 
+[testenv:upload]
+envdir = {toxworkdir}/py27-ansible25-unit
+whitelist_externals =
+    rm
+commands=
+    rm -rf dist/*
+    python setup.py sdist bdist_wheel
+    twine check dist/*
+    twine upload dist/*
+
 [testenv:pur]
 commands=
     pur -r requirements.txt


.travis.yml
```diff
@@ -287,7 +291,10 @@ jobs:
     env:
       <<: *env-functional-shard_3
       ANSIBLE_VERSION: devel
-
+  - release
```
webknjaz: This is invalid yaml.

.travis.yml
```diff
@@ -12,6 +12,10 @@ cache:
 services:
 - docker
 
+stages:
```
webknjaz: This is default ordering, no need to be explicit about it

.travis.yml
```diff
@@ -287,7 +291,10 @@ jobs:
     env:
       <<: *env-functional-shard_3
       ANSIBLE_VERSION: devel
-
+  - release
+      <<: *if-cron-or-manual-run-or-tagged
+      stage: release
+      script: tox -e upload
```
webknjaz: script step should be used for testing, not for publishing. there's deploy step for that.

.travis.yml
```diff
@@ -287,7 +291,10 @@ jobs:
     env:
       <<: *env-functional-shard_3
       ANSIBLE_VERSION: devel
-
+  - release
+      <<: *if-cron-or-manual-run-or-tagged
```
webknjaz: And cron is not a good choice. It'd trigger multiple (failing) attempts to publish the same commit/version. Unless version doesn't contain date stamp, which wouldn't make sense either because we'd end up with a number of versions with exactly same source polluting PyPI page history.

ssbarnea: What do you propose to use? We could check the output and ingore the failure to upload due to already-uploaded and keeping it as failed only for "other reasons".

How often does the cron runs? I think that weekly is ok, more often would create spam. Still, keep in mind that we can hide releases from pypi. We could easily auto-hide old pre-releases to declutter the history interface.

.travis.yml
ssbarnea: I stopped using deploy when I found how broken and unmaintained were its providers but you are right, there is a deploy script provider that can do the same thing.

I do like more the new stages as it helps organizing much better but I can fix this.

Still, here is something to think about: "tox -e upload" does also run a  twine check which maxes it bit mixed: containing elements of both testing and uploading / deploying.

Because you are not running all tests on each commit I wonder if it would not be better to migrate to a multe-branch setup when PRs are made on `devel` branch and once a day we run the heavy tests and when they pass, we merge the changes to master. 

Something like this could assure that master always contains code that passed the heavy-testing, while keeping the roundtrip for testing and merging a PRs short.

Different users may pick their own poison.

webknjaz: @ssbarnea in my implementation, check is taken care of and it runs from the proper place. And yes, I use stages.

I have a huge experience with Travis CI and I know about drawbacks as well but I also think that abusing tox this is not the answer since it limits our ability to easily distinguish between different actions happening in logs, for example.

Multi-branch sounds interesting (like the original Git flow?) but this is not the place to discuss it. Feel free to create a separate issue about it, currently cron takes care of it + the way I set up conditions facilitates the flow.

Please let me take care of CI+packaging because it's what I'm good at.

OTOH I could use some help with enabling more tests in CI, I'm struggling with #1591 so if you want to help this would be ideal if you could help with that :)

.travis.yml
webknjaz: It runs daily. I already have a good implementation and I don't see a reason to reimplement that.

webknjaz:
I've reported a bug wrt broken config here: https://travis-ci.community/t/bug-invalid-config-should-be-reported-to-github-as-an-error/956

webknjaz:
Implemented via #1593

Issue #1556, opened 2018-11-05
vagrant driver writes vagrant-console-log to installation dir instead of $TMPDIR


pescobar:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible 2.7.1
molecule, version 2.19.0
python-vagrant==0.5.15
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

# Desired Behavior

molecule should write all the logs to `$TMPDIR`. If molecule is installed in a folder where you don't have write permissons it crashes when booting a vagrant instance. 

# how to reproduce the problem

```
$> echo $TMPDIR
/tmp/escobar

$> molecule init role --driver-name vagrant --role-name testing-molecule-vagrant

$> rm  testing-molecule-vagrant/tasks/main.yml

$> cat <<EOF >> testing-molecule-vagrant/tasks/main.yml
---
# tasks file for testing-molecule-vagrant

- name: just a test
  command: uname -r
  changed_when: false
EOF

$> cd testing-molecule-vagrant

$> molecule --debug test
```

During `TASK [Create molecule instance(s)]` you get error: `  "msg": "ERROR: See log file '/tmp/escobar/molecule/testing-molecule-vagrant/default/vagrant-instance.err'"`

and in the log file:

```
There was an error while executing `VBoxManage`, a CLI used by Vagrant
for controlling VirtualBox. The command and stderr is shown below.

Command: ["startvm", "384da504-e582-41a5-8f4f-08f478b62768", "--type", "headless"]

Stderr: VBoxManage: error: RawFile#0 failed to create the raw output file /data/virtualenvs/ansible-2.7.1-molecule-2.19.0/lib/python2.7/site-packages/molecule/provisioner/ansible/playbooks/vagrant/ubuntu-xenial-16.04-cloudimg-console.log (VERR_ACCESS_DENIED)
VBoxManage: error: Details: code NS_ERROR_FAILURE (0x80004005), component ConsoleWrap, interface IConsole
### 2018-11-05 22:06:45 ###
```

in this machine ansible and molecule are installed in a virtualenv located in `/data/virtualenvs/ansible-2.7.1-molecule-2.19.0/` that is not owned by my user

pescobar:
I found I can workaround the problem if I change `default/molecule.yml` from the default value:

```yaml
platforms:
  - name: instance
    box: ubuntu/xenial64
```

to 

```yaml
platforms:
  - name: instance
    box: bento/ubuntu-16.04
[...]

PR #1600, opened 2018-11-23
Fixing #1556 VERR_ACCESS_DENIED

jonashackt:
This should fix https://github.com/ansible/molecule/issues/1556 "vagrant driver writes vagrant-console-log to installation dir instead of $TMPDIR". 

I just added `'modifyvm', :id, '--uartmode1', 'disconnected'` to the default Vagrantfile of Molecule, as this seems to fix the issue. See also https://github.com/jonashackt/molecule-ansible-docker-vagrant/blob/master/docker/molecule/vagrant-ubuntu/molecule.yml

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/molecule/driver/vagrant.py
+++ b/molecule/driver/vagrant.py
@@ -79,6 +79,7 @@ class Vagrant(base.Base):
               gui: True
             provider_raw_config_args:
               - "customize ['modifyvm', :id, '--cpuexecutioncap', '50']"
+              - "customize ['modifyvm', :id, '--uartmode1', 'disconnected']"
             provider_override_args:
               - "vm.synced_folder './', '/vagrant', disabled: true, type: 'nfs'"
             provision: True


lwm:
Any ideas how we can test this @jonashackt?

jonashackt:
@lwm not really, as TravisCI & CircleCI don't support Vagrant. But maybe AppVeyor would be a choice, I created an example project here: https://github.com/jonashackt/vagrant-ansible-on-appveyor As of what I got to know, this should work. But sadly - also the AppVeyor support doesn't know, why - it's not working...

jonashackt:
But how does the Vagrant testing work for the current implementation anyway @lwm ?

lwm:
Heyo. I don't know how it's been doing (if at all) for Vagrant here. Vagrant functionality is definitely a solid and well desired feature for this tool, so hopefully there is something in place which you can extend. If not, it would be nice to know how difficult it is to do (and perhaps create a ticket to track that also).

decentral1se:
Anyone using Vagrant can please provide some guidance here?

I don't feel confident to merge this at all.

t2d:
I need to do this all the time manually to make the official Ubuntu vagrant images run.

jonashackt:
I already tried to automatically test a Vagrant environment, but failed on TravisCI and CircleCI, since they are both blocking a VirtualBox & Vagrant installation. However I got quite far on AppVeyor: https://github.com/jonashackt/vagrant-ansible-on-appveyor - there, you can install VirtualBox and Vagrant like a charm, only thing that's left there is the Vt-x CPU support of the standard machines there - this feature needs payment. 

BUT: AFAIK there is a possibilty to request such a machine with VT-x enabled for OpenSource projects. We could try to request such a machine for Molecule and setup a exhaustive Vagrant test setup?!

decentral1se:
@jonashackt if you have any free cycles to pursue this, I would gladly support you here. I think long term, we're looking at breaking out these drivers into their own plugin / test environments but for now, it would be great to make some progress on testing this right now.

ssbarnea:
Is there a way to run ansible tests only for Vagrant? I have two hosts with vagrant installed and I could manually test it. Hopefully the modified code is covered by some of the tests.

Issue #1603, opened 2018-11-27
Molecule fails installing ansible roles from requirements.yml using git+ssh


Lowess:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version

ansible 2.7.1
  config file = None
  configured module search path = [u'/home/molecule/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.15 (default, Aug 16 2018, 14:17:09) [GCC 6.4.0]

molecule, version 2.19.0
```

Molecule installation method (one of):

- docker image

Ansible installation method (one of):

- docker image

Detail any linters or test runners used:

None

# Desired Behavior

The Docker image `quay.io/ansible/molecule` is missing the `openssh-client` package which prevents us from installing Ansible roles from private git repositories over ssh. This action is failing even though you have the ssh keys are present.

Example:

```
# requirements.yml
---

- src: git@bitbucket.org:<teamname>/<rolename>
  scm: git
  version: master
```

# Actual Behaviour

Ansible fails with return code 128:

```sh
docker run --rm -it \
  -v /Users/florian/Workspace/gumgum-ansible-roles/ansible-role-meetup:/tmp/ansible-role-meetup \
  -v /var/run/docker.sock:/var/run/docker.sock -v /Users/florian/.ssh:/home/molecule/.ssh \
  -w /tmp/ansible-role-meetup \
  --volumes-from=ssh-agent \
  -e SSH_AUTH_SOCK=/.ssh-agent/socket \
  quay.io/ansible/molecule:latest molecule dependency

--> Validating schema /tmp/ansible-role-meetup/molecule/default/molecule.yml.
Validation completed successfully.
--> Test matrix

└── default
    └── dependency

--> Inventory /tmp/ansible-role-meetup/molecule/default/../inventory/group_vars/ linked to /tmp/molecule/ansible-role-meetup/default/group_vars
--> Scenario: 'default'
--> Action: 'dependency'
 [WARNING]: - ansible-role-common was NOT installed successfully: - command
/usr/bin/git clone git clone git@bitbucket.org:<teamname>/<rolename> <rolename> failed in directory /home/molecule/.ansible/tmp/ansible-local-
9ViBtYx/tmpD_FpFB (rc=128)

ERROR! - you can use --ignore-errors to skip failed roles and finish processing the list.
[...]

PR #1604, opened 2018-11-27
Fixing #1603 Add openssh-client to Docker image for git+ssh support

Lowess:
This PR addresses the issue https://github.com/ansible/molecule/issues/1603

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -25,6 +25,7 @@ ENV PACKAGES="\
     ruby-rdoc \
     ruby-irb \
     git \
+    openssh-client
 "
 
 ENV PIP_PACKAGES="\
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -25,7 +25,7 @@ ENV PACKAGES="\
     ruby-rdoc \
     ruby-irb \
     git \
-    openssh-client
+    openssh-client \
 "
 
 ENV PIP_PACKAGES="\


lwm:
I know there has been debate on what and what not should be included in the Dockerfile but this looks like a no-brainer give the issue report which is clearly something that Molecule should be supporting. Thanks for bringing this change @Lowess.

lwm:
It would perhaps be nice to include something with a `git+ssh` URL in the functional or scenario tests for this? I am not quite sure where it should go but it should be simple enough to find somewhere. Since the package is in transition from different maintenance, it would be nice to have tests for everything :)

ragingpastry:
I think a separate functional test might be good. A large number of the current functional tests already bring in roles via ansible-galaxy so tacking on something with git+ssh might be a good solution. This would bring in external dependencies but I'm not sure how you would properly test this otherwise.

Perhaps the Docker build itself should be tested separately from Molecule proper?

vutkin:
Hi @themr0c , @Lowess , @lwm we are supporting next docker images with systemd support for testing our roles: https://cloud.docker.com/u/leandelivery/repository/docker/leandelivery/docker-systemd

Could we use them as default molecule base images for testing? They already has all needed packages and improvements.
Will them work for us?

themr0c:
@vutkin are these images publicly available ?

tgadiev:
> @vutkin are these images publicly available ?
Yes, of course.
https://hub.docker.com/r/leandelivery/docker-systemd

vutkin:
@themr0c sure, there are images for centos 7, ubuntu 16.04,18.04 with systemd support.

build/Dockerfile
```diff
@@ -25,6 +25,7 @@ ENV PACKAGES="\
     ruby-rdoc \
     ruby-irb \
     git \
+    openssh-client
```
webknjaz:
```suggestion
    openssh-client \
```

themr0c: Hi @Lowess can you please apply the change suggested by @webknjaz ?

Lowess:
Hi guys, sorry for my delayed response. I addressed the comment above but I wont have time right now to implement the test cases mentioned above by @vutkin. Can the testing improvement be addressed in an other PR ?

lwm:
Agreed @Lowess, we'll torture you no longer :) 

@vutkin, can you please raise an issue for https://github.com/ansible/molecule/pull/1604#issuecomment-463749380, thanks!

vutkin:
https://github.com/ansible/molecule/issues/1768

PR #1605, opened 2018-11-27
Make molecule support the key sysctls for docker containers.

f-roscher:
Signed-off-by: Florian Roscher <mail@florian-roscher.de>

This PR adds support for the argument 'sysctls' for docker containers. Ansible supports this, but the default create.yml was missing that dict entry. 
I testet this with a molecule create and the sysctl setting was visible the container. Not setting it worked, too, due to default(omit). 

I tried to test with pytest, but there are a lot of errors, even on branch master. If I should do some more here to get this accepted or easy your work, any hints would be appreciated on  how to continue. 
I might appear on IRC, if that helps. 

#### PR Type

- Feature Pull Request

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -65,6 +65,7 @@
         log_driver: json-file
         command: "{{ item.command | default('bash -c \"while true; do sleep 10000; done\"') }}"
         privileged: "{{ item.privileged | default(omit) }}"
+        sysctls: "{{ item.sysctls | default(omit) }}"
         security_opts: "{{ item.security_opts | default(omit) }}"
         volumes: "{{ item.volumes | default(omit) }}"
         tmpfs: "{{ item.tmpfs | default(omit) }}"


lwm:
Any test case for this @f-roscher?

lwm:
> That being said: I will look into building a test. This is what is being asked for, right, @lwm ? If someone happens to have some pointers towards the files I need to touch it would help for sure.

Yep! So, what I can here is: I looked at what is in place for another option for the docker create.yml. I did a search for `security_opts` and this is what I got: https://github.com/ansible/molecule/search?q=security_opts&unscoped_q=security_opts. You seem to be able to add this new option in  `test/resources/playbooks/docker/create.yml`, `test/unit/model/v2/test_platforms_section.py`, `molecule/model/schema_v2.py`, etc. Hope that is clear!

> Another question: Do I need to and am I able to fix the Travis CI "build errored"? At a first glance I did not get wether my line makes the difference here. But I did not have much time to look at this the last two weeks.

Yeah, I am not fully in the know but I think the CI is unstable right now. However, it is passing in current state: https://travis-ci.com/ansible/molecule. So, you could have some chance for a clean build here ;)

molecule/provisioner/ansible/playbooks/docker/create.yml
```diff
@@ -65,6 +65,7 @@
         log_driver: json-file
         command: "{{ item.command | default('bash -c \"while true; do sleep 10000; done\"') }}"
         privileged: "{{ item.privileged | default(omit) }}"
+        sysctls: "{{ item.sysctls | default(omit) }}"
```
lwm: Thanks for this @f-roscher. I think we need to add this to the schema in https://github.com/ansible/molecule/blob/master/molecule/model/schema_v2.py and include some unit tests to assure that is validation as expected.

ssbarnea: @f-roscher Can you please address that request? I will be happy to help you merge this improvement once you also update the schema.

f-roscher:
Looks like I missed some notification. Found this while doing housekeeping tasks. I will look into it.

decentral1se:
Thanks @f-roscher!

ssbarnea:
Closing due to the lack of updates.

Issue #1611, opened 2018-12-04
`molecule` entry-point script mysteriously disappears from the virtualenv (and from disk completely) during tests


webknjaz:
It is totally unclear how this happens. I suspect some of tests to have side effects like this.
It looks like @Shaps faced this too.

# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version
```

Molecule installation method (one of):

- source

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

# Desired Behavior

Tests should be passing in stable manner.

# Actual Behaviour

https://travis-ci.com/ansible/molecule/jobs/162837106#L1053

webknjaz:
Ref: https://github.com/tox-dev/tox/issues/1097
Ref: https://github.com/requests/requests/issues/4890

webknjaz:
Fixed by 18835c5008b72bd678197bb1c85167fd7038e65a

PR #1615, opened 2018-12-05
Allow docker command to be null, set default CMD in dockerfile instead

### LINKED ISSUE #1614: CMD in pre-built docker image replaced with `while true; do sleep 10000; done`
# Issue Type

- Bug report

Was kind of reported in #1441 

```
ansible 2.7.2
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/home/fabian/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /home/fabian/.local/lib/python2.7/site-packages/ansible
  executable location = /home/fabian/.local/bin/ansible
  python version = 2.7.15 (default, May 15 2018, 15:37:31) [GCC 7.3.1 20180303 (Red Hat 7.3.1-5)]
molecule, version 2.19.1.dev54

```

- source

- pip

# Desired Behavior
If I don't specify a `command` or specify `null` as the value for `command`, the existing `CMD` on the image should be used.

# Actual Behaviour
When consuming an image with the docker driver, the existing `CMD` on the image is overwritten, even if `command` is unspecified or null. 

In order to get the behavior I desire from molecule, I currently need to copy-paste the content  of the `CMD` directive into the `command` section of the `molecule.yml`, but this is inherently fragile as every time the container used changes its `CMD` I would need to reflect that in every related molecule specification.

fabianvf:
Fixes #1614

#### PR Type

- Bugfix Pull Request

Instead of setting the default command value of `while true; do sleep 10000; done` in the `docker_container` invocation, create the sample Dockerfile with the default CMD on it, and allow `command` to be nullable, in which case it will use the existing `CMD` directive on the image.


fabianvf:
Need to update the tests that are relying on the docket default command directive, will try to get that pushed this morning.

fabianvf:
I'm unable to reproduce the test failure in travis, and as far as I can tell there's no useful debug output in the travis log, can someone rekick the build to verify it's not a flake? @gundalow @webknjaz

webknjaz:
@fabianvf you can restart tests by just reopening PR if needed.

Those previous failure was related to #1611 (combo of tox-dev/tox#1097 + requests/requests#4890), which is fixed now.

fabianvf:
@webknjaz I think I actually found the issue, there was a separate Dockerfile template for tests that didn't have the CMD directive in it. Think this should work now.

webknjaz:
Let's wait for results now. I've canceled the previous build because it was blocking the queue in CI.

webknjaz:
@fabianvf build fails

PR #1639, opened 2018-12-21
Update schema_v2.py

mickelmayers:
Platforms_base_schema destroy copy.deepcopy(base_schema). Then when validating multiply molecules it's running with errors. Eg. two moleule with different drivers. One Azure second docker . If docker will be validated first from it schema we are getting 
 'image': {
                    'type': 'string',
which is not valid for azure driver but will be in schema. 
ERROR: Failed to validate.
{'platforms': [{0: [{'image': ['must be of string type']}], 1: [{'image': ['must be of string type']}], 2: [{'image': ['must be of string type']}]}]}

Moving platforms_base_schema to base_schema and changing the util.merge_dicts fixes the issue.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/molecule/model/schema_v2.py
+++ b/molecule/model/schema_v2.py
@@ -260,7 +260,31 @@ base_schema = {
             },
         }
     },
-    'platforms': {},
+    'platforms': {
+         'type': 'list',
+        'schema': {
+            'type': 'dict',
+            'schema': {
+                'name': {
+                    'type': 'string',
+                    'required': True,
+                },
+               'groups': {
+                    'type': 'list',
+                    'schema': {
+                        'type': 'string',
+                    }
+                },
+                'children': {
+                    'type': 'list',
+                    'schema': {
+                        'type': 'string',
+                    }
+                },
+            }
+        }  
+      
+    },
     'provisioner': {
         'type': 'dict',
         'schema': {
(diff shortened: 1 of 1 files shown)


lwm:
Aha! This might help with what I was tracking with https://github.com/ansible/molecule/pull/1458#issuecomment-445564794. I'll give this change a test on my branch when I am running tests and give feedback. Thanks for picking this up @mickelmayers.

Can I ask, do you have any idea how we could unit test this in isolation?

lwm:
OK, I did get a run on my PR #1458 with this commit included and saw my test failures reduced from 30+ down to only 2! Here's the 2 failures in the logs: https://travis-ci.com/ansible/molecule/jobs/167700655. Any thoughts? I think this change is definitely going in the right direction. 

By the way, the code in this commit fails all the style checks and is not signed (causes CI to fail).

Run out of time again ... :sweat:

webknjaz:
Hi @mickelmayers, could you please sign-off: https://github.com/ansible/molecule/pull/1639/checks?check_run_id=43636104

lwm:
I've transferred this change into https://github.com/ansible/molecule/pull/1458, signed it, attribute and ran the formatter over it. All the test failures have gone away and I'm finally hoping to get a green CI build. Hope that works.

PR #1640, opened 2018-12-26
sort scenarios based on directory name

duck-rh:
It is sometimes useful to run simple scenarios first and more complex
and lengthy ones only if the later were fine. Scenarios can be simply
ordered on the filesystem while avoiding ugly scenario names.

Signed-off-by: Marc Dequènes (Duck) <duck@redhat.com>

See above for how it works. I used `molecule matrix` to check the order while changing the path.

#### PR Type

- Feature Pull Request

### DIFF
--- a/molecule/scenarios.py
+++ b/molecule/scenarios.py
@@ -69,7 +69,9 @@ class Scenarios(object):
 
             return scenarios
 
-        return [c.scenario for c in self._configs]
+        scenarios = [c.scenario for c in self._configs]
+        scenarios.sort(key=lambda x: x.directory)
+        return scenarios
 
     def print_matrix(self):
         msg = 'Test matrix'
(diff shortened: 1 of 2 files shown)


lwm:
Hey @duck-rh, nice one.

This seems like a fine change to me. Although I'd like this to be configurable, I don't see how that is possible right now since the most likely place for it is the `molecule.yml` and that is already in each scenario. Adding some configurable there might not be ideal. Anyway, we see this ordering style in web server configuration world where we do things like `01_default.conf` and so on.

We'll need to add a test though, right?

duck-rh:
@lwm I added a test. I have no idea how to make it configurable. Honestly I don't think it is really needed. Currently there is no behavior specified and there is none people could rely on, so it is not breaking anything expected. And as you said yourself, it is quite a usual order in many softwares using split conffiles.

PR #1643, opened 2018-12-31
fix: EC2 create security group in correct vpc

danihodovic:
Currently the securty group is created in the default vpc. If we provide
a non-default vpc_subnet_id for the Molecule instance the security group
and instance will be in different vpcs and the instance creation will
fail with:
```
"msg": "The following group names are not valid: molecule"
```

Patch the module to create the security group in the same subnet as the
instance.

#### PR Type

- Bugfix Pull Request


danihodovic:
This is not ready for review yet... I pulled the trigger too early

PR #1644, opened 2018-12-31
fix: EC2 create security group in correct vpc

danihodovic:
Currently the security group is created in the default vpc. If we provide
a non-default vpc_subnet_id for the Molecule instance the security group
and instance will be in different vpcs and the instance creation will
fail with:
```
"msg": "The following group names are not valid: molecule"
```

Patch the module to create the security group in the same subnet as the
instance. This is done for each instance mentioned in molecule.yml
platforms.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/test/resources/playbooks/ec2/create.yml
+++ b/test/resources/playbooks/ec2/create.yml
@@ -28,12 +28,20 @@
     keypair_name: molecule_key
     keypair_path: "{{ lookup('env', 'MOLECULE_EPHEMERAL_DIRECTORY') }}/ssh_key"
   tasks:
-    - name: Create security group
+    - name: Find the vpc for the subnet
+      ec2_vpc_subnet_facts:
+        subnet_ids: "{{ item.vpc_subnet_id }}"
+      loop: "{{ molecule_yml.platforms }}"
+      register: subnet_facts
+
+    - name: Create security groups
       ec2_group:
+        vpc_id: "{{ item.subnets[0].vpc_id }}"
         name: "{{ security_group_name }}"
         description: "{{ security_group_name }}"
         rules: "{{ security_group_rules }}"
         rules_egress: "{{ security_group_rules_egress }}"
+      loop: "{{ subnet_facts.results }}"
 
     - name: Test for presence of local keypair
       stat:


lwm:
LGTM.

vutkin:
Hi @danihodovic , @lwm , why we need security group at all that differ from `default`?
We should not manage security groups on create step at all.

vutkin:
What is the case for using 2 different VPC? 
By the way we use only one vpc and multiple private subnets with `default` SG.

themr0c:
LGTM. @gundalow , @lwm, a second opinion please.

decentral1se:
Oh wait a second, this was supposed to be in https://github.com/ansible/molecule/blob/803f0959e8593f5730e9a18676f31695250b4590/molecule/cookiecutter/scenario/driver/ec2/%7B%7Bcookiecutter.molecule_directory%7D%7D/%7B%7Bcookiecutter.scenario_name%7D%7D/create.yml and not a test file!? Assume you wanted that @danihodovic? Didn't catch this on first review ...

danihodovic:
Yes that was the intention =/. Are there automated tests to catch this?

decentral1se:
Aha, well, we're not in a position, in terms of resources, to run the functional tests right now, so we're just trying to loosely maintain them until we can run them. Can you please open a PR which addes your original change to the file I've linked above? That'd be much appreciated. Will merge ASAP.

jenslauterbach:
For everyone that finds this issue and wonders how to quickly find the default `vpc_subnet_id`.

With the following command you will find the `VpcId` of your default VPC. In the example below it is `vpc-1337`.

```
$ aws ec2 describe-vpcs --filters "Name=isDefault,Values=true"
{
    "Vpcs": [
        {
            "CidrBlock": "",
            "DhcpOptionsId": "",
            "State": "",
            "VpcId": "vpc-1337",
            "OwnerId": "",
            "InstanceTenancy": "",
            "CidrBlockAssociationSet": [
                {
                    "AssociationId": "",
                    "CidrBlock": "",
                    "CidrBlockState": {
                        "State": ""
                    }
                }
            ],
            "IsDefault": true
        }
    ]
}
```

Next you need to get the available subnets of the default vpc with the id `vpc-1337` you have to run:

```
$ aws ec2 describe-subnets --filters "Name=vpc-id,Values=vpc-1337" "Name=state,Values=available"
```

This will give you a list of subnets. Probably one for every availability zone. Pick one of them and use its `SubnetId`.

See: 

https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-vpcs.html
https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-subnets.html

Issue #1637, opened 2018-12-20
Allow the specification of a custom cookiecutter template for role creation


AlexisLessard:
# Issue Type

- Feature request

# Molecule and Ansible details

```
ansible --version && molecule --version
```
Ansible: 2.7.5
Molecule: 2.19.0

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- OS package

# Desired Behavior
I tough molecule was using `ansible-galaxy init` to create a new role. I was mistaken. It's using its own sets of cookiecutter templates, which works great! However, what I like about `ansible-galaxy init` is the ability to specify a custom role skeleton. An organisation can therefore standardise the README, or implement generic tasks by default in the role. I want to be able to specify a custom cookiecutter template, and I want to be able to call `molecule init role` like so:

```
molecule init role -r role_name -d docker --template /abs/path/to/cookicutter/template
molecule init role -r role_name -d docker -t /abs/path/to/cookicutter/template
```

The option name could be changed, but I feel like this is the most appropriate. I should be working on that shortly.
Do you think there would also be a way to change the default role template? I haven't seen a config file for molecule (except molecule.yml in the role), so I'm not sure how one would got about doing that without tinkering in the source code.

PR #1647, opened 2019-01-03
Add specification of a custom cookiecutter template for role creation

AlexisLessard:
This feature has been requested in #1637 

With this simple feature, now, a user can specify a cookicutter template folder, instead of using the default one. This is similar to the `init template` command (which I discovered _after_ I began work on this ...). However, by using the `role` command, the molecule folder will also be included, meaning that an institutionnal template can be converted to use molecule instantly! 
New unit tests have been added, as I've tried coding using TDD. I created a fixture in `test_role.py`  that creates a temporary cookicutter template folder. Not sure if this is best practice or not, so feel free to recommend another way to test this. What I didn't want was to depend on fixed files in the test folder, and this was the least clutered way I found to do this.

The option ` --template` or `-t` have been added to the command. The _click_ framework validates the template path before sending it down to the code, but a failsafe for absent directory is also present. 

If someone could recommend how to handle an absent parameter that isn't required, like with this option, I'd appreciate it. Currently, this is the code before `role.execute()` is called:
```python
if template is not None:
        command_args['template'] = template
```
I feel this could be cleaner.

Unit tests have been run succesfully and I've tested the option by building the package. No issues were detected.
#### PR Type

- Feature Pull Request


lwm:
> With this simple feature, now, a user can specify a cookicutter template folder, instead of using the default one. 

Brilliant. Definitley agree with this change. Looking forward to see how this progresses.

wilmardo:
Thanks for the PR ❤️ 

I have some concerns:

First, isn't it better to add this to the `template` command instead of the role?

The following is already a valid command ([docs](https://molecule.readthedocs.io/en/latest/usage.html#molecule.command.init.template.Template)):
`molecule init template --url <url to repo with template>`

I would like to see this local template as:
`molecule init template --path <local path>`

> However, by using the role command, the molecule folder will also be included, meaning that an institutionnal template can be converted to use molecule instantly!

Don't know how I feel about this, it sounds like templating but in the background we combine another template which is not part of the initial template.
In my humble opinion, if a user is using templating he/she should provide the molecule directory in the template like:

* https://github.com/retr0h/cookiecutter-molecule
* https://github.com/LANsible/ansible-role-skeleton

> Not sure if this is best practice or not, so feel free to recommend another way to test this. What I didn't want was to depend on fixed files in the test folder, and this was the least clutered way I found to do this.

There already is a cookiecutter template in the tests, see [test/resources/templates](https://github.com/ansible/molecule/tree/master/test/resources/templates). Don't know if this is complete enough for your tests but it seems every file based dependency is defined in the `/test/resources` directory.

AlexisLessard:
Thanks for your feedback, @lwm  and @wilmardo !

> Don't know how I feel about this, it sounds like templating but in the background we combine another template which is not part of the initial template.

Well, isn't this what the role command does already? It uses a cookiecutter template in [molecule/molecule/cookiecutter/role/](https://github.com/ansible/molecule/tree/master/molecule/cookiecutter/role) to create a role, and then adds the [molecule/molecule/cookiecutter/molecule/](https://github.com/ansible/molecule/tree/master/molecule/cookiecutter/molecule) template inside the resulting folder. Is that what you mean by combining templates?
What I want to do is allow the user to override the first template. Beside the role template path, everything is the same.

> There already is a cookiecutter template in the tests, see test/resources/templates

Woops! Missed that. I could use it inside the tests instead of using a dynamicaly generated template. I'll check how the template command is tested and try to reproduce it. It could dramatically reduce the test file's sizes. I'm still not sure that using a static file inside a unit test is totally safe, but if there's already a use of it, I'll do it as well, as cleanly as possible. Thank you!

> First, isn't it better to add this to the template command instead of the role?

If I understand correctly, the template command only creates a role based on a cookiecutter template. It looks like a simple command line utility to process those templates from remote repositories. In a way, it could be used for any template, not just ansible roles. What I wanted is to have a cookiecutter template for the role, like the `template` command does, and like the `ansible-galaxy` allows with their custom role skeleton, but then add everything necessary to test the template with molecule. If I want that with the `template` command, I have to add the molecule folder manually to my template, which means I have to understand the configuration to reproduce it, which is not something I feel some users are willing to do. The default molecule configuration can work in most cases, so why not include it in my template? 
To be honest, I didn't know the template command before I started working on this, but I still stand by my choice to put it in the role command.

AlexisLessard:
I'd like to add support for templates in a remote repository, but this might be for another PR, if this one gets approved.

AlexisLessard:
Closing this and reopening it to trigger a build

AlexisLessard:
This PR is now ready for review and approval! @lwm, you might wanna check it out.

lwm:
To try and further clarify, if I wanted to use a custom role template and have molecule generate the scenario template for me, I would have to do:

```
$ molecule init template -t my-role-template  # my-role-template has no molecule folder
$ cd my-role-template
$ molecule init scenario -r my-role-template  # give me the up to date molecule folder
```

Whereas, now, with this change, all I need to do is:

```
$ molecule init role -t my-role-template
```

And I get my custom role template with the latest scenario template.

Correct?

molecule/command/init/base.py
```diff
@@ -68,3 +74,8 @@ def _resolve_template_dir(self, template_dir):
                 'cookiecutter', template_dir)
 
         return template_dir
+
+    def _validate_template_dir(self, template_dir):
+        if not os.path.isdir(template_dir):
+            util.sysexit_with_message(
+                "The specified template directory does not exist")
```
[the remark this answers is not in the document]
AlexisLessard: Seems easy enough. I'll work on it.

AlexisLessard:
> Whereas, now, with this change, all I need to do is:
`$ molecule init role -t my-role-template`
And I get my custom role template with the latest scenario template.
Correct?

Pretty much, yeah. I don't want the user to worry about what will be in the molecule folder. Molecule itself should manage that (At least, that's what I think). But if the user wants to use molecule, there's a fair chance he knows ansible, and therefore may want a custom template (with a standardised readme, or no files or templates directory, depending on what the organization requires).  `init template` can be used if the user wants to manage everything. It's a nice tool to have (altough, next, we could add support for local templates as well as url). I'm not sure how to use the `scenario` command, tough, and how my PR compliments it. 

> Now, if we're about overriding only the outer role template and not the molecule scenario template, then adding this to the init role is the right way to go.

Yep, that's it! We use the `molecule init role` command to initiate a new role, and to then test it with molecule. We do not just want to implement a template, but we want to use one, just for the role part. Just like a custom `ansible-galaxy init` squeleton, but with the bonus of the molecule testing framework!

lwm:
@wilmardo, further thoughts? @AlexisLessard, do you want to add documentation for the differences of the `init ...` commands? I can also take a stab at it if not.

AlexisLessard:
@lwm I don't mind doing a bit of documentation. but I don't know what and where I'd document it. Can I modify the official doc at molecule.readthedocs.io? If so, how? 

Also, I'll retriger a build once the master build passes on travis.

lwm:
@AlexisLessard, cool! `molecule.readthedocs.io` is made up of the files in https://github.com/ansible/molecule/tree/master/doc/source. You'll probably want to `grep` around for the right place. In the `docs/` folder, you can run `make` to see the available `sphinx` commands that you can run. You can run `tox -e docs` (or something like that) to verify like the CI will as well.

AlexisLessard:
@lwm  I've added a little paragraph explaining the option in the documentation. If this fits you, my PR would be ready.
On a fun note, I've juste learned what sphinx is!
On a less fun note, it took me an hour to understand how to generate 4 lines of documentation... All knowledge is good knowledge, I say!

molecule/command/init/role.py
```diff
@@ -38,6 +38,16 @@ class Role(base.Base):
     .. option:: molecule init role --role-name foo
 
         Initialize a new role.
+
+    .. program:: molecule init role --role-name foo --template path
+
+    .. option:: molecule init role --role-name foo --template path
+
+        Initialize a new role using a local *cookiecutter* template. This allows
```
lwm: How about:

> Initialize a new role using a local *cookiecutter* template. This allows the customization of a role while still using the default generated ``molecule`` folder. Please refer to the ``init scenario`` command in order to customise the default generated ``molecule`` folder.

AlexisLessard:
@lwm, What are scenarios in molecule, and how are they used? How do they relate to the template option? I can't seem to find it in the documentation. I could add it in another PR, but for now, I want to understand so I can make a proper doc for this PR.

lwm:
@AlexisLessard, the documentation is not the best right now but there is a PR out (I made it) with a 'Getting Started Guide' which tries to cover an introductor explanation of what scenarios are. See https://github.com/ansible/molecule/pull/1719/files#diff-87a7eaaee165971087f4caa913a88277R35. Other than that, we currently have https://molecule.readthedocs.io/en/latest/configuration.html#scenario. Hope that is enough to get you going! Any docs PR you can add would be really appreciated.

molecule/command/init/role.py
AlexisLessard: I changed it, but I still want to include the ansible-galaxy parallel. That was the whole point of this PR so I really want it to be explicit.

AlexisLessard:
Closinga and reopening to trigger a build

decentral1se:
Thanks for your patience @AlexisLessard, pushing this onto the v2.21 train :rocket:

AlexisLessard:
Perfect| Thank you @decentral1se! Was the documentation I wrote helpful? If so, will you approve the changes?

decentral1se:
I don't have cycles to take another review pass just now but I will do once v2.20 drops!

.gitignore
```diff
@@ -15,3 +15,4 @@ pytestdebug.log
 ubuntu-xenial-16.04-cloudimg-console.log
 .python-version
 coverage.xml
+.vscode/
```
wilmardo: Might be nice to clean this up, if everyone is adding their IDE to the gitignore the file will grow humongous :)

molecule/command/init/role.py
```diff
@@ -38,6 +38,17 @@ class Role(base.Base):
...
         Initialize a new role.
+
+    .. program:: molecule init role --role-name foo --template path
+
+    .. option:: molecule init role --role-name foo --template path
+
+        Initialize a new role using a local *cookiecutter* template. This
+        allows the customization of a role while still using the default
+        generated ``molecule`` folder. This is similar to an
+        ``ansible-galaxy init`` skeleton. Please refer to the ``init scenario``
```
wilmardo: Might be better to rephrase to:

Please refer to the ``init scenario`` command in order to generate a custom ``molecule`` scenario.

Since you aren't customizing the default scenario since it already exists, right?

molecule/command/init/role.py
```diff
@@ -107,8 +124,15 @@ def execute(self):
     type=click.Choice(config.molecule_verifiers()),
     default='testinfra',
     help='Name of verifier to initialize. (testinfra)')
+@click.option(
+    '--template',
+    '-t',
+    type=click.Path(
+        exists=True, dir_okay=True, readable=True, resolve_path=True),
+    help="Path to a cookiecutter custom template to initialize the role. "
+    "The molecule folder will be added to this template")
```
wilmardo: The upstream molecule folder will be added to this template

Might clarify it a bit further

molecule/command/init/role.py
AlexisLessard: I  would also change 

> default generated ``molecule`` folder

  to

> "upstream ``molecule`` folder"

like you said on your other change request. It does feel more appropriate. Thanks!

.gitignore
AlexisLessard: @wilmardo , how would you suggest we do that? A comment to introduce a section for IDE settings? Or just remove it alltogether?

wilmardo: In my opinion just remove it altogether but the Ansible guys might have something to say about that ;)

AlexisLessard: I just learned that there is a local gitignore that I can configure (`.git/info/exclude`). Feels more appropriate. I'll revert the commit.

AlexisLessard: Done

molecule/command/init/role.py
AlexisLessard: Done

molecule/command/init/base.py
```diff
@@ -52,14 +53,19 @@ def _process_templates(self,
...
+            cookiecutter.main.cookiecutter(
+                template_dir,
+                extra_context=extra_context,
+                output_dir=output_dir,
+                overwrite_if_exists=overwrite,
+                no_input=True,
+            )
+        except cookiecutter.exceptions.NonTemplatedInputDirException:
+            util.sysexit_with_message(
+                "The specified template directory is in an invalid format")
```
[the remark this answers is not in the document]
AlexisLessard: Done

molecule/command/init/base.py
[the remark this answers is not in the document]
AlexisLessard: Done

molecule/command/init/role.py
```diff
@@ -107,8 +113,15 @@ def execute(self):
     type=click.Choice(config.molecule_verifiers()),
     default='testinfra',
     help='Name of verifier to initialize. (testinfra)')
+@click.option(
+    '--template',
+    '-t',
+    type=click.Path(
+        exists=True, dir_okay=True, readable=True, resolve_path=True),
+    help="Path to a cookicutter custom template to initialize the role. "
```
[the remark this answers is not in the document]
AlexisLessard: Done

molecule/command/init/role.py
AlexisLessard: Done

AlexisLessard:
I'll reopen and close this PR to trigger a build when the main build will pass. Afterwards, it should be done.

AlexisLessard:
Closing and reopening to trigger a build.

Issue #1645, opened 2019-01-02
[Build an Ansible compatible image fails with fedora:rawhide container


ssbarnea:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible 2.7.5
  config file = /Users/ssbarnea/.ansible.cfg
  configured module search path = [u'/Users/ssbarnea/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/ssbarnea/.pyenv/versions/2.7.15/lib/python2.7/site-packages/ansible
  executable location = /Users/ssbarnea/.pyenv/versions/2.7.15/bin/ansible
  python version = 2.7.15 (default, Oct 17 2018, 20:02:57) [GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.11.45.2)]
molecule, version 2.19.1.dev55
```

# Desired Behavior

Image build should suceed on rawhide image.

# Actual Behaviour

The root cause is that `python2-dnf` package no longer exists being replaced by `python3-dnf`. 
Verbose output available at https://gist.github.com/ssbarnea/792000358390a75757b206a2cb38f943

Running molecule in debug mode does give more information
```
No match for argument: python2-dnf
```

To reproduce the problem just try to use this image:
```
platforms:
  - name: fedora-rawhide
    hostname: fedora-rawhide
    image: fedora:rawhide
```

MarkusTeufelberger:
That was one of the reasons for me to write #1642 - that way the installation could be done by Ansible instead of a shell script and it would be also maybe a bit faster to spin up containers (no need to build them first).

christiangda:
Hi everyone,

For my is failing with "Amazon Linux 1" after I update to Ansible  2.8 this morning.

```bash
 TASK [christiangda.epel_repo : Install EPEL repository on Amazon Linux] ********
    fatal: [amazonlinux-1]: FAILED! => {"changed": false, "cmd": "dnf install -y python2-dnf", "msg": "[Errno 2] No such file or directory", "rc": 2}
    changed: [amazonlinux-2]
```

To reproduce it:

**Prepare your environment**
```bash
mkdir ansible-roles
cd ansible-roles/

python3 -m venv venv
source venv/bin/activate
pip install pip --upgrade
pip install ansible">2.7" # <-- Fail, pip install ansible"<2.8"  <-- Work
pip install pytest
pip install pytest-mock
pip install pylint
pip install rope
pip install autopep8
[...]

PR #1649, opened 2019-01-04
Avoid failure to prepare docker image on fedora-rawhide

### LINKED ISSUE #1642: Proposal: Use prepare.yml instead of rendering Dockerfiles with the Docker driver
# Issue Type

- Feature request

# Desired Behavior

Similar to Vagrant it might be beneficial to let Ansible itself install Python/Bash/Sudo/??? to various containers in a (pre-rendered) `prepare.yml` playbook and instead using the images listed in "platforms" unmodified. The current approach (render https://github.com/ansible/molecule/blob/master/molecule/cookiecutter/scenario/driver/docker/%7B%7Bcookiecutter.molecule_directory%7D%7D/%7B%7Bcookiecutter.scenario_name%7D%7D/Dockerfile.j2 and build a custom image from that) seems a bit out of place, after all Vagrant boxes for example are also used unmodified instead of rendering out a Packer config and building the VM from scratch.

ssbarnea:
Fedora Rawhide dropped the python2-dnf package and now has only
python3-dnf one.

By using a wildcard we avoid the installation failure as it will
install all found versions.

Closes #1642

#### PR Type

- Bugfix Pull Request
- Feature Pull Request


MarkusTeufelberger:
Uhm, this doesn't necessarily close #1642 but probably it is a typo and you meant #1645

Issue #1657, opened 2019-01-08
test-requirements.txt must pin pytest-cov==2.6.0


singuliere:
# Issue Type

- Bug report

pytest-cov [2.6.1 was released January 7th](https://pypi.org/project/pytest-cov/2.6.1/) and [adds a pytest>=3.6 requirement](https://github.com/pytest-dev/pytest-cov/compare/v2.6.0...master#diff-2eeaed663bd0d25b7e608891384b7298R124).

pytest cannot be upgraded because it leads to other errors.

pytest-cov must be pinned to 2.6.0.

# Desired Behavior

Travis CI tests passes

# Actual Behaviour

Travis CI tests [fail](https://travis-ci.com/ansible/molecule/jobs/168821692) with:

```
pytest.vendored_packages.pluggy.PluginValidationError: Plugin 'pytest_cov' could not be loaded: (pytest 3.0.7 (/home/travis/build/ansible/molecule/.tox/ansible27-unit/lib/python2.7/site-packages), Requirement.parse('pytest>=3.6'))!
```

PR #1658, opened 2019-01-08
pin pytest-cov==2.6.0

singuliere:
Fixes: https://github.com/ansible/molecule/issues/1657

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -2,7 +2,7 @@ mock
 pur
 # pytest==3.1.0 issues with pytest-verbose-parametrize
 pytest==3.0.7
-pytest-cov
+pytest-cov==2.6.0
 pytest-helpers-namespace
 pytest-mock
 pytest-verbose-parametrize


test-requirements.txt
```diff
@@ -2,7 +2,8 @@ mock
 pur
 # pytest==3.1.0 issues with pytest-verbose-parametrize
 pytest==3.0.7
-pytest-cov==2.6.0
+pytest-cov<2.6.1; python_version < "3.6"
```
singuliere: the changed requirement is for [pytest>=3.6](https://github.com/pytest-dev/pytest-cov/compare/v2.6.0...master#diff-2eeaed663bd0d25b7e608891384b7298R124) and not python 3.6

singuliere:
Re-pushed after checking that [travis CI](https://travis-ci.com/ansible/molecule/builds/96588015) fails with https://github.com/ansible/molecule/pull/1658/commits/966e9f39f1a01cc21d4656a30f2f7b768cd2e1c9

test-requirements.txt
webknjaz: So it should be pinned with markers too

singuliere:
Two [tests fail](https://travis-ci.com/ansible/molecule/jobs/168882598) but there are no error messages:

```
----------------------------- Captured stderr call -----------------------------
ERROR:
--------------------------- Captured stdout teardown ---------------------------
```

singuliere:
@webknjaz given the current instability of Travis and the fact that most tests pass, would you consider merging this pull request? Without it everything fails so the situation is worse.

singuliere:
@webknjaz thanks for the close/open pull request trick to trigger the CI, much more convenient that repushing. Fingers crossed that travis will be faster in the morning :crossed_fingers:

singuliere:
The tests all pass based on the **two** following Travis CI runs

* https://travis-ci.com/ansible/molecule/builds/96645217
* https://travis-ci.com/ansible/molecule/builds/96714750

all tests pass. A [single job](https://travis-ci.com/ansible/molecule/jobs/169132028) fails in [this run](https://travis-ci.com/ansible/molecule/builds/96714750) but the [same job succeeds](https://travis-ci.com/ansible/molecule/jobs/168882600) in [the other run](https://travis-ci.com/ansible/molecule/builds/96645217).

webknjaz:
I've done some restarts, let's see how it goes. It looks like we'll need to split more jobs after all...

test-requirements.txt
```diff
@@ -2,7 +2,7 @@ mock
 pur
 # pytest==3.1.0 issues with pytest-verbose-parametrize
 pytest==3.0.7
-pytest-cov
+pytest-cov==2.6.0
```
webknjaz: I still want conditionals I think.

webknjaz: env markers. anyway, let's merge it as is for now. but for future I'd like to upgrade pytest, there's even a new major version out there.

singuliere: I'm not sure what env marker would be relevant in that context, hence my question. There is no dependency on a specific version of python and I can't think of any other env marker.

singuliere:
thanks for the merge :-)

Issue #1308, opened 2018-05-20
Don't fail on yamllint warnings


cam-fulcrum:
# Issue Type

- Feature request
- Bug

# Molecule and Ansible details

```
ansible 2.5.3

```

```
molecule, version 2.14.0
```

Molecule installation method (one of):

- pipenv

Ansible installation method (one of):

- pipenv

# Desired Behavior

Hi, I'd like to continue to run tests even if `yamllint` gives warnings. I understand the need for linting, but its pretty frustrating to have to fix lint warnings even to run a test. I'd prefer not to have to change settings in `.yamllint` too.

For example, when I run `molecule test`, it runs the linter and then fails.

```
    /home/me/dev/ansible-roles/role/defaults/main.yml
      9:55      warning  too few spaces before comment  (comments)
        
    /home/me/dev/ansible-roles/role/molecule/default/inventory/group_vars/all.yml
      1:1       warning  missing document start "---"  (document-start)
```
And then it just stops executing. Is this a bug or desired behaviour?
Thanks for your hard work!

retr0h:
> And then it just stops executing. Is this a bug or desired behaviour?

Yes, this is the desired functionality.  If linting is wrong, one should fix their project or ignore the rules they don't like.   Molecule is about ensuring projects conform to the best practices.

cam-fulcrum:
Thanks for your reply!
Ok, so is there a configurable option to continue with warnings? I had a look at the lint configuration and I couldn't see any option.

retr0h:
> Ok, so is there a configurable option to continue with warnings? I had a look at the lint configuration and I couldn't see any option.

Hi @cam-fulcrum the functionality to exit if lint fails cannot be changed.  You must either modify your .yamllint file to ignore the errors or correct the errors.

Hope this helps.

cam-fulcrum:
Ok cool, thanks!

vutkin:
Hello, @retr0h , could you kindly clarify why molecule test always failed if there is only warnings in yamllint *AND* official docs (http://yamllint.readthedocs.io/en/stable/configuration.html#errors-and-warnings) says:
```
By default the script will exit with a return code 1 only when there is one or more error(s).
```

TIA.

retr0h:
> I think it should not be default behavior.

Molecule is opinionated.  This is my opinion as to what it should do.  Developers should fix their linting issues or exempt them.

vutkin:
[...]

PR #1659, opened 2019-01-08
Disable YAML lint truthy rule by default

samdoran:
Ansible accepts flexible values for booleans and the default YAML lint rules should match what Ansible supports rather than enforcing strict `true/false` values. Ansible will continue to support the 1.1 pragma, e.g., `yes/no/true/false` (and more) values for boolean variables, even after the YAML 1.2 spec is released.

In the current default state, `molecule test` will fail when running `yamllint` against YAML files that Ansible considers valid. Molecule should not fail on files that Ansible considers valid.

During the [Ansible Docs meeting today](https://meetbot.fedoraproject.org/ansible-docs/2019-01-08/docs_working_group.2019-01-08-15.31.html), it was decided to use `yes/no/true/false` values in Ansible documentation to reduce confusion and be consistent.

This change aligns Molecule with what Ansible considers valid boolean values as well as Ansible documentation going forward.

There is a [feature request](https://github.com/adrienverge/yamllint/issues/150) to `yamllint` to have the `truthy` rule accept a list of valid values. If/when this is implemented in `yamllint`, we could update the Molecule defaults to only pass on the recommended boolean values, which are `yes/no/true/false` per the docs meeting earlier today.

Related:
#1308 
https://github.com/adrienverge/yamllint/issues/150

Signed-off-by: Sam Doran <sdoran@redhat.com>

#### PR Type

- Feature Pull Request

### DIFF
--- a/molecule/driver/delegated.py
+++ b/molecule/driver/delegated.py
@@ -88,7 +88,7 @@ class Delegated(base.Base):
 
     .. important::
 
-        It is the developers responsibility to configure the ssh config file.
+        It is the developer's responsibility to configure the ssh config file.
 
     .. code-block:: yaml
 
--- a/molecule/lint/yamllint.py
+++ b/molecule/lint/yamllint.py
@@ -80,10 +80,21 @@ class Yamllint(base.Base):
 
     @property
     def default_options(self):
-        return {
+        opts = {
             's': True,
-            'truthy': False,
         }
+
+        # Config file and default flags are mutually exclusive. If 'c' or
+        # 'config-file' are specified, do not override the default
+        config_file_flags = set([
+            'c',
+            'config-file',
+        ])
+        lint_options = set(self._config.config['lint'].get('options', {}).keys())
+        if not config_file_flags.intersection(lint_options):
+            opts['d'] = '{extends: default, rules: {truthy: disable}}'
+
+        return opts
 
     @property
     def default_env(self):
--- a/molecule/lint/yamllint.py
+++ b/molecule/lint/yamllint.py
@@ -80,22 +80,9 @@ class Yamllint(base.Base):
 
     @property
     def default_options(self):
-        opts = {
+        return {
             's': True,
         }
-
-        # Config file and default flags are mutually exclusive. If 'c' or
-        # 'config-file' are specified, do not override the default
-        config_file_flags = set([
-            'c',
-            'config-file',
-        ])
-        lint_options = self._config.config['lint'].get('options', {})
-        lint_option_keys = lint_options.keys()
-        if not config_file_flags.intersection(lint_option_keys):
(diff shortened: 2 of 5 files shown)


samdoran:
I just realized this isn't the correct fix. I need to disable the `truthy` rule, which must be done via rules config file not through a command line option. That's actually somewhat tricky. I'll play with `-d` to see if that will work.

samdoran:
I could use some help figuring out the test failures. Most of them seem to be related to the pinned version of `pytest`:

```
pkg_resources.VersionConflict: (pytest 3.0.7 (/home/travis/build/ansible/molecule/.tox/ansible27-functional/lib/python3.7/site-packages), Requirement.parse('pytest>=3.6'))
```

test/unit/lint/test_yamllint.py
```diff
@@ -174,6 +183,12 @@ def test_execute_bakes(_patched_get_files, patched_run_command, _instance):
     x = [
         str(sh.Command('yamllint')),
         '-s',
+        '-d',
+        "'{extends:",
```
webknjaz: It looks like you don't need most of the commas here.

webknjaz:
@samdoran it's probably because of #1658

molecule/lint/yamllint.py
```diff
@@ -31,14 +31,32 @@
...
-    lines length, trailing spaces, and indentation.
+    `yamllint`_ is a linter for YAML files. In addition to checking for syntax
+    validity it also checks for key repetition as well as cosmetic problems such as
+    line length, trailing spaces, and indentation.
 
-    Additional options can be passed to `yamllint` through the options
-    dict.  Any option set in this section will override the defaults.
+    The default ``yamllint`` settings that ship with ``molecule`` are:
+
+    .. code-block: yaml
```
lwm: I wonder if this won't be getting out-of-date very soon (extra maintenance issue for developers, misleading for users). Should it better to say here molecule spits out the defaults and also perhaps how? As far as I know, that means init'ing a scenario but maybe there is a separate command for that too?

samdoran: Very good point. I changed this to a link instead.

gundalow:
@lwm Could you please re-review and update your Review
@samdoran When you get a chance could you please rebase

molecule/driver/delegated.py
```diff
@@ -89,8 +89,8 @@ class Delegated(base.Base):
             managed: False
             login_cmd_template: 'ssh {instance} -F /tmp/ssh-config'
             ansible_connection_options:
-              ansible_connection: ssh
-              ansible_ssh_common_args -F /path/to/ssh-config
+                ansible_connection: ssh
```
webknjaz: We don't need to have extra indentation here. Two spaces is pretty typical for yaml

Issue #1669, opened 2019-01-16
The CI is too slow


decentral1se:
# Issue Type

- Gripe about the CI

On https://travis-ci.com/ansible/molecule/builds/97459293, we saw:

```
Ran for 1 hr 26 min 9 sec
Total time 4 hrs 59 min 45 sec 
```

Hopefully this ticket can be a place to gather ideas on how to speed this process up.

PS. I am thankful for all the great work that has been done to get this CI in place.

ragingpastry:
Is this something that could be fixed using a tool like detox?

ssbarnea:
I am not sure if there is much to be done about it as molecule needs to test a huge combination of python versions, ansible versions, platforms/distributions, containers/vms/clouds...

Even with a paid travis plan this could be limited. I wonder if using openstack build infrastructure would not be a better way to test changes.

ragingpastry:
Detox would help the execution of python / ansible tests as they would happen concurrently which could speed some things up. Maybe "fixed" was the incorrect word to use but I could see concurrent builds working fairly quickly.

lwm:
OK, `detox` might be a good first step then but I see in that repository link:

> detox is unmaintained and incompatible with tox > 3.6. detox was a plugin for tox to enable parallel environment execution. tox 3.7 added a native possibility to do this (tox -p|--parallel) and effectively supercedes detox.

So, perhaps we can improve by simply passing `-p`!

lwm:
OK, after some experimenting. I see that we can't get anything out of `detox/tox -p` because we run each build environment in isolation. We can't parallelize this. I thought it might be useful in that it would speed up somehow these individual builds (by using multiple cores) but that functionality doesn't seem to offered.

themr0c:
Would `sitepackages = true` in tox.ini bring some improvement?

lwm:
@themr0c, good idea about the requirements! Not sure about `sitepackages`. Let's try these.

And just to link, https://github.com/ansible/molecule/pull/1698 looks like it will definitely help!

webknjaz:
Related to: https://github.com/ansible/molecule/issues/1702#issuecomment-457395524

Once the problem with parallel Molecule execution is solved, we can try using thread-based parallelism which via pytest-xdist. It could improve things dramatically.

themr0c:
[...]

PR #1681, opened 2019-01-20
Try to pass `--parallel all` for tox tests.

lwm:
Something for https://github.com/ansible/molecule/issues/1669.


lwm:
Ah, wrong arguments. Well, I tested it locally now.  Closing out since https://github.com/ansible/molecule/issues/1669#issuecomment-455909886.

PR #1691, opened 2019-01-23
List all python distribution extras in tox env declaration

### LINKED ISSUE #1689: tox.ini should reference all extras present in setup.cfg
# Issue Type

- Bug report

# Actual Behaviour

Some extras present in setup.cfg are missing in tox.ini and should be added.

In setup.cfg:

```
[extras]
docs =
    alabaster
    Sphinx
azure =
    ansible[azure]
docker =
    docker>=2.0.0
ec2 =
    boto
    boto3
gce =
    apache-libcloud
linode =
    linode-python; python_version<"3.0"
lxc =
    lxc-python2; python_version<"3.0"
openstack =
    shade
vagrant =
    python-vagrant
windows =
    pywinrm
```

in tox.ini:

```
extras =
    azure
    docker
    gce
    lxc
    openstack
    vagrant
```

themr0c:

Closes #1689

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/test/functional/conftest.py
+++ b/test/functional/conftest.py
@@ -268,6 +268,14 @@ def get_docker_executable():
     return distutils.spawn.find_executable('docker')
 
 
+def get_linode_executable():
+    try:
+        pytest.importorskip('linode')
+        return True
+    except Exception:
+        return False
+
+
 def get_lxc_executable():
     return distutils.spawn.find_executable('lxc-start')
 
@@ -291,11 +299,8 @@ def supports_docker():
 
 @pytest.helpers.register
 def supports_linode():
-    try:
-        pytest.importorskip('linode')
-        return True
-    except Exception:
-        return False
+    # FIXME: Travis CI
+    return not IS_TRAVIS and get_lxd_executable()
 
 
 @pytest.helpers.register
--- a/test/functional/conftest.py
+++ b/test/functional/conftest.py
@@ -300,7 +300,7 @@ def supports_docker():
 @pytest.helpers.register
 def supports_linode():
     # FIXME: Travis CI
-    return not IS_TRAVIS and get_lxd_executable()
+    return not IS_TRAVIS and get_linode_executable()
 
 
 @pytest.helpers.register
--- a/tox.ini
+++ b/tox.ini
@@ -24,10 +24,14 @@ deps =
 extras =
     azure
     docker
+    docs
+    ec2
     gce
+    linode
     lxc
     openstack
     vagrant
+    windows
 commands =
     unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
     functional: pytest test/functional/ {posargs}


tox.ini
```diff
@@ -24,10 +24,14 @@ deps =
...
     azure
     docker
+    docs
+    ec2
     gce
+    linode
     lxc
     openstack
     vagrant
+    windows
```
webknjaz: Is this conditional in `setup.cfg`

webknjaz:
Also linters fail

tox.ini
themr0c: I don't understand ?

themr0c:
@webknjaz You are talking about the linting issue there,  https://travis-ci.com/ansible/molecule/jobs/172385408 : all builds are failing on lint and this PR is the attempt to fix the lint issue https://github.com/ansible/molecule/pull/1695

webknjaz:
I'd probably want to fix linting issues in a separate PR so that those changes wouldn't pollute other PRs

tox.ini
webknjaz: I'm curious whether it will try installing incompatible dep under Linux and then running tests against it

themr0c: Ah ! certainly not, the windows extra is to install pywinrm, the module to be able to connect to windows from ansible.

lwm:
Can we merge this?

themr0c:
Sorry i failed a rebase again, hence the push --force. But it is just the exact same commit rebased on master to relaunch the build which failed on a weird error.

themr0c:
Unfortunately, with this PR the  `linode` modules get installed, linode tests that were skipped get activated, and they fail.

lwm:
Riiighhhhht, OK, we can't merge this until we have a strategy for these functional tests that talk across the network.

webknjaz:
@themr0c right, that's what I thought...
You can modify https://github.com/ansible/molecule/blob/6502649/test/functional/conftest.py#L268-L274 to ignore travis as a workaround.

themr0c:
If i didn't make any enormous mistake in last commit, linode functional tests should be now skipped in travis.

test/functional/conftest.py
```diff
@@ -268,6 +268,14 @@ def get_docker_executable():
     return distutils.spawn.find_executable('docker')
 
 
+def get_linode_executable():
+    try:
+        pytest.importorskip('linode')
+        return True
+    except Exception:
```
webknjaz: What exception is actually happening here? AFAIK `pytest.importorskip` with just return `None` if there's nothing to import. Which means that this check'd always return `True`.

`pytest.importorskip` is specifically designed to trigger skipping the current test anyway so I don't know why you would wrap it like this.

webknjaz: Also, a bunch of unrelated changes makes it difficult to guess what is the purpose of the PR...

webknjaz: Oh.. I see now.

test/functional/conftest.py
```diff
@@ -291,11 +299,8 @@ def supports_docker():
 
 @pytest.helpers.register
 def supports_linode():
-    try:
-        pytest.importorskip('linode')
-        return True
-    except Exception:
-        return False
+    # FIXME: Travis CI
+    return not IS_TRAVIS and get_lxd_executable()
```
lwm: `%s/get_lxd_executable/get_linode_executable/`

test/functional/conftest.py
webknjaz: I'm not sure how this check works but it's probably not related to the scope of the current PR. So let's ignore it for now.

test/functional/conftest.py
webknjaz: oh... I missed that. btw you can suggest changes in PRs with `Ctrl+G`:
```suggestion
    return not IS_TRAVIS and get_linode_executable()
```

lwm:
@themr0c, I think https://github.com/ansible/molecule/pull/1691#discussion_r258524200 is relevant in that, `importorskip` won't actually throw an exception. Let's make sure about that.

webknjaz:
@lwm while it's relevant to the initial code, it's just copy-pasted without changes so I suggest to address it separately in order to not overload PR with extra changes

test/functional/conftest.py
themr0c: I just wanted to align on the other platforms where there is a split between def_xxx_executable and a supports_xxx.
The try and exception is just a copy paste during the split, nothing new here.

webknjaz: Yes, I noticed that already and suggested to address it separately in other comments.

lwm:
:rocket:

Issue #1575, opened 2018-11-09
[TODO] Redesign sharding in pytest


webknjaz:
# Issue Type

- Feature request

# Molecule and Ansible details

`devel`

Molecule installation method (one of):

N/A

Ansible installation method (one of):

N/A

# Desired Behavior

Based on https://github.com/pytest-dev/pytest/issues/4359#issuecomment-437342535, we need to have a slightly better sharding implementation. The one I've built in #1541 is relatively fragile.

PR #1698, opened 2019-01-23
fixes #1697 shard travis builds for python2

### LINKED ISSUE #1697: shard travis builds for python 2
# Issue Type

- Bug report

# Desired Behavior

Travis builds for python 2.7 should get sharded as the builds for python 3

# Actual Behaviour

Builds are not sharded, and sometimes it causes time limit error.

```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

Example: https://travis-ci.com/ansible/molecule/jobs/172385426

themr0c:

fixes #1697 shard travis builds for python2

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
----
 conditions: v1
 dist: xenial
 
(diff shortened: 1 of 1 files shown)


themr0c:
NB: if it appears to be impossible, there is already some ticket to propose some more deep enhancement #1575

.travis.yml
```diff
@@ -48,6 +48,16 @@ language: python
 - &env-functional-shard_3
   <<: *env-functional
   PYTEST_ADDOPTS: "'-m shard_3_of_3'"
+- &py27-env-functional-shard_1
```
webknjaz: This set of YAML anchors is not needed: those above hold exact copies of the same values as below.
Also, you call them "py27" while there's literally nothing pointing to python 2.7, they are as generic as ones without such prefix.

.travis.yml
```diff
@@ -72,164 +82,229 @@ jobs:
     env:
       ANSIBLE_VERSION: "27"
       TESTS_TYPE: unit
+    name: unit tests, Ansible 2.7, Python 3.7
```
webknjaz: I'd prefer having UI changes in a separate PR. Having a bunch of mixed patches in one PR makes it non-atomic and hard to review as it's fairly complicated to find bits which are worth checking and easy to miss important changes.
So please let's avoid having such white noise in patches. Also, this change is clearly not "fixes #1697 shard travis builds for python2"

.travis.yml
```diff
@@ -72,164 +82,229 @@ jobs:
...
   - <<: *py-36
     <<: *if-cron-or-manual-run-or-tagged
     env:
       <<: *env-functional-shard_3
       ANSIBLE_VERSION: "25"
+    name: functional tests shard 3/3, Ansible 2.5, Python 3.6
+
+  - <<: *py-27
+    env:
+      <<: *py27-env-functional-shard_1
```
webknjaz:
```suggestion
      <<: *env-functional-shard_1
```

.travis.yml
```diff
@@ -72,164 +82,229 @@ jobs:
...
+      ANSIBLE_VERSION: "27"
+    name: functional tests shard 1/3, Ansible 2.7, Python 2.7
 
   - <<: *py-27
     env:
+      <<: *py27-env-functional-shard_2
```
webknjaz:
```suggestion
      <<: *env-functional-shard_2
```

.travis.yml
```diff
@@ -72,164 +82,229 @@ jobs:
...
       ANSIBLE_VERSION: "27"
-      TESTS_TYPE: functional
+    name: functional tests shard 2/3, Ansible 2.7, Python 2.7
 
   - <<: *py-27
     env:
+      <<: *py27-env-functional-shard_3
```
webknjaz:
```suggestion
      <<: *env-functional-shard_3
```

.travis.yml
```diff
@@ -72,164 +82,229 @@ jobs:
...
+      ANSIBLE_VERSION: "27"
+    name: functional tests shard 3/3, Ansible 2.7, Python 2.7
+
+  - <<: *py-27
+    env:
+      <<: *py27-env-functional-shard_1
       ANSIBLE_VERSION: "26"
-      TESTS_TYPE: functional
```
webknjaz: Removal of such env var WILL BREAK construction of `TOXENV` var. Don't do this.

.travis.yml
webknjaz: P.S. I've extracted names into the master

.travis.yml
themr0c: I am really happy to find someone pickier than me \o/.
PYTEST_ADDOPTS: "'-m shard_3_of_3'", with single quotes, is functional wit python3, so i didn't change it. But it didn't work with python2
The attempt to remove the single quotes failed miserably PYTEST_ADDOPTS: "-m shard_3_of_3"
The current attempt to without the single quotes but with an = in place of the space appears to be functional PYTEST_ADDOPTS: "-m=shard_3_of_3".
Indeed, now, good question: is it better to keep an exception for py2, or is it better to, at least, try the "-m=shard_3_of_3" for all versions pf python?

.travis.yml
themr0c: Isn'tit done in py27-env-functional-shard_1 ?

themr0c:
@webknjaz Thank you for your comments, much appreciated. I replied to some, and I will adapt after you confirm the final direction to take.

.travis.yml
webknjaz: I'm fine with both, I just wanted to reduce duplication which leads to even more diverse semi-copypasted approaches in future

.travis.yml
webknjaz: Oh, the conflict is not yet resolved

.travis.yml
webknjaz: Oh, right.. I overlooked it

.travis.yml
themr0c: done

.travis.yml
themr0c: done

.travis.yml
```diff
@@ -244,22 +247,57 @@ jobs:
...
+      ANSIBLE_VERSION: "26"
+    name: functional tests shard 2/3, Ansible 2.6, Python 2.7
+
+  - <<: *py-27
+    env:
+      <<: *env-functional-shard_3
+      ANSIBLE_VERSION: "26"
+    name: functional tests shard 3/3, Ansible 2.6, Python 2.7
+
+ - <<: *py-27
```
webknjaz: Add a space in front of this

themr0c: i fixed this and all yamllint errors

themr0c:
Good news the errors on the last build https://travis-ci.com/ansible/molecule/builds/98812943 are errors that are already in master, unrelated to this PR.

.travis.yml
```diff
@@ -8,369 +9,407 @@ cache:
   bundler: true
   pip: true
   directories:
-  - $HOME/.cache/pre-commit
-  - $HOME/.pre-commit
-  - $HOME/.rvm
-  - $HOME/virtualenv/python$(python -c 'import platform; print(platform.python_version())')
-  - $HOME/Library/Caches/Homebrew
+    - $HOME/.cache/pre-commit
```
webknjaz: I like previous style more, I used it on purpose. It helps to maintain shorter lines even for nested struts

themr0c: This is not a matter of opinion, the changes were requested by yamllint. I can rollback no problem.

gundalow: Lets go with yamllint

gundalow:
This improves on what we have, so merging
@themr0c @lwm @webknjaz Thank you all for your work on this

webknjaz:
:+1:

Issue #1666, opened 2019-01-15
too many common errors are offuscated in ansible, requiring run with --debug


ssbarnea:
This is a generic issue with molecule which can really put many users off: a big number of very common execution errors are offuscated by crytic errors and you never find the cause till you run molecule again in debug mode.

One such example: what happens if local docker is stopped. Just try it.

Expected behaviour: get a clear error and not have to run molecule in debug mode in order to find out the cause.

Another issue that make this issue even more problematic is that `--debug` is extreamly verbose and as far as I know there is nothing in-between normal execution and debug mode. For example I almost never found useful to see the "blue" connection messages, but the other changes in verbosity did prove essential for debugging errors. It may be my lack of knowledge but I wasn't able to find something less verbose.

lwm:
Agree with all points here. We can make improvements.

> One such example: what happens if local docker is stopped. Just try it.

One idea is to introduce a validation layer right at the start of the run which asks which driver is being requested, in your example, Docker, and then trys to check if it is up and ready to run. We could bail out with some nice error messages immediately if not.

That might take some work. In the meantime, as suggested on https://etherpad.openstack.org/p/ansible-molecule, we could try and put together a simple 'troubleshooting' guide and output this with any error message.

> Another issue that make this issue even more problematic is that --debug is extreamly verbose and as far as I know there is nothing in-between normal execution and debug mode. For example I almost never found useful to see the "blue" connection messages, but the other changes in verbosity did prove essential for debugging errors. It may be my lack of knowledge but I wasn't able to find something less verbose.

AFAIK, a way in between `--debug` is to set `log: true` in your `molecule.yml`.

asmacdo:
Something along the lines of `molecule test --debug-failed` would be *very* helpful for running molecule tests with travis.

ssbarnea:
I am glad to see others feel the same pain. As someone that tries to advocate molecule use to others I find the need to run afain with debug very inconvenient.

We should aim to get meaninful error on normal runs, without asking user to add extra params in order to findout about stuff like:
- unable to connect to docker
- docker module was not importable
- docker build failed....

I am not saying that default execution should be more verbose but I am saying that failed tasks should be verbose.

decentral1se:
To be clear, I'm really +1 on something for this (and I hope I have time soon) to improve the experience with Molecule and debugging failures. It's pretty damn important and without, it can certainly be frustrating.

decentral1se:
[...]

PR #1703, opened 2019-01-25
Explicitly call out commonly needed fields to run systemd in docker

maxamillion:
Added some docs in the `docker` driver configuration section to explicitly highlight what's needed to utilize systemd inside the target container.

Signed-off-by: Adam Miller <admiller@redhat.com>

#### PR Type
- Docs Pull Request

### DIFF
--- a/molecule/driver/docker.py
+++ b/molecule/driver/docker.py
@@ -37,6 +37,7 @@ class Docker(base.Base):
     variables from ``molecule.yml`` into ``create.yml`` and ``destroy.yml``.
 
     .. _`docker_container`: http://docs.ansible.com/ansible/latest/docker_container_module.html
+    .. _`Docker Security Configuration`: https://docs.docker.com/engine/reference/run/#security-configuration
 
     .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


molecule/driver/docker.py
```diff
@@ -87,6 +88,26 @@ class Docker(base.Base):
...
+    .. code-block:: yaml
+
+        platforms:
+        - name: instance
+            image: centos:7
+            privileged: true
+            volume_mounts:
+            - "/sys/fs/cgroup:/sys/fs/cgroup:rw"
+            command: "/usr/sbin/init"
+            environment: { container: docker }
```
webknjaz: I don't like it inline, could you change it to multi-line plz?

lwm:
Related: https://github.com/ansible/community/issues/427#issuecomment-455196367, https://github.com/ansible/molecule/issues/1666.

molecule/driver/docker.py
maxamillion: @webknjaz fixed :)

vutkin:
Hi @maxamillion , We (lean_delivery org https://github.com/lean-delivery), have created separete docker images with systemd support for tests in docker: https://hub.docker.com/r/leandelivery/docker-systemd/tags

examples for usage:

https://github.com/lean-delivery/ansible-development-kit/blob/develop/%7B%7Bcookiecutter.role_name%7D%7D/molecule/default/molecule.yml

PR #1706, opened 2019-01-29
references #1669 - Parallelize unit tests. 

themr0c:
It won't be a huge gain, maybe only 5 minutes in total, because we are targeting 3 to 4 minutes jobs, and travis runners have only 2 cores.

#### PR Type

- Bugfix Pull Request


lwm:
:beers:

Issue #1573, opened 2018-11-08
[BUG] Broken concurrency: one cannot run few molecule processes in parallel


webknjaz:
# Issue Type

- Bug report

# Molecule and Ansible details

```
devel
```

Molecule installation method (one of):

N/A

Ansible installation method (one of):

N/A

Detail any linters or test runners used:

docker driver

# Desired Behavior

Molecule should create resources (test envs) with names unique to test run.
I want to be able to run concurrently the test of Molecule itself.

# Actual Behaviour

Molecule's default docker image instance name defaults to the driver name, that's it. So when you run multiple processes they mess with the same resource breaking each other in unpredictable ways.

ragingpastry:
Ran into this a while back like lwm pointed out. There is probably a much better way to solve it though.

webknjaz:
@lwm yea, kinda.
Also, see prototype of the possible fix: https://github.com/webknjaz/molecule/commit/87ccd8190064f9bdd8aa1069f8dddad8ed804f47

lwm:
Fix looks legit @webknjaz! How do you want to proceed? The CI should be able to verify this as working or we need to do some manual testing? Just to mark, I'll need to update https://github.com/ansible/molecule/pull/1683 documentation when we have this working differently.

webknjaz:
@lwm thanks, feel free to fork from that branch (https://github.com/ansible/molecule/compare/master...webknjaz:feature/molecule-parallelism) or cherry-pick it if you need this done fast because I'm not going to have time for it during the next few weeks. It seems fine but I don't like that it's a bit sketchy (and I'd normally add docstrings before merging such thing).
I think CI would probably be flaky with xdist enabled if the issue still exists but yes, if checked manually I'd be more confident.

In general, it's more about seeking the design feedback: is it fine to add such random thing to identifier names? what else would this affect? I checked with docker but what about other providers? do existing users *rely* on having the same instance name across subsequent re-runs (current patch would break this expectation because the random thing would be different every time)? what else did I miss? any UX problems?

lwm:
Related: https://github.com/ansible/molecule/pull/1716 + https://github.com/ansible/molecule/issues/1704.

hugoShaka:
It is interesting, and way cleaner that I've done on https://github.com/hugoShaka/molecule/commit/b335b0939436aa021aec6958ddf43d0fabb81e3c which was provider-specific 👏 

[...]

PR #1716, opened 2019-01-30
doc: avoid MOLECULE_EPHEMERAL_DIRECTORY with tox

goneri:
If an `MOLECULE_EPHEMERAL_DIRECTORY` environment variable is defined
before `molecule` startup, it will be used to compute the
`ephemeral_directory` variable[1].

In config, `ephemeral_directory` is used to overwrite
`MOLECULE_EPHEMERAL_DIRECTORY`[2].

As a result:
- `MOLECULE_EPHEMERAL_DIRECTORY` is different
- it can content a /, which is not a good fit in a docker or VM name

This commit promotes a new environment variables called `TOX_ENVNAME` to
avoid any conflict.

[1]: https://github.com/ansible/molecule/blob/master/molecule/scenario.py#L182-L186
[2]: https://github.com/ansible/molecule/blob/master/molecule/config.py#L193

#### PR Type

- Docs Pull Request

### DIFF
--- a/doc/source/ci.rst
+++ b/doc/source/ci.rst
@@ -183,7 +183,8 @@ To view the factor generated tox environments.
 
 If using the `--parallel functionality`_ of Tox (version 3.7 onwards), Molecule
 must be made aware of the parallel testing by setting a
-``MOLECULE_EPHEMERAL_DIRECTORY`` environment variable per environment.
+``MOLECULE_EPHEMERAL_DIRECTORY`` environment variable per environment. In addition,
+we export a ``TOX_ENVNAME`` environment variable, it's the name of our tox env.
 
 .. code-block:: ini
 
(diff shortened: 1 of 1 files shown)


lwm:
I think https://github.com/ansible/molecule/issues/1573#issuecomment-459893426 is the real fix we want which will solve all our conflict issues. I think this will also resolve issues marked in https://github.com/ansible/molecule/pull/1716#issuecomment-459257920. I'll get a PR out for this if we're in agreement.

lwm:
Hmmm, not sure now. Added https://github.com/ansible/molecule/issues/1573#issuecomment-461816683.

goneri:
Overall, I think Molecule should never rewrite the environment variable on the fly likes it do. This is really confusing and lead to misconfiguration.
In addition to this patch, what do you think if I push a patch that raises an error if MOLECULE_EPHEMERAL_DIRECTORY is not a relative path?

themr0c:
Successfully passed the CI this night. Can you have a look again @lwm?

lwm:
> Overall, I think Molecule should never rewrite the environment variable on the fly likes it do. This is really confusing and lead to misconfiguration. In addition to this patch, what do you think if I push a patch that raises an error if MOLECULE_EPHEMERAL_DIRECTORY is not a relative path?

That sounds very useful, yes please. If more feel strongly about this (once they start to see the error), then we start to discuss how to do this better. Thanks for investigating this :+1:

PR #1719, opened 2019-01-31
Reorganise documentation: Getting started + better navigation

lwm:
A big stab at making the documentation more useful!

#### PR Type

- Docs Pull Request


gundalow:
Would like @acozine to say if she's (as Ansible Docs Lead) to say if she's happy for the images to be removed before we get the text updated

lwm:
Sure thing. To clarify, I was going to work on adding text based on comments in the etherpad notes from meetings after this PR but want to make the diffs easy to review. I can also include that in this PR too (would basically be including more direction on how to navigate the TOC to get started with Molecule).

samccann:
@lwm @gundalow Near as I can tell, this isn't removing images, it's removing all the videos. I don't fully understand how the docs build works for molecule, but if you remove this from the readme.rst, isn't that also going to remove it from the actual readthedocs molecule documentation the next time that is regenerated/republished?

lwm:
@samccann, you can build the docs locally with:

```
$ pip install alabaster sphinx
$ cd docs
$ make html
$ $BROWSER build/html/index.html
```

Here's what the home page looks like after this PR:

2019-02-05-155231_1920x1080_scrot

The main value add here is to remove the visual noise from the videos. When you scroll down, you see the entire TOC rendered and can more quickly get where you need to go (intro, config, etc.).

samccann:
@lwm That's basically my point tho - this .rst file IS the home page... so removing the videos here means they disappear from https://molecule.readthedocs.io/en/latest/ as well the next time someone publishes that site.

Or does https://molecule.readthedocs.io/en/latest/ get its home page from somewhere else?

I don't think we can remove the videos entirely until they are replaced with text, even if the text is coming in a few days, that leaves users with nothing until then.

samccann:
hmm...gimme a few minutes to actually watch these videos...

samccann:
okay I haven't watched all the videos but I think we can't remove them from the docsite without replacing with the text documentation at the same time. The install video didn't add much to my understanding, but the new role and test role were much easier for me to understand than trying to find that matching content in the rest of the existing documentation.  (aka we need to replace the videos with a molecule quickstart or getting started text guide).

lwm:
Thanks @samccann!

>  (aka we need to replace the videos with a molecule quickstart or getting started text guide).

OK, I'll get on this today hopefully.

singuliere:
Having the quick start text addition in the same PR but in a different commit would be convenient for review, IMHO .Thanks for doing this @lwm, these videos are nice but they are a distraction.

lwm:
@samccann @acozine @gundalow, here's a fairly big stab at making the documentaiton more useful based on the running commentary that has been doing on in the etherpads (linked in commit). I am sorry for the big diff but it didn't seem to be possible to break it down into smaller parts given discussion above. Hopefully you can follow steps in https://github.com/ansible/molecule/pull/1719#issuecomment-460668769 to build the documentation locally and make an easier review pass. The big addition is the 'Getting Started Guide'. Overall, I think this is a much better structure that will facilitate more documentation fixes to land from other contributors because it will be clear where things need to go (it is tough now).

lwm:
Of course, I can split it all up into separate PRs but it would be nice to not have to do it :sweat_smile:

lwm:
OK, this was too optimistic. I'll be back with something easier to review!

PR #1732, opened 2019-02-08
allow overriding pytest functional arguments with posargs

singuliere:
It is useful to run a single test from the command line with, for instance:

    tox -e ansibledevel-functional -- -k 'test_command_init_role[docker]' test/functional/test_command.py

The `test/functional/` pytest argument is made a default used when there are no [posargs](https://tox.readthedocs.io/en/latest/config.html#substitutions-for-positional-arguments-in-commands) instead of being hardcoded and always used. If all pytest invocations include `test/functional/` there is no way for the user to restrict the tests being run to a single file.

#### PR Type

- Feature Pull Request

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -30,7 +30,7 @@ extras =
     vagrant
 commands =
     unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
-    functional: pytest test/functional/ {posargs}
+    functional: pytest {posargs:test/functional/}
     lint: flake8
     lint: yamllint -s test/ molecule/
 


singuliere:
@lwm thanks for the quick review :-) I see you have a green :heavy_check_mark: and @pilou- has a gray :white_check_mark: . Does that mean you have merge rights and @pilou- does not? Or is it unrelated?

webknjaz:
I think green only means a member of org review, not merge rights.

tox.ini
```diff
@@ -30,7 +30,7 @@ extras =
     vagrant
 commands =
     unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
-    functional: pytest test/functional/ {posargs}
+    functional: pytest {posargs:test/functional/}
```
webknjaz: There's no point in doing this since currently `-- -k ...` syntax works. With this, it'll match tests from unit test suite as well.

singuliere: The reason why pytest has both **-k** and positional arguments is the same reason why this patch is proposed. For instance, it allows for **tox -- test/test.py** to run all test from a single test file. Achieving the same result with the **-k** option would be impractical.

webknjaz: I see why you'd want that and it's a natural wish.

Current `tox.ini` combines too much test modes in one section. I think it's an antipattern and should be improved.

Now, `unit` and `functional` factors correspond to running unit tests and functional tests respectively. But the proposed change advertises abusing that semantics which is far from being intuitive from the UX point of view.

I propose the following (thoughts, which I was having in mind for a while now):
1) separate these sections to make a clear distinction between their purposes
2) have a dedicated section for running whatever you want with `pytest`
3) [maybe] remove that tags plugin, I haven't heard about anyone using it, it complicates things a lot and it's probably possible to achieve the same result in other ways...

P.S. If you want, I can do (1) and (2) since I've got some experience with tox internals.

webknjaz: P.P.S. @ssbarnea has submitted an unfinished attempt to improve other aspects of tox config @ 
https://github.com/ansible/molecule/commit/cb4ad3044f82fce508338f7592cbb21e6900d590 via https://github.com/ansible/molecule/pull/1547 in past. This is also worth exploring...

singuliere: > But the proposed change advertises abusing that semantics which is far from being intuitive from the UX point of view.

I respectfully disagree. This is a very common and convenient pattern.

lwm: You've raised important points @webknjaz. If I read this correctly, your comment in https://github.com/ansible/molecule/pull/1732#discussion_r255506045 shows us a more sustainable plan for our Tox usage. That's great and I would like to see that in a separate issue to share knowledge of what needs to be done and who is doing it.

For the purposes of this PR: this change might not be taking us in the direction of what you've desribed but it is making it more convenient to run specific tests for contributors in the right here and now. That's important and for that reason, I'd like to see this change merged as a 'low road' to making Tox easier to work with. WDYT?

tox.ini
```diff
@@ -30,7 +30,7 @@ extras =
...
     lint: flake8
     lint: yamllint -s test/ molecule/
```
```suggestion
 
 [testenv:pytest]
 commands =
     pytest {posargs}
```

tox.ini
webknjaz: @lwm

Feel free to overrule my judgment, for now. We can fix it later, after all.

OTOH after this change, it will be possible running unit tests from within functional factor which makes the existence of `unit` factor questionable.

P.S. on
> is making it more convenient to run specific tests

it is already possible:
```console
tox -e unit -- -k test_unit_smth
tox -e functional -- -k test_functional_smth
```

and after this change it will be
```console
tox -e functional -- -k test_unit_smth
tox -e functional -- -k test_functional_smth
tox -e functional -- test/unit/test_unit_thing.py::test_unit_smth
```

and if you apply the suggestion below (https://github.com/ansible/molecule/pull/1732/files#r257429626) if will look like

```console
tox -e pytest -- -k test_unit_smth
tox -e pytest -- -k test_functional_smth
tox -e pytest -- test/unit/test_unit_thing.py::test_unit_smth
tox -e pytest -- test/functional/test_func_thing.py::test_functional_smth
```

singuliere:
I apologize for proposing such a controversial modification. I would not want to waste any more volunteer / unpaid time from molecule contributors because of this.

lwm:
:face_with_thermometer:

webknjaz:
@singuliere hey, why didn't you just accept the suggested change I made? It's just one click and it'd achieve exactly what you wanted...

tox.ini
lwm: > Feel free to overrule

Overruling is not productive behaviour.

lwm:
@webknjaz, I can only speculate it is because you 1) Pushed a number of unrelated of concerns onto this contributor - https://github.com/ansible/molecule/pull/1732#discussion_r255506045 2) used negative language such as "there is no point in doing this" - https://github.com/ansible/molecule/pull/1732#discussion_r255320106 and "abusing that semantics" - https://github.com/ansible/molecule/pull/1732#discussion_r255506045.

Hence, this reaction. This is not the first time a contributor has reacted negatively to a review of yours (please do not ask me to reference). I recommend that you reconsider your choice of language and try to consider and prioritise volunteer contributor capacity/time/motivation when making your reviews.

tox.ini
webknjaz: Not if I explicitly allowed you to :)

In this case, it may be more productive as you need to keep going with improvements and I don't have any time involvement guarantees. I just wanted to share/document my vision but not block anyone.

PR #1738, opened 2019-02-11
allow for become_method and become_pass variables with the delegated driver

geropizarro:
When using the delegated driver, it should be possible to set different become methods and passwords for every instance if required.

### DIFF
--- a/molecule/driver/delegated.py
+++ b/molecule/driver/delegated.py
@@ -51,8 +51,8 @@ class Delegated(base.Base):
           instance: instance_name
           port: ssh_port_as_string
           user: ssh_user
-          become_method: su
-          become_pass: mypass
+          become_method: valid_ansible_become_method
+          become_pass: password_if_required
 
         - address: winrm_endpoint
           instance: instance_name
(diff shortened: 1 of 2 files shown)


lwm:
Hey @geropizarro, thanks for taking time on this. The change makes sense to me. I'd like to see a test to verify it. Our functional tests are a bit flaky right now so I would recommend adding a unit test: https://github.com/ansible/molecule/blob/master/test/unit/driver/test_delegated.py. Our CI is more red than green these days, so you might expect some spurious failures :sweat_smile:

geropizarro:
@lwm like this? all tests passing for me

p1rateg0ld:
Hello, I could really use this feature. Is there any workaround?

ssbarnea:
@geropizarro Can you please fix the DCO? It would be a pitty not to be able to merge this fix because of that.

geropizarro:
@ssbarnea DCO fixed, though the build now failed two jobs with "No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."

Issue #1734, opened 2019-02-08
Run dependency before destroy/create by default?


seandst:
# Issue Type

- Feature request

# Desired Behavior

I'd like to propose the idea of running the `dependency` sequence step before any `create` or `destroy` sequence steps so galaxy roles needed for scenario provisioning can be used in the `create` and `destroy` playbooks *by default*. I am aware that sequences can be reordered at will, and am specifically proposing that this would change molecule's default steps sequences.

# Rationale

I work on a team that writes Ansible roles, and the molecule create/destroy tasks are common across every scenario for those roles. Being that our job is to create roles, I recently attempted to create a role to encapsulate the create/destroy tasks so that we can have less duplication in our scenarios, and also so that we can change the provisioning for all scenarios without having to literally go out and change all scenarios.

I did a proof of concept for this which works perfectly, except now every scenario using the create/destroy role has to ensure that the supporting role is installed, e.g by having its sequences reconfigured to put `dependency` above `create` or `destroy`, by explicitly installing it from galaxy before running molecule, etc. These are fine solutions that work in my testing, but I think that using galaxy roles should be possible in any molecule step playbook, not just `prepare`/`converge`, and that the existing dependency handling can be leveraged to enable this by default in molecule.

# Concerns

- I think it's reasonable to consider the ordering of sequence steps to be part of molecule's "interface", and so making this change could potentially be considered a backward-incompatible change.
- Doing this conflates the dependencies of a scenario's supporting playbooks with the dependencies of the scenario itself. This is easily overcome with comments in the scenario's requirements file, but still kludgy.
- I think this would mean that `dependency` needs to run for every sequence type, which is potentially objectionable.

# Molecule and Ansible details

ansible 2.7.7, molecule 2.19.0, both installed via pip.

lwm:
Hi @seandst, thanks for raising a solid report. Looks like a topic worth discussing!

So, as a reference for all sequences: https://molecule.readthedocs.io/en/latest/configuration.html#scenario.

> I think that using galaxy roles should be possible in any molecule step playbook, not just `prepare`/`converge`, and that the existing dependency handling can be leveraged to enable this by default in molecule.

I agree. 

[...]

PR #1739, opened 2019-02-11
Modify sequences to run dependency first

seandst:
In sequences where a playbook is run, run the `dependency` step first so
roles can be declared in scenerio requirements and used in all playbook
steps.

#### PR Type

- Feature Pull Request

re #1734

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,12 @@
 History
 *******
 
+Unreleased
+==========
+
+* `dependency` step is now run by default before any playbook sequence step, including
+  `create` and `destroy`. This allows the use of roles in all sequence step playbooks.
+
 2.20
 ====
 
--- a/test/unit/test_scenario.py
+++ b/test/unit/test_scenario.py
@@ -26,6 +26,7 @@ import pytest
 
 from molecule import config
 from molecule import scenario
+from molecule import util
 
 
 # NOTE(retr0h): The use of the `patched_config_validate` fixture, disables
@@ -34,6 +35,61 @@ from molecule import scenario
 @pytest.fixture
 def _instance(patched_config_validate, config_instance):
     return scenario.Scenario(config_instance)
+
+
+def test_prune(_instance):
+    e_dir = _instance.ephemeral_directory
+    # prune data also includes files in the scenario inventory dir,
+    # which is "<e_dir>/inventory" by default.
+    # items are created in listed order, directories first, safe before pruned
+    prune_data = {
+        # these files should not be pruned
+        'safe_files': [
+            'state.yml',
+            'ansible.cfg',
+            'inventory/ansible_inventory.yml',
+        ],
+        # these directories should not be pruned
+        'safe_dirs': ['inventory'],
+        # these files should be pruned
+        'pruned_files': [
+            'foo',
+            'bar',
+            'inventory/foo',
+            'inventory/bar',
+        ],
+        # these directories should be pruned, including empty subdirectories
+        'pruned_dirs': [
+            'baz',
+            'roles',
+            'inventory/baz',
+            'roles/foo',
+        ],
+    }
+
+    for directory in prune_data['safe_dirs'] + prune_data['pruned_dirs']:
+        # inventory dir should already exist, and its existence is
+        # required by the assertions below.
+        if directory == 'inventory':
+            continue
+        os.mkdir(os.path.join(e_dir, directory))
+
+    for file in prune_data['safe_files'] + prune_data['pruned_files']:
+        util.write_file(os.path.join(e_dir, file), '')
+
+    _instance.prune()
+
+    for safe_file in prune_data['safe_files']:
+        assert os.path.isfile(os.path.join(e_dir, safe_file))
+
+    for safe_dir in prune_data['safe_dirs']:
+        assert os.path.isdir(os.path.join(e_dir, safe_dir))
+
+    for pruned_file in prune_data['pruned_files']:
+        assert not os.path.isfile(os.path.join(e_dir, pruned_file))
+
+    for pruned_dir in prune_data['pruned_dirs']:
+        assert not os.path.isdir(os.path.join(e_dir, pruned_dir))
 
 
 def test_config_member(_instance):
--- a/test/unit/test_scenario.py
+++ b/test/unit/test_scenario.py
@@ -127,9 +127,9 @@ def test_inventory_directory_property(_instance):
 
 def test_check_sequence_property(_instance):
     sequence = [
+        'dependency',
         'cleanup',
         'destroy',
-        'dependency',
         'create',
         'prepare',
         'converge',
@@ -154,6 +154,7 @@ def test_converge_sequence_property(_instance):
 
 def test_create_sequence_property(_instance):
     sequence = [
+        'dependency',
         'create',
         'prepare',
     ]
@@ -166,7 +167,7 @@ def test_dependency_sequence_property(_instance):
 
 
 def test_destroy_sequence_property(_instance):
-    assert ['cleanup', 'destroy'] == _instance.destroy_sequence
+    assert ['dependency', 'cleanup', 'destroy'] == _instance.destroy_sequence
 
 
 def test_idempotence_sequence_property(_instance):
@@ -192,9 +193,9 @@ def test_syntax_sequence_property(_instance):
 def test_test_sequence_property(_instance):
     sequence = [
         'lint',
+        'dependency',
         'cleanup',
         'destroy',
-        'dependency',
         'syntax',
         'create',
         'prepare',
--- a/molecule/command/base.py
+++ b/molecule/command/base.py
@@ -20,7 +20,6 @@
 
 import abc
 import collections
-import fnmatch
 import glob
 import os
 
@@ -56,29 +55,6 @@ class Base(object):
     @abc.abstractmethod
     def execute(self):  # pragma: no cover
         pass
-
-    def prune(self):
-        """
-        Prune the ephemeral directory with the exception of safe files and
-        returns None.
-
-        :return: None
-        """
-        safe_files = [
-            self._config.provisioner.config_file,
-            self._config.provisioner.inventory_file,
-            self._config.state.state_file,
-        ] + self._config.driver.safe_files
-        files = util.os_walk(self._config.scenario.ephemeral_directory, '*')
-        for f in files:
-            if not any(sf for sf in safe_files if fnmatch.fnmatch(f, sf)):
-                os.remove(f)
-
-        # Remove empty directories.
-        for dirpath, dirs, files in os.walk(
-                self._config.scenario.ephemeral_directory, topdown=False):
-            if not dirs and not files:
-                os.removedirs(dirpath)
 
     def print_info(self):
         msg = "Scenario: '{}'".format(self._config.scenario.name)
@@ -129,6 +105,8 @@ def execute_cmdline_scenarios(scenario_name, args, command_args):
                 LOG.warn(msg)
                 execute_subcommand(scenario.config, 'cleanup')
                 execute_subcommand(scenario.config, 'destroy')
+                # always prune ephemeral dir if destroying on failure
+                scenario.prune()
                 util.sysexit()
             else:
                 raise
@@ -156,6 +134,11 @@ def execute_scenario(scenario):
         # and is also used for reporting in execute_cmdline_scenarios
         scenario.config.action = action
         execute_subcommand(scenario.config, action)
+
+    # pruning only if a 'destroy' step was in the sequence allows for normal
+    # debugging by manually stepping through a scenario sequence
+    if 'destroy' in scenario.sequence:
+        scenario.prune()
 
 
 def get_configs(args, command_args, ansible_args=()):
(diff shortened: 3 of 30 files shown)


seandst:
tox looked good locally; here's hoping travis likes it. Please let me know if I missed something, and note that there is some related discussiong (including rationale) in the related issue.

seandst:
I think "the details of what it is, how to use it" are covered if you include the related issue, and here are the details on "how it was tested":
```
$ tox -t lint -t format -t docs -t unit
  ...
  lint: commands succeeded                                          
  py27-ansible25-unit: commands succeeded          
  py27-ansible26-unit: commands succeeded                                    
  py27-ansible27-unit: commands succeeded
  py35-ansible25-unit: commands succeeded  
  py35-ansible26-unit: commands succeeded       
  py35-ansible27-unit: commands succeeded        
  py36-ansible25-unit: commands succeeded                        
  py36-ansible26-unit: commands succeeded   
  py36-ansible27-unit: commands succeeded                            
  py37-ansible25-unit: commands succeeded
  py37-ansible26-unit: commands succeeded                                             
  py37-ansible27-unit: commands succeeded                                            
ERROR:   format-check: commands failed                     
```

Format-check did pick out some of my changes, so I fixed them and retested. The formatting it complains about now are, as far I can tell, unrelated to this PR. Functional testing was, I think understandably, something I did not do. :)

I didn't notice the the docs tests weren't tagged, so...

```
$ tox -e doc
  ...
  doc: commands succeeded
```

This originally failed because of a format literal in the docs `conf.py`. I haven't yet made py3 the default in my ansible virtualenvs, but it's delightful to see that it's used here.

seandst:
Manual functional testing has revealed some errors, so I'm going to WIP it and start pulling things apart.

test/unit/test_scenario.py
```diff
@@ -101,7 +102,11 @@ def test_dependency_sequence_property(_instance):
 
 
 def test_destroy_sequence_property(_instance):
-    assert ['destroy'] == _instance.destroy_sequence
+    x = [
```
lwm: I know this `x = ...` is everywhere in the code base but we're trying to get away from it. Could you rename? Thanks!

seandst:
Ah, I never updated here with what was actually wrong, sorry about that.

The (basic) problem is that the ephemeral dir is "pruned" on destroy:
https://github.com/ansible/molecule/blob/a580b5fbf55b61d09a2b495c7631d07dba7ead1a/molecule/command/destroy.py#L98

So when I change, say, the test sequence to put `dependency` before `destroy`, everything works fine up until the first `create` step thereafter, at which point everything explodes because `destroy` removed the `roles` dir from the ephemeral dir.

Questions that I still don't have an answer for through all this include things like "Why only prune on destroy?", and "Why prune at all?". Since I don't have answers to those questions, I feel compelled to continue emulating the current behavior as much as possible, which is basically "prune on destroy".

A wall of text follows outlining some of the solutions I've messed with (yeah, this is just a list of the highlights):
- **Expand the [`safe_files`](https://github.com/ansible/molecule/blob/a580b5fbf55b61d09a2b495c7631d07dba7ead1a/molecule/command/base.py#L64) list to include files declared by [dependency managers](https://github.com/ansible/molecule/blob/a580b5fbf55b61d09a2b495c7631d07dba7ead1a/molecule/dependency/ansible_galaxy.py).** This works, but only really makes sense for the galaxy manager, not so much gilt or shell. It also just feels wrong given that all the other "safe files" are coming out of the config, and don't require spelunking the ephemeral dir in-process to find all the roles files to delete. It also means that ansible-galaxy roles will *never* be pruned out of the ephemeral dir.
- **Take `prune` off of the [command base class](https://github.com/ansible/molecule/blob/a580b5fbf55b61d09a2b495c7631d07dba7ead1a/molecule/command/base.py#L57-L78)**, move it into the `molecule.command.base` namespace, and explicitly call it after sequences that are known to have a destroy step in them by default. This was getting somewhere, but the ability to redefine sequence orders means that any user could at any time reorder their sequences to add or remove destroy steps. The command sequence can be inspected to do this more smartly, such as "only call prune after the sequence runs, and only if destroy was in that sequence", but I was uncomfortable doing this without basically refactoring a bunch of duplicated [command calling](https://github.com/ansible/molecule/blob/a580b5fbf55b61d09a2b495c7631d07dba7ead1a/molecule/command/destroy.py#L131-L137) code. This is something I wasn't comfortable doing, having only recently started looking closely at how molecule actually works. I'm still not very comfortable with it, even after getting much more familiar with how molecule works for #1749.
- **Add a `--prune=on_destroy|never` option to the test command** that acts very much like the `--destroy=always|never` option. This was alright, but I was somewhat surprised to learn that `--destroy` is only defined on the test command, so e.g. the default check sequence would still be broken by the `['dependency', 'destroy', 'create'] step ordering. While I was trying this out, another idea presented itself...
- **Expose prune as a command, and add it to default sequences where destroy is called.** This seems to be the "best" option, in terms of dealing with customized sequences, as it now allows users to explicitly prune if/when they want to. The main issues that I really see with this are that existing custom sequences defined by users would need to be updated to include the new prune step, and I think that polluting the "top level" command list with utility commands like this can only serve to make molecule harder to understand and use. While I do consider this to be the "best" option, I think the drawbacks warrant serious consideration.

The simplest solution may be to just delete the `prune` method and remove [this line](https://github.com/ansible/molecule/blob/a580b5fbf55b61d09a2b495c7631d07dba7ead1a/molecule/command/destroy.py#L98), then add notes to the docs about how to find and safely clean the ephemeral dir.

A common problem to every solution I've considered is that I need a better understanding of molecule's strategic direction to make a more informed decision about how to proceed. So I'd love feedback on the ideas presented here, and welcome suggestions for different approaches to consider.

test/unit/test_scenario.py
seandst: I'd rather not create an inconsistency like that, and I'd also rather not clutter up this PR with a second commit making the change file-wide. My preferred approach would be to leave it like this and open up a second PR that fixes the `x = ...` problem for the entire file in one atomic action.

lwm: You can ammend your commit? As you want though!

seandst: Given the relative ease of fixing terse identifiers in the tests versus this way-more-involved PR, I think I'll just go clean up test_scenario.py real quick.

seandst: PR posted: https://github.com/ansible/molecule/pull/1760

tima:
> The simplest solution may be to just delete the `prune` method and remove [this line](https://github.com/ansible/molecule/blob/a580b5fbf55b61d09a2b495c7631d07dba7ead1a/molecule/command/destroy.py#L98), then add notes to the docs about how to find and safely clean the ephemeral dir.

I'm a fan of simple solutions. ;)

Would creating `/roles` if not present in `create` smooth things out with little disruption to how things work today?

seandst:
I brought this up in today's working group meeting at lwn's suggestion, and these are the conclusions reached:

1. prune is useful, and should stick around
1. being able to run roles in create/destroy sounds useful

So I'll keep looking at the angles on this and see about finding a way to thread the needle.

More details on the discussion can be found in the meeting [logs](https://meetbot.fedoraproject.org/ansible-molecule/2019-02-20/ansible_molecule_working_group.2019-02-20-19.01.log.html) referencing this issue.

seandst:
> Would creating `/roles` if not present in `create` smooth things out with little disruption to how things work today?

`/roles` and all the roles inside them would need to be recreated (effectively running the entire dependency step again). So when I said this:

> So when I change, say, the test sequence to put `dependency` before `destroy`, everything works fine up until the first `create` step thereafter, at which point everything explodes because `destroy` removed the `roles` dir from the ephemeral dir.

I probably should have said something more like "because `destroy` removed all of the installed roles from the ephemeral dir."

seandst:
This has been updated. Things may have gotten out of hand, and as always I apologize in advance for the following wall of text.

This has been broken up into three atomic commits, each one depending on the one before and independently tested/testable. The three commits are based on what I mentioned in the "Take `prune` off of the command base class" method above, with one minor change. Instead of putting prune into `command.base`, it ended up fitting really well directly on the `scenario.Scenario` class, since that's also where the ephemeral directory itself is defined. I did end up refactoring the sequence runner commands to centralize their behavior into `command.base`, as I suspected in my earlier comment.

It would be both reasonable and trivial to break each commit into its own pull request if desired (just let me know what's preferable). I do not think these changes should be included in the 2.20 release, so I plan to leave the WIP tag on this PR until 2.20 is released. At that point, the changelog will need to be rebased (or removed, since I think that's automated or soon going to be automated).

Before talking about the first commit, I'll just restate the problem being solved here: I would love to be able to use roles installed via the `dependency` step in `create` and `destroy` playbooks (and now `cleanup` playbooks, too, which is great!). My initial solution was to just edit the default sequences to put `dependency` above any `create` or `destroy` steps. The problem is that `destroy` calls `prune`, which blows away roles (and anything else deemed "unsafe") from the ephemeral dir. That is desirable behavior, and everything that follows attempts to accomplish both goals:

- be able to use galaxy roles in any playbook sequence step
- prune the ephemeral dir on destroy to ensure a "clean" state for testing

On to the first commit!

```
    DRY up sequence command execution
    
    Removes a lot of repeated boilerplate found in commands that run
    scenario sequences, centralizing it in the base command module.
    
    Aside from removing duplicated code, this should also help make it a
    little easier to do introspection during sequence execution and take
    actions for all sequences based on the config, as seen with the
    'destroy==always' handling. That behavior could now be added to any
    sequence by adding support for the `--destroy` option to the decorated
    command function.
```

This was needed because the `prune` method is called in the `destroy` command class, but the special handling of `destroy` in the [`test` sequence](https://github.com/ansible/molecule/pull/1739/files#diff-661fbc37fc6515dc04b4adf743f59a95) meant that I would have ended up editing every single command to inspect the current sequence and conditionally `prune`. At that point, since I was going to have to touch nearly every command anyway, I figured I'd centralize the scenario runner behaviors in `command.base`, and then work from there.

On to the second commit.

```
    Scenarios can now prune themselves
    
    This decouples scenario pruning from the `destroy` command, allowing
    scenarios to be pruned in any context where a scenario instance is
    available.
    
    This slightly modifies the pruning behavior. Previously this would
    unconditionally prune whenever a 'destroy' sequence step was executed.
    Now, `prune` is only called at the end of a sequence if that sequence
    included a `destroy` sequence step. This prevents unexpected pruning
    during sequence execution. As an example, users attempting to use
    ansible galaxy roles in the `create` and `destroy` steps would need to
    very carefully ensure the `dependency` step is run before *and after*
    every destroy step. With this change, a single dependency call is
    needed.
```

After centralizing the command execution behaviors, it was now easy to pull prune off of `command.base.Base`, put it onto `scenario.Scenario`, and then do sequence introspection to conditionally call prune when the sequence includes a `destroy` step. This emulates the existing behavior of pruning on destroy, while also not preventing the use of galaxy roles during a sequence.

The final commit is actually the original change that started this pull request, which puts `dependency` over other playbook steps by default. That commit's description is the description seen in this pull request.

Unit testing and limited functional testing has been done for each commit, and I believe I've got unit test coverage on all the new behaviors. As this is mostly refactoring, these changes should be transparent to any functional testing. I'm very interested to see what travis has to say about it.

seandst:
Rebased to latest master, WIP removed.

seandst:
> I do not think these changes should be included in the 2.20 release, so I plan to leave the WIP tag on this PR until 2.20 is released.

^ This was mainly so that this PR isn't considered as a blocker for 2.20. If it ends up in 2.20, of course, that's fine with me. :)

CHANGELOG.rst
```diff
@@ -50,6 +50,8 @@ Other
 * Molecule can now be called as a Python module (``python -m molecule``). Patch by `@ssbarnea`_.
 * Add `Travis CI integration`_ and fix related test issues.
 * Add Docker ``buildargs`` option for configuring the ``docker_image`` ``create.yml`` build step.
+* `dependency` step is now run by default before any playbook sequence step, including
```
seandst: I don't recall if there's automatic changelog creation or not. Should I remove this, throw it into an "unrelease" section above 2.20, or...?

decentral1se: Nothing automated yet, yes please, if you could add change log entry, that'd be great :+1:

seandst: Thanks, done. I left out the "Important Changes"/"Other" subsections since that looked like a special thing for 2.20 and the redhat/ansible adoption. I'm happy to make any desired changes, of course. :)

seandst:
> This is a _big_ change. It's just verging on the 'hard to review' side. I've done my best anyway and it looks good to me! I hope you'll be around to help us maintain going forward :) Any more eyeballs on this?

Yeah, that was not really how I wanted it to go, so thanks for bearing with it and sifting through the wall of text. I'll definitely be sticking around for a while, but I hope I'll never have to build a PR like this again. :)

> We're waiting on some packaging issue before we can merge. Thanks for your patience on this one.

No hurry, I'm a much bigger fan of "do it right" than I am of "do it *right now*".

seandst:
This is blocked pending integration of changes introduced in #1903, but that should be a pretty quick rebase.

seandst:
Updated to include changes from #1903, and unit tests have been updated to cover the new behavior.

PR #1740, opened 2019-02-12
LXD prepare.yml: Install ca-certificates-cacert and ca-certificates-mozilla for openSUSE

hswong3i:
openSUSE split the general `ca-certificates` package into 3 individual packages:

- ca-certificates
- ca-certificates-cacert
- ca-certificates-mozilla

Without the other 2 `ca-certificates-*` packages most operation, e.g. wget or curl will always get failed with HTTPS remote source.

Ideally LXD driver `prepare.yml` should be update accordingly.

#### PR Type

- Feature Pull Request

### DIFF
--- a/molecule/provisioner/ansible/playbooks/lxd/prepare.yml
+++ b/molecule/provisioner/ansible/playbooks/lxd/prepare.yml
@@ -3,51 +3,27 @@
   hosts: all
   gather_facts: false
   tasks:
-    - name: Test if Python is installed
-      raw: python -V  # execute Python
-      register: python_version
-      changed_when: "'not found' in python_version.stderr"
-      failed_when: false
-
-    - name: Install Python when not installed
+    - name: Install some basic packages because LXC containers are bare
       raw: |
         if [ -x "$(command -v apt-get)" ]
         then
-          apt-get update && apt-get install -y python
+          apt-get update && apt-get install -y ca-certificates curl python sudo
         elif [ -x "$(command -v yum)" ]
         then
-          yum install -y python
+          yum install -y ca-certificates curl python sudo
         elif [ -x "$(command -v zypper)" ]
         then
-          zypper -n --gpg-auto-import-keys refresh && zypper -n install -y python python-xml
+          zypper -n --gpg-auto-import-keys refresh && zypper -n install -y ca-certificates ca-certificates-cacert ca-certificates-mozilla curl python python-xml sudo
         elif [ -x "$(command -v apk)" ]
         then
-          apk update && apk add python
+          apk update && apk add ca-certificates curl python sudo
         elif [ -x "$(command -v pacman)" ]
         then
-          pacman -Syu --noconfirm python2
+          pacman -Syu --noconfirm ca-certificates curl python2 sudo
         elif [ -x "$(command -v dnf)" ]
         then
-          dnf --assumeyes install python python-devel python2-dnf
+          dnf --assumeyes install ca-certificates curl python python-devel python2-dnf sudo
         elif [ -x "$(command -v emerge)" ]
         then
-          emerge --ask n dev-lang/python-2
+          emerge --ask n ca-certificates curl dev-lang/python-2 gentoolkit sudo
         fi
-      when: "'not found' in python_version.stderr"
-
-    - name: Gathering Facts
-      setup:  # aka gather_facts
-
-    - name: Install equery on Gentoo based containers when not available
-      raw: test -e /usr/bin/equery || emerge --ask n gentoolkit
-      register: equery_status
-      changed_when: equery_status.stdout != ''
-      when: ansible_os_family == 'Gentoo'
-
-    - name: Install some basic packages because LXC containers are bare
-      package:
-        name:
-          - ca-certificates
-          - curl
-          - sudo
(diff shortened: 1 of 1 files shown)


molecule/provisioner/ansible/playbooks/lxd/prepare.yml
```diff
@@ -51,3 +51,11 @@
...
           - sudo
         state: present
+
+    - name: Install ca-certificates-cacert and ca-certificates-mozilla for openSUSE
+      package:
+        name:
+          - ca-certificates-cacert
+          - ca-certificates-mozilla
+        state: present
+      when: ansible_os_family == 'Suse'
```
lwm: Do we need to also specify `ansible_distribution_version` here? 

I assume that these `ca-certificate-*` packages would not be available on older versions?

hswong3i: Not really needed, because we just have images:opensuse/42.3 and images:opensuse/15.0 from LXD repo :-P

hswong3i:
> In general, I am wondering what will happen when all the OS family differences start to slowly arrive in these files. It will become overwhelming, I imagine. But, that is not today :) Just thinking out loud here 🐝

Yes, this is what am I thing for, too. I would like to group all OS specific bootstrap to our first Ansible task within this file, which detected by package manager.

I will rewrite this PR for simplify it :-)

molecule/provisioner/ansible/playbooks/lxd/prepare.yml
lwm: Right, but if someone is building other older images, they will get a failed task, no? We need to be careful here and try to cover all the cases we can think of for this core functionality. Molecule use case are big and wide (just like Ansible!)!

molecule/provisioner/ansible/playbooks/lxd/prepare.yml
```diff
@@ -3,51 +3,30 @@
   hosts: all
   gather_facts: false
   tasks:
-    - name: Test if Python is installed
```
hswong3i: Simply skip this check and just always confirm python and other basic packages are installed correctly.

molecule/provisioner/ansible/playbooks/lxd/prepare.yml
```diff
@@ -3,51 +3,30 @@
...
         then
-          emerge --ask n dev-lang/python-2
+          emerge --ask n ca-certificates curl dev-lang/python-2 gentoolkit sudo
         fi
-      when: "'not found' in python_version.stderr"
-
-    - name: Gathering Facts
-      setup:  # aka gather_facts
 
-    - name: Install equery on Gentoo based containers when not available
```
hswong3i: Simply merge extra per OS specific bootstrap to above per package manager based bootstrap

hswong3i:
https://software.opensuse.org/package/ca-certificates-cacert
https://software.opensuse.org/package/ca-certificates-mozilla

This is how openSUSE name their package, for many different versions :-)

molecule/provisioner/ansible/playbooks/lxd/prepare.yml
```diff
@@ -3,51 +3,27 @@
...
-          dnf --assumeyes install python python-devel python2-dnf
+          dnf --assumeyes install ca-certificates curl python python-devel python2-dnf sudo
         elif [ -x "$(command -v emerge)" ]
         then
-          emerge --ask n dev-lang/python-2
+          emerge --ask n ca-certificates curl dev-lang/python-2 gentoolkit sudo
         fi
-      when: "'not found' in python_version.stderr"
-
-    - name: Gathering Facts
```
hswong3i: no more follow up after raw, no gather_facts required

molecule/provisioner/ansible/playbooks/lxd/prepare.yml
lwm: OK, looks resolved by https://github.com/ansible/molecule/pull/1740#issuecomment-463037332.

lwm:
Hey @hswong3i, sorry to leave this sitting for so long. Can you rebase on master (and perhaps squash commits) and then we can get a green CI build (which is now looking more reliable). Thanks for patience!

webknjaz:
@lwm there's probably no need to require rebase from the contributor because currently, CI doesn't run LXD/LXC testing (I failed to set it up: https://github.com/ansible/molecule/pull/1591)

hswong3i:
PR rebased ;-)

hswong3i:
P.S. should we also port this `prepare.yml` for docker? AFAIK at least docker image for Ubuntu didn't coming with Python pre-installed, too.

lwm:
@hswong3i, I don't see a `prepare.yml` for Docker. Preqrequisite installs happen in the `Dockerfile.j2`.

Thanks for the rebase, I'll merge this when I see it go green :+1:

hswong3i:
@decentral1se only 1 failed test (https://travis-ci.com/ansible/molecule/jobs/179357481), which belongs to Docker not LXD. Should this looks good enough?

decentral1se:
Restarted the build and it's green, thanks for your patience! I'll merge this :+1:

Issue #1723, opened 2019-02-04
`molecule lint` does not match what `ansible-lint .` does


lwm:
# Issue Type
- Bug report

# Molecule and Ansible details

HEAD and latest release AFAIK.

Molecule installation method (one of):
- pip

Ansible installation method (one of):
- pip

# Desired Behavior

`molecule lint` and `ansible-lint .` behave identically.

# Actual Behaviour

See https://github.com/ansible/molecule/pull/1675#issuecomment-459451002 and https://github.com/ansible/molecule/pull/1675#issuecomment-460098613.

lwm:
Potentially related was https://github.com/ansible/molecule/pull/1722. Basically, the configuration options were not working correctly from the `molecule.yml` (not sure on details) and a `.ansible-lint` had to be introduced to configure `ansible-lint` correctly.

Rickkwa:
Potential blocker for 2.20 release, as having linting break would be a pretty big deal I think.

lwm:
Agreed.

lwm:
Thanks for that fix @pilou-!

lwm:
OK, the fix is merged, awaiting a release which is being tracked at https://github.com/ansible/ansible-lint/milestone/15.

Rickkwa:
Looks like ansible-lint 4.1.0 was released. I ran the commands from my original comment and looks good.

gundalow:
@Rickkwa Thank you for confirming.

pilou-:
@gundalow should not we exclude ansible-lint 4.0.0 and 4.0.1 (from allowed dependencies) ?

PR #1743, opened 2019-02-13
Exclude ansible-lint 4.0.1

lwm:
To cover https://github.com/ansible/molecule/issues/1723#issuecomment-463011292.

/cc @pilou-


ssbarnea:
Ansilbe lint 4.1.0 was released so if that passed the testing we can switch to it.

gundalow:
So we know the newer version of ansible-lint is causing some CI failures, though we will fix them shortly.

Issue #1744, opened 2019-02-13
[investigate]: use of molecule with multi-role repositories


ssbarnea:
This tricket should be used for documenting current challanges that we may face for enabling molecule use in multi-role repositories.

1. MOLECULE_GLOB could be make user-overridable for allowing molecule to look inside each role directory.
2. Duplicate scenario names in different folders should be allowed: `ERROR: Duplicate scenario name 'default' found.  Exiting.` -- clearly we could have "default" on each role.
3. use should be able run molecule from any subdir of a git project, just like tox.
4. molecule should auto-load its base-config from the root of the current (git) project (look up cwd for first .git dir)
5. user should be limit molecule run on a single role, like `molecule -r foo list`
6. role and scenario filters should be supported as options instead of command args because they would make sense for most commands. Running `molecule --help` shoudl list them, without needing to ask user to do a `molecule [list|lint|matrix|test|....] --help` in order to see that they add `-s foo` argument. Yep, for very few commands these do not make sense, but they can just be ignored in those cases (if we **really** want we can even add argparse logic to disalow that).

To track progress on implementing the multi-role testing with molecule we should create an empty project for testing this use.

lwm:
Hey @ssbarnea, could you explain further point number 6? I didn't quite grok it. 

The other points seem reasonable and useful! Thanks for pushing this forward.

ssbarnea:
@lwm Sure, let me try to  describe it here and later we can rephrase the ticket to make it clear.

Expected behaviour:
```
molecule --help
# output should include a line like:
-s, --scenario SCENARIO       Limit molecule execution to SCENARIO
```
Currently this information is displayed only when you run `molecule list --help` (or other command),

15/16 currently listed commands to allow -s param, only one does not. It would be better to document these shared arguments at the root-level instead of displaying them only when a command is added.

True that we need to be careful not to make then disapear from the current location when doing that.

lwm:
@ssbarnea, thanks for that. I am not sure agree that this would be better. However, I'd rather not block your progress and motivation on this multiple role functionality. This discussion on CLI parameter discovery seems out of scope for the minimal steps needed at making Molecule better at handling multiple roles, so maybe we can leave it out for now? I think we need to improve the CLI in general and this discussion will be part of that.

Lirt:
Hi,

To add my ideas to point 4.

[...]

PR #1745, opened 2019-02-13
Make default scenario name use the parent folder

ssbarnea:
#### PR Type

- Bugfix Pull Request

Changes the default scenario name from being just hardcoded 'default'
to be the directory name hosting the molecule.yml file.

This change will allow us to reuse (symlink) molecule.yml file across
scenarios and roles because we will not longer be forced to define
this field in each of them.

Related-Bug: #1744

### DIFF
--- a/molecule/command/init/scenario.py
+++ b/molecule/command/init/scenario.py
@@ -89,6 +89,11 @@ class Scenario(base.Base):
 
 
 def _role_exists(ctx, param, value):  # pragma: no cover
+    # if role name was not mentioned we assume that current directory is the
+    # one hosting the role and determining the role name.
+    if not value:
+        value = os.path.basename(os.getcwd())
+
     role_directory = os.path.join(os.pardir, value)
     if not os.path.exists(role_directory):
         msg = ("The role '{}' not found. "
@@ -137,7 +142,7 @@ def _default_scenario_exists(ctx, param, value):  # pragma: no cover
 @click.option(
     '--role-name',
     '-r',
-    required=True,
+    required=False,
     callback=_role_exists,
     help='Name of the role to create.')
 @click.option(
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -337,6 +337,8 @@ class Config(object):
             util.sysexit_with_message(msg)
 
     def _get_defaults(self):
+        scenario_name = (os.path.basename(os.path.dirname(self.molecule_file))
+                         or 'default')
         return {
             'dependency': {
                 'name': 'galaxy',
@@ -394,7 +396,7 @@ class Config(object):
             },
             'scenario': {
                 'name':
-                'default',
+                scenario_name,
                 'check_sequence': [
                     'cleanup',
                     'destroy',
--- a/test/scenarios/idempotence/molecule/raises/molecule.yml
+++ b/test/scenarios/idempotence/molecule/raises/molecule.yml
@@ -17,8 +17,9 @@ provisioner:
     destroy: ../../../../resources/playbooks/docker/destroy.yml
   lint:
     name: ansible-lint
-scenario:
-  name: raises
+# commented on purpose to validate that molecule will inherit it from parent folder name:
+# scenario:
+#   name: raises
 verifier:
   name: testinfra
   lint:
--- a/doc/source/ci.rst
+++ b/doc/source/ci.rst
@@ -236,8 +236,6 @@ conflict.
       name: ansible
       lint:
         name: ansible-lint
-    scenario:
-      name: default
     verifier:
       name: testinfra
       lint:
--- a/molecule/cookiecutter/molecule/{{cookiecutter.role_name}}/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/molecule.yml
+++ b/molecule/cookiecutter/molecule/{{cookiecutter.role_name}}/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/molecule.yml
@@ -46,8 +46,6 @@ provisioner:
   name: {{ cookiecutter.provisioner_name }}
   lint:
     name: {{ cookiecutter.provisioner_lint_name }}
-scenario:
-  name: {{ cookiecutter.scenario_name }}
 verifier:
   name: {{ cookiecutter.verifier_name }}
   lint:
--- a/molecule/scenario.py
+++ b/molecule/scenario.py
@@ -36,12 +36,15 @@ class Scenario(object):
     for testing the role in a particular way.  The default scenario is named
     ``default``, and every role should contain a default scenario.
 
+    Unless mentioned explicitly, the scenario name will be the directory name
+    hosting the files.
+
     Any option set in this section will override the defaults.
 
     .. code-block:: yaml
 
         scenario:
-          name: default
+          name: default  # optional
           create_sequence:
             - create
             - prepare
--- a/test/unit/test_config.py
+++ b/test/unit/test_config.py
@@ -448,6 +448,13 @@ def test_interpolate_raises_on_failed_interpolation(patched_logger_critical,
     patched_logger_critical.assert_called_once_with(msg)
 
 
+def test_get_defaults(config_instance, mocker):
+    mocker.patch.object(config_instance, 'molecule_file',
+                        '/path/to/test_scenario_name/molecule.yml')
+    defaults = config_instance._get_defaults()
+    assert defaults['scenario']['name'] == 'test_scenario_name'
+
+
 def test_preflight(mocker, config_instance, patched_logger_info):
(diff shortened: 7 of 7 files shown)


lwm:
Thanks for following up on https://github.com/ansible/molecule/issues/1744! We've had a lot of discussion about conflicts lately - mainly the duplication of instance identifiers (see https://github.com/ansible/molecule/issues/1573). So, I am glad we're also getting around to this conversation too!

So, the goal of this PR is described in the following (from #1744):

> 2\. Duplicate scenario names in different folders should be allowed: `ERROR: Duplicate scenario name 'default' found.  Exiting.` -- clearly we could have "default" on each role.

So, the question for me is: is using the parent directory any better than the default name. I think it quickly solves the problem of duplicate scenario names but I am concerned that the parent directory isn't a clear name for a scenario. Default is not great either: it can be misleading since most useful roles can have multiple entry points and combinations of variables which change the flow of execution - there may not be a clear 'default'.

TBH, I lean on the side of making `-s scenario-name` a mandatory option. You should provide the name for your scenario. We have collectively learned that explicit is better than implicit and I think it is relevant here.

So, my proposal is to make it mandatory and not do any implicit naming generation here. Thoughts?

molecule/config.py
```diff
@@ -392,7 +392,7 @@ def _get_defaults(self):
             },
             'scenario': {
                 'name':
-                'default',
+                os.path.split(os.path.dirname(self.molecule_file))[-1],
```
lwm: We should add some tests for changes here since this is core functionality.

ssbarnea: @lwm please, please please... mo mandadatory options on cli! So many reasons:

This would prevent us from reusing the same file using symlinks. That was the point of not having to define a scenario name. 

Making the -s mandatory would be a serious UX downgrade and mostly a bad design. A tool should almost never have mandatory arguments, as in this case there are support to be arguments.

Still, you touched an interesting aspect: why do we really need to have an implicit "default" scenario in the first place? To me I would say that default behavior should be to run all scenarios found and to be able to optionally add `-s` param to limit it to a specific scenario (or list, or glob). It would not be really bad to be able to `-s foo*:bar`.

I think that we could look at tox cli options an lear from them: by default it runs all  environments defined in env_list but user can also tell it to run a specific one(or list) via `tox -e foo` or `TOX_ENV` variable. I would really find tox as "sucks!" if it woudl always force me to mention the -e param.

Regarding testing, I guess I could change the tests in order to validate that the directory name ends-up as being used as implicit scenario name.

For the other changes I think is better to leave them to their own tickets and separated PRs as big PRs are hard to review and merge, and more likely to break things.

lwm:
OK, I can pass on mandatory arguments. Maybe I'm sending this conversation in the wrong direction with this. And I get the sense we leave ourselves open to bikeshedding on default vs. my-role-name as the default generated scenario.

How about this, how about this - we try to resolve the issue of getting 'ERROR: Duplicate scenario name 'default' found. Exiting.' internally which is our main issue. Can we change this internal logic to look for duplicate scenario names **within** the role (instead of globally across all roles). So, for example, if we have `role1/molecule/default` and `role2/molecule/default`, Molecule will run them just fine because `role1` and `role2` are different. WDYT?

lwm:
Let's discuss this at the WG meeting today if possible!

molecule/config.py
tima: +1 to no mandadatory options on cli.

themr0c:
Please fix the errors in builds an we may be able to ship it with 2.20.

molecule/config.py
```diff
@@ -393,7 +393,7 @@ def _get_defaults(self):
...
```
decentral1se: Can we use some `os.path` function to avoid any possibility of `IndexError` reaching the user?

test/scenarios/idempotence/molecule/raises/molecule.yml
```diff
@@ -17,8 +17,9 @@ provisioner:
     destroy: ../../../../resources/playbooks/docker/destroy.yml
   lint:
     name: ansible-lint
-scenario:
-  name: raises
+# commented on purpose to validate that molecule will inherit it from parent folder name:
```
decentral1se: It's not clear what this is testing. Can we see a clearer example or a unit test as well?

molecule/config.py
seandst: This should work; I'm not sure if there's an even more direct way to express it.
```python
os.path.basename(os.path.dirname(self.molecule_file))
```

```python
>>> import os
>>> x = '/path/to/some/dir/molecule.yml'
>>> os.path.basename(os.path.dirname(x))
'dir'
```

It can also be `or`-guarded to default to `default`, e.g.
```python
>>> x = '/molecule.yml'  # but probably this should never happen, right?
>>> default_scenario_name = os.path.basename(os.path.dirname(x)) or 'default'
>>> default_scenario_name
'default'
```

test/scenarios/idempotence/molecule/raises/molecule.yml
seandst: Please disregard if this is not helpful, but I love this feature and want to see it make it to 2.20. To that end, I took a quick stab at a unit test:
```diff
diff --git a/test/unit/test_config.py b/test/unit/test_config.py
index 76c6abb8..ef5e2636 100644
--- a/test/unit/test_config.py
+++ b/test/unit/test_config.py
@@ -448,6 +448,13 @@ def test_interpolate_raises_on_failed_interpolation(patched_logger_critical,
     patched_logger_critical.assert_called_once_with(msg)
 
 
+def test_get_defaults(config_instance, mocker):
+    mocker.patch.object(config_instance, 'molecule_file',
+                        '/path/to/test_scenario_name/molecule.yml')
+    defaults = config_instance._get_defaults()
+    assert defaults['scenario']['name'] == 'test_scenario_name'
+
+
 def test_preflight(mocker, config_instance, patched_logger_info):
     m = mocker.patch('molecule.model.schema_v2.pre_validate')
     m.return_value = None
```

molecule/config.py
```diff
@@ -337,6 +337,8 @@ def _interpolate(self, stream, env, keep_string):
             util.sysexit_with_message(msg)
 
     def _get_defaults(self):
+        scenario_name = \
```
webknjaz: Please use brackets instead of EOL escaping
```suggestion
        scenario_name = (
```

molecule/config.py
```diff
@@ -337,6 +337,8 @@ def _interpolate(self, stream, env, keep_string):
...
+            os.path.basename(os.path.dirname(self.molecule_file)) or 'default'
```
webknjaz:
```suggestion
            os.path.basename(os.path.dirname(self.molecule_file)) or 'default'
        )
```

decentral1se: :+1: 

https://travis-ci.com/ansible/molecule/jobs/181909265#L233

molecule/model/schema_v2.py
```diff
@@ -613,6 +613,9 @@ def pre_validate_base_schema(env, keep_string):
                 'image': {
                     'type': 'string',
                 },
+                'dockerfile': {
```
decentral1se: :bomb:

webknjaz: What kind of change are you requesting here?

decentral1se: Ah right: it's a rebase mixup of some sort. This change is from https://github.com/ansible/molecule/pull/1769, looks like. Happened a few times now so I dropped a rather short bomb ;) Point taken though!

ssbarnea:
@decentral1se I am aware about the CLI behaviour related to default scenario and I find it normal. I do want to improve it but later. At this moment they work just fine: `molecule init role -r testrolelatest` will create a "default" scenario with a molecule.yml file that does not have a hardcoded 'default' name inside, allowing user to rename it if he wants.

Regarding your second question, it is more complicated and we should implement that change on different PR as is clearly not easy. This is because user may call molecule from deep within the codebase. Also molecule has a glob pattern that is using to detect roles, how does it know where to start from -- like root?

Other tools like `tox` are able to detect parent folder by looking for the `tox.ini` file, but molecule does not have such a config file at the root of the repository. This prevents us from finding the reference folder when doing lookups. As you know the glob pattern can be customized and can match multiple folders in multiple locations. From here it only gets more complex.

molecule/command/init/scenario.py
```diff
@@ -137,7 +142,7 @@ def _default_scenario_exists(ctx, param, value):  # pragma: no cover
 @click.option(
     '--role-name',
     '-r',
-    required=True,
+    required=False,
     callback=_role_exists,
```
ssbarnea: @decentral1se I hope this solved the problem you remarked regarding calling `molecule init scenario`. Now you can freely call it with `molecule init scenario -s foo` from inside the role directory.

seandst: YES. This is great.

PR #1746, opened 2019-02-13
Allow user to define custom molecule glob patterns

ssbarnea:
#### PR Type

- Bugfix Pull Request

By defining MOLECULE_GLOB, user can now change the folders where
molecule would look for configuration files.

If not defined it will continue to use the same default value.

Related-Bug: #1744

### DIFF
--- a/molecule/command/base.py
+++ b/molecule/command/base.py
@@ -32,7 +32,7 @@ from molecule import logger
 from molecule import util
 
 LOG = logger.get_logger(__name__)
-MOLECULE_GLOB = 'molecule/*/molecule.yml'
+MOLECULE_GLOB = os.environ.get('MOLECULE_GLOB', 'molecule/*/molecule.yml')
 MOLECULE_DEFAULT_SCENARIO_NAME = 'default'
 
 


ssbarnea:
@gundalow @lwm please review.

molecule/command/base.py
```diff
@@ -32,7 +32,7 @@
 from molecule import util
 
 LOG = logger.get_logger(__name__)
-MOLECULE_GLOB = 'molecule/*/molecule.yml'
+MOLECULE_GLOB = os.environ.get('MOLECULE_GLOB', 'molecule/*/molecule.yml')
```
lwm: Do we also need to register this in `molecule/config.py`? Can we also get a test for it?

ssbarnea: @lwm I do not know how to add a test for this right now as I have problems even running the test suite locally.

If you know how to add it please feel free to update it.

My personaly take on this would be to introduce it like this (as an undocumented feature) so we  can have time to work on a more complex usage scenario which would:

- add test suite for multiple roles
- document multiple role usage
- add a working example of multi roles (may be the same used for testing it)

lwm:
> My personaly take on this would be to introduce it like this (as an undocumented feature) so we can have time to work on a more complex usage scenario

:+1: :+1: :+1:

Issue #1730, opened 2019-02-06
PyPi uploads failing


lwm:
# Issue Type
- Bug report

Full logs at https://travis-ci.com/ansible/molecule/jobs/175744500#L1716.
Configuration at https://github.com/ansible/molecule/blob/master/.travis.yml#L369-L393.

gundalow:
```
NOTE: Try --verbose to see response content.
HTTPError: 400 Client Error: File already exists. See https://test.pypi.org/help/#file-name-reuse for url: https://test.pypi.org/legacy/
```

gundalow:
If https://travis-ci.com/ansible/molecule/jobs/175827471 passes we can close this

lwm:
On the latest build, it failed again: https://travis-ci.com/ansible/molecule/builds/101423592. 

So, I'm going to see what I can do to fix this shortly.

gundalow:
```
NOTE: Try --verbose to see response content.
HTTPError: 400 Client Error: File already exists. See https://test.pypi.org/help/#file-name-reuse for url: https://test.pypi.org/legacy/

```

lwm:
I've opened up https://github.com/ansible/molecule/pull/1755 for this.

webknjaz:
FTR I've posted the full analysis of this problem @ https://github.com/ansible/molecule/pull/1755#pullrequestreview-205294580

lwm:
https://travis-ci.com/ansible/molecule/jobs/178902007#L1715 :clap: :clap: :clap:

webknjaz:
Oh, so the problem will reoccur in just 25 commits :)

PR #1755, opened 2019-02-19
Do not fail the deploy build when version exists

lwm:
#### PR Type
- CI Fix

Closes https://github.com/ansible/molecule/issues/1730.

I have a suspicion that the real cause might be something else but this works for now. See:

> 12:35 < decentral1se> webknjaz: any idea why https://travis-ci.com/ansible/molecule/jobs/178776158#L1718 is happening?
> 12:36 < decentral1se> I only see https://test.pypi.org/project/molecule/2.19.1.dev96/ on test.pypi
> 12:36 < decentral1se> IIUC, that CI run should have been uploading 2.19.1.dev99?

References:
  * https://travis-ci.com/ansible/molecule/jobs/178776158#L1718
  * https://github.com/travis-ci/dpl#pypi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -432,6 +432,7 @@ jobs:
           HPBkDwncLYLEzxirPfNB88GYFWZ/wA1jMgCd4wT7SgKQlIyRSCT6KXMfoOUVMH/uwaLggoURaGtmD/qnrfXj+bG15nBCCOaIZdXzODln/PwDFAYT8ZspzRPzQOfncdk4WTsVbwpiGgpdm+TxGGBz8yedvVXiTwmwLCGC0FsAE8Wp8krNH1Kwqp3OaZeePakIbEj0UXgCTnAol3ZgVWWy+6bfDBb/aLiGXjsAIb7sY6HzwvQUr43xFO7tReaGO23mGwgIy/tNstXarwxezlw6FlGe5KJGvE/a4yAPuWg9kuJt5MqwbCJzhP38SeH4Gc6x0o6jPT/+NbvLV1ck1Qbz5gCYmXlOzucDP+P5t8E3g+zJbNbCR00k1BpG3MIEzrHeyp/hCdDYnGB3TVnlj+L4Y/yoUdCq9FJI8xHLbj/fMe5vxBMMsvDbt2JJReavi3WB2pv2nHtivkKj/Kow4FRm+Kp3WH4x5NlZDjV5gsUbwPcN1kIpb09sPZE2wYugvCTOiExt7mdd0JMMcgxuR1jGgGxqRARQeqjz14VVNsrVIIrMh8JX8u7Rl9sqmU4UuLGTxnq2cwyBmYzJ0gmDgN3TvQ96n9D2sP0YQj35v7RY5vsdYsfLucLdTfTajP9OL4Bw+ogXMJrArIq+j+7uJyTFqNdAN+Y6nX7WV0W6rp8aA1I=
       distributions: dists
       skip-cleanup: true
+      skip_existing: true
       on:
         all_branches: true
 


gundalow:
https://github.com/travis-ci/dpl#pypi

> skip_existing: Optional. When set to true, the deployment will not fail if a file with the same name already exists on the server. It won't be uploaded and will not overwrite the existing file. Defaults to false.

So I'm confused why we are trying to use the same filename

gundalow:
Though will be interesting to see if this "fixes" it

lwm:
Ah! I see the latest successful master branch build uploaded:

> https://travis-ci.com/ansible/molecule/jobs/175075271

And in the latest build, we do try to upload this duplicate name:

> molecule-0.0.1.dev99-py2.py3-none-any.whl

However, what was confusing and still is, is that I don't see this on test.pypi: https://test.pypi.org/project/molecule/#files.

I think I have a lead on the real fix now. Although I think we should include this change anyway in the case that a developer is making a prerelease manually or so and then the build trys to do it also. Less brittle builds are good.

> Though will be interesting to see if this "fixes" it

We'll only find out on a build on the master branch unfortunately.

.travis.yml
```diff
@@ -432,6 +432,7 @@ jobs:
           HPBkDwncLYLEzxirPfNB88GYFWZ/wA1jMgCd4wT7SgKQlIyRSCT6KXMfoOUVMH/uwaLggoURaGtmD/qnrfXj+bG15nBCCOaIZdXzODln/PwDFAYT8ZspzRPzQOfncdk4WTsVbwpiGgpdm+TxGGBz8yedvVXiTwmwLCGC0FsAE8Wp8krNH1Kwqp3OaZeePakIbEj0UXgCTnAol3ZgVWWy+6bfDBb/aLiGXjsAIb7sY6HzwvQUr43xFO7tReaGO23mGwgIy/tNstXarwxezlw6FlGe5KJGvE/a4yAPuWg9kuJt5MqwbCJzhP38SeH4Gc6x0o6jPT/+NbvLV1ck1Qbz5gCYmXlOzucDP+P5t8E3g+zJbNbCR00k1BpG3MIEzrHeyp/hCdDYnGB3TVnlj+L4Y/yoUdCq9FJI8xHLbj/fMe5vxBMMsvDbt2JJReavi3WB2pv2nHtivkKj/Kow4FRm+Kp3WH4x5NlZDjV5gsUbwPcN1kIpb09sPZE2wYugvCTOiExt7mdd0JMMcgxuR1jGgGxqRARQeqjz14VVNsrVIIrMh8JX8u7Rl9sqmU4UuLGTxnq2cwyBmYzJ0gmDgN3TvQ96n9D2sP0YQj35v7RY5vsdYsfLucLdTfTajP9OL4Bw+ogXMJrArIq+j+7uJyTFqNdAN+Y6nX7WV0W6rp8aA1I=
       distributions: dists
       skip-cleanup: true
+      skip_existing: true
```
webknjaz: Can we please stick to kebab-case here? The keys are normalized by travis internally, so there's nothing to worry about
```suggestion
      skip-existing: true
```

webknjaz:
Oh, and another addition which would improve this could be checking the version of the generated dist and failing the job manually. This can be done (1) either in `before_deploy` step or (2) in ` metadata-validation` env in `tox.ini` (https://github.com/ansible/molecule/blob/72b950f/tox.ini#L177-L179). Thankfully, this case is pretty deterministic so we could probably check that the base version is not `0.0.1`

lwm:
Thanks for this explanation @webknjaz! 

I was struggling with the layers of Travis `dpl`/`setuptools`/`pbr`/`tox`!

I've opened up https://github.com/ansible/molecule/pull/1756 for the fix then :+1:

.travis.yml
lwm: >     * OTOH, adding this option will silently hide any future errors like this which is unfortunate. I'd like this to not be left unnoticed so I prefer it to fail the job. This way humans are less likely to miss it.

Fair enough! I'll leave this off.

PR #1759, opened 2019-02-20
Assure cleanup of python precompiled files with tox

ssbarnea:
Sometimes leftovers files like `.py[co]` or `__pycache__` can break tox execution.

Use of `PYTHONDONTWRITEBYTECODE=1` in `tox.ini` is not enough because this
prevents processed started by tox to create these files. Still, if user
run code without it they could also be created and affecting tox
execution later.

By doing a clean-up step we assure that this does not affect test
execution.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,14 +28,18 @@ extras =
     lxc
     openstack
     vagrant
+commands_pre =
+    find {toxinidir} -type f -not -path '{toxinidir}/.tox/*' -not -path '*/__pycache__/*' -name '*.py[c|o]' -delete
 commands =
     unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
     functional: pytest test/functional/ {posargs}
     lint: flake8
     lint: yamllint -s test/ molecule/
+whitelist_externals =
+    find
 
 [testenv:lint]
-deps = 
+deps =
     -rlint-requirements.txt
 extras = []
 skip_install = true


tox.ini
```diff
@@ -28,14 +28,18 @@ extras =
...
     unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
     functional: pytest test/functional/ {posargs}
     lint: flake8
     lint: yamllint -s test/ molecule/
+whitelist_externals =
+    find
 
 [testenv:lint]
-deps = 
+deps =
```
webknjaz: Please try to separate logical changes in the future. Even this small...

tox.ini
```diff
@@ -28,14 +28,18 @@ extras =
     lxc
     openstack
     vagrant
+commands_pre =
+    find {toxinidir} -type f -not -path '{toxinidir}/.tox/*' -not -path '*/__pycache__/*' -name '*.py[c|o]' -delete
```
webknjaz: Why not `__pycache__`? I normally also delete such folders to remove as many side effects as possible.
Any objections?

webknjaz:
@ssbarnea we need to either rebase this commit or hit `Resolve conflicts` in GitHub's UI. Whichever you prefer.

tox.ini
ssbarnea: My impression is that __pycache__ folders are not affected by the same conflicts as the other files as they allow multiple interpreters to coexists without conflicts.

I copied this sequence from another project but I have nothing against removing this folders too if anyone knows that they can really cause conflicts.

tox.ini
ssbarnea: Sure. I will try. Even better, I do know how to avoid the need for such fixes with some additional pre-commit hooks.

tox.ini
webknjaz: True. That's what I thought. Just wanted to be on the safe side :)

webknjaz:
GitHub thinks that there's a conflict for some reason :(

tox.ini
webknjaz: Well, yeah... `-n` or/and `-p` are your friends in Git :)

Issue #1761, opened 2019-02-20
molecule sdist is huge due to inclusion of assets


ssbarnea:
Current molecule sdist is >4MB in side because it includes assets/ which are not really source code.

This makes the compilation much longer and increases the footprint for no real reason.

PR #1762, opened 2019-02-20
Avoid including asset folder in sdist

ssbarnea:
Adding `asset` folder to sdist produces a >4MB file instead of ~240kb
for now reason and this is not needed for molecule usage.

Adds dist/ and build/ folders to .gitignore as it is not normal
to have untracked files after packaging the module.

Fixes: #1761

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@
 .tox/
 .vagrant/
 .pytest_cache/
+build/
+dist/
 doc/build/
 htmlcov/
 pytestdebug.log
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,6 @@
+include LICENSE README.rst
+include requirements.txt
+include test-requirements.txt
+include lint-requirements.txt
+prune asset
+prune tests


ssbarnea:
AFAIK that is the only way to avoid it from being compiled. I tried finding a setup.cfg before ending up with thus approach but failed.

themr0c:
Successfully passed the CI this night.

webknjaz:
Okay, then I suggest getting back to it once we migrate to more canonical setuptools

lwm:
Do we take on some risk by defining our own MANIFEST.in now and stop letting `pbr` generate it? I do appreciate that we want to get rid of `pbr` and therefore should leave build configuration alone. OTOH, the following release `sdist` can be considerably smaller with this change.

webknjaz:
Having a manifest file increases the number of places which we have to maintain. When we migrate to setuptools-scm it'll also take care of files tracked by Git.

webknjaz:
Current manifest file will probably be merged with rules generated by pbr (but I never checked how it's done internally)

ssbarnea:
@webknjaz Yes, this is what pbr does and for the moment I see no reason to have a different approach.

webknjaz:
Alright, let's merge this for now

PR #1766, opened 2019-02-21
Further bump the git clone depth

decentral1se:
As a result of observation in https://github.com/ansible/molecule/issues/1730#issuecomment-465264335.

Don't want to see any failing builds right now :)

Releases coming soon and then I will drop this value back down to 100 so we are getting our original performance optimisation.

Signed-off-by: Luke Murphy <lukewm@riseup.net>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ conditions: v1
 dist: xenial
 
 git:
-  depth: 150
+  depth: 200
 
 cache:
   bundler: true


decentral1se:
Same change as https://github.com/ansible/molecule/pull/1756, so will merge.

webknjaz:
I think dropping it down is not very impactful. It's not like it's thousands of commits... The more important thing to do is to have a linter clearly warning us about reaching the limit.

PR #1769, opened 2019-02-22
Allow custom dockerfile location

ssbarnea:
Instead of always looking for Dockerfile.j2 in the scenario directory
allow user to define its location relatively to that folder.

Enables reuse of file across multiple scenarios without the need
to use symlinks.

#### PR Type

- Feature Pull Request

### DIFF
--- a/molecule/driver/docker.py
+++ b/molecule/driver/docker.py
@@ -47,6 +47,7 @@ class Docker(base.Base):
           - name: instance
             hostname: instance
             image: image_name:tag
+            dockerfile: Dockerfile.j2
             pull: True|False
             pre_build_image: True|False
             registry:
--- a/molecule/model/schema_v2.py
+++ b/molecule/model/schema_v2.py
@@ -613,6 +613,9 @@ platforms_docker_schema = {
                 'image': {
                     'type': 'string',
                 },
+                'dockerfile': {
+                    'type': 'string',
+                },
                 'pull': {
                     'type': 'boolean',
                 },
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -20,7 +20,7 @@
 
     - name: Create Dockerfiles from image names
       template:
-        src: "{{ molecule_scenario_directory }}/Dockerfile.j2"
+        src: "{{ molecule_scenario_directory + '/' + (item.dockerfile | default( 'Dockerfile.j2')) }}"
         dest: "{{ molecule_ephemeral_directory }}/Dockerfile_{{ item.image | regex_replace('[^a-zA-Z0-9_]', '_') }}"
       with_items: "{{ molecule_yml.platforms }}"
       when: not item.pre_build_image | default(false)
--- a/test/resources/playbooks/docker/create.yml
+++ b/test/resources/playbooks/docker/create.yml
@@ -20,7 +20,7 @@
 
     - name: Create Dockerfiles from image names
       template:
-        src: Dockerfile.j2
+        src: "{{ item.dockerfile | default('Dockerfile.j2') }}"
         dest: "{{ molecule_ephemeral_directory }}/Dockerfile_{{ item.image | regex_replace('[^a-zA-Z0-9_]', '_') }}"
       with_items: "{{ molecule_yml.platforms }}"
       when: not item.pre_build_image | default(false)
--- a/test/unit/model/v2/test_platforms_section.py
+++ b/test/unit/model/v2/test_platforms_section.py
@@ -145,6 +145,7 @@ def _model_platforms_docker_errors_section_data():
             'hostname': int(),
             'image': int(),
             'pull': int(),
+            'dockerfile': bool(),
             'pre_build_image': int(),
             'registry': {
                 'url': int(),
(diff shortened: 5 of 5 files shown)


molecule/provisioner/ansible/playbooks/docker/create.yml
```diff
@@ -20,7 +20,7 @@
 
     - name: Create Dockerfiles from image names
       template:
-        src: "{{ molecule_scenario_directory }}/Dockerfile.j2"
+        src: "{{ molecule_scenario_directory + '/' + (item.dockerfile | default( 'Dockerfile.j2')) }}"
```
decentral1se: You'll need to update the `platforms_docker_schema` and add tests for it as well as the `test/resources/playbooks/docker/create.yml` to match this functionality under functional testing. Ideally, you could also document this in `molecule/driver/docker.py`.

molecule/model/schema_v2.py
```diff
@@ -608,6 +608,9 @@ def pre_validate_base_schema(env, keep_string):
                 'image': {
                     'type': 'string',
                 },
+                'dockerfile': {
```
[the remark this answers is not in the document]
ssbarnea: I don't know how to test that, any update would be welcomed.

molecule/driver/docker.py
```diff
@@ -90,7 +91,7 @@ class Docker(base.Base):
             buildargs:
                 http_proxy: http://proxy.example.com:8080/
 
-    If specifying the `CMD`_ directive in your ``Dockerfile.j2`` or consuming a
+    If specifying the `CMD`_ directive in your ``dockerfile`` or consuming a
```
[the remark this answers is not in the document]
ssbarnea: That change is only to make the documentation correct. If I would use Dockerfile.j2 and user had defined a different value for dockerfile, the doc will not be correct. If you want I could mention that default value is `Dockerfile.j2`  by adding `(defaults to Dockerfile.j2)`.

molecule/model/schema_v2.py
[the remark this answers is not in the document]
decentral1se:
```diff
diff --git a/test/unit/model/v2/test_platforms_section.py b/test/unit/model/v2/test_platforms_section.py
index 7cdbd3d6..ea14a095 100644
--- a/test/unit/model/v2/test_platforms_section.py
+++ b/test/unit/model/v2/test_platforms_section.py
@@ -121,6 +121,7 @@ def _model_platforms_docker_errors_section_data():
             'hostname': int(),
             'image': int(),
             'pull': int(),
+            'dockerfile': bool(),
             'pre_build_image': int(),
             'registry': {
                 'url': int(),
```

molecule/driver/docker.py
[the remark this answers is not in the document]
decentral1se: > That change is only to make the documentation correct.

Not as I understand it.

This section is about `command` and `override_command` configuration, not `dockerfile`, so it feels strange to mention it here.  And I think we should default to `Dockerfile.j2` when talking about the file because that it what users are most familiar with. Otherwise, by this logic, we should be updating multiple places in the documentation with `Dockerfile.j2` -> `dockerfile`.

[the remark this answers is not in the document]
ssbarnea: Nevermind, i removed it.

decentral1se:
Pushing this into v2.20. Seems low risk.

ssbarnea:
DO NOT MERGE, I am updating it now...

doc/source/ci.rst
```diff
@@ -232,8 +232,6 @@ conflict.
       name: ansible
       lint:
         name: ansible-lint
-    scenario:
```
decentral1se: Uhhhh, wait a second, why is this being changed now (since last review)?

What is the relation to `dockerfile`? Is this not for https://github.com/ansible/molecule/pull/1745?

test/unit/test_config.py
```diff
@@ -448,6 +448,13 @@ def test_interpolate_raises_on_failed_interpolation(patched_logger_critical,
     patched_logger_critical.assert_called_once_with(msg)
 
 
+def test_get_defaults(config_instance, mocker):
```
decentral1se:
```
    def test_get_defaults(config_instance, mocker):
        mocker.patch.object(config_instance, 'molecule_file',
                            '/path/to/test_scenario_name/molecule.yml')
        defaults = config_instance._get_defaults()
>       assert defaults['scenario']['name'] == 'test_scenario_name'
E       AssertionError: assert 'default' == 'test_scenario_name'
E         - default
E         + test_scenario_name
```

molecule/config.py
```diff
@@ -337,6 +337,8 @@ def _interpolate(self, stream, env, keep_string):
             util.sysexit_with_message(msg)
 
     def _get_defaults(self):
+        scenario_name = \
```
decentral1se: ?

molecule/config.py
```diff
@@ -394,7 +396,7 @@ def _get_defaults(self):
             },
             'scenario': {
                 'name':
-                'default',
+                scenario_name,
```
decentral1se: ?

test/unit/test_config.py
decentral1se: ?

decentral1se:
Build failure unrelated, thanks!

Issue #1763, opened 2019-02-20
Allowing empty docker command is gonna break existing roles


beenje:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible 2.7.7
  config file = /Users/benjaminbertrand/.ansible.cfg
  configured module search path = [u'/Users/benjaminbertrand/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/benjaminbertrand/miniconda3/envs/molecule-dev/lib/python2.7/site-packages/ansible
  executable location = /Users/benjaminbertrand/miniconda3/envs/molecule-dev/bin/ansible
  python version = 2.7.15 | packaged by conda-forge | (default, Nov 28 2018, 14:39:28) [GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
molecule, version 2.19.1.dev125
```

Molecule installation method (one of):

- source (pip -e .)

Ansible installation method (one of):

- pip

# Desired Behavior

We have many roles with the following structure:

```
├── molecule
│   └── default
│       ├── Dockerfile.j2
│       ├── molecule.yml
│       ├── playbook.yml
│       ├── prepare.yml
│       └── tests
│           └── test_default.py
```

No create.yml and the following Dockerfile.j2:
```
$ cat molecule/default/Dockerfile.j2
# Molecule managed

{% if item.registry is defined %}
FROM {{ item.registry.url }}/{{ item.image }}
{% else %}
FROM {{ item.image }}
{% endif %}

RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get install -y python sudo bash ca-certificates && apt-get clean; \
    elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install python sudo python-devel python2-dnf bash && dnf clean all; \
    elif [ $(command -v yum) ]; then yum makecache fast && yum install -y python sudo yum-plugin-ovl bash && sed -i 's/plugins=0/plugins=1/g' /etc/yum.conf && yum clean all; \
    elif [ $(command -v zypper) ]; then zypper refresh && zypper install -y python sudo bash python-xml && zypper clean -a; \
    elif [ $(command -v apk) ]; then apk update && apk add --no-cache python sudo bash ca-certificates; \
    elif [ $(command -v xbps-install) ]; then xbps-install -Syu && xbps-install -y python sudo bash ca-certificates && xbps-remove -O; fi
```

In molecule.yml:
[...]

Issue #1441, opened 2018-08-14
Why does docker create.yml command default to sleep ? 


rgarrigue:
Hi

I'm wondering why the command in the default `create.yml` below default to this sleep loop ? Shouldn't this specific left to the original Docker image ? 

```yaml
...
    - name: Create molecule instance(s)
      docker_container:
        name: "{{ item.name }}"
        docker_host: "{{ item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
        hostname: "{{ item.hostname | default(item.name) }}"
        image: "molecule_local/{{ item.image }}"
        state: started
        recreate: false
        log_driver: json-file
        command: "{{ item.command | default('bash -c \"while true; do sleep 10000; done\"') }}"
        privileged: "{{ item.privileged | default(omit) }}"
        security_opts: "{{ item.security_opts | default(omit) }}"
...
```

I'm asking because I find this counter-intuitive. I know the "specify command: /.../init for systemd container" is documented, but working with molecule for some times I don't read the doc as often as before... and I ended up searching a couple of hour before realizing this is why my services were broken.

wilmardo:
My guess would be to keep the container running, by default a Docker container exits when no process is running. Bash sleep keeps the container busy until Ansible comes around for provisioning.
See these two SO questions: 
- [How can I keep container running on Kubernetes?](https://stackoverflow.com/questions/31870222/how-can-i-keep-container-%20running-on-kubernetes)
- [How to keep Docker container running after starting services?](https://stackoverflow.com/questions/25775266/how-to-keep-docker-container-running-after-starting-services)

retr0h:
What @wilmardo said

PR #1771, opened 2019-02-22
Add `override_command` command to platforms definition

decentral1se:
#### PR Type
- Bugfix Pull Request

Closes https://github.com/ansible/molecule/issues/1763.

This adds a `override_command` key to the platforms definition. It
defaults to true. A user who wants to override the `CMD` directive from
their `Dockerfile.j2` can specify `override_command: False` to have it
honoured.

This allows 3 use cases to be covered:

  * Getting the zero configuration default of `bash -c ...`
  * Overriding the `CMD` directive from the `command` key
  * Overriding the `CMD` directive from the `Dockerfile.j2` and setting `override_command: false`

See also:

  * https://github.com/ansible/molecule/pull/1615
  * https://github.com/ansible/molecule/issues/1614
  * https://github.com/ansible/molecule/issues/1441

I've done some manual testing against this as well for the above use cases.

I'd really appreciate review/testing from @fabianvf and @beenje!

### DIFF
--- a/molecule/driver/docker.py
+++ b/molecule/driver/docker.py
@@ -55,6 +55,7 @@ class Docker(base.Base):
                 username: $USERNAME
                 password: $PASSWORD
                 email: user@example.com
+            override_command: True|False
             command: sleep infinity
             privileged: True|False
             security_opts:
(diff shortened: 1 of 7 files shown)


molecule/driver/docker.py
```diff
@@ -88,6 +89,12 @@ class Docker(base.Base):
             buildargs:
                 http_proxy: http://proxy.example.com:8080/
 
+    If specifying the `CMD`_ directive in your ``Dockerfile.j2``, then you must
```
fabianvf: It's also relevant if you're consuming a built image, rather than building it yourself (this is the issue I was hitting when I made the initial change)

beenje:
This looks good to me! Thanks!

decentral1se:
Thank you both! I'll fixup the docs and re-run the CI and merge if no one else has further comment :+1:

PR #1774, opened 2019-02-23
Allow to add an extra inventory source

### LINKED ISSUE #1748: Allow to pass extra hosts to the inventory
# Issue Type

- Feature request

```
ansible 2.7.7
  config file = /Users/benjaminbertrand/.ansible.cfg
  configured module search path = [u'/Users/benjaminbertrand/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/benjaminbertrand/miniconda3/envs/molecule2/lib/python2.7/site-packages/ansible
  executable location = /Users/benjaminbertrand/miniconda3/envs/molecule2/bin/ansible
  python version = 2.7.15 | packaged by conda-forge | (default, Nov 28 2018, 14:39:28) [GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
molecule, version 2.19.0
```

- pip

- pip

# Desired Behavior

Molecule automatically generates the inventory based on the hosts defined under `platforms`.
I'd like the option to define extra hosts to be added to the inventory, not for provisioning but just to access variables from `hostvars[name]`.

We use a dynamic inventory and have several roles (for dhcp, dns...) that use variables from other hosts. Testing those roles with molecule requires to have extra hosts, groups and variables defined.

I couldn't find a way to do that today.
It would be nice if we could give a base inventory with some pre-defined hosts that molecule would then update, or maybe add a `extra_hosts` section directly in molecule.yml

beenje:
#### PR Type

- Feature Pull Request

This change allows to pass a link to another inventory source when
using symlinking (with the new `hosts` key).
This source can be any inventory accepted by Ansible (ini file, yaml or
even a dynamic inventory script).

All inventory files (including host_vars and group_vars directories) are
moved to an "inventory" directory.
During converge, the ansible-playbook command is always passed this
directory as inventory (Ansible will automatically merge the multiple
inventory sources).

All hosts, groups and variables from this extra inventory are available
when running `molecule converge`.
This allows to access variables from extra hosts not created by molecule
using for example something like `hostvars["extra-host"]["extra-var"]`
Note that you probably want to restrict the playbook to the hosts
created by molecule (use the appropriate group instead of `all`).

Close #1748

Signed-off-by: Benjamin Bertrand <beenje@gmail.com>


molecule/provisioner/ansible.py
```diff
@@ -248,15 +248,19 @@ class Ansible(base.Base):
 
     An alternative to the above is symlinking.  Molecule creates symlinks to
     the specified directory in the inventory directory.  This allows ansible to
-    converge utilzing its built in host/group_vars resolution.  These two
+    converge utilizing its built in host/group_vars resolution.  These two
     forms of inventory management are mutually exclusive.
+    Symlinking even allows to pass an additional inventory file (or dynamic inventory script),
```
decentral1se: Could I nitpick on a new line above here for the purposes of readability?

beenje: You can :-)

beenje:
> > Note that you probably want to restrict the playbook to the hosts created by molecule (use the appropriate group instead of `all`).
> 
> Oh! Thanks for documenting. That sounds bad. Should we somehow make it default that Molecule doesn't target these extra hosts? I think we take on a responsibility here to not potentially bork extra hosts that are targetted when users miss this small detail (changing the group). Could we introduce a new configurable option that ensures we don't target those hosts?

There is nothing to do for `molecule create` and `molecule destroy`. They loop on `molecule_yml.platforms` so they are not impacted by the extra inventory.
1. `molecule converge` runs the `molecule/default/playbook.yml` that targets `all` by default.
I don't think it makes sense to add an option in `molecule.yml` to change that. How would that work? By adding a `--limit` option in the background? That sounds too much like hidden magic to me.
Molecule is built on top of Ansible playbooks because people are familiar with that. If you add extra hosts to your inventory, changing the playbook you run seems like the right thing to do. This playbook is not hidden somewhere.
2. Even if people use existing hosts (for testing, it can often be just dummy hosts) and someone runs a playbook on `all`, it should fail on those extra hosts.
Molecule uses some very specific connection parameters for the hosts it has created. I doubt that those parameters would work on "real" hosts. I admit that I only tested that with the docker and vagrant driver.
3. Same thing for `molecule verify`. I only know about testinfra. In your `molecule/default/tests/test_default.py` you have to target the hosts you want with:
```
testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
    os.environ['MOLECULE_INVENTORY_FILE']).get_hosts('all')
```
This is actually a line I often change to target a specific group or host.

Those files are generated from a cookiecutter template as a base to get you started. If we change something, maybe the template could be updated.
We could always target a `molecule_group` by default. But I think it makes things more complex for many users who don't care about that. 

I see using an extra inventory as an advanced option. People that need that should understand the impact.

> Thinking about backwards compatibility, this change seems fine. Can you see any issues? The only case I can think of is where there are still /tmp/molecule/... ephemoral directories created between runs but that seems like an edge case. We might want to think about destroying (if this is not already happening).

I don't see any issues. I add a new directory under the ephemeral directory. So even if there are old files present, they will just be ignored.

> 
> And finally, I saw in `ansible.py`, we have:
> 
> > An alternative to the above is symlinking.  Molecule creates symlinks to the specified directory in the inventory directory.  This allows ansible to converge utilizing its built in host/group_vars resolution.  These two forms of inventory management are mutually exclusive.
> 
> So, is it also possible to define `hosts` using the static configuration method? We should make sure that is also possible within this change and not just for the symlink method.

I thought about that. On once side, I agree. It sounds cleaner to allow the same parameters in the static configuration. On the other side, it requires some work to add an option that will probably be seldom use. I think we would have to use something like:
```
        provisioner:
          name: ansible
          inventory:
            hosts:
              all: # keys must be unique, i.e. only one 'hosts' per group
                hosts:
                  test1:
                  test2:
                    var1: value1
                vars:
                  group_var1: value2
                children:   # key order does not matter, indentation does
                   other_group:
                   last_group:
                     hosts:
                       test1 # same host as above, additional group membership
                     vars:
                       last_var: MYVALUE
            group_vars:
              foo1:
                foo: bar
              foo2:
                foo: bar
                baz:
                  qux: zzyzx
            host_vars:
              foo1-01:
                foo: bar

```

It's quite verbose and I think it's easier to create a file. OK, I can hear this is a weak argument. So I should probably implement that anyway.
Could you please confirm?

molecule/provisioner/ansible.py
decentral1se: Heroic

decentral1se:
OK, great! Makes sense to me.

For me, to finish this PR is then a documentation issue. Can we do:

* Document that if you use an inventory file option, there is an additional responsibility to consider when targetting hosts on your playbook. Something that you might traditionally ignore when using Molecule otherwise.
* That this `hosts` does not work on the static configuration. You could leave this in a RST note or something and a link to the issue tracker. If someone would like to be possible, let's do it. Otherwise, I am happy to leave as is.

Thanks for the efforts :+1:

beenje:
- I ended adding support for `hosts` in the static configuration
- Let me know if the documentation I added is enough

decentral1se:
Would be nice to get another pair of eyes on this before merging! Thanks!

decentral1se:
OK, asked at WG and not sure if anyone has spare cycles, so getting this merged now then :+1:

beenje:
Thanks!

PR #1775, opened 2019-02-24
Disallow non-unique instance names for platforms

decentral1se:
#### PR Type
- Bugfix Pull Request

I ran into an issue where I named my instances the same (by accident)
and then when I ran `molecule create` I had 2 when I expected 3 containers
created. I realised we are missing validation for this.

References:
  * http://docs.python-cerberus.org/en/stable/customize.html#custom-rules

### DIFF
--- a/molecule/model/schema_v2.py
+++ b/molecule/model/schema_v2.py
@@ -18,6 +18,7 @@
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
 
+import collections
 import copy
 import functools
 import re
@@ -25,8 +26,7 @@ import re
 import cerberus
 import cerberus.errors
 
-from molecule import interpolation
-from molecule import util
+from molecule import interpolation, util
 
 
 def coerce_env(env, keep_string, v):
@@ -269,6 +269,8 @@ base_schema = {
                 'name': {
                     'type': 'string',
                     'required': True,
+                    'unique':  # https://github.com/pyeve/cerberus/issues/467
+                    True,
                 },
                 'groups': {
                     'type': 'list',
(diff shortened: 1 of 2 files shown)


decentral1se:
Hmmm, this needs some more work then ...

decentral1se:
Opened https://github.com/pyeve/cerberus/issues/467 against this.

decentral1se:
@webknjaz, any inspiration on this one? This seems like 'a good enough' fix ...

See final comment in https://github.com/ansible/molecule/pull/1775#issuecomment-466949763.

molecule/model/schema_v2.py
```diff
@@ -944,6 +945,20 @@ class Validator(cerberus.Validator):
...
         super(Validator, self).__init__(*args, **kwargs)
 
+    def _validate_unique(self, unique, field, value):
+        """Ensure value uniqueness.
+
+        The rule's arguments are validated against this schema:
+        {'type': 'boolean'}
+        """
+        if unique:
+            root_key = self.schema_path[0]
```
decentral1se: See http://docs.python-cerberus.org/en/stable/customize.html#validator-document-path-validator-schema-path.

Seems fine to index since it will be populated with the path to the value.

molecule/model/schema_v2.py
```diff
@@ -944,6 +945,20 @@ class Validator(cerberus.Validator):
...
+            data = [doc[field] for doc in self.root_document[root_key]]
```
webknjaz: Since you don't process data twice, use a generator expression which will allow lazy iteration without the creation of an intermediate list.
```suggestion
            data = (doc[field] for doc in self.root_document[root_key])
```

test/unit/model/v2/test_platforms_section.py
```diff
@@ -115,6 +115,23 @@ def test_platforms_docker(_config):
     assert {} == schema_v2.validate(_config)
 
 
+@pytest.mark.parametrize(
+    '_config', ['_model_platforms_docker_section_data'], indirect=True)
+def test_platforms_unique_names(_config):
+    _config['platforms'] += [{'name': 'instance'}]  # duplicate platform name
```
webknjaz: how about
```suggestion
    instance_name = _config['platforms'][0]['name']
    _config['platforms'] += [{'name': instance_name}]  # duplicate platform name
```

PR #1778, opened 2019-02-26
Documentation fix

robertdebock:
#### PR Type
- Docs Pull Request

The title describes "Create" but the code showed `state: destroy`.

### DIFF
--- a/doc/source/configuration.rst
+++ b/doc/source/configuration.rst
@@ -199,7 +199,7 @@ Create instances with interfaces.
                 network_name: private_network
                 type: static
             platform_box: ubuntu/trusty64
-            state: destroy
+            state: up
 
 Create instances with additional provider options.
 


decentral1se:
Nice one! You'll need to pass the `--signoff` flag again on your commit :sweat_smile:

robertdebock:
I guess I made a mistake somewhere, I can't get the DCO test to pass, although I've signed off locally and pushed it. Any suggestions?

decentral1se:
Ah, you have 3 commits now and one of them is not signed. If you run a `git rebase -i HEAD~4` then you can mark those extra commits with `squash` and then mark the original commit with `edit` and then interatively add the `--signoff` marker :) Git!

robertdebock:
> Ah, you have 3 commits now and one of them is not signed. If you run a `git rebase -i HEAD~4` then you can mark those extra commits with `squash` and then mark the original commit with `edit` and then interatively add the `--signoff` marker :) Git!
> 
> EDIT: `git reset HEAD~3` is also an option - then just make a new single commit ...

Got it, bit of messing around. Thanks!

decentral1se:
:rocket:

PR #1782, opened 2019-02-27
🚑 Avoid using SphinX-specific text-roles

webknjaz:
Drop InterSphinX usage from README

#### PR Type

- Bugfix Pull Request


decentral1se:
:1st_place_medal:

PR #1786, opened 2019-02-27
[Docs]: point to projects instead of milestones.

decentral1se:
As agreed in the latest WG.

### DIFF
--- a/doc/source/development.rst
+++ b/doc/source/development.rst
@@ -16,8 +16,8 @@ Release Engineering
 Pre-release
 -----------
 
-* Ensure the `GitHub Milestones`_ is complete and closed.
-* Edit the :ref:`changelog`, based on Milestone and recent commits.
+* Ensure the `GitHub Project`_ has all cards in the ``Done`` column.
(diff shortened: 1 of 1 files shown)


doc/source/development.rst
```diff
@@ -63,8 +63,8 @@ Post-release
 Roadmap
 =======
 
-* From Molecule v2.20 `GitHub Milestones`_ are used to track each release
+* From Molecule v2.20 the `GitHub Project`_ tool is used to track each release.
 
 .. _`PyPI`: https://pypi.org/project/molecule
-.. _`GitHub Milestones`: https://github.com/ansible/molecule/milestones
+.. _`GitHub Project`: https://github.com/ansible/molecule/projects
```
webknjaz: Can also put it into this section https://github.com/ansible/molecule/blob/master/setup.cfg#L15 so that it'd be on PyPI as well.

decentral1se: Yep! Submitted https://github.com/ansible/molecule/pull/1792.

Issue #1780, opened 2019-02-26
All docker platforms of a single scenario get the same command set (introduced in #1771)


zeitounator:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version
ansible 2.7.8
  config file = None
  configured module search path = ['/home/olcla/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.6/dist-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.6.7 (default, Oct 22 2018, 11:32:17) [GCC 8.2.0]
molecule, version 2.19.1.dev139

```

Molecule installation method (one of):

- sudo pip install git+https://github.com/ansible/molecule

Ansible installation method (one of):

- pip

# Desired Behavior

Recent implementation of `command_override` for docker containers in #1771 is buggy.

Suppose I have more than one platform:

```yaml
platforms:

  - name: container1
    image: thoteam/ansible-centos7-apache-java:latest
    command: /usr/sbin/init
    pull: true
    pre_build_image: true
    privileged: true
    published_ports:
      - 8090:443
    groups:
      - nexus
    networks: &nexus_networks
      - name: nexus_default

  - name: container2
    image: thoteam/ldap-server-mock:latest
    pull: true
    pre_build_image: true
    exposed_ports:
      - 3004
    volumes:
      - ${MOLECULE_PROJECT_DIRECTORY}/molecule/ldap-server-mock-conf.json:/srv/ldap-server-mock-conf.json
      - ${MOLECULE_PROJECT_DIRECTORY}/molecule/ldap-server-mock-users.json:/srv/ldap-server-mock-users.json
    groups:
      - mockldap
    networks: *nexus_networks
```

When I run `molecule create` with the above config, I'm expecting to see 2 running containers:
* container1 running /usr/sbin/init
* container2 running the default command (bash -c "while true; do sleep 10000; done")

# Actual Behaviour

Both containers are running the default command.

This is related to the new task introduced in #1771 (https://github.com/decentral1se/molecule/blob/d5255917c11d10ed161f6ee346e645222f22af69/molecule/provisioner/ansible/playbooks/docker/create.yml#L58).

The `set_fact` calculating `command_directive` runs on localhost and loops other the molecule platforms. So the var gets overwritten on each pass and only the last value for the last platform gets recorded and used for all platforms when creating the containers in the next task

My 2 cent towards the solution:
[...]

PR #1793, opened 2019-02-28
Handle missing `command_directives_dict` gracefully

decentral1se:
#### PR Type
- Bugfix Pull Request

Will await review from @zeitounator :+1: 

Pushing onto v2.20 as a further bug fix following on from https://github.com/ansible/molecule/issues/1780.

Turns out the fix for this original issue was not so simple after all :sweat_smile:

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -59,7 +59,7 @@
       set_fact:
         command_directives_dict: >-
           {{ command_directives_dict | default({}) |
-            combine({ item.name: item.command | default('bash -c "while true; do sleep 10000; done"') })
+             combine({ item.name: item.command | default('bash -c "while true; do sleep 10000; done"') })
           }}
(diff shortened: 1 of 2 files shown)


zeitounator:
:+1: 

I just took the time to correctly install molecule in editable mode. No more excuses not to provide a PR next time :)

Running from this branch, everything looks ok in all situations. Tested with two platforms and the following scenari:
* Both with override_command: false => OK (both use command defined at docker build time)
* One with override_command: false and other with command defined => OK (docker build time command and defined command)
* One with override_command. false and other without any command => OK (docker build time command and default molecule command).

Thanks for your prompt reaction.

decentral1se:
>  I just took the time to correctly install molecule in editable mode. No more excuses not to provide a PR next time :)

The more the merrier! :rocket: :rocket: :rocket: 

Also caught in (thanks @fabianvf for the report!):

> https://travis-ci.org/operator-framework/operator-sdk/jobs/499919307#L1874

Think we're on track with this fix then. I'll merge when the CI passes.

PR #1795, opened 2019-02-28
Assure removal of __pycache__ before each tox run

ssbarnea:
Avoid bug where presence of __pycache__ folders in source tree could
break test execution.

Now __pycache__ is wiped before each run, previously only .py[co]
files were removed.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -33,7 +33,7 @@ extras =
     vagrant
     windows
 commands_pre =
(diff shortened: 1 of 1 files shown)


ssbarnea:
@decentral1se @gundalow @fabianvf ^ easy fix.

webknjaz:
I'm curious if `tox` would allow us passing `--no-compile` to `pip install`. It would be nice improvement as well.

Issue #1596, opened 2018-11-17
devel: consolidate linting into a single job (tox target)


ssbarnea:
# Issue Type

- Feature request

# Molecule and Ansible details

At this moment in order to preform only linting a developer would have to run this command:
```
tox -e py27-lint,py36-lint,py37-lint,format-check
```

This is insanely long command and I really doubt anyone is actively using it. 

Also is takes a huge amount of disk space as it creates 4 different environments without any practical reasons, especially as it makes no sense to run linting with 3 different intepretes, using the latest supported version is more than enough. 

PS.  Please note that the alternative of using the oldest interpreter (py27) is not as good as the newest one because it is known that the parsing ability of the newer ones is greatlly improved. 

Also CI job should be updated to assure the same execution, matching local development usage.

Proposed command
```
tox -e lint
```

This approach is also future proof as we could add new linters or swap exiting ones without changing the user interface, the job would still have the same name regardless what internal tools are used at a moment.

webknjaz:
I'd like to see separate envs for different checks. I'd default to the current `python` executable, not bound to version. In CI we use Python 3.7, specified on the config level.

webknjaz:
Oh, and another thought: I tend to use pre-commit.com tool to consolidate linters in one place and then, additionally to CI, users may optionally set that up as a pre-commit or pre-push hook in their local repo with just a single one-time command.

decentral1se:
PR out for review on this: https://github.com/ansible/molecule/pull/1800.

webknjaz:
> py27-lint,py36-lint,py37-lint,format-check

you can do:
```ini
[toxenv:lint]
commands =
    {[py37-lint]commands}
    {[format-check]commands}
```

To achieve this.

ssbarnea:
Done, long time ago.

PR #1800, opened 2019-03-04
RFC: Use only one tox environment for linting: lint

ssbarnea:
Simplifies linting changes and avoid confusions around what needs to
be run before unittests.

Any further linting improvements should be included inside the same
tox environment.

Last command verifies that no untracked files are left or not uncommited
files are left, assuring that lint succeeds only if all changes
were commited.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,11 +113,6 @@ jobs:
       TOXENV: lint
 
   - <<: *lint
-    name: checking the code style
-    env:
-      TOXENV: format-check
-
-  - <<: *lint
     name: checking the docs build
     env:
       TOXENV: doc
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -112,11 +112,6 @@ jobs:
     env:
       TOXENV: lint
 
-  - <<: *lint
```
webknjaz: I'm "-1" here: collapsing multiple separate distinct checks into one results in the undebuggable log. We already have this problem, no need to make it worse.

decentral1se:
https://github.com/ansible/molecule/issues/1596#issuecomment-470517259 looks like the light at the end of the tunnel then!

Issue #1673, opened 2019-01-17
`MOLECULE_DEBUG` env var is not honoured


ssbarnea:
# Issue Type

- Feature request

# Desired Behavior

Like Ansible, molecule should load environment variable declared with `MOLECULE_` prefix and and use them as default values for command line arguments.

At this moment to enable debug more, user needs to add `--debug` after molecule and before the commands. This may prove quite inconvenient as the CLI call could be in a file or even other ci repository.

By regonnizing MOLECULE_DEBUG=1 it would be much easier to trigger the debug mode without having to alter the source code and the entire build chain.

Hint: Ansible has quite a nice implementaiton regarding parsing environment variables and translating them directly into default values, avoiding the need to implement this feature for each command line option.

lwm:
OK, `MOLECULE_DEBUG` is already being honoured. 

See https://github.com/ansible/molecule/blob/master/molecule/config.py#L189.

I assume you already tried this and it didn't work, so, is it that it needs to be `=True/False`?

lwm:
Hey @ssbarnea, I've played around with this with `MOLECULE_DEBUG=1` and `MOLECULE_DEBUG=True` and it didn't work. So, this looks like a bug!

ssbarnea:
We need to be sure that we correctly recognize:
* true: =1 =[tT]rue =yes
* false: =0 =[Ff]false =no
Just to be sure people are not surprised when they define it. If we cover for these we are safe.

PS. I am not sure which should be the behavior when variable is defined but empty.

webknjaz:
I've looked at the source and the root of the issue is that when merging `os.environ` with default hardcoded config default takes precedence over OS-provided env vars.

PR #1787, opened 2019-02-27
WIP: Make OS-provided env vars override defaults

webknjaz:
Fixes #1673

#### PR Type

- Bugfix Pull Request

PR #1802, opened 2019-03-05
Assure MOLECULE_DEBUG value is used

ssbarnea:
Previously the value of MOLECULE_DEBUG environment variable was ignored
because configuration was taking precedence.

Fixes: https://github.com/ansible/molecule/issues/1673

### DIFF
--- a/molecule/shell.py
+++ b/molecule/shell.py
@@ -30,6 +30,7 @@ import click_completion
 import molecule
 from molecule import command
 from molecule import util
+from molecule.config import MOLECULE_DEBUG
 
 click_completion.init()
 
@@ -87,7 +88,7 @@ def _allowed(ctx, param, value):  # pragma: no cover
 @click.group()
 @click.option(
     '--debug/--no-debug',
-    default=False,
+    default=MOLECULE_DEBUG,
     callback=_allowed,
     help='Enable or disable debug mode. Default is disabled.')
 @click.option(
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -21,6 +21,7 @@
 import os
 
 import anyconfig
+from ansible.module_utils.parsing.convert_bool import boolean
 import six
 
 from molecule import interpolation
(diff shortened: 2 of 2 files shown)


decentral1se:
Oh nice one! How does this relate to https://github.com/ansible/molecule/pull/1787?

Don't have time to take a pass now but will do later ... would be great to fix this :+1:

ssbarnea:
@decentral1se the other patch is "WIP" so it does not make sense to review it. It may be better/more generic but while is WIP I cannot estimate when this bug will be fixed,.... so I prefer to fix what I know is broken first.

webknjaz:
@ssbarnea feel free to cherry-pick that branch. It fixes one part of the problem.

decentral1se:
What @webknjaz said. That PR is also a fix for the issue you're trying to fix. Please check it out.

Also, we'll need some tests for this change. If you can manually test as well, that'd be great ...

decentral1se:
Oooh nice, when I cherrypick https://github.com/ansible/molecule/pull/1802/commits/832d1ade02b7fcbe40c27b0f408a2c462907948b and https://github.com/ansible/molecule/pull/1787/commits/c3abca5783820fee88c2d681a7e78d4df4d6004f into a branch, I can pass `MOLECULE_DEBUG=True` and `MOLECULE_DEBUG=0` and see the correct results. I haven't tried other env vars but we'd need to also make sure they're available too ...

molecule/shell.py
```diff
@@ -87,7 +88,7 @@ def _allowed(ctx, param, value):  # pragma: no cover
 @click.group()
 @click.option(
     '--debug/--no-debug',
-    default=False,
+    default=MOLECULE_DEBUG,
```
webknjaz: @ssbarnea @decentral1se I just realized that it's Click. And it supports env vars out of the box. There's `envvar='MOLECULE_DEBUG'` argument to the decorator and also there's a global prefix you can set up for options/args.
https://click.palletsprojects.com/en/7.x/options/#values-from-environment-variables

Issue #1806, opened 2019-03-06
PyYAML update CVE-2017-18342


gundalow:
# Issue Type

- Bug report

# summary

Currently we have
https://github.com/ansible/molecule/blob/9d1dec85ed2f995e2ebd24e234fd8fe4334ecb62/requirements.txt#L15

[CVE-2017-18342](https://nvd.nist.gov/vuln/detail/CVE-2017-18342) defines

**Vulnerable versions:** < 4.2b1
**Patched version:** 4.2b1
In PyYAML before 4.1, the `yaml.load()` API could execute arbitrary code. In other words, `yaml.safe_load` is not used.

Do we need to update?

decentral1se:
Following the note from jborean93 that we should make sure we're using `safe_load`, I grepped around and can't find an instance of the plain 'ol `load` call. Someone might want to check that again more solidly but I think we're OK for now

seandst:
greppage: find files with 'yaml' in them (to get `import yaml`, `from yaml.module import foo`, etc...), then print lines with 'load' in them:

```
$ for file in `grep -rl yaml molecule/`
> do if grep -q load $file
> then echo --
> echo $file:
> grep load $file
> fi
> done
--
molecule/config.py:
                    defaults, util.safe_load(interpolated_config))
                                        util.safe_load(interpolated_config))
        d = util.safe_load_file(env_file)
--
molecule/driver/ec2.py:
        instance_config_dict = util.safe_load_file(
--
molecule/driver/vagrant.py:
        instance_config_dict = util.safe_load_file(
--
molecule/driver/gce.py:
        instance_config_dict = util.safe_load_file(
--
molecule/driver/openstack.py:
        instance_config_dict = util.safe_load_file(
--
molecule/driver/delegated.py:
        instance_config_dict = util.safe_load_file(
--
molecule/driver/linode.py:
        instance_config_dict = util.safe_load_file(
--
molecule/driver/azure.py:
        instance_config_dict = util.safe_load_file(
--
molecule/provisioner/ansible.py:
    The playbook loading order is:
--
molecule/provisioner/ansible/plugins/filters/molecule_core.py:
    return util.safe_load(interpolated_data)
--
molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py:
  vagrant_config_hash = YAML::load_file('{{ vagrantfile_config }}')
--
molecule/util.py:
def safe_load(string):
        return yaml.safe_load(string) or {}
def safe_load_file(filename):
[...]

PR #1808, opened 2019-03-06
Mitigate against PyYaml CVE-2017-18342

ssbarnea:
While we already use only safe_load on molecule we are better-of
forcing use of a version on PyYaml which is not affected by this
security issue.

Fixes: #1806

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ Jinja2==2.10
 pbr==5.1.1
 pexpect==4.6.0
 psutil==5.4.6; sys_platform!="win32" and sys_platform!="cygwin"
-PyYAML==3.13
+PyYAML>=5.1  # MIT
 sh==1.12.14
 six==1.11.0
 tabulate==0.8.2


requirements.txt
```diff
@@ -12,7 +12,7 @@ Jinja2==2.10
 pbr==5.1.1
 pexpect==4.6.0
 psutil==5.4.6; sys_platform!="win32" and sys_platform!="cygwin"
-PyYAML==3.13
+PyYAML>=5.1b3  # MIT
```
webknjaz: We cannot depend on beta versions.

decentral1se:
I'd be inclined to leave this off based on https://github.com/ansible/molecule/pull/1808#discussion_r263610826 and https://github.com/ansible/molecule/issues/1806#issuecomment-470742420.

requirements.txt
ssbarnea: Please add some comments on https://github.com/yaml/pyyaml/issues/259 to raise awareness of the needed release of pyyaml.

decentral1se:
Awaits https://github.com/yaml/pyyaml/issues/259 :+1:

requirements.txt
webknjaz: Come to #pyyaml at Freenode to push this forward. Ingy is planning the release AFAIK but he is extremely cautious this time. I can't even push a testing PR hanging there since around the summer...
I can only say that everyone knows about this issue but what's needed is probably more collaboration and helping the maintainer. He even sometimes asks for changes feedback before creating actual PRs in the project he maintains, afraid to cause a shitload of useless negativity.
I wish I could dedicate some time to that project myself :(
Anyway, Ingy is quite responsive on IRC. I'm sure that if you offer him some help he'll be glad to accept it.

ssbarnea:
I am not sure if that closing is the right way to deal with it because is not really resolved, we are just waiting for an external dependency but our core is still affected by the use of an unsafe library.

decentral1se:
> I am not sure if that closing is the right way to deal with it because is not really resolved

Sorry if my meaning didn't come across well (or, it always feels bad to get a PR closed!) but https://github.com/ansible/molecule/issues/1806#issuecomment-470742420 remains open and is IMHO sufficient to track the progress of the new release of the library. I prefer to minimise noise in the PR/issues list where possible.

webknjaz:
@ssbarnea we can always reopen/resubmit this once the external requirement is fulfilled. Besides, I'd like to emphasize that **it is not a security issue** so no need to rush.

Oh, and also it might be wise first ensure that any other dists which are usually installed along with Molecule itself will not suffer the dependencies conflict. I can imagine that if some of our (direct or transitive) deps is following typical semver practices and has `pyyaml >=3,<4` to avoid unexpected breaking changes we can get pretty much unusable (impossible to install) software by locking `>5`.
On Molecule side we already use safe API.

ssbarnea:
cc

webknjaz:
cc https://github.com/ansible/molecule/issues/1806#issuecomment-472506822

requirements.txt
```diff
@@ -12,7 +12,7 @@ Jinja2==2.10
 pbr==5.1.1
 pexpect==4.6.0
 psutil==5.4.6; sys_platform!="win32" and sys_platform!="cygwin"
-PyYAML==3.13
+PyYAML>=5.1  # MIT
```
webknjaz: Why MIT? Also, this might deserve a major release.

cc @decentral1se

ssbarnea: @webknjaz The comment is only informative, I have seen documenting licenses of dependencies as comments as a common practice in OpenStack and even outside.

I think it does not hurt to start documentig licenses this way.

ssbarnea: I doubt we need a major release only for bumping some requirments. As long we do not break our API everything would count as minor.

decentral1se: The API remains unchanged, so think we're looking at a minor release here. Shall we add a change log entry for this? For the comments: it may confuse contributors to see comments with licenses on some dependencies and not on others (is this optional? What is this for? Should I do this?), so I'd rather just avoid this additional practice for our dependency management on this project.

Let's merge this soon!

webknjaz: changelog entry: yes
merge: move to `setup.cfg`

ssbarnea: I would refrain from migrating more into setup.cfg until packaging is fixed. I am considering proposing revert of the famous packaging change from last week https://github.com/ansible/molecule/issues/1966. 

I think it was too early as it was not tested enough.

webknjaz: It's not broken. Your use-case is not supported and you didn't provide steps to reproduce so it doesn't count. It works as long as you use it as designed.

webknjaz:
Closing as it should be done via https://github.com/ansible/molecule/issues/1806#issuecomment-486711379. Feel free to start doing that.

PR #1809, opened 2019-03-07
Fix custom dockerfile usage

ssbarnea:
Fixes recent bug which prevented the use of the new dockerfile param
that pointed to custom locations of Dockerfile.

It was not identified initially because we did not use the dockerfile param in any test.

#### PR Type

- Bugfix Pull Request

Fixes bug introduced in https://github.com/ansible/molecule/pull/1769

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -39,7 +39,7 @@
         path: "{{ molecule_ephemeral_directory }}"
         name: "molecule_local/{{ item.item.image }}"
         docker_host: "{{ item.item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
-        dockerfile: "{{ item.item.dockerfile | default(item.invocation.module_args.dest) }}"
+        dockerfile: "{{ item.invocation.module_args.dest }}"
         force: "{{ item.item.force | default(true) }}"
         pull: "{{ item.item.pull | default(omit) }}"
         buildargs: "{{ item.item.buildargs | default(omit) }}"
--- a/test/resources/playbooks/docker/create.yml
+++ b/test/resources/playbooks/docker/create.yml
@@ -39,7 +39,7 @@
         path: "{{ molecule_ephemeral_directory }}"
         name: "molecule_local/{{ item.item.image }}"
         docker_host: "{{ item.item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
-        dockerfile: "{{ item.item.dockerfile | default(item.invocation.module_args.dest) }}"
+        dockerfile: "{{ item.invocation.module_args.dest }}"
         force: "{{ item.item.force | default(true) }}"
         pull: "{{ item.item.pull | default(omit) }}"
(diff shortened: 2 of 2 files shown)


molecule/provisioner/ansible/playbooks/docker/create.yml
```diff
@@ -39,7 +39,7 @@
         path: "{{ molecule_ephemeral_directory }}"
         name: "molecule_local/{{ item.item.image }}"
         docker_host: "{{ item.item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
-        dockerfile: "{{ item.item.dockerfile | default(item.invocation.module_args.dest) }}"
```
ssbarnea: .dockerfile is the template not the generated file, this was the mistake. this is mainlt reverting it to old value.

decentral1se:
For testing, I would take a look at https://github.com/ansible/molecule/tree/master/test/scenarios/overrride_driver/molecule/default, which you copy and specify a custom `dockerfile` in the `molecule.yml`. Then, write a test in https://github.com/ansible/molecule/blob/master/test/functional/docker/test_scenarios.py to drive it and somehow devise a way to check it gets overrided (some `assert ... in` from the bake run might be all that is needed)?

test/scenarios/overrride_driver/molecule/default/playbook.yml
```diff
@@ -5,3 +5,11 @@
     - name: Dummy task
       command: /bin/true
       changed_when: false
+    - name: check presence of from-custom-dockerfile
```
decentral1se: This looks like a good strategy then! However, the `override_driver` is now testing two things: custom dockerfile and overriding the driver. Can we separate these tests out into separate scenarios? Let's ignore the fact that our functional tests take 10000000 years to run and assume we'll get that improved soon :) You'll also need to add a a test in https://github.com/ansible/molecule/blob/master/test/functional/docker/test_scenarios.py to drive it as well.

ssbarnea: @decentral1se I am really concerned about the time needed to run the test and if we do not combine them we will be doomed not to be able to run the tests.

I am more pragmatic here, i do care to test multiple execution paths and provide a good code (branch) coverage. I do not care if a specific test is testing more than one aspect, if it fails it should be easy to stop what went wrong.

On the other hand "override" seems like a good candidate, we are also overriding the default template ;)

ssbarnea: @decentral1se I tried to follow your recomandation and add a new scenarion and this broke the doors of hell... take a look at how the tests are written on https://github.com/ansible/molecule/blob/0627ca213048a5bd5ea671f6b14eee2a7624f867/test/functional/test_command.py#L394

I have to modify half of them to allow a new scenario to be added because they check for the list of expected scenarios.

I have the impression that that stuff with test isolation makes no sense, we cannot really add new scenarios without affecting almost everythign else.

Isolation is good for unittesting but for functional, i doubt is doable.

gundalow:
2019-03-13 IRC Meeting. We need someone to test this before we can merge

ssbarnea:
@gundalow @decentral1se some logs of running with this change https://seashells.io/v/JrsznDjz --- code used https://github.com/ssbarnea/molex/blob/master/molecule/default/molecule.yml#L25

PR #1810, opened 2019-03-07
Update Dockerfile.j2

MadZimbo:
Added rsync package 
Without rsync the Ansible synchronize module fails with 
""msg": "protocol version mismatch -- is your shell clean?\n(see the rsync man page for an explanation)\nrsync error: protocol incompatibility (code 2)"
when testing with docker 
Would be nice if this could be integrated

#### PR Type
- Feature Pull Request

### DIFF
--- a/test/resources/playbooks/docker/Dockerfile.j2
+++ b/test/resources/playbooks/docker/Dockerfile.j2
@@ -9,9 +9,9 @@ FROM {{ item.image }}
 ARG testarg=fail
 ENV envarg=$testarg
 
-RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get install -y python sudo bash ca-certificates && apt-get clean; \
-    elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install python sudo python-devel python2-dnf bash && dnf clean all; \
-    elif [ $(command -v yum) ]; then yum makecache fast && yum install -y python sudo yum-plugin-ovl bash && sed -i 's/plugins=0/plugins=1/g' /etc/yum.conf && yum clean all; \
-    elif [ $(command -v zypper) ]; then zypper refresh && zypper install -y python sudo bash python-xml && zypper clean -a; \
-    elif [ $(command -v apk) ]; then apk update && apk add --no-cache python sudo bash ca-certificates; \
-    elif [ $(command -v xbps-install) ]; then xbps-install -Syu && xbps-install -y python sudo bash ca-certificates && xbps-remove -O; fi
+RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get install -y python sudo bash ca-certificates rsync && apt-get clean; \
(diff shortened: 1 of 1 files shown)


ssbarnea:
While this is valid it opens a multitude of questions: what needs to be in the default template and what needs to be out?

If we include rsync, we make the ansible module work but at the same time we would fail to validate a role that is supposed to install rsync if needed.

What is a an ansible-enabled image? Sadly the answer is not a boolean: you are likely going to have some modules working and some not working. 

For example shade is needed in order to use openstack modules, do we include them or not? We need to decide what goes in and what goes out. What fixed a problem for someone may break for someone else.

decentral1se:
Hi @MadZimbo, thanks for the pull request.

You've stumbled on a rather touch area :sweat_smile: and to continue from @ssbarnea: I don't think it's for Molecule core to decide what goes in or out but to allow to configure. We just need to provide the minimal dependencies (I would say we are already going beyond that ...). 

The solution for this problem so far has been to add it to your generated `molecule/default/Dockerfile.j2`.  This isn't ideal but this is how I would solve this issue for now. What do you think @MadZimbo?

To note, there is a related discussion happening over at https://github.com/ansible/molecule/issues/1642. There is some motivation to see if we can get these images built (or not even build them) in a more "Ansible" way. We might imagine a `extra_packages: [rsync]` configuration in our `molecule.yml` then?

MarkusTeufelberger:
If you need to use a module (e.g. synchronize) with external dependencies, make sure that your role installs these dependencies or modify the `Dockerfile.j2` file or the default `playbook.yml` locally.

It is really not feasible to install all dependencies of all modules by default.

ssbarnea:
As noted by others this is sensitive issue without any clear solution. Still, in order to improve the user experience we could consider adding some optional parameters for provisioning additional requirements. 

I already have a custom dockerfile that adds extra packages so we may be able to build a more generic solution for enabling additional ansible modules on the builded images.

Default should be minimal in order to enable ansible to connect to the machine, but optionally we can boostrap additional ansible dependencies (optionals).

MadZimbo:
> Hi @MadZimbo, thanks for the pull request.
> 
> You've stumbled on a rather touch area 😅 and to continue from @ssbarnea: I don't think it's for Molecule core to decide what goes in or out but to allow to configure. We just need to provide the minimal dependencies (I would say we are already going beyond that ...).
> 
> The solution for this problem so far has been to add it to your generated `molecule/default/Dockerfile.j2`. This isn't ideal but this is how I would solve this issue for now. What do you think @MadZimbo?
> 
> To note, there is a related discussion happening over at #1642. There is some motivation to see if we can get these images built (or not even build them) in a more "Ansible" way. We might imagine a `extra_packages: [rsync]` configuration in our `molecule.yml` then?

Hi @decentral1se, thanks for the reply. I see your point and based on the other comments it makes quite sense. Thanks for your time!

decentral1se:
:+1:

Issue #1791, opened 2019-02-28
Add hetzner driver


decentral1se:
# Issue Type
- Feature request

See implementation in https://github.com/ansible/ansible/pull/53062.

We should implement a driver here in Molecule to support Hetzner driver automation.

gundalow:
Given we have a PR https://github.com/ansible/molecule/pull/1811 can we close this?

decentral1se:
Fine by me!

PR #1811, opened 2019-03-07
Add Hetzner Cloud driver

decentral1se:
Closes https://github.com/ansible/molecule/issues/1791.


gundalow:
@LKaemmerling FYI

LKaemmerling:
Could you please name it something like `hcloud` or `Hetzner Cloud`? The name `Hetzner` would mean everything from Hetzner (included Dedicated Servers and so on) We have the specific `Hetzner Cloud` Modules, which only contains functionality for the Hetzner Cloud :)

decentral1se:
"Hetzner Cloud" it is :+1:

vutkin:
@gundalow why we need to inject so many drivers?
Many clouds->many drivers->who will be support them?
Would it be better to use delegated driver with different create/destroy files?
What do you think?

decentral1se:
@vutkin if you have any issue with project direction, please raise at the working group meeting.

decentral1se:
FYI raised https://github.com/ansible/ansible/pull/55713 while working on this.

decentral1se:
OK, this is ready for review.

I've done quite a bit of manual testing for this and am quite happy with it.

All reviews will be much appreciated.

References:
  * https://docs.ansible.com/ansible/devel/modules/hcloud_server_module.html

/cc @LKaemmerling

decentral1se:
CI failures are unrelated. Please see https://github.com/ansible/molecule/issues/2006.

It all runs green on my local machine with the work-around listed in #2006.

decentral1se:
How to manual test this:

```
$ git clone -b add-hetzner-driver https://github.com/ansible/molecule
$ cd molecule

# Then a workaround for the issue in #2006
$ rm pyproject.toml + set usedevelop=False in [testenv] in tox.ini

$ pip install -e .'[hetznercloud]'
$ molecule init -r hcloudtest -d hetznercloud
$ cd hcloudtest
$ export HCLOUD_TOKEN=<MYTOKEN>

$ molecule test
```

webknjaz:
https://github.com/ansible/molecule/pull/2009 fixes tox.

decentral1se:
OK, CI is running on green now :rocket: :rocket: :rocket:

LKaemmerling:
LGTM. How do you get the API token for the tests?

decentral1se:
Thanks for the review. I used my personal account to test this with an API token. FYI: we're interested in getting something organised (cc @gundalow) to arrange continuous integration testing with this driver against real hetzner cloud machines in the future.

LKaemmerling:
For this we have developed a little system for creating/deleting temporary API tokens. You need to get a fresh token for each run and afterward delete the token.  Basically you will only need to perform two curl requests:
https://github.com/terraform-providers/terraform-provider-hcloud/blob/master/scripts/get-token.sh (for getting a token)
https://github.com/terraform-providers/terraform-provider-hcloud/blob/master/scripts/delete-token.sh (for deleting a token)

@decentral1se could you please write me a mail to lukas.kaemmerling(at)hetzner-cloud.de with the following information:
How many tests will run concurrently? (Like how many travis CI runs may be run concurrently?)
How long does one run will take maximal? 
I could send you the token then.

webknjaz:
@LKaemmerling currently each Travis CI build has 45 jobs which can potentially interact with Hetzner but they currently only test docker provider. It can take 20-45 min for each job to finish, adding extra provider tests will increase this time and we may need to split tests further meaning more jobs.
I believe that Travis CI itself currently sets overall jobs limit to 10.
So I'd assume we'll probably need 10–20 concurrent temporary tokens.

webknjaz:
@decentral1se I believe you need to also add extras to these lines: https://github.com/ansible/molecule/blob/766dc35/Dockerfile#L16 + https://github.com/ansible/molecule/blob/766dc35/Dockerfile#L54

decentral1se:
Thanks @webknjaz.

Sent you a mail @LKaemmerling! Let's get this moving ASAP :)

decentral1se:
Rebased and now seeing a CI failure which appears to be unrelated. 

I've added a ticket to track that in https://github.com/ansible/molecule/issues/2022.

Related: Made a short topic annoucement about the CI developments on this PR over at https://meetbot.fedoraproject.org/sresults/?group_id=ansible_molecule_working_group&type=team for anyone interested.

.travis.yml
```diff
@@ -611,3 +612,5 @@ before_script:
   - gem install inspec rubocop
 script:
   - tox
+after_script:
```
decentral1se: AFAICT, this will always run regardless of failure. See https://docs.travis-ci.com/user/job-lifecycle/.

bin/get-hetzner-cloud-token
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
decentral1se: Scripts adapted from ones mentioned in https://github.com/ansible/molecule/pull/1811#issuecomment-487449174.

bin/get-hetzner-cloud-token
```diff
@@ -0,0 +1,18 @@
...
+
+set -eux
+
+curl \
+  -A 'travis-ansible-provider' \
```
decentral1se: @LKaemmerling, is this OK? I switched it from "travis-terraform-provider" ...

test/scenarios/driver/hetznercloud/molecule/default/molecule.yml
```diff
@@ -0,0 +1,37 @@
...
+lint:
+  name: yamllint
+  options:
+    config-file: ../../../resources/.yamllint
+
+platforms:
+  - name: instance
+    server_type: cx11
+    ssh_keys:
+      # FIXME(decentral1se): change when we get a CI account
```
decentral1se: @LKaemmerling, molecule needs to connect to these machines but we're going to be creating them with a TTS/HCLOUD_TOKEN flow and it is not clear to me how to create SSH keys programmatically that are not first pre-authorized with the Hetzner account.

Currently, the driver implementation assumes that the SSH key referenced here is loaded with the ssh-agent. So the CI has to have control of the private part of the SSH key. That could be available from the Travis CI configuration but we'd have to agree on a key ... unsure of how to proceed ...

bin/get-hetzner-cloud-token
decentral1se: @webknjaz, I am sure you will have something to say regarding these scripts, so I ping you now.

bin/get-hetzner-cloud-token
LKaemmerling: Yep is okay :)

test/scenarios/driver/hetznercloud/molecule/default/molecule.yml
LKaemmerling: You can Send me a ssh key (having a fix key) Or you could add it via the hcloud_ssh_key ansible Module

.travis.yml
webknjaz: Yes.

bin/get-hetzner-cloud-token
webknjaz: I'm normally putting a space for readability after `#!` and use another invocation style. But no strong reason to changing this.
```suggestion
#! /usr/bin/env bash
```

bin/get-hetzner-cloud-token
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -eux
```
webknjaz: This is a stricter mode:
```suggestion
set -euxo pipefail
```

bin/get-hetzner-cloud-token
webknjaz: In general, I'd prefer this written in Python. And also it looks like there's already a `contrib/` folder for this so no reason to create another one.

bin/get-hetzner-cloud-token
```diff
@@ -0,0 +1,18 @@
...
+
+curl \
+  -A 'travis-ansible-provider' \
+  --header 'Authorization: Bearer '"$HETZNER_CLOUD_TTS_TOKEN"'' \
```
webknjaz: You don't need this many quotes
```suggestion
  --header "Authorization: Bearer ${HETZNER_CLOUD_TTS_TOKEN}" \
```

.travis.yml
```diff
@@ -603,6 +603,7 @@ addons:
 before_install:
   - export TOXENV=$(echo $TOXENV_TMPL | envsubst)
   - pip install tox-venv tox-tags
+  - export HCLOUD_TOKEN=$(./bin/get-hetzner-cloud-token)
```
webknjaz: You'd also need to put `${HETZNER_CLOUD_TTS_TOKEN}` into travis config

.travis.yml
```diff
@@ -611,3 +612,5 @@ before_script:
   - gem install inspec rubocop
 script:
   - tox
+after_script:
+  - ./bin/rm-hetzner-cloud-token $HCLOUD_TOKEN
```
webknjaz: Looking at this, we might want to use `commands_pre` and `commands_post` in `tox.ini` for this to keep the only source for truth.

molecule/cookiecutter/scenario/driver/hetznercloud/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/prepare.yml
```diff
@@ -0,0 +1,11 @@
+---
+{% raw -%}
+- name: Prepare
+  hosts: all
+  gather_facts: false
+  tasks:
+    - name: Install python for Ansible
+      raw: test -e /usr/bin/python || (apt -y update && apt install -y python-minimal python-zipstream)
```
webknjaz: Maybe we could multiline this?
```suggestion
      raw: >-
        test -e /usr/bin/python ||
        (apt -y update && apt install -y python-minimal python-zipstream)
```

molecule/driver/hetznercloud.py
```diff
@@ -0,0 +1,190 @@
+#  Copyright (c) 2018-2019 Red Hat, Inc.
```
webknjaz: Usually, copyright is a year of content creation. Ans this module has been created in 2019.
```suggestion
#  Copyright (c) 2019 Red Hat, Inc.
```

molecule/driver/hetznercloud.py
```diff
@@ -0,0 +1,190 @@
...
+
+    Provide the files Molecule will preserve upon each subcommand execution.
+
+    .. code-block:: yaml
+
+        driver:
+          name: hetznercloud
+          safe_files:
+            - foo
+    """  # noqa
```
webknjaz: Please only ignore one specific warning, not everything wholesale.

webknjaz:
```suggestion
    """
```

molecule/driver/hetznercloud.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def __init__(self, config):
+        super(HetznerCloud, self).__init__(config)
+        self._name = 'hetznercloud'
+
+    @property
+    def name(self):
+        return self._name
+
+    @name.setter
```
webknjaz: Why do you need a setter here? Is anything outside supposed to assign thing to the public attribute?

molecule/driver/hetznercloud.py
```diff
@@ -0,0 +1,190 @@
...
+
+            return {
+                'ansible_user': d['user'],
+                'ansible_host': d['address'],
+                'ansible_port': d['port'],
+                'connection': 'ssh',
+                'ansible_ssh_common_args':
+                ' '.join(self.ssh_connection_options),
+            }
+        except StopIteration:
```
webknjaz: Instead of this, use a fallback in next.

molecule/driver/hetznercloud.py
```diff
@@ -0,0 +1,190 @@
...
+            # Instance has yet to be provisioned , therefore the
+            # instance_config is not on disk.
+            return {}
+
+    def _get_instance_config(self, instance_name):
+        instance_config_dict = util.safe_load_file(
+            self._config.driver.instance_config)
+
+        return next(item for item in instance_config_dict
+                    if item['instance'] == instance_name)
```
webknjaz: You probably want to have a default value to prevent `StopIteration`:
```suggestion
                    if item['instance'] == instance_name, ())
```

test/functional/conftest.py
```diff
@@ -387,6 +388,15 @@ def supports_gce():
     return _env_vars_exposed(env_vars) and HAS_GOOGLE_AUTH
 
 
+@pytest.helpers.register
+def supports_hetznercloud():
+    pytest.importorskip('hcloud')  # Ansible provides no import
```
webknjaz: I don't understand this comment.
```suggestion
    pytest.importorskip('hcloud')
```

bin/get-hetzner-cloud-token
decentral1se: Thanks!

test/scenarios/driver/hetznercloud/molecule/default/molecule.yml
decentral1se: > hcloud_ssh_key ansible Module

Right, looks like this is the solution for it! Thanks for the guidance :+1:

.travis.yml
decentral1se: Already done but it did not seem to be picking it up ... :/

.travis.yml
decentral1se: Right, that seems like a much better approach ... can hopefully avoid the scripts, maybe ...

molecule/cookiecutter/scenario/driver/hetznercloud/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/prepare.yml
decentral1se: Unrelated since this is a copy/pasta and all other files do that. I'm not refactoring here.

test/functional/conftest.py
decentral1se: All other functions try to import driver dependencies from Ansible, this does not. This comment will stop others wondering why we did not follow the same approach and trying to go luck at Ansible modules ...

decentral1se: Suggest an improvement but let's not delete this context.

molecule/driver/hetznercloud.py
decentral1se: Unrelated since this is a copy/pasta and all other files do that. I'm not refactoring here.

molecule/driver/hetznercloud.py
decentral1se: Unrelated since this is a copy/pasta and all other files do that. I'm not refactoring here.

test/functional/conftest.py
webknjaz: When I read this I don't see any other approaches. Plz put the explanation there

decentral1se:
OK. Ready for another review round. Status is:

  * Hetzner Cloud modules only available on Ansible 2.8+
  * Ansible 2.8+ broken in Molecule due to testinfra issue

Therefore, I can't run the functional tests right now on the CI because all Ansible 2.8 tests are failing. However, all the rest is in place with this PR (see new token manager module which avoids needs for bash scripts). Hence, I propose to submit this for final reviews, let's merge the work done and then I'll be moving onto fixing the Ansible 2.8 / testinfra issue so I can get the CI in place for this driver.

It ain't clean but work with me.

contrib/tokenmgmr
```diff
@@ -0,0 +1,146 @@
...
+@main.command()
+@click.option(
+    '--driver-name',
+    '-d',
+    type=click.Choice(['hetznercloud']),
+    help='Name of driver to get token for.')
+def get(driver_name):
+    """Retrieve a temporary CI token."""
+    if driver_name == 'hetznercloud':
+        return print(_get_hetzner_cloud_token())
```
webknjaz: This will return `None`. If that's what you expect, the return statement is completely unnecessary.
```suggestion
        print(_get_hetzner_cloud_token())
```
Also, I'm pretty sure that you should use `click.echo` instead.

contrib/tokenmgmr
```diff
@@ -0,0 +1,146 @@
...
+            'Unable to remove a Hetzner Cloud API token '
+            'via a curl invocation. Saw the following '
+            'error:\n{}'
+        ).format(str(exception)))
+
+
+@click.group()
+def main():
+    """Retrieve tokens for interacting with driver providers."""
+    pass
```
webknjaz: The docstring itself is a statement, no need to have explicit `pass`.
```suggestion
```

contrib/tokenmgmr
```diff
@@ -0,0 +1,146 @@
...
+    try:
+        token = os.getenv('HCLOUD_TOKEN')
+        return sh.curl(
+            'https://tt-service.hetzner.cloud/token?token={}'.format(token),
+            A='travis-ansible-provider',
+            H='Authorization: Bearer ${HETZNER_CLOUD_TTS_TOKEN}',
+            X='DELETE',
+        )
+    except sh.ErrorReturnCode as exception:
+        log.error((
```
webknjaz: maybe [`log.exception('...')`](https://docs.python.org/3/library/logging.html#logging.Logger.exception)?
or `log.error(..., exc_info=exception)`?

contrib/tokenmgmr
```diff
@@ -0,0 +1,146 @@
+#! /usr/bin/env python
```
webknjaz: Please use `.py` file extension. Btw, remember to commit the executable bit.

tox.ini
```diff
@@ -56,10 +57,15 @@ commands_pre =
     # TODO: move this to more generic deps once stable 2.8.0 is out
     ansible28: pip install --pre 'ansible >= 2.8b1,<2.9'
     find {toxinidir} -type f -not -path '{toxinidir}/.tox/*' -path '*/__pycache__/*' -name '*.py[c|o]' -delete
+    functional: sh -c 'export HCLOUD_TOKEN=$(./contrib/tokenmgmr get -d hetznercloud)'
```
webknjaz: I'm pretty sure this won't work: you invoke sh subshell which will probably lose env vars in its internal session.
It looks like this should probably be done on the CI level.

decentral1se:
@LKaemmerling, The TTS/API token handshake is in place now.

However, I'm seeing a number of failures relating to failed SSH connections after the instances are created and the role is being run. For example, see the logs in https://travis-ci.com/ansible/molecule/jobs/203120479#L559-L598. The implementation polls at port 22 for the SSH service and when it is up it then connects to run the roles against the new machine. When it tries to log in, the connection times out. That feels like a firewall issue to me but I am unsure of how to debug this since those machines are not accessible to me.

A full job did pass (!) which did make connections to newly created machines and execute roles on them. See https://travis-ci.com/ansible/molecule/jobs/203120480. Nearly there it seems ...

The full latest build logs are at https://travis-ci.com/ansible/molecule/builds/113216198.

Also to note: a lot of tokens and boxes are not left not destroyed due to all this debugging and false positives etc. with running the CI right now and possibly going forward. I hope things are being cleaned up afterwards!

LKaemmerling:
On the latest travis run i can not see any issue directly related to the cloud (https://travis-ci.com/ansible/molecule/jobs/203141639). 

All resources (servers, volumes, tokens and so on) will be deleted from our system after a hour.

decentral1se:
Aha, this is blocked by https://github.com/ansible/molecule/issues/2070 and https://github.com/ansible/molecule/issues/2069.

decentral1se:
Oh, now it appears to be blocked by  https://github.com/ansible/molecule/issues/2077 as well :bomb:

decentral1se:
Current status is then: there are endless issues with running the full functional test suite on the CI. SSH connection issues, machines not responding, docker networks randomly failing, incorrect passwords (really unsure how this is happening). And all the while, all the tests pass locally. I think I will abandon this for now.

LKaemmerling:
Really sad to hear this. @decentral1se is there something we could do to help you finish this? We plan to add more modules to ansible in the near future and also use molecule internal, so we would like to push this forward :)

decentral1se:
@LKaemmerling I'm glad Hetzner intend to use Molecule. I don't use Hetzner Cloud services as I did when I first started this work. So, I have no reason to continue to work on it since I won't use it for the forseeable future. Probably some paid staff from Hetzner can pick up this work and fight the CI issues.

decentral1se:
Also, Hetzner could just pay me.

LKaemmerling:
@decentral1se would it be possible that you rerun the tests? I guess I have found something. We have set the Temporary Token Service to a max token lifetime of 60min. After this 60min all resources created will be destroyed. This could be a reason because I saw the last run takes ~1h 2 min. (https://travis-ci.com/ansible/molecule/jobs/215355197)

ssbarnea:
I propose making a fork and enabling travis with hertzner credentials and running on a daily cron, so we can see it works reliably enough.

LKaemmerling:
I guess i have found the problem.

@decentral1se you said that the tests are working locally. I can confirm this. The tests on the CI fail because of the parallelism from travis.

You have something like the instance name (`instance`, `instance-1`  and `instance-2`) hardcoded.  This will cause problems when running the same tests multiple times in parallel. 

Because the modules work primarily with the instance name, you will have the following behavior when running the same tests in parallel. Hopefully, I can show it to you with the following sample:

Travis CI:
Job 1 (functional tests on Python 3.7): Started at 10:00
 ---> Will create server `instance`
 ---> Will work with this server
Job 2 (functional tests on Python 2.7): Started at 10:05
--> Will create server `instance` -->**Attention**: It will get the Server from Job1! (Because the name need to be unique and the ansible modules follow the idempotency principle)
--> Will work with this server
Job 1: At 10:20
 --> All tests are done, will remove the server
--> Finished
Job 2: ERR! The server was removed, no connection (or another error)

As far as I see it, it is not a problem with our Hetzner Cloud, it is rather a problem with the CI and the errors only occur during these tests because we are the first cloud provider where the tests are done in the CI.

decentral1se:
There's our `--parallel` functionality and then there is also the Travis CI job parallelization. You've probably nailed the issue here. The solution might then be: isolate all hetzner functional tests and put them in a single job that only runs once (no parallel stuff at all on any layer). A fork is no solution IMHO. I'm not starting this unless @ssbarnea is OK with the proposed direction.

ssbarnea:
@decentral1se what kind of changes are you expecting? I am already scared about the current complexity related to sharding.

The reality is that we need to make resources unique for each travis job. Mainly the hertz driver should run even when running with "tox -p auto".

decentral1se:
Perhaps we follow your direction, and a single cron job that runs nightly. We do this "for now" and can at least see what is failing. Then the infrastructure will be in place when more CI resources are available to really lock it down. We keep the API keys and users for the driver to rely on real quality Q&A for the driver. Later, we get it running rock solid. So, TLDR; an additional build in the cron job to run them all in serial. I will use a pytest marker to group them. At the end of the day, all the other drivers don't even run, so we'll be just as good as them.

PS. Note to safe: some fix attempts over at https://github.com/LKaemmerling/molecule/commits/add-hetzner-driver to watch out for.

decentral1se:
OK, I've added fixes from @LKaemmerling and enabled the tests on only a single job (due to lack of CI resources. All the tests pass, see https://travis-ci.com/ansible/molecule/jobs/217346476). The CI is very unreliable right now, so I re-running the other jobs which failed for unrelated issues (that's off-topic for this PR). Can I get final reviews and then let's merge this.

decentral1se:
CI failing due to https://github.com/ansible/molecule/issues/2177 (unrelated).

decentral1se:
OK, final approach is then to https://github.com/ansible/molecule/pull/2191 and get a clean CI build here.

decentral1se:
@ssbarnea let's merge this, it's green :fire: :fire: :fire:

Issue #1812, opened 2019-03-07
Docs should not state that we support Ansible 2.4 which is EOL


webknjaz:
# Issue Type

- Bug report

# Molecule and Ansible details

N/A

# Desired Behavior

Docs should say we support Ansible>=2.5

# Actual Behaviour

Docs say we support Ansible>=2.4

decentral1se:
Improvement to be made: https://github.com/ansible/molecule/pull/1814#discussion_r264278773.

decentral1se:
Fixes for this have lead to https://github.com/ansible/molecule/issues/1822 being raised.

PR #1814, opened 2019-03-10
Do not advertise support for EOL Ansible version

decentral1se:
#### PR Type
- Docs Pull Request

Closes https://github.com/ansible/molecule/issues/1812.

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -14,9 +14,9 @@ Requirements
 Depending on the driver chosen, you may need to install additional OS packages.
 See ``INSTALL.rst``, which is created when initializing a new scenario.
 
-* :std:doc:`Ansible <ansible:index>` >= 2.4
+* :std:doc:`Ansible <ansible:index>` >= 2.5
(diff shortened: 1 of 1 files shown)


doc/source/installation.rst
```diff
@@ -14,9 +14,9 @@ Requirements
 Depending on the driver chosen, you may need to install additional OS packages.
 See ``INSTALL.rst``, which is created when initializing a new scenario.
 
-* :std:doc:`Ansible <ansible:index>` >= 2.4
+* :std:doc:`Ansible <ansible:index>` >= 2.5
 * Python 2.7
```
webknjaz: @decentral1se it looks confusing: in which cases Python 2.7 would work that Python 3.6 wouldn't? It looks like this could be collapsed to just one line saying smth like `Python ==2.7 or >= 3.6` and `Ansible >=2.5`.

decentral1se: Yep, right. Submitted https://github.com/ansible/molecule/pull/1820.

Issue #1813, opened 2019-03-08
REGRESSION 2.19 -> 2.20 Docker image fails with "Failed to import docker or docker-py - No module named docker"


alainchiasson:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version

docker run --rm -it -v "$(pwd)":/root -v /var/run/docker.sock:/var/run/docker.sock -w /root quay.io/ansible/molecule:latest sudo molecule --version
molecule, version 2.19.0

docker run --rm -it -v "$(pwd)":/root -v /var/run/docker.sock:/var/run/docker.sock -w /root quay.io/ansible/molecule:latest sudo ansible --version
ansible 2.7.8
  config file = None
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.15 (default, Jan 24 2019, 16:32:39) [GCC 8.2.0]


```

Molecule installation method (one of):

- NONE: using docker image 

Ansible installation method (one of):

- NONE: using docker image 

docker images | grep molecule

```
quay.io/ansible/molecule   latest              76c463a83726        3 days ago          516MB
molecule_local/centos      7                   d5efb4c1540b        3 days ago          226MB
quay.io/ansible/molecule   2.19                0a943e169bf0        4 months ago        580MB
````

Detail any linters or test runners used:

# Desired Behavior

On repo https://github.com/alainchiasson/role-test-duplicates.git

The following works : 

```
docker run --rm -it -v "$(pwd)":/root -v /var/run/docker.sock:/var/run/docker.sock -w /root quay.io/ansible/molecule:2.19 sudo molecule test
```

Please give some details of the feature being requested or what
should happen if providing a bug report.

# Actual Behaviour

On repo https://github.com/alainchiasson/role-test-duplicates.git

The following fails:
```
docker run --rm -it -v "$(pwd)":/root -v /var/run/docker.sock:/var/run/docker.sock -w /root quay.io/ansible/molecule:latest sudo molecule test
```
Running from a macbook pro. - debug of prepare. 

```
MBP-de-Alain:role-test-duplicates alain$ docker run --rm -it -v "$(pwd)":/root -v /var/run/docker.sock:/var/run/docker.sock -w /root quay.io/ansible/molecule:latest sudo molecule --debug create
[...]

PR #1815, opened 2019-03-10
Re-add openssh-client to docker image

decentral1se:
#### PR Type
- Bugfix Pull Request

Follows https://github.com/ansible/molecule/issues/1813#issuecomment-471035576 where I took a look at https://github.com/ansible/molecule/commit/bce7bfebcaa66c9417f653778f7e4bf03a554af5 to see if we're dropping any packages we shouldn't be. I think this is the only one! Original report for this change was in https://github.com/ansible/molecule/pull/1604.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,8 +29,10 @@ LABEL maintainer "Ansible <info@ansible.com>"
 
 ENV PACKAGES="\
     docker \
+    openssh-client \
     ruby \
     "
+
 ENV BUILD_DEPS="\
     gcc \
     libc-dev \


decentral1se:
Thanks for reviewing @webknjaz, good catch. I've fixed that up and will merge based on https://github.com/ansible/molecule/issues/1813#issuecomment-471577888.

Dockerfile
```diff
@@ -29,8 +29,10 @@ LABEL maintainer "Ansible <info@ansible.com>"
 
 ENV PACKAGES="\
     docker \
+    openssh-client \
     ruby \
     "
+
```
webknjaz: FTR I intentionally didn't add this line, because both vars are packages installed with apk so I wanted to group them.

PR #1820, opened 2019-03-11
Move Ansible/Python requirements into one line

decentral1se:
Addresses https://github.com/ansible/molecule/pull/1814#discussion_r264278773.

Closes https://github.com/ansible/molecule/issues/1812.

#### PR Type
- Docs Pull Request

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -14,9 +14,7 @@ Requirements
 Depending on the driver chosen, you may need to install additional OS packages.
 See ``INSTALL.rst``, which is created when initializing a new scenario.
 
-* :std:doc:`Ansible <ansible:index>` >= 2.5
-* Python 2.7
-* Python >= 3.6 with Ansible >= 2.5
+* Python == 2.7 or >= 3.6 and :std:doc:`Ansible <ansible:index>` >= 2.5
 
 CentOS 7
 --------


doc/source/installation.rst
```diff
@@ -14,9 +14,7 @@ Requirements
 Depending on the driver chosen, you may need to install additional OS packages.
 See ``INSTALL.rst``, which is created when initializing a new scenario.
 
-* :std:doc:`Ansible <ansible:index>` >= 2.5
-* Python 2.7
-* Python >= 3.6 with Ansible >= 2.5
+* Python == 2.7 or >= 3.6 and :std:doc:`Ansible <ansible:index>` >= 2.5
```
ssbarnea: As someone that knows well the defaupt python3 on centos-7 is python 3.5, I find these requirements bit to worry because they take Python 3.5 out of the list of support matrix.

Shortly based on https://github.com/ansible/ansible/blob/stable-2.5/setup.py#L243 it think the change is wrong.

decentral1se: You know what @ssbarnea, I think you're right.

Why aren't we supporting 2.6 and 3.5? I know people have said "we have to follow Ansible" as a reason to *not* do this ... I can't find a link to the conversation. Anyway, I'll see if this is a hellish task.

I'll open up a new issue :+1:

Issue #1724, opened 2019-02-04
Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!


daniel-wtd:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version

ansible 2.7.6
  config file = None
  configured module search path = ['/home/ds/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/ds/.venv-molecule-p3/lib/python3.7/site-packages/ansible
  executable location = /home/ds/.venv-molecule-p3/bin/ansible
  python version = 3.7.2 (default, Jan 16 2019, 19:49:22) [GCC 8.2.1 20181215 (Red Hat 8.2.1-6)]
molecule, version 2.19.0
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip
- OS package

Detail any linters or test runners used:

# Desired Behavior

`molecule test` should run locally on Fedora 29 in virtualenv

# Actual Behaviour

`molecule test` brakes due to missing libselinux-python at Step `Create Dockerfiles from image names`

Creating an env with site-packages is not able to perform due to pyyaml version.
Creating an env without site packages is not able to perform due to missing selinux python bindings.
Copying the selinux packages in the venv does not help.

Output of `molecule test` without site packages.

```
    TASK [Create Dockerfiles from image names] *************************************
    task path: /home/ds/.venv-molecule-p3/lib/python3.7/site-packages/molecule/provisioner/ansible/playbooks/docker/create.yml:21
    <127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: ds
    <127.0.0.1> EXEC /bin/sh -c 'echo ~ds && sleep 0'
    <127.0.0.1> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /home/ds/.ansible/tmp/ansible-tmp-1549316301.9148593-163859707096071 `" && echo ansible-tmp-1549316301.9148593-163859707096071="` echo /home/ds/.ansible/tmp/ansible-tmp-1549316301.9148593-163859707096071 `" ) && sleep 0'
    Using module file /home/ds/.venv-molecule-p3/lib/python3.7/site-packages/ansible/modules/files/stat.py
[...]

PR #1823, opened 2019-03-11
Avoid python libselinux bindings being hidden by tox

ssbarnea:
Adds libselinux python extensions to the default Dockerfile template
used by molecule in order to avoid failure to build it on selinux
systems.

Fixes: https://github.com/ansible/molecule/issues/1724

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -40,7 +40,7 @@ Pip
 :std:doc:`pip <pip:usage>` is the only supported installation method.
 
 Keep in mind that on selinux supporting systems, if you install into a virtual
-environment, you may face https://github.com/ansible/ansible/issues/34340 even
+environment, you may face :gh:`issue <ansible/ansible/issues/34340>` even
 if selinux is not enabled or is configured to be permissive.
 
 It is your reponsability to assure that soft dependencies of Ansible are
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -39,6 +39,13 @@ Pip
 
 :std:doc:`pip <pip:usage>` is the only supported installation method.
 
+Keep in mind that on selinux supporting systems, if you install into a virtual
+environment, you may face https://github.com/ansible/ansible/issues/34340 even
+if selinux is not enabled or is configured to be permissive.
+
+It is your reponsability to assure that soft dependencies of Ansible are
+available on your controller or host machines.
+
 Requirements
 ------------
 
--- a/tox.ini
+++ b/tox.ini
@@ -41,6 +41,11 @@ commands =
     lint: yamllint -s test/ molecule/
 whitelist_externals =
     find
+# Enabling sitepackages is needed in order to avoid encountering exceptions
+# caused by missing selinux python bindinds in ansible modules like template.
+# Selinux python bindings are binary and they cannot be installed using pip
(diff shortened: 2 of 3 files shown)


decentral1se:
> Adds libselinux python extensions to the default Dockerfile template

I don't see any addition to the `Dockerfile` template!?

tox.ini
```diff
@@ -41,6 +41,10 @@ commands =
     lint: yamllint -s test/ molecule/
 whitelist_externals =
     find
+# sitepackages=true <- template module <- libselinux bindings which cannot be
+# be installed with pip. Also this lowers virtualenvs footprints and install
+# time
+sitepackages=true
```
webknjaz: I blocked this change earlier if you remember.

Can we just recommend using `tox --sitepackages` to folks who want this?

ssbarnea: @webknjaz adding new guidelines and recomandation would play against making the experience slimless, it would only make it worse. Even those (few) reading the docs would forget to add the extra parameters and run into troubles.

If we are not careful on to easy the first time experience, we may endup flooded by recurring newbe questions, and tickets.

webknjaz: wrap it with whitespaces to match the common style 
```suggestion
sitepackages = true
```

tox.ini
```diff
@@ -41,6 +41,10 @@ commands =
     lint: yamllint -s test/ molecule/
 whitelist_externals =
     find
+# sitepackages=true <- template module <- libselinux bindings which cannot be
```
webknjaz: We need to have way more comprehensive explanation then

webknjaz:
cc https://github.com/ansible/molecule/issues/1724#issuecomment-472173834

gundalow:
2019-03-13 IRC Meeting. Agreed this isn't 2.20 blocking

fabianvf:
What information do you feel is missing? The current comment seems quite clear to me.

tox.ini
```diff
@@ -41,6 +41,11 @@ commands =
...
```
webknjaz: Could you please style it in a form
```ini
# This is needed because:
# 1) ...
# 2) ...
# 3) ...
# 
# Side-effects are:
# 1) ...
# 2) ...
# 3) ...
```

webknjaz:
@fabianvf things tend to seem "obvious" once you already know a lot of details about them. I just want it to become more obvious to a person who has no context, sees it for the first time and tries to acquire such context. "Look at these 100+ comments" is not very helpful here. I think, this kind of comments should follow the same rules/considerations as good commit messages and docstings.

fabianvf:
I'm not sure how many people would be building context from the tox.ini. I think it would be much more helpful to add a section to the docs explaining why Molecule doesn't work well in a normal virtualenv on Fedora-based distributions. If we wanted we could then link to it from here as well.

doc/source/installation.rst
```diff
@@ -39,6 +39,13 @@ Pip
 
 :std:doc:`pip <pip:usage>` is the only supported installation method.
 
+Keep in mind that on selinux supporting systems, if you install into a virtual
+environment, you may face https://github.com/ansible/ansible/issues/34340 even
```
webknjaz: You can use extlink ref here (https://github.com/ansible/molecule/blob/6f14f0e/doc/source/conf.py#L76):
```suggestion
environment, you may face :gh:`issue <ansible/ansible/issues/34340>` even
```

Issue #1799, opened 2019-03-04
Fix flaky functional docker tests


decentral1se:
# Issue Type
* CI Issue

Nearly all CI builds fail with some unknown issue from the functional tests which I end up restarting manually (and they regularly pass then). These functional tests are currently only running the Docker tests but some of them are flaky. I'll list some below that I've seen and let's see what we can do to fix them.

* https://travis-ci.com/ansible/molecule/jobs/181549828
* https://travis-ci.com/ansible/molecule/jobs/181549832
* https://travis-ci.com/ansible/molecule/jobs/181549833
* https://travis-ci.com/ansible/molecule/jobs/180778543
* https://travis-ci.com/ansible/molecule/jobs/180778551

And so on ... just no decent logs to know what is the issue.

It would be nice to resolve this. First step is probably to add some debugging to these logs ...

decentral1se:
OK, let's see what https://github.com/ansible/molecule/pull/1855 shows us.

decentral1se:
OK, https://github.com/ansible/molecule/pull/1855#issuecomment-474935769 should cover us :fireworks:

decentral1se:
https://github.com/ansible/molecule/issues/1819 is also related :+1:

decentral1se:
OK, this is a duplicate of https://github.com/ansible/molecule/issues/1819.

PR #1828, opened 2019-03-13
Assure no_log is false on failed tasks

ssbarnea:
Improves user experience by displaying the errors that caused key
tasks to fail.

Avoids adding extra verbosity by using no_log on successes.

Fixes: #1666

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/molecule/cookiecutter/scenario/driver/azure/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
+++ b/molecule/cookiecutter/scenario/driver/azure/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
@@ -67,6 +67,7 @@
       until: azure_jobs.finished
       retries: 300
       with_items: "{{ server.results }}"
+      no_log: not item.failed
 
 
     # Mandatory configuration for Molecule to function.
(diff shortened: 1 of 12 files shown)


decentral1se:
btw, if we don't have energy for the `no_log:False` debate now, I am happy to accept this patch.

decentral1se:
Adding more data points for `no_log:False` default in https://github.com/ansible/molecule/issues/1666 :+1:

decentral1se:
This also addresses the first step of https://github.com/ansible/molecule/issues/1799

seandst:
> linking

s/linking/leaking/ ?

decentral1se:
Oh hmmmmm ... :(

seandst:
An idea I've been toying with is adding something alongside `MOLECULE_DEBUG` and the `--debug` flag like `MOLECULE_PROVISION_LOG`, that can be derived from `MOLECULE_DEBUG`, altered by a `--provision-log` flag, and stored as the default config value for `provisioner.log`. That could then be used instead of the current `no_log: "{{ not (lookup('env', 'MOLECULE_DEBUG') | bool or molecule_yml.provisioner.log|default(false) | bool) }}"` handling. That could also be combined with the log on failure stuff seen in this PR, I think.

I don't actually like the envvar or flag name, as they're a bit long, but that's just a cosmetic thing.

webknjaz:
Yes, "leaking", thanks :)

decentral1se:
I think https://github.com/ansible/molecule/pull/1839 is another attempt at this from a different angle.

Can we close here?

ssbarnea:
Superceeded by another PR.

PR #1830, opened 2019-03-13
Add a GitHub Action for linting test playbooks

webknjaz:
Experimenting with https://github.com/marketplace/actions/ansible-lint to have GitHub Actions based checks

#### PR Type

- Testing Pull Request
- Feature Pull Request

### DIFF
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -6,7 +6,7 @@ workflow "On PR" {
 action "Run Ansible Lint" {
   uses = "ansible/ansible-lint-action@master"
   env = {
-    ACTION_PLAYBOOK_NAME = "."
+    ACTION_PLAYBOOK_NAME = "test/"
   }
 }
 
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -1,9 +1,9 @@
 workflow "On PR" {
   on = "pull_request"
-  resolves = ["ansible/ansible-lint-action@master"]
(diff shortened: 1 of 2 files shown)


webknjaz:
It's not harmful, although it doesn't add much value.
Consider using this Action as dogfooding.

webknjaz:
@ssbarnea I think `set -x` should address this in https://github.com/ansible/ansible-lint-action

ssbarnea:
> @ssbarnea I think `set -x` should address this in https://github.com/ansible/ansible-lint-action

This is why I raised https://github.com/ansible/ansible-lint-action/issues/2 -- hopefully this will be addressed.

webknjaz:
@ssbarnea resolved

decentral1se:
This looks cool. However, I just remembered that I also ran the linter on our test files and saw a whole slew of issues that are not present here? Or am I missing something. Please see https://github.com/ansible/molecule/issues/1867.

webknjaz:
@decentral1se this check only looks for all of `playbook.yml` in the dir omitting others or a full path to only one playbook. It's more or less fine for the start. For more sophisticated playbook guessing we'll need to improve the action which is on my mind but will have to wait until I have time to actually do that.
I'm thinking about wrapping it with docs.octomachinery.dev but that's a discussion which is a bit off-topic in this repo.

Issue #1831, opened 2019-03-13
molecule requirementes should not pinned


ssbarnea:
# Desired Behavior

Molecule is a development/testing tool and it should not have pinned runtime or testing requirements.

Instead we need to use ranges that prevent usage of dependencies what **we know** are bugged or broken.

This also should avoid possible conflicts between a molecule dependency would come into conflict with the version mentioned by ansible or another dependency. 

Sooner or later we will clearly have a different version of `foo` package when using one version of anasible and another one when using a newer version.

When implementing this we should also use `pip check` to validate that after installation there are no reported conflicts reported by pip.

ssbarnea:
@webknjaz @decentral1se @gundalow Please use up/down to confirm the desrcribed behavior.

webknjaz:
@ssbarnea in general, you're right. I'd even suggest using pyup bot or dependabot to automate keeping things up-to-date.

I'm not sure whether @gundalow or someone recorded this somewhere, but current dependency spec is messed up: it has both runtime and dev deps pinned and mixed in the single unmanageable file. It's one of the caveats of `pbr` which I hope to solve by removing it.

Projects usually have different types of dependencies:
1) runtime deps, they should be open and have a lower limit + maybe a higher limit to next major if a certain dependency follows semver.
2) env deps, they describe and pin all packages including direct dev/test deps and their transitive dependencies. They can also specify hashes for deps, which makes the thing even more resilient.

These two categories should be dealt with separately. It's fine to pin dev deps versions as long as we have an automated process of bumping them. It's better than keeping them without any limits since automatic bumping triggers PR+CI meaning that we'd have recorded logs of what happens with the project with new deps instead of cryptic bug reports post-factum.
We just need to be more careful with runtime deps and try to not shoot in own legs...

webknjaz:
Ref: #1710

micheelengronne:
Merge request: https://github.com/ansible/molecule/pull/2024

decentral1se:
I think we largely dealt with the worst of this.

PR #1832, opened 2019-03-13
Unpin molecule requirements

ssbarnea:
Also adds a `pip check` execution for testing that there are no
conflicts after instalation of all requirements.

Temporary the exit code returned by `pip check` is ignored as we have
a known conflict caused by `ansible[azure]` dependency.

Avoids installing lxc on MacOS(Darwin) to avoid test failures as lxc
does not even compile on that platform.

Forces pytest<4.0.0 because one plugin does not support it yet but
includes comment so we can fix that as soon the plugin is updated.

Fixes: #1831

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,7 +57,7 @@ gce =
 linode =
     linode-python; python_version<"3.0"
 lxc =
-    lxc-python2; python_version<"3.0"
+    lxc-python2; python_version<"3.0" and platform_system=="Linux"
 openstack =
     shade
 vagrant =
--- a/tox.ini
+++ b/tox.ini
@@ -34,6 +34,9 @@ extras =
     windows
 commands_pre =
     find {toxinidir} -type f -not -path '{toxinidir}/.tox/*' -path '*/__pycache__/*' -name '*.py[c|o]' -delete
+    # temporary ignoring the exit code of pip check until we address azure
+    # extra conflicts: https://github.com/ansible/ansible/issues/53761
+    - pip check
 commands =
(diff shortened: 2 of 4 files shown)


setup.cfg
```diff
@@ -57,7 +57,7 @@ gce =
 linode =
     linode-python; python_version<"3.0"
 lxc =
-    lxc-python2; python_version<"3.0"
+    lxc-python2; python_version<"3.0" and platform_system!="Darwin"
```
webknjaz: I'd say linux only. LXC is *Linux* Containerization. It cannot work anywhere else

tox.ini
```diff
@@ -34,6 +34,9 @@ extras =
     windows
 commands_pre =
     find {toxinidir} -type f -not -path '{toxinidir}/.tox/*' -path '*/__pycache__/*' -name '*.py[c|o]' -delete
+    # temporary ignoring the exit code of pip check until we address azure
+    # extra conflicts: https://github.com/ansible/ansible/issues/53761
+    - pip check
```
webknjaz: I think linting doesn't belong to the testing section. It's a separate thing. It's closer to metadata validation.

ssbarnea: `pip check` is not linting, is detecting conficts between python packages. We cannot do it in linting because it is possible to have conflicts on each specific environment - really possible to discover that we have a conflict only on python27 with ansible26. 

Obviously that we cannot afford to add extra environments only for this purpose.'

Also this test is very quick, like less than 1s.

webknjaz: Okay. Agreed.

PR #1833, opened 2019-03-13
Assure that we deploy only on master pushes

ssbarnea:
Avoids case where deploy would fail because a previous build already
deployed a package with the same name.

PYPI servers never allow a published packages from being overriden due
to security concerns. We need to be sure that we never attempt to upload
the same version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -427,7 +427,7 @@ jobs:
     <<: *reset-prerequisites
     stage: Deploy
     name: Publishing current Git tagged version of dist to PyPI
-    if: repo == "ansible/molecule" AND tag IS present AND type == push
+    if: repo == "ansible/molecule" AND tag IS present
     env:
       TOXENV: metadata-validation
     deploy: &deploy-step
--- a/.travis.yml
+++ b/.travis.yml
@@ -427,7 +427,7 @@ jobs:
     <<: *reset-prerequisites
     stage: Deploy
     name: Publishing current Git tagged version of dist to PyPI
-    if: repo == "ansible/molecule" AND tag IS present
+    if: repo == "ansible/molecule" AND tag IS present AND type == push
     env:
       TOXENV: metadata-validation
     deploy: &deploy-step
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -439,15 +439,20 @@ jobs:
...
-    if: repo == "ansible/molecule" AND type NOT IN (cron, pull_request)  # Always run, except if PR or cron
+    if: type == push  # publish only pushes
     name: Publishing current (unstable) Git revision of dist to Test PyPI
     deploy:
       <<: *deploy-step
       server: https://test.pypi.org/legacy/
-
+      on:
+        tags: true
+        repo: ansible/molecule
```
gundalow: Apparently these should be added to the `if: ` line.

.travis.yml
```diff
@@ -427,7 +427,7 @@ jobs:
     <<: *reset-prerequisites
     stage: Deploy
     name: Publishing current Git tagged version of dist to PyPI
-    if: repo == "ansible/molecule" AND tag IS present
```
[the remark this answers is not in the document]
webknjaz: Exactly. Except that it'll fail in the end because forks can't access secrets from upstream.

.travis.yml
```diff
@@ -427,7 +427,7 @@ jobs:
...
+    if: repo == "ansible/molecule" AND tag IS present AND type == push
```
webknjaz: No, this is not needed. Those deploy jobs appear there because of push events in branches since you did pushes to branches in upstream. Tag builds are only triggered by pushes so it doesn't matter here.

```suggestion
    if: repo == "ansible/molecule" AND tag IS present
```

.travis.yml
```diff
@@ -442,7 +442,7 @@ jobs:
         all_branches: true
 
   - <<: *deploy-job
-    if: repo == "ansible/molecule" AND type NOT IN (cron, pull_request)  # Always run, except if PR or cron
+    if: repo == "ansible/molecule" AND type == push AND branch == "master"  # publish only pushes to master
```
webknjaz: So why did you decide to exclude `api` event?

webknjaz: On the second thought, it's probably the right thing to do...

PR #1835, opened 2019-03-14
Avoid duplicate builds on PRs

ssbarnea:
Avoid building branches other than master and tags in order to avoid
duplicate builds on PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@
 conditions: v1
 dist: xenial
 
+# avoid building other branches to avoid duplicate builds on PRs
+branches:
+  only:
+    - master
+    - /^\d+\.\d+(\.\d+)?(-\S*)?$/
 git:
   depth: 200
 


.travis.yml
```diff
@@ -2,6 +2,11 @@
 conditions: v1
 dist: xenial
 
+# avoid building other branches to avoid duplicate builds on PRs
+branches:
+  only:
+    - master
+    - /^\d+\.\d+(\.\d+)?(-\S*)?$/
```
decentral1se: :fearful:

"fearful"

ssbarnea:
Please make a decision on this:
A) we avoid duplicated builds on PRs by losing ability to test branches withour a PR (can still be done by people on their own forks)
B) Keep current behaviour with doubled builds on PRs.

I just want to know if I should merge or close.

webknjaz:
C) don't create branches in upstream.

MarkusTeufelberger:
Yeah, what's with all these single-commit branches in this repository here? If you want to do private experiments, please create WIP pull requests and keep them in your own repository like everyone else without commit rights has to do.

ssbarnea:
Closing this due to lack of support. Regarding the question of making changes in private branches: these have a serious issue: it does not allow multiple people to contribute to the same change. This is why I find more productive to use temporary upstream branches where any core can fix them, regardless which one created it in the first place (something bit similar with how gerrit works).

PS. As long we are careful to cleanup the temporary branches we should be ok.

webknjaz:
@ssbarnea you actually can push to other people's PR branches given that they didn't untick that checkbox ☒ to allow maintainers to push there. You can even do it from the GitHub UI

gundalow:
I personally vote for:
C) don't create branches in upstream

* This is the process for ansible/ansible
* If you enable other people to commit to your fork (checkbox on right of PR) 
* Only a few people have commit and therefor could create these branches
* branches created in ansible/molecule will be pulled down by everybody

decentral1se:
I push to my fork as per usual ....

PR #1836, opened 2019-03-14
Also publish to Test PyPI on tag in Travis CI

webknjaz:
Travis CI CD should publish to Test PyPI on tag event as well, otherwise, we'll have gaps there.

#### PR Type

- Bugfix Pull Request


webknjaz:
@ssbarnea we missed this in the previous PR

PR #1839, opened 2019-03-16
Added ability to control no_log behavior

ssbarnea:
User can override this behaviour by defining MOLECULE_NO_LOG variable to true of false. When defined that value would override any value of `provider.log` mentioned inside molecule.yaml.

When variable is not defined, the value from molecule.yml file will be used.

When none of these are defined, logging will be disables (keeps behavior consistent with old versions).

This matches Ansible behavior where environment variables do take precedence over values defined in playbooks.

Fixes: #1666

### DIFF
--- a/molecule/cookiecutter/scenario/driver/azure/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
+++ b/molecule/cookiecutter/scenario/driver/azure/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
@@ -4,7 +4,7 @@
   hosts: localhost
   connection: local
   gather_facts: false
-  no_log: "{{ not (lookup('env', 'MOLECULE_DEBUG') | bool or molecule_yml.provisioner.log|default(false) | bool) }}"
+  no_log: "{{ molecule_no_log }}"
   vars:
     resource_group_name: molecule
     location: westus
--- a/molecule/cookiecutter/scenario/driver/azure/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/destroy.yml
+++ b/molecule/cookiecutter/scenario/driver/azure/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/destroy.yml
@@ -4,7 +4,7 @@
   hosts: localhost
   connection: local
   gather_facts: false
-  no_log: "{{ not (lookup('env', 'MOLECULE_DEBUG') | bool or molecule_yml.provisioner.log|default(false) | bool) }}"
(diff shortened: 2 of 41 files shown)


ssbarnea:
As this change can risk leaking credentials we need extra attention so please try to use it locally first and see how it behavers, especially if you do use some providers needing credentials. 

It would be extreamly useful if someone can find an example where these credentials are exposed so we can investigate that. We have no credentials on Travis so we cannot really test/risk anything with CI.

Still, if we start testing new providers things could change. My impression is that if the user already used the recommended/safe ways to access the providers we probably do not risk leaking anything because molecule would not have to deal with any credentials by itself.

.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
 exclude = .venv/,.tox/,dist/,build/,doc/,.eggs/,molecule/provisioner/ansible/plugins/libraries/goss.py
 format = pylint
-ignore = E741,W503,W504
+ignore = E741,W503,W504,H
```
webknjaz: What's `H` and how is it related to this PR?

molecule/cookiecutter/scenario/driver/azure/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
```diff
@@ -4,7 +4,7 @@
   hosts: localhost
   connection: local
   gather_facts: false
-  no_log: "{{ not (lookup('env', 'MOLECULE_DEBUG') | bool or molecule_yml.provisioner.log|default(false) | bool) }}"
+  no_log: "{{ molecule_no_log }}"
```
webknjaz: I like the attempt to move this logic out of playbooks :+1:

decentral1se:
> My impression is that if the user already used the recommended/safe ways to access the providers we probably do not risk leaking anything because molecule would not have to deal with any credentials by itself.

Not sure what you mean by this?

.flake8
ssbarnea: Not directly related, mainly if you install flake8-docstrings into sytem python, our tests will start to fail because we started to use --site-system-packages. This is disabling the docstrings plugin errors. CI not affected but very easy to test. I was too lazy to create a new CR only for this nit.

ssbarnea: Already fixed as it was merged in different PR.

ssbarnea:
@decentral1se Fixed by keeping old behavior, we can decide another time the change of default value.

decentral1se:
> Fixed by keeping old behavior, we can decide another time the change of default value.

Oh yes, great idea :+1: :+1: :+1: 

CI is unfortunately failing (example: https://travis-ci.com/ansible/molecule/jobs/187030475).

Issue #1757, opened 2019-02-19
Document and add check for pbr/dpl setup


decentral1se:
# Issue Type
- Documentation Request
- CI Fix

* We should document the explanation in https://github.com/ansible/molecule/pull/1755#pullrequestreview-205294580 for how the prerelease versions are being created to make this more accessible.
* We should implement the check in https://github.com/ansible/molecule/pull/1755#issuecomment-465193956 to future proof an error like we saw in #1730.

decentral1se:
I've submitted https://github.com/ansible/molecule/pull/1853 for the user end of prereleases.

decentral1se:
Re-reading this ticket, I think the pbr/dpl stack is too brittle to document against: we want to get rid of pbr and there have been some changes against the pre-release triggering (I think it is at least understood by more than one person lurking the project, which is good!). So, I'm just going to close this off and do other things.

PR #1853, opened 2019-03-20
Document pre-release installation

decentral1se:
Something for https://github.com/ansible/molecule/issues/1757.

#### PR Type
- Docs Pull Request

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -69,6 +69,25 @@ some benefits:
 * allows molecule installation at user level without even needing to have
   the script in ``PATH``.
 
+.. note::
+
+    We also have a continuous pre-release process which is provided for early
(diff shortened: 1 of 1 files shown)


doc/source/installation.rst
```diff
@@ -69,6 +69,25 @@ some benefits:
 * allows molecule installation at user level without even needing to have
   the script in ``PATH``.
 
+.. note::
+
+    We also have a continuous pre-release process which enables early feedback.
+    This pre-release is provided for testing purposes only. It is available
```
gundalow: "testing" is a loaded term when we are talking about Molecule which is a testing tool, though my brain is failing to come up with something better.

decentral1se: Fair enough! Pushed a rewording which uses 'early adoption and feedback purposes only'.

Issue #1859, opened 2019-03-21
Document setuptools upgrade for CentOS7


robertvolkmann:
Installation described for CentOS 7 failed on a fresh GenericCloud image with:
```
$ sudo yum install -y epel-release
$ sudo yum install -y gcc python-pip python-devel openssl-devel libselinux-python
$ pip install --user molecule
[...]
Collecting testinfra==1.19.0 (from molecule==2.20)
  Using cached https://files.pythonhosted.org/packages/2b/96/3c0eb74894acf3e026e437ee1b1161a200adb9c9a189c36bc4a92ed2db79/testinfra-1.19.0.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-asndpD/testinfra/setup.py", line 59, in <module>
        'six>=1.4',
      File "/usr/lib64/python2.7/distutils/core.py", line 112, in setup
        _setup_distribution = dist = klass(attrs)
      File "/usr/lib/python2.7/site-packages/setuptools/dist.py", line 269, in __init__
        _Distribution.__init__(self,attrs)
      File "/usr/lib64/python2.7/distutils/dist.py", line 287, in __init__
        self.finalize_options()
      File "/usr/lib/python2.7/site-packages/setuptools/dist.py", line 302, in finalize_options
        ep.load()(self, ep.name, value)
      File "/tmp/pip-build-asndpD/testinfra/setuptools_scm-3.2.0-py2.7.egg/setuptools_scm/integration.py", line 10, in version_keyword
      File "/tmp/pip-build-asndpD/testinfra/setuptools_scm-3.2.0-py2.7.egg/setuptools_scm/version.py", line 66, in _warn_if_setuptools_outdated
    setuptools_scm.version.SetuptoolsOutdatedWarning: your setuptools is too old (<12)
```

The version bump of testinfra from 1.16.0 to 1.19.0 by #1751 causes this issue.

decentral1se:
Hi @robertvolkmann,

Ah yes, that makes sense. CentOS 7 ships with an old setuptools, no? I would recommend to install Molecule in a virtual environmen then. Or if you have some system wide friendly way to upgrade setuptools? We should document that this is a limitation on CentOS7.

Thoughts?

Also note, we're considering https://github.com/ansible/molecule/issues/1849 too.

robertvolkmann:
Hi @decentral1se,

[...]

PR #1869, opened 2019-03-22
[DNM] Added minimal setuptools version setup requirements

ssbarnea:
Some of our dependencies require a minimal version of setuptools and
we need to be sure that this is installed before requirements are
installed.

Fixes: #1859

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,5 +23,6 @@ import setuptools
 setuptools.setup(
     packages=setuptools.find_packages(),
     pbr=True,
-    setup_requires=['pbr']
+    setup_requires=['pbr', 'setuptools>=12']
+    # setuptools due https://github.com/ansible/molecule/issues/1859
 )


doc/source/installation.rst
```diff
@@ -44,6 +44,20 @@ if selinux is not enabled or is configured to be permissive.
 It is your reponsability to assure that soft dependencies of Ansible are
 available on your controller or host machines.
 
+.. warning::
```
decentral1se: Includes a commit that shouldn't be there?

ssbarnea: @decentral1se this if f... weird! I have no idea where this code came from but I have the impression that somehow github combine my change with another change made by you related to setuptools. I guess you used the same branch name `fix/setuptools`?

When I pushes there was not such branch open, still the final result is a combined change. Not sure what cause it. My change was a single line and this is what I was expecting to see raise.

I will fix it now, hopefully it does not become a habit of me raising random code.

ssbarnea:
@decentral1se I am not sure that this change is fixing the reported problem before we merge this lest find a way to reproduce the error we seen with testinfra.

doc/source/installation.rst
seandst: That's one of many good reasons to open PRs from forks, I think.

setup.py
```diff
@@ -23,5 +23,6 @@
 setuptools.setup(
     packages=setuptools.find_packages(),
     pbr=True,
-    setup_requires=['pbr']
+    setup_requires=['pbr', 'setuptools>=12']
```
webknjaz: I'm here to tell you that this doesn't make sense because of the way it works. `setuptools.setup()` installs this only for setup-time and then discards those.
But the problem is that it will not pick up a newer version because the older version is already in runtime and module cannot substitute itself.
The proper place for adding build deps nowadays is `pyproject.toml`->`[build-system]`->`requires'.

decentral1se:
Oh dear, I am so used to clicking this delete branch button that I now did it on someones elses! Sorry!

decentral1se:
:smile:

webknjaz:
@decentral1se :) BTW we've set up https://github.com/apps/delete-merged-branch so that it'd take care of merged things if they are in our repo...

webknjaz:
@ssbarnea your "DNM" tag got me puzzled. I couldn't understand what it means for a while.

I want to inform you that we have https://github.com/marketplace/wip GitHub App integrated and it only understands `WIP`, `Work in progress` and `🚧` marks.
See https://github.com/ansible/molecule/pull/1869/checks?check_run_id=83024113 for details.

ssbarnea:
@webknjaz I will remember this. Mainly in OpenStack DNM is more popular than WIP but with very similar meanings (DoNoMerge). The different is that DNM is usually added to stuff that should not be merged even if the work on it is done (conflicts, or further debates). WIP usually means "i am working on it". Now you know the difference and why both are useful.... I had to google first time myself to find the meaning of DNM.

webknjaz:
@ssbarnea FWIW I wrote a tutorial reimplementing WIP app with my framework: https://tutorial.octomachinery.dev/en/latest/octomachinery-with-checks-api.html
Maybe at some point, we'd deploy actual app where we'd control extra conditions and stuff. It's also possible to have more illustrative feedback :) as in https://github.com/webknjaz/comment-reactor-pyconsk-324/pull/3/checks?check_run_id=84343706

Issue #1758, opened 2019-02-20
rewrite def supports_linode()


themr0c:
# Issue Type

- Feature request

# Desired Behavior

As stated in https://github.com/ansible/molecule/pull/1691#discussion_r258524200
in test/functional/conftest.py, `def supports_linode()` should be rewritten in a more functional way.

Issue #1821, opened 2019-03-11
molecule testing should skip any drivers that are not configured by default


ssbarnea:
Currently running functional testing on a machine takes forever even if the user has only docker configured. 

We need to assure that all testing drivers that are not configured (azure, lxc, openstack,ec2,...) are skipped almost instantly by default.

decentral1se:
Agreed!

Kinda related: https://github.com/ansible/molecule/issues/1758.

decentral1se:
https://github.com/ansible/molecule/pull/1870 closes this for now.

decentral1se:
https://github.com/ansible/molecule/pull/1870 is merged! Gonna close this off for now.

Issue #1752, opened 2019-02-16
improve testing experience for developers


ssbarnea:
Currently running molecule test suite locally seems to be a real challenge because the old approach of cloning and just running `tox` does not work at all. 

If you are lucky this will run succefully linting and unit tests but will likely give you 99% failure rase on functional tests and not without waiting many minutes for them to run.

Functional testing should be made smart: like detecting which drivers are available and configured on current platform and use only these, skipping all the tests that cannot be performed because your current system does not have the prerequisites for testing them (like missing tools or credentials).

On CI the auto-detection would be disabled as we do not want to accidentaly skip tests just because some credentials were changed or something similar. Still, for development machines the default behaviour should be to tests "whatever can be tested".

I am fully aware that user can define his own `TOXENV` variable to limit what is run, still this does not cover for different drivers, which may not exist on his machine (example lxc is not available on MacOS).

Maybe we should also consider a different matrix split of testing by driver instead of having unit/functional. We could run unti and functional in the same environment (sequential?) but we should probably split by driver. As a user I find extremely useful to do something like `tox -e py27-ansible25-docker`.

Most developers withh not have credentials for all the supported platforms, likely they will have *none* and just rely on the two non cloud drivers.

- [ ] Coverage report is kinda useless/spammy for developer usage, should not be disabled on non CI usage
- [ ] Why we really need rubocop? Can we avoid it as it introduces a testing dependency on ruby which is quite problematic for at least two platforms MacOS and CentOS-7 where is far from easy to install.
- [ ] Consider using a Makefile or bash script for testing because this could be used to test and eventually install build-prerequisites for each platform. While writing bash could be easier I am inclined to believe that a Makefile could make easier to maintain long term (can provide help here).
- [ ] Replace complex development environment instructions with a simple command like "make bootstrap". The script could replace documentation and users could make PRs to improve it.
- [ ] Consired recording system package dependencies inside [bindep.txt](https://pypi.org/project/bindep/) instead of docs.

Undocumented tricks for easing local testings:
```
export TRAVIS=1
export CI=1
export PYTEST_ADDOPTS="-x --no-cov"
```
First two lines would disable any non-docker tests and the last one will fail on fast on first failure and avoid doing coverage and its reports. Maybe we can find a way to make these implicit for developers.

lwm:
Important points! Thanks for raising. We'll have to try and make this work better for ourselves indeed.

webknjaz:
[...]

PR #1870, opened 2019-03-22
Improve supports_* driver checks (use Ansible logic)

decentral1se:
#### PR Type
- CI Improvement

Something for https://github.com/ansible/molecule/issues/1752.
Closes https://github.com/ansible/molecule/issues/1758.

It's now possible to just run `tox -e functional` and get a clean docker run.

Since we're tied to Ansible, let's use the logic there too for checking dependencies.
Regarding environment variables, let's just check for the ones that our `create.yml` require.

Also, let's show logs for failed functional docker tests. It's just more useful and there are no credentials there, we control them and we need better feedback when they fail on the CI.

### DIFF
--- a/test/functional/conftest.py
+++ b/test/functional/conftest.py
@@ -40,6 +40,11 @@ LOG = logger.get_logger(__name__)
 IS_TRAVIS = os.getenv('TRAVIS') and os.getenv('CI')
 
 
+def _env_vars_exposed(env_vars, env=os.environ):
+    """Check if environment variables are exposed."""
+    return all(var in env for var in env_vars)
+
+
 @pytest.fixture
 def with_scenario(request, scenario_to_test, driver_name, scenario_name,
                   skip_test):
@@ -268,14 +273,6 @@ def get_docker_executable():
     return distutils.spawn.find_executable('docker')
 
 
-def get_linode_executable():
-    try:
-        pytest.importorskip('linode')
-        return True
-    except Exception:
-        return False
-
-
 def get_lxc_executable():
     return distutils.spawn.find_executable('lxc-start')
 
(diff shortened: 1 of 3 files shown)


test/functional/conftest.py
```diff
@@ -299,8 +291,15 @@ def supports_docker():
...
 @pytest.helpers.register
 def supports_linode():
-    # FIXME: Travis CI
-    return not IS_TRAVIS and get_linode_executable()
+    from ansible.modules.cloud.linode.linode import HAS_LINODE
+
+    required_env_vars = ('LINODE_API_KEY', )
+
+    for required_env_var in required_env_vars:
+        if os.getenv(required_env_var) is None:
```
webknjaz: `required_env_var in os.environ` is probably more reliable in this case.

test/functional/conftest.py
```diff
@@ -343,26 +342,72 @@ def demands_delegated():
 
 @pytest.helpers.register
 def supports_azure():
-    # FIXME: come up with an actual check
-    return not IS_TRAVIS  # FIXME: Travis CI
+    from ansible.module_utils.azure_rm_common import HAS_AZURE
+
+    required_env_vars = (
```
webknjaz: You've got a log of copy-paste in this PR. Put it into a function.

test/functional/conftest.py
```diff
@@ -343,26 +342,72 @@ def demands_delegated():
...
+        'AZURE_SUBSCRIPTION_ID',
+        'AZURE_CLIENT_ID',
+        'AZURE_SECRET',
+        'AZURE_TENANT',
+    )
+
+    for required_env_var in required_env_vars:
```
webknjaz: could be
```suggestion
    return all(v in os.environ for v in required_env_vars) and HAS_AZURE
```
```suggestion
    for required_env_var in required_env_vars:
```
```suggestion
    return all(v in os.environ for v in required_env_vars) and HAS_AZURE
```

decentral1se:
OK, taken another pass to clean this up.

decentral1se:
Closes https://github.com/ansible/molecule/issues/1821.

decentral1se:
Solid feedback :100: 

> The openstack skip logic could be improved to work with os-client-config/clouds.yml, but I think it would be preferable to open that as a separate enhancement issue once this is merged (and I'm happy to do that).

That'd be great!

decentral1se:
Right, gonna merge this then. Looks like getting it out the way will clear up other work.

Issue #1819, opened 2019-03-11
Tests failing due to network unavailability


ssbarnea:
# Issue Type

- Bug report

# Molecule and Ansible details

To reproduce this, try running:
```
 # wget -vv https://github.com/yatesr/ansible-timezone/archive/1.1.0.tar.gz
Resolving github.com (github.com)... 192.30.253.113, 192.30.253.112
Connecting to github.com (github.com)|192.30.253.113|:443... failed: No route to host.
Connecting to github.com (github.com)|192.30.253.112|:443... failed: No route to host.
```
 
While this worked from some machines is seems to fail from others, or it may work from the browser but from fromthe CLI.

At this moment this is needed during testing but we need to find a way to avoid this issue.
```
MOLECULE_DEBUG=1 tox -e functional -- -k 'test_command_check[driver/docker-docker-default]'
```

decentral1se:
Hmmmm, it seems `test/scenarios/dependency/molecule/gilt/playbook.yml` does some downloading. Maybe we should change all of these network invocations into local role installations instead. We could add a dummy role of some sort. Good catch!

decentral1se:
Closed https://github.com/ansible/molecule/issues/1799 in favour of this. We've got network access in:

```
➜  molecule (cifix/remove-flaky-network-calls) ✗ g s
## cifix/remove-flaky-network-calls
 M test/scenarios/dependency/molecule/ansible-galaxy/requirements.yml
 M test/scenarios/plugins/molecule/default/gilt.yml
 M test/scenarios/side_effect/molecule/default/requirements.yml
 M test/scenarios/verifier/molecule/goss/verify.yml
 M test/scenarios/verifier/molecule/inspec/verify.yml
```

I'll try to remove them and then we should be getting reliable functional test builds ...

ssbarnea:
@decentral1se Very good idea! I think that we should not rely on external repositories. If we need code for testing, we can always host it in our repository.

decentral1se:
OK, I do think https://github.com/ansible/molecule/pull/1871 is now doing a pretty good job at resolving this as much possible (compromise between still accessing network to perform good functional tests vs. not accessing network).

PR #1871, opened 2019-03-22
Reduce network access in functional tests

decentral1se:
#### PR Type
- CI Improvement

Closes https://github.com/ansible/molecule/issues/1819.

OK, here we go. I've tried to diagnose all network access in the functional tests.

Please see each commit for the reasoning behind each change.

Points and questions (all answered below ...):

* ~~We now vendor 70mb (binary + RPM package) in the repository~~
  * ~~How to better deal with these binary/RPM packages without network access?~~
* ~~We still make network access calls with dependency resolvers.~~
  * ~~Should we just remove the dependency scenario? We already unit test it!~~


decentral1se:
> NO WAY!

@ssbarnea, this is not helpful. I would appreciate more positive responses for my efforts ... I expect that in return since I aim to be respectful on this project. For you and others ...

Please read what I said in the initial description:

>> How to better deal with these binary/RPM packages without network access?

I am asking for recommendations on what to do here. 

This change makes the functional tests *much* more reliable if we don't do these downloads.

Can I please get some constructive feedback?

seandst:
Is this intended to help out people running functional tests locally? In other words, is it okay if travis (or $whateverCI) still downloads things? I have some ideas, but would like to understand the motivations a little better before commenting.

The issue link in the description doesn't point to an issue, so I'm not sure if this is in reference to an open issue that's already got some of these discussion points or not.

decentral1se:
Ah, I've updated the description. The ticket was https://github.com/ansible/molecule/issues/1819. So, this PR fixes (or makes much less likely) the fact that the functional tests for the CI are breaking nearly every run due to flaky network access.

RE: big binary sizes, I saw that https://git-lfs.github.com/ is the offered Github solution for storing pointers to files ... however, it doesn't stop us having to download a thing from Github (the original problem, in some ways ...).

What to do here? My growing feeling is that we should remove anything that speaks over the network and replace with unit tests and mocks. And leave testing core functionality of the things that do the downloading to the test suite of the tool that does the downloading ...

However, if we're going to be adding tests for drivers (Linode, GCP, AWS, etc.) they could all potentially speak over the network ... ideally, we'd have a mock server for all of those (original idea of @webknjaz which I still think is relevant)

This is a big topic, I'm going to add a note to the WG agenda ...

MarkusTeufelberger:
Please don't vendor binaries. Neither in git-lfs nor in the repo directly. Many/most CI systems allow you to have a caching directory where you could load these binaries from (after the first successful download) instead of hitting the network each time.

In the long run I think a "driver" should just be the name of any Ansible module that's then used to spin up resources (with `state: present`) or destroy them (with `state: absent`) - with the resources defined in the `instances:` object.

decentral1se:
>  Many/most CI systems allow you to have a caching directory where you could load these binaries from

OK, that seems fine. I'll aim for that.

decentral1se:
OK, I won't bother with cache since both the binary and the RPM are downloaded into the docker container which molecule manages. I'd rather not deal with having to download it on the host, cache it and then copy it over. I think this PR has reduced the network access enough for now. Let's see what reliability issues we face afterwards. Thanks for reviews :+1:

test/resources/roles/testplugin/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License
+
+Copyright (c) 2017 John Dewey
```
webknjaz: RH?

decentral1se:
> If you don't want them, why not just skip those on pytest level?

Well the whole point of this PR is to remove them since they fail often during network access.

webknjaz:
@decentral1se my point is that you don't need to remove those *from code* but add more code controlling whether they get to run in CI (or wherever they fail). Removing parameters completely makes it impossible to run them when they will be needed.

ssbarnea:
> Can I please get some constructive feedback?
Sure, almost forgot that this is not IRC. Altered the message. Sadly I do not know exactly what needs to be done but it should not involve adding binary files to repository and external network needs for functional should be minimized.

decentral1se:
> Sure, almost forgot that this is not IRC.

I don't know what kind of excuse this is.

>  Altered the message. 

Not much better to be honest.

Moving on ...

ssbarnea:
I was trying to get this PR locally to test it but somehow hub cli fails with it:
```
$ hub checkout https://github.com/ansible/molecule/pull/1871                                                                                                          From https://github.com/ansible/molecule
! [rejected]          refs/pull/1871/head -> cifix/remove-flaky-network-calls  (non-fast-forward)
FAIL: 1
```

It is a bit strange because I am sure that I used it to get other PRs locally and test them, via this syntax or alternative `hub pr checkout 1234` way.

test/resources/roles/testplugin/LICENSE
decentral1se: This code is copy/pasta'd from https://github.com/retr0h/ansible-plugins to avoid network acess.

My guess is we need to retain the license.

decentral1se:
OK, I've taken another pass at this. Summary is:
  * Leave binary/package download as is as per https://github.com/ansible/molecule/pull/1871#issuecomment-476793145.
  * Use `pytest.skip` to avoid extra dependency downloading as per https://github.com/ansible/molecule/pull/1871#issuecomment-477140374.

decentral1se:
OK, I've addressed comments and wanted to get this in since it will give us much guarantees to see the CI runs for functional testing actually pass instead of failing with all sorts of flaky network issues. Thanks for the reviews folks.

test/functional/test_command.py
```diff
@@ -183,8 +183,12 @@ def test_command_create(scenario_to_test, with_scenario, scenario_name):
         'driver_name',
         'scenario_name',
     ])
-def test_command_dependency_ansible_galaxy(scenario_to_test, with_scenario,
-                                           scenario_name):
+def test_command_dependency_ansible_galaxy(request, scenario_to_test,
+                                           with_scenario, scenario_name):
+    # FIXME(decentral1se): skipped due to failures on network access
+    if request.getfixturevalue('driver_name') != 'docker':
```
webknjaz: I'm pretty sure that you could've just added `driver_name` as an argument of the test function...

decentral1se: Nah, didn't work ...

PR #1883, opened 2019-03-26
Remove manual version checks for Ansible/Python

decentral1se:
Please see https://github.com/ansible/molecule/pull/1890.

### DIFF
--- a/molecule/shell.py
+++ b/molecule/shell.py
@@ -46,41 +46,28 @@ def _get_ansible_version():  # pragma: no cover
     return ansible.__version__
 
 
-def _supported_python2_version():  # pragma: no cover
-    return _get_python_version()[:2] == (2, 7)
+def _supported_python_version():  # pragma: no cover
+    py_version = _get_python_version()
+    return py_version[:2] == (2, 7) or py_version >= (3, 5)
 
 
-def _supported_python3_version():  # pragma: no cover
-    return _get_python_version() >= (3, 6)
-
-
-def _supported_ansible_version():  # pragma: no cover
-    if (distutils.version.LooseVersion(_get_ansible_version()) <=
-            distutils.version.LooseVersion('2.4')):
+def _supported_ansible_and_python_version():  # pragma: no cover
+    if (distutils.version.LooseVersion(_get_ansible_version()) <
+            distutils.version.LooseVersion('2.5')):
         msg = ("Ansible version '{}' not supported.  "
                'Molecule only supports Ansible versions '
-               "'>= 2.4'.").format(_get_ansible_version())
+               "'>= 2.5'.").format(_get_ansible_version())
         util.sysexit_with_message(msg)
 
-    if _supported_python2_version():
-        pass
-    elif _supported_python3_version():
-        if (distutils.version.LooseVersion(_get_ansible_version()) <
-                distutils.version.LooseVersion('2.4')):
-            msg = ("Ansible version '{}' not supported.  "
-                   'Molecule only supports Ansible versions '
-                   "'>=2.5' with Python version '{}'").format(
-                       _get_ansible_version(), _get_python_version())
-            util.sysexit_with_message(msg)
-    else:
+    if not _supported_python_version():
         msg = ("Python version '{}' not supported.  "
                'Molecule only supports Python versions '
-               "'2.7' and '>= 3.6'.").format(_get_python_version())
+               "'2.7' and '>= 3.5'.").format(_get_python_version())
         util.sysexit_with_message(msg)
 
 
 def _allowed(ctx, param, value):  # pragma: no cover
-    _supported_ansible_version()
+    _supported_ansible_and_python_version()
 
     return value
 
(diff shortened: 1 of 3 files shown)


molecule/shell.py
```diff
@@ -46,41 +46,28 @@ def _get_ansible_version():  # pragma: no cover
...
         msg = ("Python version '{}' not supported.  "
                'Molecule only supports Python versions '
-               "'2.7' and '>= 3.6'.").format(_get_python_version())
+               "'2.7' and '>= 3.5'.").format(_get_python_version())
         util.sysexit_with_message(msg)
 
 
 def _allowed(ctx, param, value):  # pragma: no cover
-    _supported_ansible_version()
+    _supported_ansible_and_python_version()
```
webknjaz: Please call this method as an action/effect it actually does. `supported_` or `_is_supported` assumes that it'd return `bool`. Whereas it's not true: it has side-effect which is misleading.

A better way would be smth like
```suggestion
    _terminate_if_unsupported_env()
```
which makes it possible to understand what's going on in the scope of `_allowed()` function without having to look up the implementation of this other one. This is because having proper names *carries context* which augments readers understanding when they check this function separately.
Readability matters and using only concepts (via var and func names) of only that one abstraction layer which is represented by the current scope (function body) improves it dramatically.

molecule/shell.py
```diff
@@ -46,41 +46,28 @@ def _get_ansible_version():  # pragma: no cover
     return ansible.__version__
 
 
-def _supported_python2_version():  # pragma: no cover
-    return _get_python_version()[:2] == (2, 7)
+def _supported_python_version():  # pragma: no cover
```
webknjaz:
```suggestion
def _is_python_version_supported():  # pragma: no cover
```

molecule/shell.py
```diff
@@ -46,41 +46,28 @@ def _get_ansible_version():  # pragma: no cover
...
+    py_version = _get_python_version()
```
webknjaz: Just use
```suggestion
    py_version = sys.version_info
```

Having a separate one-liner doesn't really make it better since it's just a proxy for accessing one value.

decentral1se:
> On the second thought, just remove the Python version check. This should be done by the dist installer (Pip) and there's metadata in dist about the Python version restrictions.

Right, I was thinking this when I saw the code! Thought this PR might force a decision on this :)

Good!

I'll just leave the Ansible check then ...

molecule/shell.py
decentral1se: Great, thanks.

molecule/shell.py
```diff
@@ -46,41 +46,28 @@ def _get_ansible_version():  # pragma: no cover
...
-def _supported_ansible_version():  # pragma: no cover
-    if (distutils.version.LooseVersion(_get_ansible_version()) <=
-            distutils.version.LooseVersion('2.4')):
+def _supported_ansible_and_python_version():  # pragma: no cover
+    if (distutils.version.LooseVersion(_get_ansible_version()) <
+            distutils.version.LooseVersion('2.5')):
         msg = ("Ansible version '{}' not supported.  "
                'Molecule only supports Ansible versions '
-               "'>= 2.4'.").format(_get_ansible_version())
+               "'>= 2.5'.").format(_get_ansible_version())
```
webknjaz: maybe save this in a var?

decentral1se:
@ssbarnea Do not "request changes" if you haven't taken the time to understood the issue: title and description are already included and another reviewer understood well what it is about.

webknjaz:
@decentral1se

I've looked up the dependency tree and we actually have Ansible as a transitive dependency via ansible-lint. But it doesn't declare version limitations.

Reiterating the problem, I think that we could deal with Ansible the same way as with Python: let's just specify proper requirements in the dist metadata and Pip will take care of it. The latest versions of Pip are even capable of warning users about incompatible things in env...

This way we won't have to think about checking this manually in our code. And I'd prefer Ansible to be a direct dependency anyway: it's kinda weird that it's not now.

ssbarnea:
@decentral1se Maybe I was not clear but I am trying to assure that we do endup creating commit messages that are not clear, as these would endup in the changelist of the release. Currently by looking at the PR title and comment I doubt a molecule user could understand what was the purpose of the change.

I only see that is marked as a bugfix PR, but one without a bug. If is a bugfix, it needs to have a a "Fixes: #<num>" line on it which would link to the bug that is fixed by it. On other project having a meaningful commit message test is done before even looking at the modified code, commit message is part of the review too, and believe me that I was asked to update mine lots of time, even for trivial reasons like typos.

Now, back to this change, I kinda find weird that we need to have extra checks in our code for python version and so when this should be handled by packaging tools alone. Do we really need all this extra code inside molecule or we can avoid it?

webknjaz:
> Now, back to this change, I kinda find weird that we need to have extra checks in our code for python version and so when this should be handled by packaging tools alone. Do we really need all this extra code inside molecule or we can avoid it?

that's basically what I said earlier :laughing:

webknjaz:
@decentral1se @ssbarnea I've got a collection of articles about writing meaningful commit messages. I'll post it as a separate issue so that anyone could go through this and we could be on the same page...
Sounds fair?

decentral1se:
Good idea @webknjaz! New attempt in https://github.com/ansible/molecule/pull/1890!

PR #1887, opened 2019-03-28
tox.ini rework: regroup things in the testenv section

themr0c:
This is a proposal to tidy up the `tox` configuration.

The rationale I followed was following: to avoid duplication of statements, regroup everything that can be regrouped in the `testenv` section. I hope this is also increasing readability of the `tox.ini` file.

I also moved the call to a second requirement file inside of the `test-requiremenst.txt` file.

I am not sure if this approach is consensual among `tox` users, so please tell me if this is going in the right direction, or absolutely not.

#### PR Type

- Feature Pull Request

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,3 +1,4 @@
+-r requirements.txt
 mock==2.0.0
 pur==5.2.1
 pytest==3.6.0
--- a/tox.ini
+++ b/tox.ini
@@ -54,14 +54,14 @@ commands =
     unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
 skip_install =
   # install Python dist dependencies in following environments:
-  {functional,unit}: false
+  {doc,functional,unit}: false
   # don't install any Python dist dependencies in following environments:
-  {build-docker,doc,format,lint,metadata-validation,pur}: true
+  {build-docker,format,lint,metadata-validation,pur}: true
 usedevelop =
   # don't install Molecule in following environments:
-  {build-docker,doc,format,lint,metadata-validation,pur}: false
+  {build-docker,format,lint,metadata-validation,pur}: false
   # install Molecule in following environments:
-  {unit,functional}: true
+  {doc,unit,functional}: true
 whitelist_externals =
     find
 # Enabling sitepackages is needed in order to avoid encountering exceptions
--- a/tox.ini
+++ b/tox.ini
@@ -3,8 +3,7 @@ minversion = 3.7.0
 envlist =
     build-docker
     doc
-    format-check
-    format-enforce
+    format-{check,enforce}
     lint
     metadata-validation
     py{27,35,36,37}-ansible{25,26,27}-{functional,unit}
@@ -23,7 +22,7 @@ deps =
     ansible27: ansible>=2.7,<2.8
     ansibledevel: git+https://github.com/ansible/ansible.git
     build-docker: pbr
-    format: yapf>=0.25.0,<2
+    format-{check,enforce}: yapf>=0.25.0,<2
     {functional,unit}: -rtest-requirements.txt
     lint: -rlint-requirements.txt
     metadata-validation: collective.checkdocs
--- a/tox.ini
+++ b/tox.ini
@@ -1,44 +1,67 @@
 [tox]
 minversion = 3.7.0
 envlist =
-    lint
-    py{27,35,36,37}-ansible{25,26,27}-{functional,unit}
-    format-check
+    build-docker
     doc
+    format-check
+    format-enforce
+    lint
+    metadata-validation
+    py{27,35,36,37}-ansible{25,26,27}-{functional,unit}
 skipdist = True
 skip_missing_interpreters = True
 
 [testenv]
-usedevelop = True
+
 passenv = *
 setenv =
     PYTHONDONTWRITEBYTECODE=1
 deps =
-    -rrequirements.txt
-    -rtest-requirements.txt
     ansible24: ansible>=2.4,<2.5
     ansible25: ansible>=2.5,<2.6
     ansible26: ansible>=2.6,<2.7
     ansible27: ansible>=2.7,<2.8
     ansibledevel: git+https://github.com/ansible/ansible.git
+    build-docker: pbr
+    format: yapf>=0.25.0,<2
+    {functional,unit}: -rtest-requirements.txt
+    lint: -rlint-requirements.txt
+    metadata-validation: collective.checkdocs
+    metadata-validation: twine
 extras =
-    azure
-    docker
-    docs
-    ec2
-    gce
-    linode
-    lxc
-    openstack
-    vagrant
-    windows
+    {functional,unit}: azure
+    {functional,unit}: docker
+    doc: docs
+    {functional,unit}: ec2
+    {functional,unit}: gce
+    {functional,unit}: linode
+    {functional,unit}: lxc
+    {functional,unit}: openstack
+    {functional,unit}: vagrant
+    {functional,unit}: windows
 commands_pre =
     find {toxinidir} -type f -not -path '{toxinidir}/.tox/*' -path '*/__pycache__/*' -name '*.py[c|o]' -delete
 commands =
-    unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
+    doc: python setup.py build_sphinx -n -W --builder=html
+    format-check: yapf -d -r molecule/ test/
+    format-enforce: yapf -i -r molecule/ test/
     functional: pytest test/functional/ {posargs}
     lint: flake8
     lint: yamllint -s test/ molecule/
+    metadata-validation: python -m setup checkdocs check --metadata --restructuredtext --strict --verbose
+    metadata-validation: twine check .tox/dist/*
+    pur: pur -r requirements.txt
+    unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
+skip_install =
+  # install Python dist dependencies in following environments:
+  {functional,unit}: false
+  # don't install any Python dist dependencies in following environments:
+  {build-docker,doc,format,lint,metadata-validation,pur}: true
+usedevelop =
+  # don't install Molecule in following environments:
+  {build-docker,doc,format,lint,metadata-validation,pur}: false
+  # install Molecule in following environments:
+  {unit,functional}: true
 whitelist_externals =
     find
 # Enabling sitepackages is needed in order to avoid encountering exceptions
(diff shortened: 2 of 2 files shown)


decentral1se:
Hmmm, not sure. I find the separate testenvs more readable? Not willing to stop initiative here if the others find this to be more useful though. Let's see and thanks for the efforts to improve this!

themr0c:
@decentral1se that is exactly my point: have a discussion on the right direction to take. Experimenting here with "regrouping everything in testenv, tox.ini  now under 100 loc".

tox.ini
```diff
@@ -1,44 +1,67 @@
...
 deps =
-    -rrequirements.txt
-    -rtest-requirements.txt
     ansible24: ansible>=2.4,<2.5
     ansible25: ansible>=2.5,<2.6
     ansible26: ansible>=2.6,<2.7
     ansible27: ansible>=2.7,<2.8
     ansibledevel: git+https://github.com/ansible/ansible.git
+    build-docker: pbr
+    format: yapf>=0.25.0,<2
```
seandst: It's not clear to me how `format-check`/`format-enforce` actually end up getting `yapf` installed. Maybe define deps with format-{check,enforce} to make it explicit? Not sure what else that might break, though. Also `format-enforce` is great.

```diff
diff --git a/tox.ini b/tox.ini
index e81580db..386ca54c 100644
--- a/tox.ini
+++ b/tox.ini
@@ -3,8 +3,7 @@ minversion = 3.7.0
 envlist =
     build-docker
     doc
-    format-check
-    format-enforce
+    format-{check,enforce}
     lint
     metadata-validation
     py{27,35,36,37}-ansible{25,26,27}-{functional,unit}
@@ -23,7 +22,7 @@ deps =
     ansible27: ansible>=2.7,<2.8
     ansibledevel: git+https://github.com/ansible/ansible.git
     build-docker: pbr
-    format: yapf>=0.25.0,<2
+    format-{check,enforce}: yapf>=0.25.0,<2
     {functional,unit}: -rtest-requirements.txt
     lint: -rlint-requirements.txt
     metadata-validation: collective.checkdocs
```

tox.ini
```diff
@@ -1,44 +1,67 @@
...
 commands =
-    unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
+    doc: python setup.py build_sphinx -n -W --builder=html
+    format-check: yapf -d -r molecule/ test/
+    format-enforce: yapf -i -r molecule/ test/
     functional: pytest test/functional/ {posargs}
     lint: flake8
     lint: yamllint -s test/ molecule/
+    metadata-validation: python -m setup checkdocs check --metadata --restructuredtext --strict --verbose
+    metadata-validation: twine check .tox/dist/*
```
seandst: I think that once `deps` or `commands` for a testenv break out to multiple lines, a dedicated `testenv:section` should probably be used. I love all the collapsed one-line dep/command testenvs though.

tox.ini
themr0c: nice, adopted

tox.ini
```diff
@@ -1,44 +1,66 @@
...
 skip_missing_interpreters = True
 
 [testenv]
-usedevelop = True
+
 passenv = *
 setenv =
     PYTHONDONTWRITEBYTECODE=1
 deps =
-    -rrequirements.txt
```
webknjaz: Please keep this line

tox.ini
```diff
@@ -1,44 +1,66 @@
 [tox]
 minversion = 3.7.0
 envlist =
+    build-docker
```
webknjaz: This is a list of envs executed when a user runs `tox` without env spec. I'd say that it should be a minimal needed thing for them to verify that a few main things work rather than wait for hours for all of the unneeded things to complete.
So they definitely don't need to check whether docker build works because in 99% cases they won't touch it.

tox.ini
```diff
@@ -1,44 +1,66 @@
 [tox]
 minversion = 3.7.0
```
webknjaz: let's bump to have the newest bugfixes and a few nice improvements..
```suggestion
minversion = 3.8.1
```

test-requirements.txt
```diff
@@ -1,3 +1,4 @@
+-r requirements.txt
```
ssbarnea: I find this as an anti-pattern. Test requirements should not also include requirments.

While I find the refactoring of tox.ini very useful, I am afraid that the scope of the refactoring keeps growing and instead we should make smaller incremental improvements on it.

tox.ini
ssbarnea: Adding a requirement on tox version that was released 60 minutes ago is not safe.

Let things rest before bumping requirements to bleading edge releases.

seandst: I think there's no reason to distrust tox's release standards. Unless there's a specific problem that comes from using 3.8, we should feel free to track the latest tox. tox claims to follow semantic versioning, which we also have no reason to distrust. I'm not generally in favor of upgrading for upgrading's sake, but improvements to parallel testing in 3.8 would likely help get `tox .. -p auto` working.

seandst: Though, in line with the idea of incremental changes, it makes sense to only bump the tox minimums if tests are updated to require features introduced in higher versions.

tox.ini
seandst: If the focus is on the user here, would it make sense to limit this a little more, e.g.
```
envlist =
    doc 
    format-check
    lint
    metadata-validation
    py{27,35,36,37}-ansible{25,26,27}-unit
```

?

test-requirements.txt
webknjaz: I don't think that it's an anti-pattern in general. It depends on the use-case. In this case, it's probably fine for now since I'm going to fix dependency management along with packaging anyway...

tox.ini
webknjaz: It's only a patch release which fixed a minor bump. It doesn't matter when it was released, only proves that fixes get released fast.
And yes, tox's standards are high and current maintainers are very responsible.
The reason I want this update is that it ships certain features and bugfixes which I'd like to rely on.

tox.ini
webknjaz: Let's re-iterate:
1) Do users change docs often? Will they benefit on rebuilding docs all the time? Probably not.
2) format-check? It's probably worth it.
3) metadata-validation? Do they change packaging all the time? Probably not.
4) unit-tests? definitely yes.
5) functional tests? I think so.

seandst: Agreed on all counts. I left functional testing out because they don't "just work" for me without adding some pytest filtering. Functional testing should probably be in the default list of tests run, but not if those test runs require, say, working credentials for all drivers. I don't know if this is a common problem or only local to me, though.

seandst:
I'm going to attempt to sum up what's been discussed so far, please correct me if I missed something.

* The preference appears to be to have explicit `[testenv:*]` blocks.
* For consistency, existing one-command envs should probably be broken out into testenv blocks.
* The default envlist should probably be `format-check`, `lint`, and `py{27,35,36,37}-ansible{25,26,27}-{functional,unit}`, based on the discussion in https://github.com/ansible/molecule/pull/1887#discussion_r270032776. This effectively means leaving out `doc` by default and running the relatively inexpensive `format-check` before unit/functional tests.
* Adding `-r requirements.txt` to `test_requirements.txt` _is not_ preferred, specifying that explicitly via `deps` in the tox config _is_ preferred.
* Running functional testing is a pain right now, but @decentral1se's PR #1870 addresses those concerns and makes functional testing properly skip unsupported providers in local testing. Leaving functional testing in is good, in anticipation of #1870 being merged.
* While bumping the tox minversion is not something to be afraid of, nothing in this PR requires tox 3.8 features (that I can see, at least), so it is left alone.

After applying all of those bullets to the current master tox.ini, here's the diff:
```diff
diff --git a/tox.ini b/tox.ini
index 3e1a5580..c8ac1c58 100644
--- a/tox.ini
+++ b/tox.ini
@@ -2,9 +2,8 @@
 minversion = 3.7.0
 envlist =
     lint
-    py{27,35,36,37}-ansible{25,26,27}-{functional,unit}
     format-check
-    doc
+    py{27,35,36,37}-ansible{25,26,27}-{functional,unit}
 skipdist = True
 skip_missing_interpreters = True
 
@@ -34,11 +33,6 @@ extras =
     windows
 commands_pre =
     find {toxinidir} -type f -not -path '{toxinidir}/.tox/*' -path '*/__pycache__/*' -name '*.py[c|o]' -delete
-commands =
-    unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
-    functional: pytest test/functional/ {posargs}
-    lint: flake8
-    lint: yamllint -s test/ molecule/
 whitelist_externals =
     find
 # Enabling sitepackages is needed in order to avoid encountering exceptions
@@ -50,6 +44,9 @@ sitepackages = true
 [testenv:lint]
 deps =
     -rlint-requirements.txt
+commands =
+    flake8
+    yamllint -s test/ molecule/
 extras =
 skip_install = true
 usedevelop = false
@@ -72,7 +69,7 @@ usedevelop = false
 
 [testenv:doc]
 # doc requires py3 due to use of f'' strings and using only python3 as
-# basepython risks using python3.4 which is not supported.
+# basepython risks using python3.5, which does not support them.
 basepython = python3.7
 passenv = *
 commands =
@@ -119,3 +116,9 @@ commands =
     '
 whitelist_externals =
     sh
+
+[testenv:unit]
+commands = pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
+
+[testenv:functional]
+commands = pytest test/functional/ {posargs}
```

This results in a more focused diff that I think addresses the points discussed here.

I did throw in a quick comment change to the doc testenv basepython requirement to better reflect the current situation as I understand it.

webknjaz:
@seandst you've perfectly summarized it. I just want to add that I've got #1895 to merge first which will address some of the concerns like the very existence of `requirements.txt`.

tox.ini
```diff
@@ -47,63 +69,15 @@ whitelist_externals =
...
-commands =
-    yapf -d -r molecule/ test/
-deps = {[testenv:format]deps}
-extras = {[testenv:format]extras}
-skip_install = true
-usedevelop = false
-
-[testenv:doc]
+[doc]
+# basepython can't be put in the testenv section
```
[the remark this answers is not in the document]
ssbarnea: The reason why I do not support bumping tox too fast is because tox is a system requirement which means that if a new version of tox is released that breaks its use on one python based project and molecule is forcing me to use that breading-edge new version I may endup with a case where I cannot run tox with both projects.

To avoid this case we only need to be carefult to wait at least two weeks before using a newer version of tox. I am confident about tox is tested well but nobody can guarantee that a new release does not break anything.

The same issue does not apply for most other test requirements because they are installed inside virtualenvironments managed by tox, so no risk of conflict.

Shortly, lets bump it when we need new features but after a cooldown period so we avoid getting burned ;)

decentral1se:
> Running functional testing is a pain right now, but @decentral1se's PR #1870 addresses those concerns and makes functional testing properly skip unsupported providers in local testing. Leaving functional testing in is good, in anticipation of #1870 being merged.

Merged!

themr0c:
I guess it's better to close this PR and do a new one with the changes proposed in the wrapup https://github.com/ansible/molecule/pull/1887#issuecomment-478728006

themr0c:
@seandst are you ready to propose a new PR based on your wrapup, or how should we proceed?

seandst:
#1870 was a major step in the right direction, so I'll keep an eye on #1895 and put up a new PR once those changes get settled.

PR #1526, opened 2018-10-28
Add user arg for docker driver

logan2211:
Allow 'user' to be specified when providing arguments to the docker
driver.

Issue #1880, opened 2019-03-26
Improve module argument handling


decentral1se:
The current situation is that we are increasingly getting requests to add keys from our schema definition in the `molecule.yml` that correspond to the default Docker driver `create.yml`. See https://github.com/ansible/molecule/issues/1879 and https://github.com/ansible/molecule/issues/1878 for more recent examples. This stands for all the driver plays that we maintain.

The problem here is that we are duplicating the effort of what Ansible is doing. We are defining a schema using http://docs.python-cerberus.org/en/stable/ and validating what we get as input. Ansible already does this with the argspec validation. Our schema's can always be be behind in terms of what key/vals they offer ...

Following the idea from https://github.com/ansible/proposals/issues/39#issuecomment-467030338, I would propose the idea for discussion: we somehow move to replace the Cerberus schema definitions with Ansible argspec validation (yet to be finished for external use, but fine to discuss for now).

This would allow us to:
  * Not maintain schemas for Ansible module key/vals
  * Not maintain validation for these schemas
  * Rely on Ansible module argument validation all the way down ...

Questions:
  * Is it feasible to use this argspec validation API here in Molecule?
  * How to manage importing and multiple Ansible versions!?

I'm writing this a bit fast and off the cuff but hopefully it's clear ...

decentral1se:
Example of our Molecule schema: https://github.com/ansible/molecule/blob/master/molecule/model/schema_v2.py#L601. Example of what Ansible is doing: https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/cloud/docker/docker_container.py#L2859

rockandska:
Not really related as discuss on IRC but think it is complementary, please find an example [here](https://github.com/rockandska/repro-issues/tree/master/molecule/molecule/issue_1880) of how i see a docker molecule.yml who is:
- more readable by avoiding some repetitions regarding the registries / hosts / network parameters
- not stick to the args predefined in molecule, the user have a full access to the module args but molecule could override some parameters (like in the example). (solve #1879 , #1878 ...)
- images managed by molecule have a unique tag (uuid based on image description + dockerfile generated) and we'll be not rebuilt each time a test is launch. (solve #1847, should solve #1688, and partially #1573 )
- dynamic instances names is here but not functional since it is not the parameter "name" use with docker_container that is used in the inventory but "platforms[x].name", same problem if we want to add to the name an uuid corresponding to the current scenario ( partially solve #1573, #1704  )
[...]

PR #1889, opened 2019-03-29
Add user arg for docker driver

ssbarnea:
Allow 'user' to be specified when providing arguments to the docker driver.

Originally raised at https://github.com/ansible/molecule/pull/1526 but it got outdated and I was not able to rebase it there.

#### PR Type

- Feature Pull Request


ssbarnea:
@logan2211 Please check ^^

rockandska:
This PR is one among another concerning missing module arguments that i solved by only rewrite a create/destroy who could be found here https://github.com/ansible/molecule/issues/1880#issuecomment-477522088 if there's some people interesting to test it and comment

molecule/driver/docker.py
```diff
@@ -56,6 +56,7 @@ class Docker(base.Base):
                 username: $USERNAME
                 password: $PASSWORD
                 email: user@example.com
+			user: root
```
logan2211: Possible indentation issue?

ssbarnea:
Please review this.

ssbarnea:
> Please add a test for the schema validation.

Schema validation is going to be dropped as we cannot keep-up with docker module changes. Instead we will pass all arguments to docker modules, and making molecule more future proof.

test/unit/model/v2/test_platforms_section.py
```diff
@@ -140,6 +141,7 @@ def _model_platforms_docker_errors_section_data():
                 'exposed_ports': [bool()],
                 'published_ports': [int()],
                 'ulimits': [int()],
+                'user': str(),
```
decentral1se: Missing the test that checks that an incorect type is validated.

rockandska:
> None of that is set in stone and no one has the strategy in place or has commited to the time needed to make such a wide sweeping change. Therefore, that remains a "hope". We need to stick to the coverage / testing approach that is in place now.

Off topic:
@decentral1se / @ssbarnea : I submit a proposal with a working example in #1923 for this kind of implementation but it seems no one has tried it or take the time to review it

Regards,

PR #1890, opened 2019-03-29
Remove manual version checks for Ansible/Python. Rely on dist metadata.

decentral1se:
#### PR Type
- Bugfix Pull Request

As detailed in https://github.com/ansible/molecule/pull/1883#issuecomment-477924577, let's just rely on dist metadata for telling the user if they have mismatching versions of Python or Ansible when using Molecule. Therefore, I remove all the manual version checking code and add a `ansible>=2.5,<3` constraint via `pbr`. We already specify Python versioning with `requires-python`.

### DIFF
--- a/molecule/shell.py
+++ b/molecule/shell.py
@@ -18,18 +18,13 @@
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
 
-import distutils
-import distutils.version
 import os
-import sys
 
-import ansible
 import click
 import click_completion
 
 import molecule
 from molecule import command
-from molecule import util
 from molecule.config import MOLECULE_DEBUG
 
 click_completion.init()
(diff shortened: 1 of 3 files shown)


molecule/shell.py
```diff
@@ -38,58 +33,10 @@
...
-    _supported_ansible_version()
-
-    return value
-
-
 @click.group()
 @click.option(
     '--debug/--no-debug',
     default=MOLECULE_DEBUG,
-    callback=_allowed,
```
webknjaz: Oh, I didn't realize that it's invoked on debug... If you want to preserve similar behavior you could probably invoke `pip check` here. Given that you *want* it.

requirements.txt
```diff
@@ -1,4 +1,5 @@
 -r lint-requirements.txt
+ansible>=2.5,<3
```
webknjaz: I'm not sure whether we want to set the higher limit + Ansible doesn't follow semver: the change in the second number in version designates a major change. So 2.6 and 2.7 are major versions while 2.6.1 would be minor and so on.

I think for apps like Molecule it's more sustainable in the long run to use an open higher limit.

webknjaz: So I'd just do
```suggestion
ansible>=2.5
```

tox.ini
```diff
@@ -16,7 +16,6 @@ setenv =
 deps =
     -rrequirements.txt
     -rtest-requirements.txt
-    ansible24: ansible>=2.4,<2.5
```
webknjaz: Good catch!

molecule/shell.py
decentral1se: I'm fine to follow the logic of "let dist metadata solve this".

requirements.txt
decentral1se: Fair enough, will amend!

webknjaz:
Tests failed probably because of the version conflict as pip might get confused when there's a few entries for the same dist because `tox` uses requirements + extra spec for this. It will be gone once the packaging will be done right and runtime deps are specified in the correct place (`setup.cfg`).

So I suggest freezing this for a bit to get packaging improvements on board first.

decentral1se:
Ah right then. Will keep this in mind then for later, nice one :+1:

webknjaz:
@decentral1se  You'll need to rebase this once #1895 is merged

decentral1se:
Awesome @webknjaz, thanks, I'll keep an eye and get that done.

PR #1894, opened 2019-04-01
Remove context-less comment from test requirements

seandst:
I went looking through the git history to try to better understand what this is and why it's here, but honestly did not spend a lot of time on that. If this comment should be expended to provide meaningful context, let me know. Otherwise, it looks like the sort of thing that should be deleted.

#### PR Type

- Remove a Comment With No Context Pull Request


seandst:
And of course I forgot to sign-off. :)

webknjaz:
@seandst it's why the next line is locked AFAIR. But let's forget about it for now as my PR #1895 includes this removal.

seandst:
Awesome!

Issue #1891, opened 2019-03-29
Learn to write good commit message and description


webknjaz:
It's a follow-up for:
> @decentral1se @ssbarnea I've got a collection of articles about writing meaningful commit messages. I'll post it as a separate issue so that anyone could go through this and we could be on the same page...
Sounds fair?

_Originally posted by @webknjaz in https://github.com/ansible/molecule/pull/1883#issuecomment-477972258_

---------------------

                                                 ![](http://software-engineer-training.com/wp-content/uploads/2009/08/overlord59-tux-samourai.png)

According to common commit style guides, it is required to write messages in an imperative manner. Make them actionable and atomic. Also **avoid using `-m`** as you often need to express more detail in a long way. Here's several articles sharing best practices:
* http://chris.beams.io/posts/git-commit/
* http://alistapart.com/article/the-art-of-the-commit
* https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message
* https://github.com/erlang/otp/wiki/writing-good-commit-messages
* https://wiki.openstack.org/wiki/GitCommitMessages
* http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
* https://dev.to/gonedark/when-to-make-a-git-commit

N.B. This all applies to issues and PRs as well. If you submit a patch, everything in diff should refer to only one logical change making it atomic. Ideally, applying a single PR should transition a project from one working state to another completely working state, meaning there shouldn't be a change submitted via two PRs and there shouldn't be anything more than needed to fulfill what's in description and title of the PR.

P.S. Don't scare (other) maintainers with huge changes. Reviewing a PR is hard work. The bigger it is the more chance is that somebody will postpone it until better times (trying to find an appropriate slot in their calendars to the review in one pass).
Smaller changes attract more viewers and people, in general, are more likely to understand tiny patches and their impact.

webknjaz:
There's also a number of things I encourage people to learn about Git. Maybe you will find something for yourselves.

![](http://2.darkroom.shortlist.com/1200/5cda8b34a6d9c6dc2c088ac50354347c:1f86c8a88fce5865c68734c9b6e73b34/samuraiimagehero.jpg)

* [ProGit 2](https://git-scm.com/book/en/v2) [:ukraine:](https://git-scm.com/book/uk/v2)
* [Try Git](https://try.github.io/)
* [Try Git Branching](http://learngitbranching.js.org/)
[...]

PR #1895, opened 2019-04-01
Change packaging toolkit to use setuptools-scm

webknjaz:
#### PR Type

- Feature Pull Request


tox.ini
```diff
@@ -14,7 +14,6 @@ passenv = *
 setenv =
     PYTHONDONTWRITEBYTECODE=1
 deps =
-    -rrequirements.txt
     -rtest-requirements.txt
     ansible24: ansible>=2.4,<2.5
```
decentral1se: Oh, mind removing this too, while we're here :smile:

molecule/__init__.py
```diff
@@ -17,8 +18,13 @@
 #  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
+"""Ansible-lint version information."""
```
decentral1se: Uh, wrong name?

molecule/__init__.py
```diff
@@ -17,8 +18,13 @@
...
 
-import pbr.version
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
 
-version_info = pbr.version.VersionInfo('molecule')  # noqa
-__version__ = version_info.release_string()
+try:
+    import pkg_resources
+    __version__ = pkg_resources.get_distribution('ansible-lint').version
```
decentral1se: `%s/ansible-lint/ansible-molecule/`

setup.py
```diff
@@ -17,11 +19,156 @@
...
 
 import setuptools
 
-setuptools.setup(
-    packages=setuptools.find_packages(),
-    pbr=True,
-    setup_requires=['pbr']
-)
+
+try:
```
decentral1se: This all looks quite bleeding edge ... what are the risks here? Is this standard practice right now?

tox.ini
```diff
@@ -103,19 +103,50 @@ usedevelop = False
...
+
+
+[testenv:build-dists-local]
+description =
+    Generate dists which may be not ready
+    for upload to PyPI because of
+    containing PEP440 local version part
+# `usedevelop = true` overrides `skip_install` instruction, it's unwanted
+usedevelop = false
+# don't install octomachinery itself in this env
```
decentral1se: `octomachinery`?

tox.ini
decentral1se: Ah, can also do it as part of https://github.com/ansible/molecule/issues/1896 (which was where I orginally caught it), if needed.

molecule/__init__.py
webknjaz: oh.. :)

setup.py
webknjaz: It's a compatibility shim. Basically, I just emulate what newer setuptools have out of the box. So that ancient installations wouldn't explode. I do this in a few projects I maintain + `ansible-lint`.

webknjaz: And new installations won't hit it at all. Because we provide a wheel which their pip will just unpack. So it only affects ancient installations in which pip doesn't know about preferring wheels

tox.ini
webknjaz: Yeah, maybe. I'll need to stabilize CI first

tox.ini
webknjaz: :D 
that's from my frameworks for writing GitHub Apps and Actions...
just artifacts of copy-pasting from different places. i'll clean that up

molecule/__init__.py
webknjaz: @decentral1se you know you can send Suggested Changes, right?
https://thenextweb.com/dd/2018/10/16/github-launches-suggested-changes-to-make-it-easier-to-collaborate-within-pull-requests/

P.S. It's just `molecule`, it's what is specified in dist metadata...

decentral1se: I promise I tried clicking it a bunch of times but it wasn't working ... ;)

setup.py
decentral1se: OK, thanks!

webknjaz:
@ssbarnea `pbr` itself is a bug. We've been talking about it a lot already. If you want a specific case: it currently generates a totally broken version number from Git tag. And replacing it with something actually working was a public plan for months.

Back to why CI is failing now: we are hitting https://github.com/pypa/pip/issues/6264 now because you've enabled `--system-site-packages` in tox. I'm trying to make it more resilient, though.

`setuptools-scm` is battle-tested and `pbr` is broken by design. I've used both. Does anyone here want to argue about it based on _actual_ experience?

gundalow:
> As long we can make molecule-with-setuptools-scm work on all platforms where molecule would be used (without having to add extra documentation to ask people to "tune" their systems) I am ok with it.

Do we have a full list of all platforms which we can test this PR against?

tox.ini
```diff
@@ -81,8 +93,11 @@ extras =
     docs
 
 [testenv:pur]
+deps =
+    pur==5.2.1
 commands=
-    pur -r requirements.txt
+    pur -r lint-requirements.txt
+    pur -r test-requirements.txt
```
seandst: These commands fail now that `lint-requirements.txt` and `test-requirements.txt` no longer exist.

ssbarnea:
Without having a full list I think this makes sense:
* CentOS 7
* Fedora 29 (this is base for RHEL 8/CentOS 8)
* Fedora Rawhide
* Ubuntu latest stable

On these it need testing with:
- without virtualenv
- with isolated virtualenv (not sure if it will work all the way due to missing selinux)
- with virtualenv with site-packages

Checkpoints:
- install passes
- pip check does not report conflicts
- molecule can be called and is not missing files like templates

tox.ini
```diff
@@ -49,7 +60,8 @@ sitepackages = true
 
 [testenv:lint]
 deps =
-    -rlint-requirements.txt
+    flake8>=3.6.0,<4
+    yamllint>=1.11.1,<2
```
seandst: The `lint` and `flake8` deps are already in the `testenv.deps` list, so one set of them is redundant and should be removed either from `testenv.deps` or `testenv:lint.deps`, as appropriate.

seandst:
The multi-platform testing matrix would be useful for testing in any case, both for `pbr` and for `setuptools-scm`. That effort would probably warrant its own issue for discussion and a separate pull request.

tox.ini
```diff
@@ -1,22 +1,33 @@
...
+
+    mock==2.0.0
+    pytest==3.6.0
+    pytest-cov==2.6.0
+    pytest-helpers-namespace==2019.1.8
+    pytest-mock==1.10.0
+    pytest-verbose-parametrize==1.4.0
+    pytest-xdist==1.26.0
+    shade==1.22.2
+    yapf>=0.25.0,<2
```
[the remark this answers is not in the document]
webknjaz: It conflicts with the current use of isolated build feature.

tox.ini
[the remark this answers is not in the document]
webknjaz: it's related to software conflicts which lead me to removing requirements file (which is not needed as having a unified place for deps in preferred.

tox.ini
```diff
@@ -103,19 +118,51 @@ usedevelop = False
...
+    pep517 >= 0.5.0
+setenv =
+commands =
+    python -m pep517.build \
+      --source \
+      --binary \
+      --out-dir {toxinidir}/dist/ \
+      {toxinidir}
+
+[testenv:build-dists]
```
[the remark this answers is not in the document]
webknjaz: It's for building things, not deploying. If you want publishing code, we could add another env and use now supported feature of env dependencies for that. But currently publishing is not done by tox.
And also tox envs are mostly bound to Travis jobs

webknjaz:
> As long we can make molecule-with-setuptools-scm work on all platforms

That's what I'm working on now. We'll have a shim emulating missing features, just like I did it for `ansible-lint`. That's why I set WIP status on this PR. I've hit a few bugs with new tox+setuptools+pip but soon I'll be able to address those.

tox.ini
webknjaz: I thought so as well, but it turned out that these are also runtime deps of Molecule and are used not only for linting of the Molecule itself but also in scenarios and tests of those scenarios. We can clean up this mess later in further refactoring. I just want to focus on finishing with the packaging stuff.

webknjaz:
> The multi-platform testing matrix

If you are suggesting more test jobs, they can be added to Travis to be run on cron or we can use other resources (like GitHub Actions) just for that.

tox.ini
webknjaz: Oh.. I forgot to remove this env. Long-term we'll just enable pyup or dependabot, it's not a human job.

setup.cfg
```diff
@@ -20,28 +15,80 @@ project_urls =
     Documentation = https://molecule.readthedocs.io
     Mailing lists = https://docs.ansible.com/ansible/latest/community/communication.html#mailing-list-information
     Source Code = https://github.com/ansible/molecule
-classifier =
-    Development Status :: 4 - Beta
```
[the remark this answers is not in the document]
webknjaz: I think it's kinda related to packaging + I've agreed this with @gundalow. Anyway, my main focus now is to patch the shim properly.

seandst:
> If you are suggesting more test jobs, they can be added to Travis to be run on cron or we can use other resources (like GitHub Actions) just for that.

Yeah, I'm mainly suggesting that while more test jobs is a good idea, I don't think it's something that should block this pull request.

decentral1se:
> That effort would probably warrant its own issue for discussion and a separate pull request.

Exactly. Let's move this discussion off this PR and focus on supporting with reviews etc.

molecule/__init__.py
```diff
@@ -17,8 +18,13 @@
 #  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
+"""Molecule version information."""
 
-import pbr.version
+from __future__ import absolute_import, division, print_function
```
decentral1se:
```suggestion
from __future__ import absolute_import print_function
```

?

setup.cfg
```diff
@@ -20,28 +15,80 @@ project_urls =
...
+    python-gilt >= 1.2.1, < 2
+    Jinja2 == 2.10
+    pexpect >= 4.6.0, < 5
+    psutil == 5.4.6; sys_platform!="win32" and sys_platform!="cygwin"
+    PyYAML == 3.13
+    sh == 1.12.14
+    six == 1.11.0
+    tabulate == 0.8.2
+    testinfra == 1.19.0
+    tree-format==0.1.2
```
decentral1se:
```suggestion
    tree-format == 0.1.2
```

setup.cfg
```diff
@@ -65,19 +112,14 @@ vagrant =
...
-[pbr]
-skip_authors = True
-skip_changelog = True
-warnerrors = True
-
-[entry_points]
+[options.entry_points]
 console_scripts =
-    molecule = molecule.shell:main
+    molecule = molecule.__main__:main
```
decentral1se: Didn't see `molecule.shell` renamed to `molecule.__main__` in this PR?

setup.cfg
```diff
@@ -65,19 +112,14 @@ vagrant =
...
+
+[options.packages.find]
+where = .
 
 [build_sphinx]
 all_files = 1
 build-dir = doc/build
 source-dir = doc/source
```
decentral1se: We can also add `warning-is-error = True` here too?

setup.py
```diff
@@ -17,11 +19,194 @@
...
+    class ShimConfigOptionsHandler(ConfigOptionsHandler):
+        """Extension class for ConfigOptionsHandler."""
+
+        @property
+        def parsers(self):
+            """Return an option mapping with default data type parsers."""
+            _orig_parsers = super(ShimConfigOptionsHandler, self).parsers
+            return dict(use_scm_version=self._parse_bool, **_orig_parsers)
+
+        def parse_section_packages__find(self, section_options):
```
decentral1se:
```suggestion
        def parse_section_packages_find(self, section_options):
```

?

setup.py
```diff
@@ -17,11 +19,194 @@
...
+        from ConfigParser import ConfigParser, NoSectionError
+        ConfigParser.read_file = ConfigParser.readfp
+
+    def maybe_read_files(d):
+        """Read files if the string starts with `file:` marker."""
+        d = d.strip()
+        if not d.startswith('file:'):
+            return d
+        descs = []
+        for fname in map(str.strip, str(d[5:]).split(',')):
```
decentral1se:
```suggestion
        for fname in map(str.strip, str(d[MARKER_LENGTH:]).split(',')):
```

?

setup.py
```diff
@@ -17,11 +19,194 @@
...
+            with io.open(fname, encoding='utf-8') as f:
+                descs.append(f.read())
+        return ''.join(descs)
+
+    def cfg_val_to_list(v):
+        """Turn config val to list and filter out empty lines."""
+        return list(filter(bool, map(str.strip, str(v).strip().splitlines())))
```
decentral1se:
```suggestion
        return list(filter(None, map(str.strip, str(v).strip().splitlines())))
```

Nitpick! I think it's more readable but maybe you have good reason ..

setup.py
```diff
@@ -17,11 +19,194 @@
...
+    def cfg_val_to_dict(v):
+        """Turn config val to dict and filter out empty lines."""
+        return dict(
+            map(lambda l: list(map(str.strip, l.split('=', 1))),
+                filter(bool, map(str.strip, str(v).strip().splitlines())))
+        )
+
+    def cfg_val_to_primitive(v):
+        """Parse primitive config val to appropriate data type."""
+        return json.loads(v.strip().lower())
```
decentral1se: Should probably be wrapped with a `try/except` block?

setup.py
```diff
@@ -17,11 +19,194 @@
...
 import setuptools
+import setuptools.command.dist_info
 
-setuptools.setup(
-    packages=setuptools.find_packages(),
-    pbr=True,
-    setup_requires=['pbr']
-)
+
+ALL_STRING_TYPES = tuple(map(type, ('', b'', u'')))
```
decentral1se: Just to note, this would all be very scary and signal some major maintenance headaches but it seems to be doing just fine over at https://github.com/ansible/ansible-lint/blob/master/setup.py! So, please take that into consideration when reviewing. Maybe this shim itself needs to be packaged up ... ;)

tox.ini
```diff
@@ -1,22 +1,33 @@
...
 skipdist = True
 skip_missing_interpreters = True
+isolated_build = True
 
 [testenv]
 usedevelop = True
 passenv = *
 setenv =
     PYTHONDONTWRITEBYTECODE=1
 deps =
```
decentral1se: Is using a `test` entry in `[options.extras_require]` an option in replacement for this? Then we can rely on a single source file for the package listing. I've been doing this a little in my recent experiments and it works quite nicely ... it could be an abuse of the `extra_requires` though ...

setup.py
```diff
@@ -17,11 +19,194 @@
...
+            )
+            opt['packages'] = setuptools.find_packages(**opt_packages_find)
+        return {'metadata': md, 'options': opt}
+
+
+def cut_local_version_on_upload(version):
+    """Generate a PEP440 local version if uploading to PyPI."""
+    import os
+    import setuptools_scm.version  # only present during setup time
+    IS_PYPI_UPLOAD = os.getenv('PYPI_UPLOAD') == 'true'
```
decentral1se: Perhaps comment that this is set from the `tox.ini` to help later reviews?

setup.cfg
webknjaz: It's not renamed. It's just a common entry point for invoking it `python -m molecule` style. Just wanted to wire it through the same invocation point.

setup.cfg
webknjaz: Yes, but it's not related to this packaging change. So you can send another PR if you want.

setup.py
webknjaz: No, this must match the parent class:
https://github.com/pypa/setuptools/blob/7db0102/setuptools/config.py#L587-L606

setup.py
webknjaz: Actually, this wouldn't work AFAICS

webknjaz: Oh, I didn't know about `None` special meaning here...

And yet, I'd argue that it's less readable. Callable is expected there by an average reader.
With `bool` it follows "explicit is better than implicit" rule.

setup.py
webknjaz: No. It should fail on any errors. And again, it's not a full-featured implementation, it's just a shim. If it fails, it shouldn't hide this. Because in this case it'll eat all warning signs and we'll ship a broken sdist.

setup.py
webknjaz: This shim is mostly copy-pasted across projects so I don't really want to change it a lot...

setup.py
webknjaz: I'm thinking that I'll maybe extend the shim to support `attr: ` in `name` and will re-wire version generator as I did in another recent project of mine: https://github.com/sanitizers/octomachinery/blob/ce217ba/setup.cfg#L9.
This seems to be more explicit. Let's see if I can do it...

tox.ini
webknjaz: Yes, I usually do this as well. Calling it `testing` though: https://github.com/cherrypy/cheroot/blob/92490dd/setup.cfg#L83-L104.

There's a downside to this, though. You cannot install extras only without installing a dist itself. It's one thing I've been discussing with one of the tox maintainers in October... It seems like this should be implemented on the pip/setuptools side.

webknjaz: I'd prefer to postpone this change and do in in a separate tox refactoring PR.

setup.py
webknjaz: Nope, you cannot package this shim. Just because of a chicken/egg problem :)
Of course, you could do even more nasty stuff but I urge you not to.

The right thing to do is to actually upgrade pip/setuptools/python to something better maintainable and if future you can even expect something as magical as https://twitter.com/webKnjaZ/status/1113530472808763394

webknjaz: ** But I was thinking about making a repo with an example like this...

webknjaz: Oh, and this version is way more improved over the one used in ansible-lint...

.travis.yml
```diff
@@ -429,7 +429,9 @@ jobs:
     name: Publishing current Git tagged version of dist to PyPI
     if: repo == "ansible/molecule" AND tag IS present
     env:
-      TOXENV: metadata-validation
+      TOXENV: metadata-validation,build-dists
+    before_deploy:
+    - echo > setup.py
```
webknjaz: FTR: deployment provider calls `setup.py` no matter what, but we build things manually so I'm wiping its contents and it will not affect pre-created dists this way but will still use twine to upload things to PyPI :)

(also a battle-tested hack)

molecule/__init__.py
webknjaz: It's just a snippet used across Ansible. I don't want to customize it.

setup.py
webknjaz: Turned out to be more complicated than I thought... Maybe another time.

webknjaz:
Please ignore linter job failure. It's unrelated: yapf got updated within limits which brought us its new view on code formatting. Let's deal with it separately.

webknjaz:
FTR: interesting things about compat — https://python3statement.org/practicalities/

webknjaz:
Blocking this for a bit, because wheels don't get package data included for some reason.

ssbarnea:
> # Please ignore linter job failure. It's unrelated: yapf got updated within limits which brought us its new view on code formatting. Let's deal with it separately.

Please fix linting as it does make no sense to review code that does not pass CI. 

I do not see master as failing and I was not able to replicate any yapf related issue with master code (even with latest 0.26). 

Feel free to make a preparatory change related to yapf, if needed.

seandst:
> I do not see master as failing and I was not able to replicate any yapf related issue with master code (even with latest 0.26).

0.26 is not the latest, 0.27 is.

I was able to reproduce the failures on master:
```
$ git checkout master
$ git pull upstream master
$ pip install -U yapf
$ yapf -d -r molecule/ test/
<observe tons of format changes>
```

The fixes to this don't belong in this PR, imo.

webknjaz:
> I do not see master as failing

I do:
* https://travis-ci.com/ansible/molecule/jobs/191637217
* https://travis-ci.com/ansible/molecule/jobs/191645608
* https://travis-ci.com/ansible/molecule/jobs/191470671
* https://travis-ci.com/ansible/molecule/jobs/191464880
* https://travis-ci.com/ansible/molecule/jobs/191324757
* https://travis-ci.com/ansible/molecule/jobs/191001247

webknjaz:
> Feel free to make a preparatory change related to yapf

Nope. I'm not raising the number of changed files from 12 to 100 here. Sorry. As per explanation @ #1891 I refuse to do non-atomic changes.

seandst:
> > 
> 
> Nope. I'm not raising the number of changed files from 12 to 100 here. Sorry. As per explanation @ #1891 I refuse to do non-atomic changes.

Also, as #1911 demonstrates, making those changes would break lint. It's a larger problem, now captured in #1916.

webknjaz:
It seems like everyone who wanted to comment on this change already did it. So I'm merging it with red CI because master is broken and this change doesn't bring it new failures.

decentral1se:
Great.

PR #1897, opened 2019-04-02
Adopt use of module invocation method for tools

ssbarnea:
Avoid annoying warnigns raised by tox about externals when sitepackages
is enabled, mainly because these tools where already found installed
as system packages.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -52,7 +52,6 @@ commands =
     lint: python -m yamllint -s test/ molecule/
 whitelist_externals =
     find
-    rm
 # Enabling sitepackages is needed in order to avoid encountering exceptions
 # caused by missing selinux python bindinds in ansible modules like template.
 # Selinux python bindings are binary and they cannot be installed using pip
@@ -157,3 +156,6 @@ setenv =
 commands =
     rm -rfv {toxinidir}/dist/
     {[testenv:build-dists-local]commands}
+whitelist_externals =
+    rm
+    {[testenv]whitelist_externals}
--- a/tox.ini
+++ b/tox.ini
@@ -46,12 +46,13 @@ extras =
 commands_pre =
     find {toxinidir} -type f -not -path '{toxinidir}/.tox/*' -path '*/__pycache__/*' -name '*.py[c|o]' -delete
 commands =
-    unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
-    functional: pytest test/functional/ {posargs}
-    lint: flake8
-    lint: yamllint -s test/ molecule/
+    unit: python -m pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
+    functional: python -m pytest test/functional/ {posargs}
+    lint: python -m flake8
+    lint: python -m yamllint -s test/ molecule/
 whitelist_externals =
     find
+    rm
 # Enabling sitepackages is needed in order to avoid encountering exceptions
 # caused by missing selinux python bindinds in ansible modules like template.
 # Selinux python bindings are binary and they cannot be installed using pip
@@ -68,7 +69,7 @@ usedevelop = false
 
 [testenv:format]
 commands =
-    yapf -i -r molecule/ test/
+    python -m yapf -i -r molecule/ test/
 deps = yapf>=0.25.0,<0.27  # pyup: < 0.27 # disable updates, conflicts with flake8 as per https://github.com/ansible/molecule/pull/1915
 extras =
 skip_install = true
@@ -76,7 +77,7 @@ usedevelop = false
 
 [testenv:format-check]
 commands =
-    yapf -d -r molecule/ test/
+    python -m yapf -d -r molecule/ test/
 deps = {[testenv:format]deps}
 extras = {[testenv:format]extras}
 skip_install = true
@@ -100,7 +101,7 @@ usedevelop = False
 # Ref: https://twitter.com/di_codes/status/1044358639081975813
 commands =
     python -m setup checkdocs check --metadata --restructuredtext --strict --verbose
-    twine check .tox/dist/*
+    python -m twine check .tox/dist/*
 
 [testenv:build-docker]
 # skip under Windows
@@ -156,6 +157,3 @@ setenv =
 commands =
     rm -rfv {toxinidir}/dist/
     {[testenv:build-dists-local]commands}
-whitelist_externals =
-    rm
-    {[testenv]whitelist_externals}


decentral1se:
> I assume this suggests using things outside of tox-managed env which is wrong.

But I thought we'd already gone down that path with https://tox.readthedocs.io/en/latest/config.html#conf-sitepackages and this is only reducing warnings because of this (I don't make any comment whether this is "right" but it makes sense based on the existing tox configuration).

And it wasn't removed in https://github.com/ansible/molecule/pull/1895? See https://github.com/ansible/molecule/pull/1895/files#diff-b91f3d5bd63fcd17221b267e851608e8R59. Oh wait, that selinux bug ...

webknjaz:
@decentral1se that's just to facilitate selinux but I think that relying on something outside of tox envs (except for POSIX binaries like `rm`) is bad. If we say "tox doesn't manage this", install it yourself, then we have no way of controlling what user will set up and they will get back to us asking to "fix" their 100% custom env just because they have a different global flake8 version or so.
Why even have tox if we have to manually set up each bit outside of it? Also, encouraging people to have something global for all their projects doesn't sound right either.
Currently, tox is a glue for standardizing everything and it should remain this way. I still think that if users want site-packages, they'd better use a CLI arg for that. This feature comes with a price and even tox maintainers were talking about the possibility to remove it...

webknjaz:
We need to check whether we can inject something like `--force-reinstall` into the `deps` section...

ssbarnea:
@webknjaz We cannot inject force reinstall and even if we could I would be against as I do not want to reinstall deps on each execution. The lack of dependency refresh on tox is a well known issue which is *not* limited to use of system site packages. Lucky for us the current maintainer started to work on a solution but it will take many months before we would be able to consuime it (it is a very complex issue related to tox design).

For the moment I see no real reason for not merging this as the only side effect it has is to avoid some really annoying warnings.

webknjaz:
@ssbarnea solve this with `python -m` style invocations then.

ssbarnea:
@webknjaz I used the `python -m` for years, spending a lot of time explaining people why was better or making PRs on lots of tools in order to implement module calling convention (you cannot guess how many authors are totally unaware of this feature).

Still, `pur` cannot be called as a module and worse it was somethign I read yesterday on https://discuss.python.org/t/pep-582-python-local-packages-directory/963/11 probbly which said that current implementation of `-m` calling method does a lookup in current folder first, which means that if you happen to have some files/modules there with the same name they will be loaded instead of the tool you want. Never happened to me but.

To summarize, I will use module calling whenever is possible.

webknjaz:
@ssbarnea local folder lookup is kinda expected and I use it all the time. You can even use `python -m .` which is pretty handy. I don't see any problem here.

Oh, and `pur` has been removed. I was hoping to go forward now and enable some bot for that.

tox.ini
```diff
@@ -46,12 +46,16 @@ extras =
...
-    lint: yamllint -s test/ molecule/
+    unit: python -m pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
+    functional: python -m pytest test/functional/ {posargs}
+    lint: python -m flake8
+    lint: python -m yamllint -s test/ molecule/
+# avoid possible undesired warnings when used with site-packages:
 whitelist_externals =
     find
+    # https://github.com/alanhamlett/pip-update-requirements/issues/24
+    pur
```
webknjaz: This one is not relevant anymore
```suggestion
    pur
```

tox.ini
```diff
@@ -46,12 +46,16 @@ extras =
...
-    lint: flake8
-    lint: yamllint -s test/ molecule/
+    unit: python -m pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
+    functional: python -m pytest test/functional/ {posargs}
+    lint: python -m flake8
+    lint: python -m yamllint -s test/ molecule/
+# avoid possible undesired warnings when used with site-packages:
 whitelist_externals =
     find
+    # https://github.com/alanhamlett/pip-update-requirements/issues/24
```
```suggestion
```

tox.ini
```diff
@@ -46,12 +46,16 @@ extras =
...
 commands =
-    unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
-    functional: pytest test/functional/ {posargs}
-    lint: flake8
-    lint: yamllint -s test/ molecule/
+    unit: python -m pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
+    functional: python -m pytest test/functional/ {posargs}
+    lint: python -m flake8
+    lint: python -m yamllint -s test/ molecule/
+# avoid possible undesired warnings when used with site-packages:
```
webknjaz: This is also not needed with `pur` removal

tox.ini
```diff
@@ -46,12 +46,13 @@ extras =
...
-    functional: pytest test/functional/ {posargs}
-    lint: flake8
-    lint: yamllint -s test/ molecule/
+    unit: python -m pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
+    functional: python -m pytest test/functional/ {posargs}
+    lint: python -m flake8
+    lint: python -m yamllint -s test/ molecule/
 whitelist_externals =
     find
+    rm
```
webknjaz: If you add it here, remove `whitelist_externals` in the last section.

webknjaz: By the way, what causes a call to `rm` here?

ssbarnea: `build-dists` does use `rm`.

ssbarnea: But true, if I fix it I fix it completly. Already pushed new version.
BTW, pur already released a fix after I raised the bug https://github.com/alanhamlett/pip-update-requirements/issues/24 --- we no longer need it but it was really nice have it fixed and released so quickly.

tox.ini
```diff
@@ -123,9 +125,6 @@ commands =
         --pull --rm \
         {posargs:-t ansible/molecule:$TAG} .\
     '
-whitelist_externals =
-    sh
```
webknjaz: Why did you do this?

webknjaz: Only this env needs `sh` so there's no need to move it to the global one.

ssbarnea: To simplify setup. What is the great feature of keeping these in several places? ... unless we do take a special pleasure in having more complex tox.ini files. Clearly is not related to security or so, in the end any code change would need to be reviewed.

webknjaz: This is related to envs being obvious and avoiding a godobject antipattern.

When you look at the env you shouldn't have to look up all the other envs to guess what else may influence it. Explicit is better than implicit.

Also, dumping everything into one env makes it unreadable/unmaintainable as I've explained somewhere else already.

webknjaz:
Hey, I've changed my mind (again, sorry) about `whitelist_externals`.
`rm` is only used in `build-dists` and shouldn't be put into the base env therefore.

There's been some refactoring by @themr0c where I explained why it should move in this direction.

PR #1899, opened 2019-04-02
Execution speed improvement for functional tests

themr0c:
Some rework on tests to enable performance and performance measurements.

* improve measurements of performance with profiling callbacks
* improve tests performance by disabling gather_facts
* improve tests performance by replacing commend module by debug module

#### PR Type

- Feature Pull Request

### DIFF
--- a/test/scenarios/host_group_vars/molecule/default/molecule.yml
+++ b/test/scenarios/host_group_vars/molecule/default/molecule.yml
@@ -16,6 +16,9 @@ platforms:
       - example_1
 provisioner:
   name: ansible
+  config_options:
+    defaults:
+      callback_whitelist: profile_roles,profile_tasks,timer
   playbooks:
     create: ../../../../resources/playbooks/docker/create.yml
     destroy: ../../../../resources/playbooks/docker/destroy.yml
--- a/test/resources/roles/molecule/tasks/main.yml
+++ b/test/resources/roles/molecule/tasks/main.yml
@@ -6,9 +6,9 @@
         mode=0755
         state=directory
 
-- name: Create /etc/molecule/{{ ansible_hostname }}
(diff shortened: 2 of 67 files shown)


test/scenarios/host_group_vars/molecule/default/molecule.yml
```diff
@@ -16,6 +16,9 @@ platforms:
       - example_1
 provisioner:
   name: ansible
+  config_options:
```
decentral1se: Note to ourselves: let's make a `molecule.yml` in `test/functional/scenarios` which holds this default configuration and change `molecule` invocations to include `--base-config ...`.

decentral1se:
> Note to ourselves: let's make a molecule.yml in test/functional/scenarios which holds this default configuration (config_options: ...) and change molecule invocations to include --base-config ... so we do not have to repeat ourselves like this.

As an idea on how to do this, again, doesn't have to be this PR - we could add a 'default options' fixture somewhere around https://github.com/ansible/molecule/blob/master/test/functional/conftest.py#L46 and pass it into every test with the `base_config` options set to the correct path. We could then change the tests from `options = { ...` to `options['...']` etc. This would help us set more general purpose options across the entire test suite.

decentral1se:
Ah, I did introduce a bug in https://github.com/ansible/molecule/pull/1871/files#diff-8a0941ded1a1bf5411a8a1e5562a82d7R6 which is fixed here. 

We're skipping all those tests, so I missed it :sweat_smile:

themr0c:
* Removed from PR some deeper changes that should not happen right now and did break the builds.
* Squashed everything in a  single commit with a nicer comment.
* Didn't fix the codestyle errors, I guess it belongs to a dedicated PR.

themr0c:
\o/ all checks passed, except the code style! ready to merge?

PR #1901, opened 2019-04-02
Update Dockerfile

cleveritcz:
Changed version of python docker image with the lowest vulnerabilities (hardening)
#### PR Type

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # This is a multi-stage build which requires Docker 17.05 or higher
-FROM python:3.7-alpine as molecule-builder
+FROM python:alpine3.8 as molecule-builder
 
 WORKDIR /usr/src/molecule
 
@@ -24,7 +24,7 @@ RUN \
 # ✄---------------------------------------------------------------------
 # This is an actual target container:
 
-FROM python:3.7-alpine
+FROM python:alpine3.8
 LABEL maintainer "Ansible <info@ansible.com>"
 
 ENV PACKAGES="\


cleveritcz:
pardon, I saw it bad. It showed me yours build from master not with edited Dockerfile. just misunderstood.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # This is a multi-stage build which requires Docker 17.05 or higher
-FROM python:3.7-alpine as molecule-builder
+FROM python:alpine3.8 as molecule-builder
```
MarkusTeufelberger:
```suggestion
FROM python:3.7-alpine3.8 as molecule-builder
```

Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN \
 # ✄---------------------------------------------------------------------
 # This is an actual target container:
 
-FROM python:3.7-alpine
+FROM python:alpine3.8
```
MarkusTeufelberger:
```suggestion
FROM python:3.7-alpine3.8
```

Dockerfile
MarkusTeufelberger: This would make it more clear which version of Python and alpine is used (since the version numbers are quite close).

Also: Why alpine 3.8 and not 3.9?

cleveritcz: Alpine 3.9 has more and higher vulnerabilities.

cleveritcz: I don't know too how to edit pull request and I am lazy to solve it out. :)

MarkusTeufelberger: Could you be a bit more specific and show some sources for this statement please?

MarkusTeufelberger: https://hub.docker.com/_/python/ doesn't seem to show any data about vulnerabilities to me. Which ones are the ones relevant to molecule?

About editing pull requests: You'll need to amend your commit and do a force push of your branch (you'll need to do that anyways, since `--signoff` is required to accept the DCO).

Dockerfile
cleveritcz: you need to swith it to tags tab. Official images has vulnerabilties scanner from Docker.

https://hub.docker.com/_/python?tab=tags&page=1

MarkusTeufelberger: Ok, seems like that's only available when having and being logged into a Dockerhub account.

Anyways: Which vulnerabilities that would affect molecule are in Alpine 3.9 but not 3.8?

Dockerfile
webknjaz: @cleveritcz just hit "Commit suggestions" on GitHub UI.

MarkusTeufelberger: Good luck with the signoff requirement...

webknjaz: You can manually type it anyway

ssbarnea:
`python:3.7-alpine` is the imagine we want as py37 us the newest stable python we support. If you are not pleased about the alpine versions used, feel free to address it with python foundation.

Do not confuse python version with alpine version.

Issue #1902, opened 2019-04-03
Lint errors cause molecule to skip cleanup


ghislainbourgeois:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version

ansible 2.7.9
  config file = None
  configured module search path = [u'/Users/ghibourg/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python2.7/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 2.7.15 (default, Sep 18 2018, 20:16:18) [GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)]
molecule, version 2.20.0
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

- yaml-lint
- ansible-lint

- cleanup playbook defined

# Desired Behavior

Given `molecule test` with linters and a cleanup step configured, when the lint step errors out,
molecule should call the cleanup step and then the destroy step. 

# Actual Behaviour

Linting errors directly call `destroy`, never calling the `cleanup` step.

```
$ molecule init role -r test && cd test

$ cat molecule/default/molecule.yml
---
dependency:
  name: galaxy
driver:
  name: docker
lint:
  name: yamllint
platforms:
  - name: instance
    image: centos:7
provisioner:
  name: ansible
  lint:
    name: ansible-lint
  playbooks:
    cleanup: cleanup.yml
verifier:
  name: testinfra
  lint:
    name: flake8

$ cat molecule/default/cleanup.yml
---
- name: Cleanup
  hosts: localhost
  tasks:
    - name: Debug
      debug:
        msg: "Cleanup"

$ molecule test
--> Validating schema /Users/ghibourg/Projects/test/molecule/default/molecule.yml.
Validation completed successfully.
--> Test matrix
                                                                                                                                                                                                                                              └── default
    ├── lint
    ├── cleanup
    ├── destroy
    ├── dependency                                                                                                                                                                                                                                ├── syntax
    ├── create
    ├── prepare
    ├── converge
    ├── idempotence                                                                                                                                                                                                                               ├── side_effect
    ├── verify
    ├── cleanup
    └── destroy
                                                                                                                                                                                                                                              --> Scenario: 'default'
--> Action: 'lint'
--> Executing Yamllint on files found in /Users/ghibourg/Projects/test/...
Lint completed successfully.
[...]

PR #1903, opened 2019-04-03
Execute cleanup on lint failure (#1902)

ghislainbourgeois:
Explicitly call `cleanup` on linting errors right before calling `destroy`.
This fixes bug #1902.

This could be improved by somehow linking the destroy task to the
cleanup task, making sure that calling destroy always calls cleanup
first it is defined.

The test just validates that the `Action: cleanup` appears in the output 
on linting error. It leverages the existing test by adding an additional
assert. 

#### PR Type

Bugfix Pull Request

### DIFF
--- a/molecule/command/test.py
+++ b/molecule/command/test.py
@@ -110,8 +110,8 @@ class Test(base.Base):
           'Molecule run (always).'))
 def test(ctx, scenario_name, driver_name, __all, destroy):  # pragma: no cover
     """
-    Test (lint, destroy, dependency, syntax, create, prepare, converge,
-          idempotence, side_effect, verify, destroy).
+    Test (lint, cleanup, destroy, dependency, syntax, create, prepare,
+          converge, idempotence, side_effect, verify, cleanup, destroy).
     """
 
     args = ctx.obj.get('args')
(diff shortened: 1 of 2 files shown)


decentral1se:
OK, tracing back to the original PR/docs, it looks like it is definitely the case that cleanup was meant to be called everytime before destroy: https://github.com/ansible/molecule/pull/1676/files#diff-fd0ed9e86ebd26cf708da3d42fbf2eacR178.

molecule/command/test.py
```diff
@@ -110,8 +110,8 @@ def execute(self):
           'Molecule run (always).'))
 def test(ctx, scenario_name, driver_name, __all, destroy):  # pragma: no cover
     """
-    Test (lint, destroy, dependency, syntax, create, prepare, converge,
-          idempotence, side_effect, verify, destroy).
+    Test (lint, cleanup, destroy, dependency, syntax, create, prepare,
```
ragingpastry: lint, cleanup, destroy, dependency, syntax, create, prepare, converge, idempotence, side_effect, verify, cleanup, destroy.

ghislainbourgeois: > Thanks for this fix! LGTM +1
> 
> As far as I understand, the `cleanup` -> `destroy` steps are in the sequence already but because there is a failure which stops execution, we have to manually call the `cleanup` -> `destroy` steps ourselves? Am I on the right track?

Yes, that is exactly what was happening. Because lint was directly calling destroy on errors, and not trying to follow the sequence. The better design I feel would be to use the sequence in some way, but that would require a lot more work.

ragingpastry: I think that this is good for now because it uses a similar pattern to the rest of molecule. Utilizing the sequence would indeed be a lot more work and would make this section different than how the rest of the molecule commands work internally.

decentral1se:
Thanks all! I'll be merging this when the CI passes ...

molecule/command/test.py
decentral1se: Agreed :+1:

PR #1911, opened 2019-04-09
Update format for yapf 0.27

seandst:
This is the result of `tox -e format` having applied yapf changes from
version 0.26 to 0.27.

re #1895

### DIFF
--- a/molecule/command/base.py
+++ b/molecule/command/base.py
@@ -89,8 +89,8 @@ def execute_cmdline_scenarios(scenario_name, args, command_args):
     :returns: None
 
     """
-    scenarios = molecule.scenarios.Scenarios(
-        get_configs(args, command_args), scenario_name)
+    scenarios = molecule.scenarios.Scenarios(get_configs(args, command_args),
+                                             scenario_name)
     scenarios.print_matrix()
     for scenario in scenarios:
         try:
--- a/molecule/command/check.py
+++ b/molecule/command/check.py
@@ -73,12 +73,11 @@ class Check(base.Base):
 
 @click.command()
 @click.pass_context
-@click.option(
-    '--scenario-name',
-    '-s',
-    default=base.MOLECULE_DEFAULT_SCENARIO_NAME,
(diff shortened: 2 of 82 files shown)


seandst:
> ... if it passes CI.

Indeed. I'll briefly point out that I did *zero* local testing or review with this, with the thought being "we can either rely on yapf or we can't". While I am taking responsibility for making the changes in this PR, I am not necessarily taking responsibility for the changes themselves. yapf gets the credit there, even if I end up with the `git blame`. :)

seandst:
Travis failures appear to be network-related. There are two failures in separate tests, both were caused by this:
```
 [WARNING]: - timezone was NOT installed successfully: Failed to get data from
the API server (https://galaxy.ansible.com/api/): ('The read operation timed
out',)
```

I recommend ignoring the travis failure, unless I've misdiagnosed the issue here.

decentral1se:
Nice one. Restarted those failed network builds (still happening ...) but the linter is complaining https://travis-ci.com/ansible/molecule/jobs/191585216.

decentral1se:
Seems this is pretty damn important now since all the CI failures are related.

Thanks for picking this up!

seandst:
Yeah, totally missed the linter failures before, sorry about that.

If yapf is writing out files that fail lint, then my faith in yapf has been eroded. I'm definitely not interested in weeding through the yapf changes and finding the middle ground between the changes yapf makes and the formatting flake8 will accept. It looks like there's an [open bug](https://github.com/google/yapf/issues/694) that is at least related to this, so it might be that the solution for now is to pin yapf to 0.26 to cut down on the CI noise related to 0.27's changes and figure out what to do later.

decentral1se:
> so it might be that the solution for now is to pin yapf to 0.26 to cut down on the CI noise related to 0.27's changes and figure out what to do later.

Let's do that!!!

seandst:
Closing this for now, then. I'll put up a different PR summing up the problem and pinning yapf.

PR #1912, opened 2019-04-09
[WIP] Use pre-commit to run linters

ssbarnea:
Runs linters via pre-commit, which provides few extra benefits:
* linters are pinned (also avoiding site-packages related bugs)
* easy bumping via `pre-commit autoupdate`
* speed-up on subsequent executions
* avoid the need to run tox -r when updating linter versions

Side effects:
* Fixes inconsequent yaml identation
* Fixes missordering of dependecies
* Fixes extra newlines on few files

#### PR Type

- Feature Pull Request


.yamllint
```diff
@@ -10,4 +10,7 @@ rules:
   brackets:
     max-spaces-inside: 1
     level: error
+  document-start: disable
```
ssbarnea: almost nobody is using the `---` prefix and adding it would touch a huge amount of files. we can do it, but in another change.

.yamllint
```diff
@@ -10,4 +10,7 @@ rules:
...
   line-length: disable
+  key-duplicates: enable
+  truthy: disable
```
ssbarnea: truthy is debatable too and does not play well with "on" use on travis config, better off than adding exclude for travis file.

MarkusTeufelberger:
Why run the linters through yet another external tool (that requires extra config checked in as well) instead of adding a few jobs that run each linter in parallel? It's not like travis-ci is going to install pre-commit hooks...

.readthedocs.yml
```diff
@@ -3,5 +3,5 @@ build:
 python:
   version: 3.6
   extra_requirements:
-  - docs
+    - docs
```
webknjaz: I'd like to avoid indenting this.

.yamllint
webknjaz: I think this could go to a separate discussion in a separate PR.

.flake8
```diff
@@ -1,4 +1,5 @@
 [flake8]
+# keep exclude in-sync with ones in .pre-commitconfig.yaml or better remove them
```
webknjaz:
```suggestion
# keep exclude in-sync with ones in .pre-commit-config.yaml or better remove them
```

webknjaz: Better keep them in just one place.

.github/ISSUE_TEMPLATE/proposal.md
```diff
@@ -1,7 +1,7 @@
 ---
 name: Proposal
 about: Suggest an large change for the Molecule project
-labels: proposal 
+labels: proposal
```
webknjaz: Separate this plz

.travis.yml
```diff
@@ -9,471 +9,471 @@ cache:
   bundler: true
   pip: true
   directories:
-  - $HOME/.cache/pre-commit
-  - $HOME/.pre-commit
-  - $HOME/.rvm
-  - $HOME/virtualenv/python$(python -c 'import platform; print(platform.python_version())')
-  - $HOME/Library/Caches/Homebrew
+    - $HOME/.cache/pre-commit
```
webknjaz: I don't like this indentation.

webknjaz:
@decentral1se `pre-commit` is a de-facto standard among popular pythonic repos. It's a good tool and I'm using it for years.

Among the advantages I like:
- consolidation of various linters in one place;
- ability to set it up as a pre-commit git hook which would invoke it whenever you make a commit or push locally;
- ability to run exactly the same checks as in CI locally. it might sound as it's conflicting with tox but in this case, "tox" is just a glue for and a unified interface for invocation of different tasks.

.readthedocs.yml
MarkusTeufelberger: ...why? It is the default in yamllint and more readable too (since the list is a child item it should be indented). Also the yaml files used by molecule are styled like this.

MarkusTeufelberger:
> in this case, "tox" is just a glue for and a unified interface for invocation of different tasks

`pre-commit` is different from this description how exactly? Also in Ansible it seems to be only used in `ansible-lint` - and there also not really to run/configure any hooks. They use tox instead to run tests which is a de-facto standard among popular pythonic repos.

.readthedocs.yml
webknjaz: I'd argue that too much indentation creates more distraction and more ugliness. Also, since the files changed are mostly maintained by me I'll be insisting.
You'll see when you end up with 10 levels of indentation someday.

It's fine to lint that it's parsable, though.

webknjaz:
@MarkusTeufelberger tox is a runner for different targets (tests, builds, linters). pre-commit is a thing which nicely consolidates linters and tox will run it.
@ssbarnea brings in another ~15 checks/linters: would you prefer to configure each of them separately in a ton of different places? And then pollute tox envs with that?

> use tox instead to run tests which is a de-facto standard among popular pythonic repos.

By the way, since you brought it up, one of the `tox` maintainers actually wrote `pre-commit` tool and `tox project itself uses it.

Look here:
- https://github.com/tox-dev/tox/blob/bcdd3c8/tox.ini#L59-L63
- https://github.com/pytest-dev/pytest/blob/13a9d87/tox.ini#L71-L72
- https://github.com/pypa/virtualenv/blob/124d0ed/tox.ini#L93-L97
- https://github.com/ambv/black/blob/master/.pre-commit-config.yaml

Are any of these projects popular and pythonic enough for you?

ssbarnea:
@MarkusTeufelberger You need to play a little bit with pre-commit to understand why is so good at what it does. I could say that the only issue I had with it was its name. Due to its name most people assume is just a git-commit-hook, which historically proved to be quite controversial. 

In fact the hook part is totally optional but the real value is the ability to manage all linters in a single place, with pinning, with one command bumping (that is not affected by tox virtenv bugs), with super caching (it will reuse a specific linter version across multiple projects, lowering the footprint on developer machines considerably). Also it has the ability to lint only modified files which is of HUGE benefit for slow tools.

Now I can  only hope that @webknjaz can be convinced to accept consistent identation on all yaml files in the repo, which has the side effect of fixing the travis config file.

ssbarnea:
I will rework this once I get two preparatory changes merged: https://github.com/ansible/molecule/pull/1918 and https://github.com/ansible/molecule/pull/1919

MarkusTeufelberger:
@webknjaz - As I wrote: the only Ansible project that has a config for this tool (and an extremely minimal at that) is ansible-lint, which is actually linted and tested using tox.

My main concern is that it gets harder to know which linter or check actually fails if some tool runs a lot of them at once in a single step.

webknjaz:
> Now I can only hope that @webknjaz can be convinced to accept consistent identation on all yaml files in the repo, which has the side effect of ~fixing~ the travis config file.

s/fixing/breaking/

> My main concern is that it gets harder to know which linter or check actually fails if some tool runs a lot of them at once in a single step.

@MarkusTeufelberger: as @ssbarnea wrote, if you play with it yourself you'll understand. It actually very clearly outputs the info about what failed. So you may cross this concern out from your list :)
- Example successful output: https://travis-ci.org/cherrypy/cherrypy/jobs/512712187#L483-L501
- Example failure: https://travis-ci.org/cherrypy/cheroot/jobs/518115081#L239-L241

> the only Ansible project that has a config for this tool

Just because I didn't get to it :)

MarkusTeufelberger:
> It actually very clearly outputs the info about what failed.

Yeah, but you need to view the output for that. I am talking about the different checks that are displayed in GitHub above (the "Some checks were not successful - 2 action required, 1 in progress, and 1 successful checks" part below). There it would be just a generic failed task just like now, ideally there would be a list of different linters so it is easy to see if yamllint or ansible-lint or flake8 or whatever were the one complaining. These jobs also would be spawned by Travis in parallel.

Having to check the build logs for which linter actually is the one that failed doesn't sound "very clearly" to me at all.

webknjaz:
@MarkusTeufelberger I agree with that in general and you could separate things if you wanted. But the point here is that if you bring these things into `tox` directly you'll pollute it.
And also take into account that most of the integrations (and Travis in particular) don't report separate checks to GitHub as they should.
I think I may want to write a plugin for tox to improve that (I've started https://tutorial.octomachinery.dev because of this after all). But now this concern isn't relevant as we cannot address it. Not yet.

tox.ini
```diff
@@ -37,8 +37,7 @@ commands_pre =
 commands =
     unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
     functional: pytest test/functional/ {posargs}
-    lint: flake8
-    lint: yamllint -s test/ molecule/
+    lint: python -m pre_commit run -a
```
webknjaz: Put this under lint env.

webknjaz: (and please use verbose `--all-files`)

.readthedocs.yml
decentral1se: So we need a "files that webknjaz maintains" exclude for yamllint? This is the second person who wants to indent this vs. your "I maintain (own!?) this file" mentality. Don't stop imparting your wisdom on this project, please, but your subjectivities around indentation may be having less interesting effects. I only speak for myself ...

webknjaz: Well, you were caring about contributor experience. So from this PoV it's not logical to make things unreadable for the contributor who looks there often vs. the possibility of somebody else looking at it once or twice in the next decade. I wouldn't care if I were to fully defer the related responsibilities to someone else.

> exclude for yamllint

a separate config would do

P.S. In the internal Ansible Core team Slack it looks like the majority of folks prefer non-indented sequences (as per recent discussions). We also mostly disagree with a lot of what ansible-lint pushes by default. I think this may lead to PoV diversity.

decentral1se: I see a number of comments telling me about experience now. Please do not use this as a bat to beat me with. My personal experience is fine except for trying to guess at what others might find easier and then hearing the fall out from it ... but I am going to stop doing that now.

MarkusTeufelberger:
> And also take into account that most of the integrations (and Travis in particular) don't report separate checks to GitHub as they should.

Once they do (or a different CI tool is used, once Travis is killed properly by their new owners), are you commiting to remove this `pre-commit` stuff from the repository again?

.readthedocs.yml
MarkusTeufelberger: The travis file is also heavily (and unnecessarily in my opinion) parametrized with tons of single line anchors etc.

I wouldn't see the indentation of the file as the main source of its distractiveness and ugliness.

> P.S. In the internal Ansible Core team Slack it looks like the majority of folks prefer non-indented sequences (as per recent discussions). We also mostly disagree with a lot of what ansible-lint pushes by default. I think this may lead to PoV diversity.

Then the internal slack group is free to comment here as well and give their reasons. I already wrote mine above (having the `-`on the same indentation level while being a child item makes it less obvious that this is a child element). Also this is about yamllint, not ansible-lint.

As a data point: https://yaml.org/spec/1.2/2009-07-21/spec.html#id2559116 indents child lists.

webknjaz:
@MarkusTeufelberger nope. Do you have any other standard and easy to set up tool to hook up with Git?

webknjaz:
> why has this been done

It's just a config for GitHub so that it'd auto-request a review from me. I hope you're not confusing the term "owner". In software projects, it often is synonymous with "responsible party". The reason is that I want to get notified about changes in certain files more granularly.

MarkusTeufelberger:
GitLab-CI for example. If you want a place to run several checks or types of checks at once, you can use tox, it is already set up in this repository... This feels like talking in circles though. Integrate `pre-commit` if you like, still it would be a better experience if the individual linters get pulled apart in Travis-CI into separate tasks that can fail separately (and ensuring that all are run! Does `pre-commit` stop on the first failing linter?) instead of having one single large `lint` job/stage.

webknjaz:
> Does `pre-commit` stop on the first failing linter?

No, it runs all of them.

> a better experience if the individual linters get pulled apart in Travis-CI into separate tasks that can fail separately

I know, but the second part of the coin is polluting Travis' per-repo and per-org queues which in turn make the experience worse :(
If only we had unlimited resources, I'd gladly vote for one fail per job, though.

I think I may achieve this with GitHub Actions, btw. Let me think about it separately.

------------


I think the main point here is the ability to have pre-commit hooks wired locally with a battle-tested tool.

webknjaz:
Superseded by #1996

PR #1914, opened 2019-04-09
Point to stable documentation in the README

decentral1se:
#### PR Type
- Docs Pull Request

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,8 +6,8 @@ Molecule
    :target: https://badge.fury.io/py/molecule
    :alt: PyPI Package
 
-.. image:: https://readthedocs.org/projects/molecule/badge/?version=latest
-   :target: https://molecule.readthedocs.io/en/latest/
+.. image:: https://readthedocs.org/projects/molecule/badge/?version=stable
+   :target: https://molecule.readthedocs.io/en/stable/
    :alt: Documentation Status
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -6,8 +6,8 @@ Molecule
    :target: https://badge.fury.io/py/molecule
    :alt: PyPI Package
 
-.. image:: https://readthedocs.org/projects/molecule/badge/?version=latest
-   :target: https://molecule.readthedocs.io/en/latest/
+.. image:: https://readthedocs.org/projects/molecule/badge/?version=stable
```
webknjaz: @decentral1se I think that the picture itself should point to latest because it's a visual indication of whether master is broken. Just like with Travis CI.

decentral1se: While I do partially agree, I know users are clicking on this icon to get to the documentation and I'd rather avoid any confusion: there have been multiple tickets mixing up the documentation ...

webknjaz: It's not about where the link leads the user, it's about what's shown on the picture.

decentral1se: I disagree. It's ambiguous to users and I never remember a concensus on the badges being "for developers only". If you want to change it, please feel free but I won't be happy to deal with further tickets regarding mismatching documentation ...

webknjaz: Another way to do this is to have two separate badges.

PR #1915, opened 2019-04-09
Pin yapf to <0.27 (0.26)

seandst:
Format checking is failing with yapf 0.27, but if yapf 0.27's changes are
applied, linting fails. We need to deal with that, but CI should not
continue to be broken while that discussion takes place, so I suggest
pinning yapf to keep CI working while we come to a consensus on what to
do about yapf.

#### PR Type

- Bugfix Pull Request

I haven't filed the bug about yapf yet, that's the next thing...

Before this change:

```
$ tox -r -e format-check
ERROR:   format-check: commands failed
```

After this change:

```
$ tox -r -e format-check
  format-check: commands succeeded
  congratulations :)
```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -69,7 +69,7 @@ usedevelop = false
 [testenv:format]
 commands =
     yapf -i -r molecule/ test/
-deps = yapf>=0.25.0,<2
+deps = yapf>=0.25.0,<0.27
 extras =
 skip_install = true
 usedevelop = false


webknjaz:
This will conflict with #1895. Let's merge it after that one is done.

decentral1se:
> This will conflict with #1895. Let's merge it after that one is done.

OK, sure thing. Waiting on you to push the big green button in the sky then :)

webknjaz:
@decentral1se I'm fine merging right now if nobody objects. The thing is that this PR edits a file which is going to be removed anyway. It's not worth spending time.

seandst:
I'm happy to just change the tox.ini since test-requirements is going away to prevent the conflict, or defer to #1895 if that change is imminent. I think that since this is a very focused change it probably makes sense to merge it first, even if the changes to test-requirements is going to go away in a matter of days, or hours.

webknjaz:
Feel free to rebase and merge this. But keep in mind that `master` is been broken by merges from the bunch of other PRs which have been merged throughout the day. So you may want to ignore that.

seandst:
Rebased. I'm assuming/hoping that the lint and format-check targets are green with this, and all other travis results can be safely ignored.

PR #1919, opened 2019-04-10
Increase yamllint coverage

ssbarnea:
Makes yamllinting execution to cover files that previously were not
linted in order to assure consistent formatting.

This also enables people to run yamllint without extra parameters
and avoid getting errors.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/.github/dco.yml
+++ b/.github/dco.yml
@@ -1,2 +1,3 @@
+---
 require:
   members: false
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,3 +1,4 @@
+---
 version: 2
 
 formats: all
@@ -8,7 +9,7 @@ build:
 python:
   version: 3.7
   install:
-  - method: pip
-    path: .
-    extra_requirements:
-    - docs
+    - method: pip
+      path: .
+      extra_requirements:
+        - docs
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,587 +9,586 @@ cache:
   bundler: true
   pip: true
   directories:
-  - $HOME/.cache/pre-commit
-  - $HOME/.pre-commit
-  - $HOME/.rvm
-  - $HOME/virtualenv/python$(python -c 'import platform; print(platform.python_version())')
-  - $HOME/Library/Caches/Homebrew
+    - $HOME/.cache/pre-commit
+    - $HOME/.pre-commit
+    - $HOME/.rvm
+    - $HOME/virtualenv/python$(python -c 'import platform; print(platform.python_version())')
+    - $HOME/Library/Caches/Homebrew
 
 services:
-- docker
+  - docker
 
 language: python
 
 .mixtures:  # is not used by Travis CI, but helps avoid duplication
-- &if-cron-or-manual-run-or-tagged
-  if: type IN (cron, api) OR tag IS present
-- &py-27
-  python: "2.7"
-- &py-36
-  python: "3.6"
-- &py-37
-  python: "3.7"
-- &reset-prerequisites
-  addons: {}
-  before_install:
-  - pip install tox-venv tox-tags
-  before_script: []
-  services: []
-- &lint
-  <<: *py-37
-  <<: *reset-prerequisites
-- &env-functional
-  TESTS_TYPE: functional
-- &env-functional-shard_1
-  <<: *env-functional
-  PYTEST_ADDOPTS: >-
-    '-m shard_1_of_3'
-- &env-functional-shard_2
-  <<: *env-functional
-  PYTEST_ADDOPTS: >-
-    '-m shard_2_of_3'
-- &env-functional-shard_3
-  <<: *env-functional
-  PYTEST_ADDOPTS: >-
-    '-m shard_3_of_3'
+  - &if-cron-or-manual-run-or-tagged
+    if: type IN (cron, api) OR tag IS present
+  - &py-27
+    python: "2.7"
+  - &py-36
+    python: "3.6"
+  - &py-37
+    python: "3.7"
+  - &reset-prerequisites
+    addons: {}
+    before_install:
+      - pip install tox-venv tox-tags
+    before_script: []
+    services: []
+  - &lint
+    <<: *py-37
+    <<: *reset-prerequisites
+  - &env-functional
+    TESTS_TYPE: functional
+  - &env-functional-shard_1
+    <<: *env-functional
+    PYTEST_ADDOPTS: >-
+      '-m shard_1_of_3'
+  - &env-functional-shard_2
+    <<: *env-functional
+    PYTEST_ADDOPTS: >-
+      '-m shard_2_of_3'
+  - &env-functional-shard_3
+    <<: *env-functional
+    PYTEST_ADDOPTS: >-
+      '-m shard_3_of_3'
 
 jobs:
   fast_finish: true
 
   allow_failures:
-  - <<: *py-37
-    env:
+    - <<: *py-37
+      env:
       ANSIBLE_VERSION: "28"
       TESTS_TYPE: unit
-  - <<: *py-37
-    env:
-      <<: *env-functional-shard_1
-      ANSIBLE_VERSION: "28"
-  - <<: *py-37
-    env:
-      <<: *env-functional-shard_2
-      ANSIBLE_VERSION: "28"
-  - <<: *py-37
-    env:
-      <<: *env-functional-shard_3
-      ANSIBLE_VERSION: "28"
-  - <<: *py-36
-    env:
-      ANSIBLE_VERSION: "28"
-      TESTS_TYPE: unit
-  - <<: *py-36
-    env:
-      <<: *env-functional-shard_1
-      ANSIBLE_VERSION: "28"
-  - <<: *py-36
-    env:
-      <<: *env-functional-shard_2
-      ANSIBLE_VERSION: "28"
-  - <<: *py-36
-    env:
-      <<: *env-functional-shard_3
-      ANSIBLE_VERSION: "28"
-  - <<: *py-27
-    env:
-      ANSIBLE_VERSION: "28"
-      TESTS_TYPE: unit
-  - <<: *py-27
-    env:
-      <<: *env-functional-shard_1
-      ANSIBLE_VERSION: "28"
-  - <<: *py-27
-    env:
-      <<: *env-functional-shard_1
-      ANSIBLE_VERSION: "28"
-  - <<: *py-27
-    env:
-      <<: *env-functional-shard_3
-      ANSIBLE_VERSION: "28"
-
-  - <<: *py-37
-    env:
-      ANSIBLE_VERSION: devel
-      TESTS_TYPE: unit
-  - <<: *py-37
-    env:
-      <<: *env-functional-shard_1
-      ANSIBLE_VERSION: devel
-  - <<: *py-37
-    env:
-      <<: *env-functional-shard_2
-      ANSIBLE_VERSION: devel
-  - <<: *py-37
-    env:
-      <<: *env-functional-shard_3
-      ANSIBLE_VERSION: devel
-  - <<: *py-36
-    env:
-      ANSIBLE_VERSION: devel
-      TESTS_TYPE: unit
-  - <<: *py-36
-    env:
-      <<: *env-functional-shard_1
-      ANSIBLE_VERSION: devel
-  - <<: *py-36
-    env:
-      <<: *env-functional-shard_2
-      ANSIBLE_VERSION: devel
-  - <<: *py-36
-    env:
-      <<: *env-functional-shard_3
-      ANSIBLE_VERSION: devel
-  - <<: *py-27
-    env:
-      ANSIBLE_VERSION: devel
-      TESTS_TYPE: unit
-  - <<: *py-27
-    env:
-      <<: *env-functional-shard_1
-      ANSIBLE_VERSION: devel
-  - <<: *py-27
-    env:
-      <<: *env-functional-shard_1
-      ANSIBLE_VERSION: devel
-  - <<: *py-27
-    env:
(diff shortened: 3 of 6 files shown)


ssbarnea:
@webknjaz Please don't take this personally by the whole point of the change is to avoid file exceptions in the linter. That is not a personal preference on indentation, that's the linter rule.

webknjaz:
> that's the linter rule.

I don't accept it because linter is wrong. And if you want to apply wrong rules to other files it's fine as long as it doesn't break files I care about.

webknjaz:
> how you're going to persuade

That's very easily addressed with having `pre-commit` tool integrated and having a separate linter

webknjaz:
> This also enables people to run yamllint without extra parameters

While this is nice we shouldn't advertise "you can build/manage your envs manually without `tox`" approach. This may create "now support my extra weird config" type requests on the issue tracker which is kinda distracting.

MarkusTeufelberger:
> I don't accept it because linter is wrong.

Please cite the part of the specification that is violated by writing a list in a mapping with indentation before the `-`. Otherwise it is just a matter of personal taste.

decentral1se:
Yes, please spare us personal subjectivities of "wrong" ...

webknjaz:
> Please cite the part of the specification

Can you cite the part where it says that indented thing is more readable then?

MarkusTeufelberger:
I never made a claim about readability, but you made a claim that `yamllint` is wrong.

Here's the spec on indenting sequences:

https://yaml.org/spec/1.2/spec.html#id2759963 in Example 2.3, 2.9 and 2.10 indents sequences with 2 spaces.

Also from there (emphasis mine):
"YAML’s block collections **use indentation for scope** and begin each entry on its own line." As the scope of a list in a mapping is the key of the mapping, not the mapping itself, the list should therefor be indented.

webknjaz:
Those examples aren't etalons of readability. Especially not that one next to your snippet example 2.4.

Your snippet was too generic.

There's a section about paragraphs https://yaml.org/spec/1.2/spec.html#-%20block%20sequence%20entry// saying: `A block sequence is simply a series of nodes, each denoted by a leading “-” indicator.` with no mention of the indentation required.

Plus we all know that parsers do not require it.
If it was true, we'd apply indentation on the document level as well following instructions about that (because a document is also a scope):
```yaml
---

  - item1:

      - with:

          - 
            some: subitem2

  - item3:

      - with:

          - 
            some: subitem4

---

  - item5:

      - with:

          - 
            some: subitem6

  - item7:

      - with:

          - 
            some: subitem8

---
```
this example doesn't use too much nesting, but for more complex things it turns YAML from format for humans into yet another format for robots.

MarkusTeufelberger:
> Plus we all know that parsers do not require it.

The linter requires it. Every other YAML file in this repository conforms to it. It is valid and well-formed YAML. You claim that the linter is wrong. Prove it.

decentral1se:
Please fix the CI errors and let's get this merged. I think making an exception for specific maintainers in our configuration is not reasonable and I see a majority of community memebers arguing for this change to be accepted.

decentral1se:
> Merging can be performed automatically once the requested changes are addressed.

Please unblock this @webknjaz.

webknjaz:
Majority of community members will not end up reading it. I'm not convinced.

MarkusTeufelberger:
So... what would it take to convince you otherwise?

webknjaz:
@MarkusTeufelberger I've already explained my position. Let's listen to each other instead of just pushing for own vision.
I don't care about files I mostly don't look at (actually I care, but less, so, as a compromise, I don't even try telling you what to do there).
But forcing me to format things you never even contributed (and most likely never will) is selfish.

Readability matters. But what does this actually mean?
I think that things should be readable to ones who are actually going to read them. It's just so happened that in this case, it's me.
I maintain tons of similar configs (`.travis.yml`, in particular!) in a lot of other repos. I often even copy-paste chunks between them when I find new/better approaches. It is PITA to have it in different formats all over the Internet.
And even if not for easy and less error-prone code merging, maintainability is also an important factor here. It's probably connected with the power of habit, however, by having familiar things looking familiar you reduce the time for understanding of what's going on along with the number of errors connecting to it.

MarkusTeufelberger:
Great, so you just need me to take over maintaining the .travis-ci file until travis finally dies.

It takes ~500 lines to run only 28 jobs anyways and contains more anchors than Popeye's village, so I'll get started there right away.

decentral1se:
> Let's listen to each other instead of just pushing for own vision.

As I understand this: your position within this thread is "I maintain this file more than other people and I want it to be formatted the way I prefer". Every other community member on this thread has argued for "the project has a general configuration and it is more consistent to apply it across the entire project". So this is about a "project vision" and not our "own vision". This is not the first time that you have argued obstinately against other community members opinions and now you are in actual fact blocking that point of view from progressing. I find this both worrying and demoralising.

decentral1se:
Please see https://github.com/ansible/community/issues/427#issuecomment-484019934.

MarkusTeufelberger:
Ok, after looking into it: Using more than one variable env-var with Travis is a mess. There would be other options out there too (e.g. Circle-CI, which has nicer integration with Github checks too or Azure devops), but according to the latest meeting notes, molecule might switch to a different CI system anyways (`ansible-core-ci`) after 2.8 drops so it's probably not worth investing any more time in this.

webknjaz:
@decentral1se

> As I understand this: your position within this thread is "I maintain this file more than other people and I want it to be formatted the way I prefer".

Again: it's just an end-effect of underlying reasoning everyone seems to have chosen to ignore. I've got actual supporting arguments there rather than just "this other guy also thinks so".

You seem to be advocating to democracy (which evidently doesn't work well; seealso: trump/brexit) where the majority is considered to know better than everyone else. While I argue that quality is more important than quantity.

> I find this both worrying and demoralising.

I'm sorry to hear that. I understand that discussions may become heated sometimes but it's never my intention to make anyone feel bad.
OTOH I fully understand your feelings. I myself feel like I've suddenly got into the middle of a quite unwelcoming place. Proving points to folks who are not interested in hearing any reason is exhausting.


@MarkusTeufelberger 

> It takes ~500 lines to run only 28 jobs anyways and contains more anchors than Popeye's village

Before judging first take time to actually understand the reasoning behind the solution. I don't even know what else to say here. You can unwrap that into 1500 lines if that seems better maintainable to you...

Yes, more explicit job configs are easier to read (this is where you're 100% correct) but unfortunately hard to maintain. At one point I even was going to have a CI config generator but never got to implement that...

> Using more than one variable env-var with Travis is a mess

Yes and no. There's advantages and drawbacks in all CIs. I've tried the majority of them. Travis' upside is that it has probably the best integration with GitHub while disadvantages include poor flexibility of the config file in particular. There's more. It's easy to judge what's wrong and it's hard to do what's right. Writing that something is bad is discouraging. Not having improvement suggestions while doing so is also making such criticism more or less useless.

I did the best I could do with this which is having an explicit list of jobs with a few explicit variables for readability.

gundalow:
* `yamllint` by definition is opinionated, that's the purpose of linters
* Coding standard is what's enforced by CI. (Though this is an update to that standard)
* It was previously confusing to have different coding standards for different parts of the code base
* I'd like the Molecule Community to move away from all these formatting discussions:
  * It's at best busy work
  * At worst we are actively annoying contributors (this *has* already happened)
  * I believe we all have better things to do with our time.
* Therefore, as interim [BDFL](https://en.wikipedia.org/wiki/Benevolent_dictator_for_life) I approve this update.

I will work on adding a `governance` section to the Contributing Guide to make the above clearer.

# Travis-CI
Just to confirm the comment made by @MarkusTeufelberger https://github.com/MarkusTeufelberger
* We are currently using (and paying) for travis-ci.com
* Molecule, along with other Ansible related projects will be looking at moving to Zuul

I know the current Travis setup isn't the best, though I want us to avoid the switching cost of moving from Travis to X to Zuul. If the capacity currently allocated (10 concurrent jobs) in Travis-CI.com isn't enough I can fix that by throwing more money at it.

.travis.yml
```diff
@@ -9,473 +9,473 @@ cache:
...
+        - echo > setup.py
+      deploy: &deploy-step
+        provider: pypi
+        user: ansible-molecule
+        password:
+          secure: >
+            HPBkDwncLYLEzxirPfNB88GYFWZ/wA1jMgCd4wT7SgKQlIyRSCT6KXMfoOUVMH/uwaLggoURaGtmD/qnrfXj+bG15nBCCOaIZdXzODln/PwDFAYT8ZspzRPzQOfncdk4WTsVbwpiGgpdm+TxGGBz8yedvVXiTwmwLCGC0FsAE8Wp8krNH1Kwqp3OaZeePakIbEj0UXgCTnAol3ZgVWWy+6bfDBb/aLiGXjsAIb7sY6HzwvQUr43xFO7tReaGO23mGwgIy/tNstXarwxezlw6FlGe5KJGvE/a4yAPuWg9kuJt5MqwbCJzhP38SeH4Gc6x0o6jPT/+NbvLV1ck1Qbz5gCYmXlOzucDP+P5t8E3g+zJbNbCR00k1BpG3MIEzrHeyp/hCdDYnGB3TVnlj+L4Y/yoUdCq9FJI8xHLbj/fMe5vxBMMsvDbt2JJReavi3WB2pv2nHtivkKj/Kow4FRm+Kp3WH4x5NlZDjV5gsUbwPcN1kIpb09sPZE2wYugvCTOiExt7mdd0JMMcgxuR1jGgGxqRARQeqjz14VVNsrVIIrMh8JX8u7Rl9sqmU4UuLGTxnq2cwyBmYzJ0gmDgN3TvQ96n9D2sP0YQj35v7RY5vsdYsfLucLdTfTajP9OL4Bw+ogXMJrArIq+j+7uJyTFqNdAN+Y6nX7WV0W6rp8aA1I=
+        distributions: dists
+        skip-cleanup: true
+        "on":
```
webknjaz: This is broken.

gundalow:
I just triggered a fresh CI run:
```
./.pyup.yml
  4:1       warning  missing document start "---"  (document-start)
  9:6       warning  truthy value should be true or false  (truthy)
  13:2      warning  missing starting space in comment  (comments)
  23:9      warning  truthy value should be true or false  (truthy)
  30:1      error    wrong indentation: expected 2 but found 0  (indentation)
  33:10     warning  truthy value should be true or false  (truthy)
  38:10     warning  truthy value should be true or false  (truthy)
  49:1      error    wrong indentation: expected 2 but found 0  (indentation)
  59:2      warning  missing starting space in comment  (comments)
  63:12     warning  truthy value should be true or false  (truthy)
./.readthedocs.yml
  12:3      error    wrong indentation: expected 4 but found 2  (indentation)
  15:5      error    wrong indentation: expected 6 but found 4  (indentation)
```

.travis.yml
ssbarnea: That is a valid yaml syntax trick for making yaml syntax compliant. Mainly unquoted strings like "on", "true"... are considered booleans when loaded. Adding quotes assures that remain strings, like it is expected by travis. I totally agreet hat is ugly, but is the price we pay for and old travis decision of not using a linter when they wrote their yaml configs.

webknjaz: indentation.

.pyup.yml
```diff
@@ -6,11 +7,11 @@ update: all
 # configure dependency pinning globally
 # default: True
 # allowed: True, False
-pin: True
+pin: true
```
webknjaz: Have you checked that pyup supports it like this?

ssbarnea: @webknjaz In fact I just raised https://github.com/pyupio/pyup/issues/346 after testing it on my own repository. I guess that until they adress it, I will still have to keep one exclude.

webknjaz: What I used was copy-pasted from the official example. So yeah. Such things shouldn't be blindly changed. Also, I can see why they might've chosen title-case: it matches Python literals...

.github/dco.yml
```diff
@@ -1,2 +1,3 @@
+---
```
ssbarnea: Tested in my own fork to assure that it does not break RTD service.

.readthedocs.yml
```diff
@@ -8,7 +9,7 @@ build:
...
   version: 3.7
   install:
-  - method: pip
-    path: .
-    extra_requirements:
-    - docs
+    - method: pip
+      path: .
+      extra_requirements:
+        - docs
```
ssbarnea: Tested in my own fork to assure that it does not break RTD service.

.yamllint
```diff
@@ -2,6 +2,9 @@ extends: default
 ignore: |
   *.molecule/
   molecule/cookiecutter/
+  .tox
+  # HACK: https://github.com/pyupio/pyup/issues/346
+  .pyup.yml
```
ssbarnea: PYUP is the only tool that seems to choke on "hardened" yaml files, so we skip it.

tox.ini
```diff
@@ -51,7 +51,7 @@ commands =
     unit: python -m pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
     functional: python -m pytest test/functional/ {posargs}
     lint: python -m flake8
-    lint: python -m yamllint -s test/ molecule/
+    lint: python -m yamllint -s .
```
ssbarnea: This the the major goal of this change: so we avoid adding YAML files to other locations that do not have standardized formatting.
Also this makes much eaiser to run run yamllint outside tox or from other tools like IDEs, hooks,...

.yamllint
webknjaz: @ssbarnea you can add a separate lint once you finish pre-commit one

.travis.yml
webknjaz: Resolving as it turned out that `?w=1` is buggy and lies to viewers.

cc @gundalow we shouldn't rely on that view.

PR #1925, opened 2019-04-10
Release molecule v2.20.post0

### LINKED ISSUE #1922: Fixing the released version metadata
It's a follow-up for https://github.com/ansible/molecule/pull/1895#pullrequestreview-224227936:
> I haven't done any manual testing of this but it looks good to me now! I know that this has the possibility to resolve a number of issues for the package (and also to push forward re-release 2.20.0 -> 2.20, right!?), so am happy to see this move forward.

1) It's not possible to replace dists in a published version @ PyPI
2) It's possible to remove certain version but not re-upload things in the same place
3) It's also bad to do such replacements
4) Maybe it's possible to upload just `2.20`, maybe not, but we shouldn't do this. For `pip install molecule==2.20` it matches both `2.20` and `2.20.0`. But this will create collisions for users who have allowed hashes saved mapped to the version.
5) We should upload something slightly higher than `2.20`, I think `2.20.post0` would do. See PEP 440 for Python versioning recommendations.
6) We can also remove `2.20.0` but it will break things for people who have it pinned like this. Maybe we should do announcements before such removal or maybe just keep it there as it seems harmless.

```python
In [1]: from packaging import version         
                                                                                          
In [5]: version.parse("2.20.0") == version.parse("2.20")                                                                                
Out[5]: True

In [7]: version.parse("2.20.0") < version.parse("2.20.post0")                                                                           
Out[7]: True
```

For publishing `2.20.post0` I will make a branch from the tag `v2.20.0` and then I will cherry-pick 5314ebd1ae6e5413137eef8d3f7be505e0c847f1 on top of it. Finally, I'll be able to tag that commit to initialize the release process.

webknjaz:
This is on top of the last tagged version with a few packaging-related commits cherry-picked.

I'll have to tag the head of this branch to trigger publishing of `v2.20.post0`.

It will resolve #1922.

#### PR Type

- Bugfix Pull Request


decentral1se:
I'm cool with a v2.21 release, no? Missed the meeting last night though ...

webknjaz:
@decentral1se: @gundalow told me that everyone agreed on `v2.21`. I was about to close this PR but forgot :)

seandst:
Since the plan here was to generate the updated 2.20 from a cherry-pick branch, a post-release of 2.20 is independent of a normal release of 2.21. If, for some reason that I can't imagine at the moment, a post-release of 2.20 becomes necessary even after the release of 2.21, I think that can still be done. But it's probably simpler to move ahead with 2.21 and leave 2.20 alone.

webknjaz:
yep

PR #1927, opened 2019-04-10
updated doc references from 'volume_mounts' to 'volumes'

dericcrago:
#### PR Type

- Docs Pull Request

### DIFF
--- a/molecule/driver/docker.py
+++ b/molecule/driver/docker.py
@@ -101,7 +101,7 @@ class Docker(base.Base):
 
     When attempting to utilize a container image with `systemd`_ as your init
     system inside the container to simulate a real machine, make sure to set
-    the ``privileged``, ``volume_mounts``, ``command``, and ``environment``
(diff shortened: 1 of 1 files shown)


molecule/driver/docker.py
```diff
@@ -101,7 +101,7 @@ class Docker(base.Base):
 
     When attempting to utilize a container image with `systemd`_ as your init
     system inside the container to simulate a real machine, make sure to set
-    the ``privileged``, ``volume_mounts``, ``command``, and ``environment``
+    the ``privileged``, ``volumes``, ``command``, and ``environment``
```
decentral1se: Is this dependant on some version of Docker? 

Do we need to include both or `volumes` is deprecated (or just wrong!?)?

dericcrago: I was *assuming* it was just wrong, this is the only mention / example in the entire repo with `volume_mounts` every other mention / example is `volumes`. Even in this same file (lines ~66-67) it's `volumes`:

```yaml
    volumes:
```

Issue #1955, opened 2019-04-10
[Announcement] pyup robot has been enabled


webknjaz:
Hi everyone,

I've enabled [`pyup` bot](https://pyup.io/) on the Molecule repo. It will help humans to keep dependencies up-to-date. This integration has been chosen over dependabot because it supports `tox.ini` (ref: https://github.com/ansible/molecule/issues/1710#issuecomment-481874323).

UI: https://pyup.io/account/repos/github/ansible/molecule/
Config: [`.pyup.yml`](https://github.com/ansible/molecule/blob/master/.pyup.yml)
Docs:
- https://pyup.io/docs/bot/config/
- https://pyup.io/docs/bot/filter/

I expect that it may be a bit annoying at first and _we need to_ adjust in more granularly. So please feel free to send PRs against config and filter comment directives against specific lines in deps declarations.

Thanks.

<details>
<summary>--:robot:</summary>

![](https://orig00.deviantart.net/f38d/f/2018/256/7/d/robot_by_turtlemurmurs-dcmqphz.gif)
</details>

ssbarnea:
I know that this bot and I can tell you that it can be bit annoying, but weekly should keep us safe.

I checked config and seems good, pinning test deps being a good practice and not pinning runtime ones (setup.cfg), as we don't want to create conflicts for consumers.

Lets use this ticket for follow-ups on that manner.

decentral1se:
Can we close this off? Seems to well integrated now.

webknjaz:
Fair enough.

PR #1958, opened 2019-04-10
Pin another instance of yapf to <0.27

seandst:
I missed this when I put together #1915

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -26,7 +26,7 @@ deps =
     pytest-verbose-parametrize==1.4.0
     pytest-xdist==1.26.0
     shade==1.22.2
-    yapf>=0.25.0,<0.27
+    yapf>=0.25.0,<0.27  # pyup: < 0.27 # disable updates, conflicts with flake8 as per https://github.com/ansible/molecule/pull/1915
 
     ansible25: ansible>=2.5,<2.6
     ansible26: ansible>=2.6,<2.7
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -26,7 +26,7 @@ deps =
     pytest-verbose-parametrize==1.4.0
     pytest-xdist==1.26.0
     shade==1.22.2
-    yapf>=0.25.0,<2
+    yapf>=0.25.0,<0.27
```
webknjaz: https://github.com/ansible/molecule/issues/1955 add filtering as per doc mentioned here

seandst: I'd normally have no problem with it, but don't have the cycles to learn a new tool today. I assume it's simple, but I'm probably not going to get a chance to look at the pyup docs til next week.

webknjaz: Alright, just hit apply here:
```suggestion
    yapf>=0.25.0,<0.27  # pyup: < 0.27 # disable updates, conflicts with flake8 as per https://github.com/ansible/molecule/pull/1915
```

seandst: Oh, that's perfect, thanks!

seandst: There are two places where this dep is declared in the file (with good reason, imo), so rather than hit the button I'm going to steal and give credit.

webknjaz: @seandst Yeah... Well, the difference is that GitHub recognizes `Co-Authored-By: User <email>` trailer in commits which in this case you should add manually. Otherwise, GitHub doesn't have metadata for a co-author.

tox.ini
```diff
@@ -69,7 +69,7 @@ usedevelop = false
 [testenv:format]
 commands =
     yapf -i -r molecule/ test/
-deps = yapf>=0.25.0,<0.27
+deps = yapf>=0.25.0,<0.27  # pyup: < 0.27 # disable updates, conflicts with flake8 as per https://github.com/ansible/molecule/pull/1915
```
webknjaz: oh.. this one seems wrong because it's parsed as a single line. we should turn it into a list

PR #1959, opened 2019-04-11
Add fail on warning for RTD

decentral1se:
Same as our other configuration.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,6 +5,9 @@ formats: all
 build:
   image: latest
 
+sphinx:
+  fail_on_warning: true
+
 python:
   version: 3.7
   install:


.readthedocs.yml
```diff
@@ -5,6 +5,9 @@ formats: all
 build:
   image: latest
 
+sphinx:
+  fail_on_warning: true
```
webknjaz: I believe this will be picked up from `setup.cfg`. So we probably don't need duplication...

decentral1se: Explicit better than blah blah blah blah blah ;)

webknjaz: Yes and no. The same thing duplicated in a lot of places makes it hard to track updates. Next time you'll update one place but not the other and it'll behave differently without any prompt why and where to look at.

It's a common knowledge that RTD invokes sphinx. And that sphinx has configs.
Sometimes implicit is better than explicit. You just need to be careful when making related decisions :)

decentral1se: Heresy!

Issue #1896, opened 2019-04-01
Once setuptools-scm is in place, remove manual Python/Ansible version checks


decentral1se:
See https://github.com/ansible/molecule/pull/1890.

Just creating this so I don't forget to get around to it ...

webknjaz:
@decentral1se I believe this is unblocked now.

decentral1se:
OK, cool, opened https://github.com/ansible/molecule/pull/1961 against this.

PR #1961, opened 2019-04-11
Remove Python/Ansible manual version checks

decentral1se:
Closes https://github.com/ansible/molecule/issues/1896.

We rely on setuptools dist metadata.

### DIFF
--- a/test/functional/conftest.py
+++ b/test/functional/conftest.py
@@ -422,15 +422,3 @@ needs_inspec = pytest.mark.skipif(
 needs_rubocop = pytest.mark.skipif(
     not has_rubocop(),
     reason='Needs rubocop to be pre-installed and available in $PATH')
-
-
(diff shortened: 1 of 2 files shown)


test/functional/conftest.py
```diff
@@ -422,15 +422,3 @@ def has_rubocop():
...
-
-
-@pytest.helpers.register
-def is_supported_ansible_python_combo():
-    ansible_below_25 = LooseVersion(ansible.__version__) < LooseVersion('2.5')
-    max_py = (3, 6) if ansible_below_25 else (3, 7)
-    return sys.version_info[:2] <= max_py
-
-
-skip_unsupported_matrix = pytest.mark.skipif(
```
webknjaz: @decentral1se apparently, something else was using it: https://travis-ci.com/ansible/molecule/jobs/192120697#L332

PR #1963, opened 2019-04-11
Add change log with towncrier

decentral1se:
Not working yet but we can probably bikeshed on stuff already.

### DIFF
--- a/molecule/shell.py
+++ b/molecule/shell.py
@@ -18,18 +18,13 @@
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
 
-import distutils
-import distutils.version
 import os
-import sys
 
-import ansible
 import click
 import click_completion
 
 import molecule
 from molecule import command
-from molecule import util
 from molecule.config import MOLECULE_DEBUG
 
 click_completion.init()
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -17,3 +17,4 @@ pytestdebug.log
 ubuntu-xenial-16.04-cloudimg-console.log
 .python-version
 coverage.xml
+pip-wheel-metadata
```
webknjaz: I believe that this should probably go to a separate PR. (Personally, though, I'd prefer having it based on the gitignore.io templates as it's easier to keep up with standards this way...)

tox.ini
```diff
@@ -126,7 +126,6 @@ commands =
 whitelist_externals =
     sh
 
-
```
webknjaz: (unrelated change)

tox.ini
```diff
@@ -159,3 +158,12 @@ commands =
 whitelist_externals =
     rm
     {[testenv]whitelist_externals}
+
+[testenv:changelog]
+skip_install = false
+usedevelop = false
+description = Generate the changelog
+extras =
+  changelog
```
webknjaz: I believe that for towncrier to work installing dist is not required. So, if you remove that, this task will complete faster.

Currently, there's no way to tell pip "install extras but not molecule itself" which means that it should probably be left out of dist metadata.

tox.ini
```diff
@@ -159,3 +158,12 @@ commands =
...
+commands =
+  towncrier
```
webknjaz: Please use four spaces to follow the style in other envs.

setup.cfg
```diff
@@ -112,6 +112,8 @@ vagrant =
     python-vagrant
 windows =
     pywinrm
+changelog =
```
webknjaz: https://github.com/ansible/molecule/pull/1963/files#r274377937

setup.cfg
```diff
@@ -112,6 +112,8 @@ vagrant =
...
+    towncrier
```
webknjaz: Also, take into account that it's not a runtime dependency meaning that there's no benefit of exposing this to users.

I didn't state it anywhere in Molecule forums, _I think_, but I usually consider extras as a public API and try to apply other public APIs considerations to extras.

changelog/_template.rst
```diff
@@ -0,0 +1,40 @@
...
+{% if sections[section] %}
+{% for category, val in definitions.items() if category in sections[section] %}
+
+{{ definitions[category]['name'] }}
+{{ underline * definitions[category]['name']|length }}
+
+{% if definitions[category]['showcontent'] %}
+{% for text, values in sections[section][category]|dictsort(by='value') %}
+{% set issue_joiner = joiner(', ') %}
+- {% for value in values|sort %}{{ issue_joiner() }}`{{ value }} <https://github.com/ansible/molecule/issues/{{ value[1:] }}>`_{% endfor %}: {{ text }}
```
webknjaz: I think, this should use extlinks declared in the sphinx config — https://github.com/ansible/molecule/blob/6b6b2bd/doc/source/conf.py#L72-L75
This is how:
```suggestion
- {% for value in values|sort %}{{ issue_joiner() }}:issue:`{{ value }} <{{ value[1:] }}`_{% endfor %}: {{ text }}
```
There's: `:issue:`, `:pr:`, `:commit:` and `:gh:` defined.
`:gh:` one is already used in the changelog, for example: https://github.com/ansible/molecule/blame/master/CHANGELOG.rst#L81

changelog/_template.rst
```diff
@@ -0,0 +1,40 @@
+{% for section in sections %}
+{% set underline = "-" %}
```
webknjaz: I think we should use indentation in this jinja template for better readability. For HTML-like source, it's usually 2 spaces.

.gitignore
decentral1se: See https://github.com/ansible/molecule/pull/1978.

decentral1se:
Uhhh!? Bizarre ... I forced pushed to remove unrelated changes and it is now closed and the "re-open" button cannot be clicked .... well, I'll open a new PR when I get time then. Thanks for review on this.

webknjaz:
Ah.. Happens :)

PR #1967, opened 2019-04-12
Backported fix that prevented run with python 3.5

ssbarnea:
Contains cherrypicks from master for fixing the CI so we can release a hotfix that brings back the ability to run with python3.5

### DIFF
--- a/test/functional/conftest.py
+++ b/test/functional/conftest.py
@@ -22,10 +22,7 @@
 import distutils.spawn
 import os
 import shutil
-import sys
-from distutils.version import LooseVersion
 
-import ansible
 import pexpect
 import pytest
 import sh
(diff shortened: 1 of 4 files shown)


webknjaz:
@ssbarnea I saw you published a new version. For the version generation tooling to pick up this tag it should be part of `master` which means that this branch now must be merged to `master` with a true merge commit (no squash or rebase!).

cc @decentral1se @gundalow @themr0c

webknjaz:
P.S. This'd also need a changelog update before pushing a tag.

decentral1se:
Hmmmm, we should probably not make releases without following some process

@webknjaz, would you mind to update the release documentation?

webknjaz:
@decentral1se I've created an issue from your comment. I've got a slightly different focus now as I've been asked to implement Ansible Collection init in the CLI ASAP...

decentral1se:
Thank you and good luck!

PR #1969, opened 2019-04-12
Remove registry password regex

### LINKED ISSUE #1662: Check on docker registry password too restrictive
# Issue Type

- Bug report

```
molecule, version 2.19.0
ansible 2.6.3
  config file = /home/rtp/.ansible.cfg
  configured module search path = [u'/home/rtp/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.13 (default, Sep 26 2018, 18:42:22) [GCC 6.3.0 20170516]
```

- pip

- OS package

I'm using vagrant on my system and docker for my local gitlab ci. For the gitlab-ci side, I'm using a container hosted by this gitlab server. To do this, I'm using two scenarii with only differences are the driver and platform used.

My molecule platforms setting for the ci job is:
```
platforms:
- name: instance
  image: ${CI_REGISTRY}/stretch/systemd:latest
  registry:
    url: https://${CI_REGISTRY}
    credentials:
      username: gitlab-ci-token
      password: $CI_JOB_TOKEN
```

# Desired Behavior

It would be easier if the regexp used to validate the password allows:
- environment variables with underscores
- syntax like ${CI_JOB_TOKEN:-invalidpassword}

# Actual Behaviour

- When running the scenario with the vagrant driver, I'm getting:
```
ERROR: Failed to validate.

{'platforms': [{0: [{'registry': [{'credentials': [{'password': ['null value not allowed']}]}]}]}]}
```

- If I create an environment variable, It now complains about the underscores :
```
ERROR: Failed to validate.

{'platforms': [{0: [{'registry': [{'credentials': [{'password': ["value does not match regex '^[{$]+[a-z0-9A-Z]+[}]*$'"]}]}]}]}]}
```
This one happens also when running the scenario on gitlab ci.

1efty:
#### PR Type
- Bugfix Pull Request

Resolves #1662.

EDIT: See https://github.com/ansible/molecule/pull/1969#discussion_r275685148.

### DIFF
--- a/molecule/model/schema_v2.py
+++ b/molecule/model/schema_v2.py
@@ -104,8 +104,7 @@ def pre_validate_base_schema(env, keep_string):
                                 'type': 'dict',
                                 'schema': {
                                     'password': {
-                                        'type': 'string',
-                                        'regex': '^[{$]+[a-z0-9A-Z]+[}]*$',
+                                        'type': 'string'
                                     },
                                 }
                             },
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ Unreleased
 
 * `dependency` step is now run by default before any playbook sequence step, including
   `create` and `destroy`. This allows the use of roles in all sequence step playbooks.
+* Removed validation regex for docker registry passwords, all ``string`` values are now valid.
 
 2.20
 ====
--- a/doc/source/configuration.rst
+++ b/doc/source/configuration.rst
@@ -4,6 +4,8 @@ Configuration
 .. autoclass:: molecule.config.Config()
    :undoc-members:
 
+.. _variable substitution:
+
 Variable Substitution
 ---------------------
 
--- a/molecule/driver/docker.py
+++ b/molecule/driver/docker.py
@@ -124,14 +124,14 @@ class Docker(base.Base):
 
         $ pip install molecule[docker]
 
-    When pulling from a private registry, the username and password must be
-    exported as environment variables in the current shell. The only supported
-    variables are $USERNAME and $PASSWORD.
+    When pulling from a private registry, it is the user's discretion to decide
+    whether to use hard-code strings or environment variables for passing
+    credentials to molecule.
 
-    .. code-block:: bash
+    .. important::
 
-        $ export USERNAME=foo
-        $ export PASSWORD=bar
+        Hard-coded credentials in ``molecule.yml`` should be avoided, instead use
+        `variable substitution`_.
 
     Provide a list of files Molecule will preserve, relative to the scenario
     ephemeral directory, after any ``destroy`` subcommand execution.
--- a/test/unit/model/v2/test_pre_validate.py
+++ b/test/unit/model/v2/test_pre_validate.py
@@ -58,7 +58,7 @@ platforms:
   - name: instance
     registry:
       credentials:
-        password: foo
+        password: 123
 """.strip()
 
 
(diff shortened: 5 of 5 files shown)


decentral1se:
Thanks for this @1efty!

The CI failures look unrelated and due to internal code churn ... I will try to investigate.

molecule/model/schema_v2.py
```diff
@@ -105,7 +105,7 @@ def pre_validate_base_schema(env, keep_string):
                                 'schema': {
                                     'password': {
                                         'type': 'string',
-                                        'regex': '^[{$]+[a-z0-9A-Z]+[}]*$',
+                                        'regex': '^[{$]+[a-z0-9A-Z_]+[}]*$',
```
decentral1se: I must ask myself why we have this regex in the first place? Can we drop it altogether!?

1efty: Looks like it forces that value to an environment variable, safe-guarding developers from putting secrets there.  I'm in alignment with you, should be developer's discretion and it slims down the number of test cases if the regex were to be expanded.  If anything a warning in the docs should be enough of a warning about not using secrets in config.

1efty: Made the update, removing the regex validation.  Failed test seems unrelated.

decentral1se: Cool! I'm re-running the failed CI build.

>  If anything a warning in the docs should be enough of a warning about not using secrets in config.

Feel like adding it? We'd also need a changelog entry for this (manual at this point).

1efty: Changelog and docs addition look good ?

decentral1se: Great :+1:

decentral1se:
Requesting further reviews to bump this along!

test/unit/model/v2/test_platforms_section.py
```diff
@@ -44,7 +44,7 @@ def _model_platforms_docker_section_data():
                 'url': 'registry.example.com',
                 'credentials': {
                     'username': 'username',
-                    'password': '$PASSWORD',
+                    'password': '$PASS_WORD',
```
webknjaz: why?

1efty: Originally this MR was for allowing shell variables with underscores.  That specific line was updated to represent a variable with an underscore.  Now it's superfluous, but still a valid test example.

webknjaz: Ah, alright. So please add this information as a comment at the end of this line.

decentral1se: Really? It's fine as is IMHO ...

webknjaz: It wasn't obvious to me at all wrt the motivation of this change. Somebody may decide to refactor this and remove that underscore back in future. Most folks wouldn't check the history of this change. And also if they do the PR title contains literally zero clues of what it's about meaning that somebody would probably go further and connect the dots based on 2-3 extra places on the Internet. Now, would you go for all of this hassle if the refactoring reasoning is already obvious to you? (it is usually obvious at the point when you've decided to do it)

This is why I would like to see some clue at least somewhere. Comment should only explain "why", implementation details can be looked up if necessary but it should facilitate making the motivation immediately obvious.

1efty: I just removed it instead.  It's unnecessary since the scope of this MR changed.

PR #1571, opened 2018-11-08
Pass etc_hosts through to docker_container module if set in molecule config

jzafran:
#### PR Type
- Feature Pull Request

Set `etc_hosts` parameter when using `docker_container` Ansible module in `create.yml`, which makes it possible to add entries to `/etc/hosts` in the container (it's not possible to modify that file once the container is started) if `etc_hosts` is set in the `platforms` item.

For example:

```yml
platforms:
  - name: xenial
    image: ubuntu:xenial
    etc_hosts:
      'test.local': '127.0.0.1'
```

Results in an entry in `/etc/hosts`

> root@xenial:/# grep test /etc/hosts
> 127.0.0.1	test.local

PR #1972, opened 2019-04-14
Pass etc_hosts through to docker_container module if set in molecule config

benjah1:
Continue #1571

Different from the origin one, the etc_hosts is string type. See [here](https://stackoverflow.com/questions/39224679/ansible-docker-container-etc-hosts-with-variable-key)
For example,
```
platforms:
  - name: xenial
    image: ubuntu:xenial
    etc_hosts: >
      {
        "{{ domain_1 }}": "{{ domain_1_ip }}",
        "domain_2" : "{{ domain_2_ip }}",
        "domain_3" : "{{ domain_3_ip }}"
      }
```

#### PR Type

- Feature Pull Request

### DIFF
--- a/molecule/driver/docker.py
+++ b/molecule/driver/docker.py
@@ -80,6 +80,7 @@ class Docker(base.Base):
               - nofile:262144:262144
             dns_servers:
               - 8.8.8.8
+            etc_hosts: "{'host1.example.com': '10.3.1.5'}"
             networks:
               - name: foo
               - name: bar
--- a/test/unit/model/v2/test_platforms_section.py
+++ b/test/unit/model/v2/test_platforms_section.py
@@ -83,6 +83,8 @@ def _model_platforms_docker_section_data():
             'dns_servers': [
                 '8.8.8.8',
             ],
+            'etc_hosts':
+            "{'host1.example.com': '10.3.1.5'}",
             'env': {
                 'FOO': 'bar',
                 'foo': 'bar',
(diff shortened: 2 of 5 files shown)


tace:
Wait a minute, you have done almost same pull request at same time as I have, see https://github.com/ansible/molecule/pull/1974

The difference is I propose it to be dictionary and you have string. 
I think the dictionary is convenient for the user.

benjah1:
Hi @tace, I agree dictionary is more convenient for the user until I took an arrow to the knee (saw this post https://stackoverflow.com/questions/39224679/ansible-docker-container-etc-hosts-with-variable-key).

Basically, to the etc_host specifically, we may want to set the host dynamically, if we are using automation pipeline.

benjah1:
* Added the dict type as well. So this etc_hosts support both string and dict.
* rebased to master.

molecule/driver/docker.py
```diff
@@ -80,6 +80,7 @@ class Docker(base.Base):
               - nofile:262144:262144
             dns_servers:
               - 8.8.8.8
+            etc_hosts: "{'host1.example.com': '10.3.1.5'}"
```
tace: Maybe 2 host/ip values could be used in the example?

test/unit/model/v2/test_platforms_section.py
```diff
@@ -190,6 +192,7 @@ def _model_platforms_docker_errors_section_data():
             'dns_servers': [
                 int(),
             ],
+            "etc_hosts": int(),
```
tace: should this be str() instead of int() ?

benjah1: This is a test intended to trigger the error and compare the error message. Which, in this case, int() will trigger the error.

tace:
hi,
is this ready for merge?

Issue #1779, opened 2019-02-26
Use Towncrier/Reno to automate the change log


decentral1se:
See https://github.com/ansible/molecule/issues/1772#issuecomment-467204991 :+1: 

It's still open for discussion but I defer to @webknjaz on this one. If no one else objects, we can go with it!

ssbarnea:
AFAIK pbr can also deal with changelogs, something we need to check.

If pbr does not cover it maybe `reno` sounds better because: we leverage a tool already used by ansible. Even if towncrier is better we may be better of leveraging the already known/used tools and avoid introducing new potention source of issues.

One exception would be if we already get signals that a migration is happening from ansible too.

webknjaz:
@ssbarnea I've mentioned this in that comment that is linked: reno can be integrated with sphinx and pbr. But I dislike it because `reno` is hard/unobvious to integrate and I still want to get rid of `pbr` anyway.

FTR I think towncrier is better known in Python community in general (except for openstack folks maybe).

decentral1se:
OK, based on latest WG, we've got some weight behind using Reno. Please check the logs.

@pabelanger has kindly offered to do a small show and tell and let's see what is what :)

webknjaz:
@pabelanger we'll need to see how it works with `pbr` removed.

decentral1se:
I'm going to take a stab at towncrier integration. Pytest/tox use it.

webknjaz:
Good luck. You'll need it. I've failed to configure `reno` once a per required expectations. So if you have another 10 minutes to spare please also look at `towncrier` config just to feel the difference in approaches and expectations it brings :)

webknjaz:
Oh, and ping me once you decide on the changelog fragments location. I want to integrate a check for PRs. It looks like: https://github.com/sanitizers/browntruck/pull/3/checks?check_run_id=78558955

decentral1se:
Read it again :)

webknjaz:
Oh... My eyes are failing me :)

webknjaz:
But anyway it's a good experience to try both for the person doing the integration anyway.

decentral1se:
Bikeshed alert https://github.com/ansible/molecule/pull/1963 !

webknjaz:
https://github.com/ansible/molecule/pull/1979#discussion_r278106413

decentral1se:
Won't have time to chase this.

PR #1979, opened 2019-04-16
WIP: add towncrier integration

decentral1se:
Closes https://github.com/ansible/molecule/issues/1779.

Follows from Git blunder in https://github.com/ansible/molecule/pull/1963.

~~Will be implementing changes requested in https://github.com/ansible/molecule/pull/1963#pullrequestreview-225475659.~~ Done!

### DIFF
--- a/changelog/README.rst
+++ b/changelog/README.rst
@@ -1,0 +1,43 @@
+Managing the Changelog
+======================
+
+This directory contains "newsfragments" which are short files that contain a
+small **ReStructuredText**-formatted text that will be added to the next
+``CHANGELOG``.
+
+The ``CHANGELOG`` will be read by users, so this description should be aimed to
+molecule users instead of describing internal changes which are only relevant
+to the developers.
+
+Make sure to use full sentences with correct case and punctuation, for example:
+
+.. code-block:: rst
+
+    Fix broken output of non-ascii messages produced by the ``warnings`` module.
+
+Each file should be named like ``<ISSUE>.<TYPE>.rst``, where ``<ISSUE>`` is an
+issue number, and ``<TYPE>`` is one of:
+
+* ``feature``: new user facing features, like new command-line options and new behavior.
+* ``bugfix``: fixes a reported bug.
+* ``doc``: documentation improvement, like rewording an entire session or adding missing docs.
+* ``deprecation``: feature deprecation.
+* ``removal``: feature removal.
+* ``trivial``: fixing a small typo or internal change that might be noteworthy.
+
+So for example: ``123.feature.rst``, ``456.bugfix.rst``.
+
+If your PR fixes an issue, use that number here. If there is no issue, then
+after you submit the PR and get the PR number you can add a changelog using
+that instead.
+
+If you are not sure what issue type to use, don't hesitate to ask in your
+change request.
+
+``towncrier`` preserves multiple paragraphs and formatting (code blocks, lists,
+and so on), but for entries other than ``features`` it is usually better to
+stick to a single paragraph to keep it concise. 
+
+You can run ``tox -e changelog-check`` to see the drafted output of what
+``towncrier`` will generate (no file changes made). You can run the real
+generation with ``tox -e changelog``.
--- a/changelog/_template.rst
+++ b/changelog/_template.rst
@@ -1,0 +1,45 @@
+{% for section in sections %}
+  {% set underline = "-" %}
+
+  {% if section %}
+    {{section}}
+    {{ underline * section|length }}{% set underline = "~" %}
+
+  {% endif %}
+
+  {% if sections[section] %}
+
+    {% for category, val in definitions.items() if category in sections[section] %}
+
+      {{ definitions[category]['name'] }}
+
+      {{ underline * definitions[category]['name']|length }}
+
+      {% if definitions[category]['showcontent'] %}
+        {% for text, values in sections[section][category]|dictsort(by='value') %}
+          {% set issue_joiner = joiner(', ') %}
+          {% for value in values|sort %}{{ issue_joiner() }}:issue:`{{ value }} <{{ value[1:] }}`_{% endfor %}: {{ text }}
+        {% endfor %}
+
+      {% else %}
+        - {{ sections[section][category]['']|sort|join(', ') }}
+
+
+      {% endif %}
+
+      {% if sections[section][category]|length == 0 %}
+        No significant changes.
+
+
+      {% else %}
+
+      {% endif %}
+
+    {% endfor %}
+
+  {% else %}
+    No significant changes.
+
+
+  {% endif %}
+{% endfor %}
--- a/doc/source/development.rst
+++ b/doc/source/development.rst
@@ -10,6 +10,13 @@ Branches
 * The ``master`` branch is stable.  Major changes should be performed
   elsewhere.
 
+Managing the Changelog
+======================
+
+Please see `changelog/README.rst` for instructions.
+
+.. _changelog/README.rst: ./changelog/README.rst
+
 Release Engineering
 ===================
 
@@ -17,7 +24,7 @@ Pre-release
 -----------
 
 * Ensure the `GitHub Project`_ has all cards in the ``Done`` column.
-* Edit the :ref:`changelog`, based on the project ``Done`` column.
+* Ensure :ref:`changelog` entries present, based on the project ``Done`` column.
 * Follow the :ref:`testing` steps.
 
 Release
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,3 +6,41 @@ requires = [
   "wheel",
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.towncrier]
+package = "molecule"
+package_dir = "molecule"
+filename = "CHANGELOG.rst"
+directory = "changelog/"
+title_format = "molecule {version} ({project_date})"
+template = "changelog/_template.rst"
+
+  [[tool.towncrier.type]]
+  directory = "removal"
+  name = "Removals"
+  showcontent = true
+
+  [[tool.towncrier.type]]
+  directory = "deprecation"
+  name = "Deprecations"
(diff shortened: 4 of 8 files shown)


changelog/README.rst
```diff
@@ -0,0 +1,39 @@
+*********************
+Manging the Changelog
+*********************
+
+This directory contains "newsfragments" which are short files that contain a
+small **ReST**-formatted text that will be added to the next ``CHANGELOG``.
```
webknjaz: Can we please use `ReStructuredText` term? https://en.wikipedia.org/wiki/ReStructuredText
Google search for `ReST` gives a lot of misleading results. Even better: maybe link it to do SphinX docs page.

changelog/README.rst
```diff
@@ -0,0 +1,39 @@
...
+
+The ``CHANGELOG`` will be read by users, so this description should be aimed to
+molecule users instead of describing internal changes which are only relevant
```
webknjaz: I think internal changes can still be useful. I saw others putting those under a separate category, like `misc` or `internal`.

OTOH wrt the style, we should encourage fragment authors to follow the same principles as are useful for docstrings in Python or commit messages in Git. For example, use imperative sentences. Describe things on the high-level outlining effects of the changes, not implementation details.

changelog/README.rst
```diff
@@ -0,0 +1,39 @@
...
+to the developers.
+
+Make sure to use full sentences with correct case and punctuation, for
+example::
+
+    Fix issue with non-ascii messages from the ``warnings`` module.
```
webknjaz: This example doesn't give the reader an understanding of what the issue was. It's kinda blurry. I think it might be better to re-phrase it as:
```suggestion
    Fix broken output of non-ascii messages produced by the ``warnings`` module.
```

In some projects (tox?) they also encourage tagging authors of the change in their template.

webknjaz: I'd also like @acozine about the best wording as it's docs-related.

changelog/README.rst
```diff
@@ -0,0 +1,39 @@
...
+
+This directory contains "newsfragments" which are short files that contain a
+small **ReST**-formatted text that will be added to the next ``CHANGELOG``.
+
+The ``CHANGELOG`` will be read by users, so this description should be aimed to
+molecule users instead of describing internal changes which are only relevant
+to the developers.
+
+Make sure to use full sentences with correct case and punctuation, for
+example::
```
webknjaz: Maybe have explicit markup highlighting?
```suggestion
example:

.. code:: rst
```

changelog/README.rst
```diff
@@ -0,0 +1,39 @@
...
+If your PR fixes an issue, use that number here. If there is no issue, then
+after you submit the PR and get the PR number you can add a changelog using
+that instead.
+
+If you are not sure what issue type to use, don't hesitate to ask in your PR.
+
+``towncrier`` preserves multiple paragraphs and formatting (code blocks, lists,
+and so on), but for entries other than ``features`` it is usually better to
+stick to a single paragraph to keep it concise. You can install ``towncrier``
+and then run ``towncrier --draft`` if you want to get a preview of how your
```
webknjaz: So for this, we'll probably need to have some tox envs instead of asking user to do manual steps

changelog/_template.rst
```diff
@@ -0,0 +1,40 @@
...
+{% if sections[section] %}
+{% for category, val in definitions.items() if category in sections[section] %}
+
+{{ definitions[category]['name'] }}
+{{ underline * definitions[category]['name']|length }}
+
+{% if definitions[category]['showcontent'] %}
+{% for text, values in sections[section][category]|dictsort(by='value') %}
+{% set issue_joiner = joiner(', ') %}
+- {% for value in values|sort %}{{ issue_joiner() }}`{{ value }} <https://github.com/ansible/molecule/issues/{{ value[1:] }}>`_{% endfor %}: {{ text }}
```
webknjaz: https://github.com/ansible/molecule/pull/1963#discussion_r274381471

pyproject.toml
```diff
@@ -6,3 +6,41 @@ requires = [
   "wheel",
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.towncrier]
+package = "molecule"
+package_dir = "molecule"
+filename = "CHANGELOG.rst"
+directory = "changelog/"
+title_format = "molecule {version} ({project_date})"
```
webknjaz: Do we need `v` prefix for the version? Some people do this.

pyproject.toml
```diff
@@ -6,3 +6,41 @@ requires = [
   "wheel",
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.towncrier]
+package = "molecule"
+package_dir = "molecule"
+filename = "CHANGELOG.rst"
+directory = "changelog/"
```
webknjaz: We need to decide on the location: some projects have it under `docs/` to keep the project root cleaner. I personally don't care much, just wanted to point out the possibilities.
I'm also planning to use this value for a GitHub Action checking the presence of change fragments in PRs.

tox.ini
```diff
@@ -159,3 +159,12 @@ commands =
 whitelist_externals =
     rm
     {[testenv]whitelist_externals}
+
+[testenv:changelog]
+skip_install = false
+usedevelop = false
+description = Generate the changelog
+extras =
```
webknjaz: I'd rather not install all Molecule's dependencies when just looking at fragments by a third-party tool is needed...

changelog/README.rst
decentral1se: Fair points. 

However, for the first attempt, let's just deal with our main priority: a user changelog.

pyproject.toml
decentral1se: I am not sure. We have not been exactly consistent on the release conventions. Once the change log is available, it will quickly become clear what should be done. I leave this as-is for now.

pyproject.toml
decentral1se: I just followed prior art of pytest, so will just leave it as is.

decentral1se:
FYI: still a draft but has addressed all previous comments and advanced to "nearly done".

tox.ini
```diff
@@ -198,3 +199,19 @@ commands =
 whitelist_externals =
     rm
     {[testenv]whitelist_externals}
+
+[testenv:changelog]
+description = Generate the changelog
+skip_install = False
```
decentral1se: @webknjaz, this is required because of the way `towncrier` works, as far as I can tell. It makes an import of the package and if it cannot, it cannot find the version. The alternative is passing `--version=<VERSION>` using the `setuptools_scm` hack we have for the docker build tag but I'd rather avoid. This seems to be the simplest solution.

decentral1se:
I'd like to declare what I will not be doing on this PR (out of scope):

  * Merging the existing change log with the new generated one
  * Making the new changelog look "just right"
  * Backporting existing unreleased changes into news fragments
  * Bikeshedding on further types we can include

Things I'll be focussing on:

  * Make tox changelog commands work correctly
  * Making sure all fragments generate correctly

tox.ini
webknjaz: ah... I see. OTOH supplying the version explicitly seems faster. It probably makes sense to post an FR for towncrier to allow specifying a callable...

tox.ini
```diff
@@ -6,6 +6,7 @@ envlist =
     py{27,35,36,37}-ansible{25,26,27,28}-{functional,unit}
     format-check
     doc
```
webknjaz: @decentral1se can we integrate changelog generation with doc generation? In such a case, we'd not need `changelog-check` here.

doc/source/development.rst
```diff
@@ -10,14 +10,21 @@ Branches
 * The ``master`` branch is stable.  Major changes should be performed
   elsewhere.
 
+Managing the Changelog
+======================
+
+Please see `changelog/README.rst` for instructions.
```
webknjaz: You should be able to do
```suggestion
Please see `<changelog/README.rst>`_ for instructions.
```

Ref: https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#external-links

tox.ini
webknjaz: @decentral1se after thinking more about it I think that it's quite challenging to make it work with version autodiscovery.
We should generate a changelog before tagging git commit but then it'll auto-discover `.dev` version, not the one we're going to release.

So we should probably have a `release` tox env accepting the desired (next) version as a positional argument and
1) generating a changelog for that version
2) committing it
3) adding a Git tag to that commit

webknjaz: Filed some ideas in upstream https://github.com/hawkowl/towncrier/issues/146

tox.ini
decentral1se: That's a fine idea for if the changelog generation was working smoothly but it is not even in place yet and there are a lot of steps to get it fully integrated. Meanwhile, documentation changes must remain easy to do. Let's do this later.

tox.ini
decentral1se: Hmm, looks tricky. I don't think I'll have time to finish this.

decentral1se:
I think I'll have to drop this for now. Hopefully someone can pick it up.

tox.ini
webknjaz: @decentral1se I've had a small chat with the maintainer of tox and virtualenv. They just don't care about shipping it in the dist now. They use draft mode and that's it. So maybe we could just do the same for now?

webknjaz:
@decentral1se maybe I'll pick it up later. I'm also coming to the PyCon US where I plan to have a series of discussions about release automation and GitHub automation in general. Also looking forward to promoting http://docs.octomachinery.dev which dramatically simplifies those.

Let's keep it open for now, shall we?

decentral1se:
> Let's keep it open for now, shall we?

It's tracked in https://github.com/ansible/molecule/issues/1779. 

I use my list of open issues/PRs to stay organised, so I prefer to close.

Issue #1727, opened 2019-02-06
Compatibility with Ansible devel (2.8) broken


pilou-:
# Issue Type

- Bug report

Since [7e92ff823e2bf97ee33c1b7ff87bb8acecb5ae77](https://github.com/ansible/ansible/commit/7e92ff823e2bf97ee33c1b7ff87bb8acecb5ae77), molecule doesn't work with ansible devel version.

Error is:
```
        ==================================== ERRORS ====================================
        ____________________ ERROR collecting tests/test_default.py ____________________
        tests/test_default.py:6: in <module>
            os.environ['MOLECULE_INVENTORY_FILE']).get_hosts('all')
        /home/travis/build/ansible/molecule/.tox/ansibledevel-functional/lib/python3.7/site-packages/testinfra/utils/ansible_runner.py:153: in __init__
            self.cli.options = self.cli.base_parser(
        E   AttributeError: 'PlaybookCLI' object has no attribute 'base_parser'
```

Issue was reported by @maxamillion on #ansible-molecule channel:
```
@maxamillion | can anyone advise what I'm doing wrong here? https://paste.fedoraproject.org/paste/6Q4OJP4CQxfzGU0HzqTpog
@maxamillion | or what I might be missing                                                                               
@maxamillion | I *think* it's a bug in testinfra, but I'm not certain                                                   
       Pilou | maxamillion: it seems base_parser has been removed (7e92ff823e2bf97ee33c1b7ff87bb8acecb5ae77)            
      @bcoca | maxamillion: tanner was saying this last week, ansible changed the internal options parser               
      @bcoca | and molecule uses it directly, which now breaks with devel                                               
      @bcoca | you can still use with <=2.7                                                                             
@maxamillion | oh right, I vaguely remember that conversation                                                           
```

singuliere:
I'll give this a try. Not sure to succeed but announcing anyways ;-)

singuliere:
Reproduced locally with

```
TOXENV='ansibledevel-functional' tox -- -k 'test_command_init_role[docker]' test/functional/test_command.py
```

singuliere:
The corresponding testinfra issue is https://github.com/philpep/testinfra/issues/401

lwm:
Thanks for investigating @singuliere. I've submitted https://github.com/ansible/molecule/pull/1733 to stop us getting failing builds on master during the cron runs that test against `ansible-devel`. Let's hope to see a fix in `testinfra` soon.

[...]

PR #1983, opened 2019-04-16
👷 Add basic testing against Ansible 2.8

webknjaz:
#### PR Type

- Testing Pull Request

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -44,7 +44,8 @@ extras =
     vagrant
     windows
 commands_pre =
-    ansible28: pip install --pre ansible==2.8b1
+    # TODO: move this to more generic deps once stable 2.8.0 is out
+    ansible28: pip install --pre 'ansible >= 2.8b1,<2.9'
     find {toxinidir} -type f -not -path '{toxinidir}/.tox/*' -path '*/__pycache__/*' -name '*.py[c|o]' -delete
 commands =
     unit: pytest test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {posargs}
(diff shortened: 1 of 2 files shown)


decentral1se:
Related: https://github.com/ansible/molecule/issues/1727.

webknjaz:
FTR: I've triggered the full test run in my fork.
It will actually run those hidden jobs with Ansible 2.8b1: https://travis-ci.com/webknjaz/molecule/builds/108597824

webknjaz:
@decentral1se jobs are failing as expected but they don't affect the overall build status and only fire on cron or manual trigger for now. I believe this is ready and we'll be able to follow-up and remove "allowed failures" mark from them once tests are fixed.

webknjaz:
P.S. anyone setting approved status next — feel free to merge immediately.

tox.ini
```diff
@@ -44,6 +44,7 @@ extras =
     vagrant
     windows
 commands_pre =
+    ansible28: pip install --pre ansible==2.8b1
```
ssbarnea: Replace it with `ansible>=2.8b1,<2.9` and it will be ok, as we do not want to bump it on each new pre-release.

ssbarnea: I also hope that final merge does not contain emoji. I don't mind them on Github, but I do not want to see them in commit messages.

webknjaz: FTR https://github.blog/2018-10-16-future-of-software/#suggested-changes-public-beta

webknjaz:
@gundalow it's the best you can do with tox now

ssbarnea:
@gundalow Please be more careful not to allow emoji into commit messages. That one already went in and we will never be able to get it out. I don't want to see git log looking like the xmas tree.

PR #1984, opened 2019-04-16
Scheduled weekly dependency update for week 15 / partial

webknjaz:
This is a re-submission of #1975 with a few partial reverts

#### PR Type

- Deps Pull Request


webknjaz:
We'd probably need to also revert `pytest-verbose-parametrize`...

webknjaz:
@ssbarnea that repo isn't private, it's upstream

webknjaz:
@ssbarnea complain @ https://github.com/manahl/pytest-plugins/issues/137

Issue #1548, opened 2018-11-03
please remove the signoff requirement (DCO)


ssbarnea:
# Issue Type

- Bug report

The DCO check is a  huge PITA as it requires everyone to use a very hard un unreliable workflow. Even the Ansible core team is not using it and I do not think that is molecule is such a high-profile / sensitive tools that really needs such an extra protection.

Not only that the automating it requires adding aliases for git-commit but these are NOT working when git is called from a non shell, like any git GUI client.

I tried the DCO github extension myself on another project and removed it after less than a day.

ssbarnea:
Update: I was able to find a way to make the Sign-off to kida work by adding a default commit template as mentioned on https://stackoverflow.com/questions/15015894/git-add-signed-off-by-line-using-format-signoff-not-working/34687806

Still, I think that if we still have to require the DCO, we need to clearly document how to configure it, and not by asking user to add aliases, as they work only on a small number of conditions. 

Git template seems more generic as based on my tests it seems that most git clients do respect them.

webknjaz:
I think it's a legal requirement. We'll keep it. As for docs, they could be improved, yes. Mind sending a PR?

cc @gundalow ^

ssbarnea:
Should I also mention that creating PRs from GitHub interface triggers the same DCO error?

See https://github.com/ansible/molecule/pull/1574 which was created by using github edit inteface. I am sure that I would not be the first to try this.

MarkusTeufelberger:
This came up in #1892 too, probably the discussion should be continued here.

MarkusTeufelberger:
> I think it's a legal requirement.

It might be a requirement by the RedHat legal team, but it is not a legal requirement. Legally (in the sense of: in front of a judge/mandated by law) it doesn't matter if a certain string shows up in a commit message or not. Also I could just add the string `Signed-off-by: John Doe <jdoe@example.com>` to any commit manually, the DCO doesn't mandate that I sign off with my actual name...

jelaplan:
I found this project when looking into whether others use this DCO bureaucratic nonsense. The statement "We'll keep it" is just so icky, yucky, gross. It things are a PITA then so what, you loser users can get over it.

ssbarnea:
@gundalow Can we sort this our once for all because the only outcome of the DCO implementation here is nothing else than a deterrent for getting contributions to the project. I prefer to avoid starting conspiratorial theories but if someone would want to kill a project, that is how it would start. 

[...]

PR #1989, opened 2019-04-17
Make sure that the DCO is applied to every commit

### LINKED ISSUE #1988: DCO requirement isn't checked for members of the Ansible organization
# Issue Type

- Bug report

# Desired Behavior

According to the "contribution guidelines", every contribution must be signed off.

# Actual Behaviour

Current Red Hat employees are not checked for including that stuff in their commit messages:
https://github.com/ansible/molecule/blob/master/.github/dco.yml

See https://github.com/probot/dco#skipping-sign-off-for-organization-members

MarkusTeufelberger:
Fixes #1988

#### PR Type

- Bugfix Pull Request

So far members of the "Ansible" organization were not always signing the DCO.

### DIFF
--- a/.github/dco.yml
+++ b/.github/dco.yml
@@ -1,2 +1,2 @@
 require:
-  members: false
+  members: true


webknjaz:
@MarkusTeufelberger I'm not sure whether it's needed legally: making commits as Red Hat employees seem to automatically cover that... (yet I may be wrong). So it's probably up to @gundalow to decide

MarkusTeufelberger:
I think this actually applies only to people in https://github.com/orgs/ansible/people - and they probably are not all employed by RedHat or have a contract with RH that all their contributions (at work and during off-work hours) are covered by this.

Even Linus Torvalds signs-off his commits...

MarkusTeufelberger:
How is this a feature that make contributions easier if it only affects 57 Github accounts?

I'm all for removing the DCO by the way. Until then, then existing rules for contributing should be followed though - they currently are not.

webknjaz:
@MarkusTeufelberger fair enough.

FTR, I've suggested @gundalow to implement a CLA bot (there's a few ready to use and easy to set up available in the wild) and he was going to check this with legal. Hopefully, we'll be able to replace DCO by a one-time signature which will address all those pain points...

MarkusTeufelberger:
Neither DCO nor CLA is required to contribute to https://github.com/ansible/ansible by the way. https://github.com/ansible/awx seems to require the DCO, but I don't find many signed-off commits there and I'm not sure if it is even possible to contribute as an external there or if anyone even tries to do that these days. The difference between external engagement between these 2 projects is striking and I partially also blame the DCO or similar restrictions.

A CLA would cover different things than a DCO, but it is a bit easier to set up at least. A bit harder to do in a corporate setting though (depending on the scope of the CLA). I also wouldn't feel comfortable to give up rights of my contributions to a place ultimately owned by IBM.

gundalow:
As discussed in IRC meeting 2019-04-17 I will discuss this with Red Hat's Legal Team.
Until then things will continue as-in.

decentral1se:
@jctanner can you by any chance follow this up to resolve it now? As far as I understand, gundalow will not be around due to holidays but we'd still like to be able to move forward on "paper work" issues like this ...

ssbarnea:
I am personally against this ticket as I am looking towards making occasional contribution as simple as posible. In fact I am for matching the Ansible contributuon policy, which has no such condition. I bet that the Ansible community would be quite unhappy if it would also be introduced there.

One major PITA with this is that it renders github web-edits useless as they would not have the DCO signature, making very hard to fix minor issues in existing PRs, or even trivial fixes which can be proposed from inside the browser (like bumping requirement versions).

MarkusTeufelberger:
I'm personally strongly against the DCO too, but this PR is about actually applying it as outlined in the DCO... There are other places where abandoning the signoff requirement can be discussed, this PR is about making sure that contributors (no matter from which organization!) are actually signing off their commit in accordance to the DCO. People leave companies and organizations all the time, it is often hard after the fact to research if someone was an employee of Red Hat for example and wrote the contribution in their capacity as employee.

decentral1se:
> There are other places where abandoning the signoff requirement can be discussed

Yes. Let's call that off-topic here and push that out to https://github.com/ansible/molecule/issues/1548.

ssbarnea:
I will close this as this PR is not actionable right now. We already have a ticket to track it and we can re-open it once we can make a decision.

MarkusTeufelberger:
What's the tracking ticket?

This PR is about enforcing the DCO properly (there are commits in this repo made after Red Hat took over without the DCO being signed!), not getting rid of it.

ssbarnea:
@MarkusTeufelberger  AFAIK https://github.com/ansible/molecule/issues/1548 is tracking that subject and I will do my best to get this DCO non-removed. Lets continue there, @gundalow is finally back and I hope we fix it next week.

MarkusTeufelberger:
There are still/again very recent commits in this repository that are not correctly signed-off. Please either reopen + merge this PR to make sure this doesn't happen without the bot complaining or remove the DCO as discussed in #1548 

The current situation is that some select people (everyone in the "Ansible" organization on GitHub) don't have their DCO requirements checked by the bot, so they seem to assume that the DCO is not applicable to them. This PR fixes this. This is perfectly actionable until a decision about the DCO as such has been taken.

PR #1990, opened 2019-04-17
Revert "Fix the CI: remove comment that Tox cannot parse correctly"

### LINKED ISSUE #1973: [BUG] tox currently doesn't parse deps with comments well
I'll need to check pyup docs on whether we could post that comment one line above or smth.
I've also filed a bug in tox: https://github.com/tox-dev/tox/issues/1260

TODO:
* [ ] restore filters once it's fixed

---------------------
oh.. this one seems wrong because it's parsed as a single line. we should turn it into a list

_Originally posted by @webknjaz in https://github.com/ansible/molecule/pull/1958_

webknjaz:
Reverts ansible/molecule#1977

Fixes #1973

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -26,7 +26,7 @@ deps =
     pytest-verbose-parametrize==1.4.0
     pytest-xdist==1.26.0
     shade==1.22.2
-    yapf>=0.25.0,<0.27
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-minversion = 3.8.6
+minversion = 3.9.0
```
gundalow: For other reviewers, `3.9.0` is the first release of `tox` that contains the https://github.com/tox-dev/tox/pull/1262 fix

PR #1995, opened 2019-04-18
Refactor tox envs

webknjaz:
* Linters can be run separately but also all at once

* Envs are logically separated by the main command/action they execute

#### PR Type

- Refactoring Pull Request

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -15,8 +15,8 @@ usedevelop = True
 passenv = *
 setenv =
     PYTHONDONTWRITEBYTECODE=1
-    unit: PYTEST_ADDOPTS=test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail
-    functional: PYTEST_ADDOPTS=test/functional/
+    unit: PYTEST_ADDOPTS=test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail {env:PYTEST_ADDOPTS}
+    functional: PYTEST_ADDOPTS=test/functional/ {env:PYTEST_ADDOPTS}
 deps =
     flake8>=3.6.0,<4
     yamllint>=1.11.1,<2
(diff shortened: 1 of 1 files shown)


seandst:
This is effectively what was decided in #1887, which I've been holding off on doing while @webknjaz is working in this space. Thanks for taking care of this.

webknjaz:
I agree that `pre-commit` is a good thing but it's unrelated + not all linters are wrapped with pre-commit configs which is a blocker (maybe temporary).

This PR doesn't attempt to change any workflows but to make `tox.ini` cleaner and better readable.

tox.ini
```diff
@@ -14,6 +15,8 @@ usedevelop = True
 passenv = *
 setenv =
     PYTHONDONTWRITEBYTECODE=1
+    unit: PYTEST_ADDOPTS=test/unit/ --cov={toxinidir}/molecule/ --no-cov-on-fail
```
ssbarnea: I am totally against defining the PYTEST_ADDOPTS variable which has a very different purpose: of allowing the *user* to customize/override the pytest execution at runtime. For example on my machine I do have this defined in shell rc file PYTEST_ADDOPTS='-s -x --no-cov' --- and works fine with hundreads of projects using tox.

webknjaz: fixed

tox.ini
```diff
@@ -74,17 +70,55 @@ usedevelop = false
...
     python -m yapf -d -r molecule/ test/
 deps = {[testenv:format]deps}
-extras = {[testenv:format]extras}
-skip_install = true
-usedevelop = false
+extras = {[testenv:lint-base]extras}
+skip_install = {[testenv:lint-base]skip_install}
+usedevelop = {[testenv:lint-base]usedevelop}
+
+[testenv:flake8]
```
ssbarnea: https://github.com/ansible/molecule/pull/1996 is the reson why I do not fancy this direction. Having one extra tox environment for each linter does not scale well. That change allows us to have only one tox environment for linting while being able to isolate each linter and still be able to run only one of them if we want.

webknjaz: This is unrelated to current PR.

PR #1996, opened 2019-04-19
Enable use of pre-commit for linting

ssbarnea:
pre-commit is framework used to orchestrating execution of many small
utilities (usually linters) which is also used by tox itself and should
not be seen as a replacement for it.

Despite its name pre-commit does not require installation of a git hook
in order to be used, this being an optional feature of the tool.

Among benefits we can count:

* lower disk footprint as cached linters are reused across repos
* must faster (even on CI due to ~/.cache/pre-commit)
* avoids conflicts between linters, each of them being isolated
* can run only changed files (w/o --all-files flag)
* pinned dependencies by design
* one line bumping support `pre-commit autoupdate`
* optional git hook (pre-commit install)
* allows user to run a single linter if they want

This change does not require any workflow changes for developers, as
`tox -e lint` will still run the linters.

#### PR Type
- Feature Pull Request

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,23 @@
+---
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.1.0
+    hooks:
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+      - id: mixed-line-ending
+      - id: check-byte-order-marker
+      - id: check-executables-have-shebangs
+      - id: check-merge-conflict
+      - id: debug-statements
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.7.7
+    hooks:
+      - id: flake8
+  - repo: https://github.com/adrienverge/yamllint.git
+    rev: v1.15.0
+    hooks:
+      - id: yamllint
+        files: \.(yaml|yml)$
+        types: [file, yaml]
+        entry: yamllint --strict
--- a/molecule/cookiecutter/scenario/driver/digitalocean/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
+++ b/molecule/cookiecutter/scenario/driver/digitalocean/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
@@ -4,7 +4,7 @@
   hosts: localhost
   connection: local
   gather_facts: false
-  no_log: "{{ molecule_no_log }}" 
+  no_log: "{{ molecule_no_log }}"
   vars:
     ssh_user: root
     ssh_port: 22
--- a/tox.ini
+++ b/tox.ini
@@ -1,8 +1,7 @@
 [tox]
 minversion = 3.9.0
 envlist =
-    flake8
-    yamllint
+    lint
     py{27,35,36,37}-ansible{25,26,27,28}-{functional,unit}
     format-check
     doc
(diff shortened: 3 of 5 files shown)


tox.ini
```diff
@@ -64,8 +62,10 @@ sitepackages = true
 # f'' is used inside docs
 basepython = python3
 deps =
-    flake8>=3.6.0,<4
-    yamllint>=1.15.0,<2
+    pre-commit>=1.15.2
+commands =
+    # to run a single linter you can do "pre-commit run flake8"
```
decentral1se: Maybe one for the contributing documentation? People will most likely go there first ...

webknjaz:
> I know next to nothing about this tool.

I hope you'll like it as we do :)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.7.7
+    hooks:
+      - id: flake8
+  - repo: https://github.com/adrienverge/yamllint.git
+    rev: v1.15.0
+    hooks:
+      - id: yamllint
+        files: \.(yaml|yml)$
+        types: [file, yaml]
```
webknjaz: I believe this option is unnecessary as the hook's own config should address this.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+        entry: yamllint --strict
```
webknjaz: Does this actually work? I thought this should go to the hook repo config unless it's local...

molecule/cookiecutter/scenario/driver/digitalocean/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
```diff
@@ -4,7 +4,7 @@
   hosts: localhost
   connection: local
   gather_facts: false
-  no_log: "{{ molecule_no_log }}" 
+  no_log: "{{ molecule_no_log }}"
```
webknjaz: Could you please submit these changes separately?

tox.ini
webknjaz: Not exactly. The documentation should mention tox as in `tox -e pre-commit -- flake8`.
Anyway, this needs updating because of a merge conflict. So @ssbarnea should get a chance to rethink it.

tox.ini
```diff
@@ -64,8 +62,10 @@ sitepackages = true
...
+    python -m pre_commit run {posargs:--all}
```
webknjaz: @ssbarnea please add this as well: https://github.com/tox-dev/tox/blob/a338734/tox.ini#L61

.pre-commit-config.yaml
```diff
@@ -0,0 +1,23 @@
+---
```
webknjaz: I believe we need an extra config for `yapf` check and maybe other "heavy" ones. e.g. `.pre-commit-config-heavy.yaml` and an additional tox env for heavy checks.

ssbarnea: Well... arguably, we could do anything in a one linee of `bash -c "..."` but this would not make it easy to use. If the user needs to add extra parameters (especiall long ones) is likely for use to make the UX worse than better.

Just to clarify: I am not againt the idea, just that I do not see many benefits. The good part is that we can easily postpone the yapf work for another change.

webknjaz: I have a strong feeling that you didn't get the idea... I'm talking about a separate tox env with `pre-commit --config heavy.yaml`

PR #1997, opened 2019-04-19
Change selinux workaround to avoid use of sitepackages during normal testing

ssbarnea:
Instead of using sitepackages in order to avoid the missing selinux inside them, we now install the selinux pure-python shim wheel which dynamically loads the bindings from their system location.

This allows us to avoid problems related to use of sitepackages with tox, where conflicts can occur or less easy to see use of wrong version of the tools.

This change does not change molecule runtime requirements as selinux is soft-dependecy of ansible which requires it only when using docker.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -65,6 +65,13 @@ available on your controller or host machines.
   highly recommended to upgrade user setuptools even when using a proper
   virtualenv as shown above.
 
+Keep in mind that on selinux supporting systems, if you install into a virtual
+environment, you may face https://github.com/ansible/ansible/issues/34340 even
+if selinux is not enabled or is configured to be permissive.
+
+It is your reponsability to assure that soft dependencies of Ansible are
+available on your controller or host machines.
+
 Requirements
 ------------
 
--- a/tox.ini
+++ b/tox.ini
@@ -27,6 +27,8 @@ deps =
     #git+https://github.com/manahl/pytest-plugins.git@54ca3c0#egg=pytest-verbose-parametrize&subdirectory=pytest-verbose-parametrize
     pytest-xdist<1.28.0  # pyup: < 1.28 # blocked by `pytest`
     shade==1.31.0
+    # https://github.com/ansible/molecule/issues/1724
+    selinux; platform_system=='Linux'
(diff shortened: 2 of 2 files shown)


doc/source/installation.rst
```diff
@@ -65,6 +65,13 @@ available on your controller or host machines.
   highly recommended to upgrade user setuptools even when using a proper
   virtualenv as shown above.
 
+Keep in mind that on selinux supporting systems, if you install into a virtual
+environment, you may face https://github.com/ansible/ansible/issues/34340 even
```
decentral1se: Probably better to: 1) make this a link with RST syntax 2) explain the issue briefly.

doc/source/installation.rst
```diff
@@ -65,6 +65,13 @@ available on your controller or host machines.
...
+if selinux is not enabled or is configured to be permissive.
+
+It is your reponsability to assure that soft dependencies of Ansible are
```
decentral1se:
```suggestion
It is your responsibility to assure that soft dependencies of Ansible are
```

tox.ini
```diff
@@ -27,6 +27,8 @@ deps =
     #git+https://github.com/manahl/pytest-plugins.git@54ca3c0#egg=pytest-verbose-parametrize&subdirectory=pytest-verbose-parametrize
     pytest-xdist<1.28.0  # pyup: < 1.28 # blocked by `pytest`
     shade==1.31.0
+    # https://github.com/ansible/molecule/issues/1724
+    selinux; platform_system=='Linux'
```
decentral1se: Concerns:

* https://github.com/ssbarnea/selinux/blob/master/setup.py#L32
* https://github.com/ssbarnea/selinux/issues/1
* Missing some setuptools configuration like `python_requires` etc.

Can you give some guarantees on this package before we adopt it?

doc/source/installation.rst
webknjaz: [1] FWIW it should be auto-linked anyway. So not a big deal to me.

doc/source/installation.rst
webknjaz: // side note: integrate a spell checker.
The one we use @ aiohttp is based on libenchant: https://github.com/aio-libs/aiohttp/blob/9504fe2/.travis.yml#L54

ssbarnea:
Obsoleted by another change.

PR #2007, opened 2019-04-24
Documentation for Docker With Non-Privileged User

ibeex:

#### PR Type

- Docs Pull Request

make.bat always showed changes so I did
https://help.github.com/en/articles/dealing-with-line-endings


doc/make.bat
```diff
@@ -1,263 +1,263 @@
-@ECHO OFF
```
decentral1se: Ah, you'll have to drop this change (from https://github.com/ansible/molecule/pull/2004).

webknjaz:
FTR #2008 is moving files one dir level up.

doc/source/examples.rst
```diff
@@ -30,6 +30,59 @@ follows.
 
 .. _`quay.io`: https://quay.io/repository/ansible/molecule
 
+Docker With Non-Privileged User
+===============================
+
+Default Molecule Docker driver execute Ansible playbooks with root user. If your
+workflow require non-privileged user, some simple changes to ``molecule.yaml``
```
webknjaz: Just curious: what could require such use-case?

ibeex: In our organisation after initial configuration `root` user is disabled so `sudo` is only way. So if I don't test with `become: yes` it is only partial test since later I need to add `become` to some tasks. We are not only one after some short googling [random example](https://www.rubytreesoftware.com/resources/ansible-bootstrap-playbook/) there is more.

webknjaz: Okay, fair enough. Initially, I thought it's about the image we ship with the molecule itself :)

webknjaz:
@ibeex thanks for your contribution! As a result of #2008 we now have docs moved. Would you mind updating this PR accordingly by resolving the merge conflict?

PR #2008, opened 2019-04-24
Reduce docs dir nesting

webknjaz:
This also follows practices used in tox/virtualenv and other well-maintained projects and prints out how to view the docs afterward.

#### PR Type

- Docs Pull Request


tox.ini
```diff
@@ -122,12 +121,23 @@ commands =
...
+    python -m sphinx \
+      -a -n -W \
+      -b html --color \
+      -d "{toxworkdir}/docs_doctree" \
+      docs/ "{toxworkdir}/docs_out"
+
+    # Print out the output docs dir and a way to serve html:
+    python -c \
+    'import pathlib; '\
+    'docs_dir = pathlib.Path(r"{toxworkdir}") / "docs_out"; index_file = docs_dir / "index.html"; print(f"\nDocumentation available under `file://\{index_file\}`\n\nTo serve docs, use `python3 -m http.server --directory \{docs_dir\} 0`\n")'
```
decentral1se: I would rather see this in the documentation instead than here for maintenance purposes.

decentral1se: Perhaps this would be a good candidate for a more readable `./bin/output-docs-path` script? We have some other stuff in this file (like the setuptools_scm version generation) that might be more easy to maintain if it lived in a script directory. Maybe for another change ...

webknjaz: I have to think about this. Ideally, I'd not want to maintain this away from the context which is concentrated right here. I with `tox` had better support for multiline commands...
Currently, I'd want to avoid introducing yet another directory...

PR #2014, opened 2019-04-26
Add a CLI command for collections initialization

webknjaz:
This is focused on the collection template generation only.
Everything else is out of scope now and should be added separately.

#### PR Type

- Feature Pull Request


MarkusTeufelberger:
Shouldn't the cookiecutter paths be `molecule/cookiecutter/collection/{{cookiecutter.collection_name}}/README.md ` etc. instead of `role_name`?

molecule/command/init/collection.py
```diff
@@ -0,0 +1,191 @@
...
+        'dependency_name': dependency_name,
+        'driver_name': driver_name,
+        'lint_name': lint_name,
+        'provisioner_name': provisioner_name,
+        'scenario_name': command_base.MOLECULE_DEFAULT_SCENARIO_NAME,
+        'subcommand': __name__,
+        'verifier_name': verifier_name,
+    }
+
+    verifier_linter_map = {
```
MarkusTeufelberger: Is testinfra missing here on purpose?

molecule/cookiecutter/collection/{{cookiecutter.role_name}}/README.md
```diff
@@ -0,0 +1,28 @@
...
+│
+├─ `plugins/`: all ansible plugins and modules go here, each in its own subdir
+│  ├─ `modules/`: ansible modules
+│  ├─ `lookups/`: lookup plugins
+│  ├─ `filters/`: Jinja2 filter plugins
+│  └─ `...`: rest of plugins
+│
+├─ `README.md` or `README.rst`: this file
+├─ `roles/`: directory for ansible roles
+└─ `tests/`: tests for the collection's content
```
MarkusTeufelberger: Tests meaning:

* Unit tests for modules/filters/lookups
* Integration tests for applying more than one of the roles in the collection
* Both?

MarkusTeufelberger:
The things that are still missing from collections seem to be inventories, vaults, host/group_vars and (maybe) Ansible config files. I guess this is by design ("Take a collection, add an inventory of hosts you control and you're good to go!").

At least group variables and something like inventory templates would be useful to have though, since in many cases these are not really specific and a bit weird to define otherwise. Looking at https://github.com/kubernetes-sigs/kubespray (a collection of Ansible roles + some glue code to set up a Kubernetes cluster) for example, they had to write their own inventory builder script and their roles often do awkward things like checking if they are currently running on a host that's in group x but not y. This mapping (which role runs on which host) should imho happen in playbooks, not roles themselves, because they then become too brittle.

Is there a place where collections are being discussed/designed that these concerns fit better than this implementation PR?

molecule/cookiecutter/collection/{{cookiecutter.role_name}}/README.md
webknjaz: Everything + `molecule.yml` will go there as well. I just didn't do this in current PR

molecule/command/init/collection.py
webknjaz: It's copied from role init and collapsed into a dict. I didn't invent anything different...

webknjaz:
> Shouldn't the cookiecutter paths be `molecule/cookiecutter/collection/{{cookiecutter.collection_name}}/README.md ` etc. instead of `role_name`?

Yes and no. This comment explains why I did this: https://github.com/ansible/molecule/pull/2014/files#diff-269ffcf909865f5963a7c062f4b741f3R61.
Long story short, it combines multiple cookiecutter templates overlaying each other. And other templates rely on the `role_name` var. So I've decided that refactoring of that should be separated.

webknjaz:
> I guess this is by design ("Take a collection, add an inventory of hosts you control and you're good to go!").

It's more like "Install collections from Galaxy, use contents from them". Collections are turned into artifacts (tarballs with metadata) and uploaded to Galaxy, unlike roles with just pointers to GitHub. One collection can contain different content types. Yet, this doesn't mean that the user has to use everything from the collection. Think of it as a new way of distributing the content where there's a lot more of the first-class citizens.

Here's a few demo recordings mentioning Collections: https://asciinema.org/a/241619 / https://www.ansible.com/resources/webinars-training/ansible-developer-ecosystem-story

> Is there a place where collections are being discussed/designed that these concerns fit better than this implementation PR?

Collections have a spec and a lot of discussions in our internal Slack but I'm not sure whether there's a public doc about them. Basically, the intention is to make Molecule collection template generator the source of truth + build dev/test/publish workflows around it.

The structure for an Ansible Collection is standardized and Ansible Engine 2.8.0a0 currently understands how to pick up content from collections as per spec.
As for the Collection development workflow we still need to figure out how to best integrate it with Molecule.

This PR is based on https://github.com/bcoca/collection.

decentral1se:
Concern from altest WG meeting: how is this going to get into circulation. Will there be a "collections" branch or will it be a separate code path that can live on master but not potentially break things? There were some concerns regarding having this code on master and not having collections functionality in the latest Ansible release (just on a branch too). Hope I've capture that ...

Thanks for this work and I will review when it gets past draft status.

MarkusTeufelberger:
Could be behind a feature flag I guess, so `molecule init collection` requires a (secret/undocumented?) `--experimental-enable-collections` parameter to be set until it is out of alpha.

jctanner:
@MarkusTeufelberger for general discussion of collections and their implementation, you can use #ansible-devel on irc.freenode.net or the ansible development mailing list https://groups.google.com/forum/#!forum/ansible-devel
https://docs.ansible.com/ansible/latest/community/communication.html

If you find bugs or want to submit general feature requests, please file them at https://github.com/ansible/ansible/issues

PR #2015, opened 2019-04-27
Improve Vagrant prepare.yml as LXD

hswong3i:
Clone LXD prepare.yml for Vagrant.

See https://github.com/ansible/molecule/blob/master/molecule/provisioner/ansible/playbooks/lxd/prepare.yml

#### PR Type

- Feature Pull Request

### DIFF
--- a/molecule/cookiecutter/scenario/driver/vagrant/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/prepare.yml
+++ b/molecule/cookiecutter/scenario/driver/vagrant/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/prepare.yml
@@ -2,10 +2,31 @@
 {% raw -%}
 - name: Prepare
   hosts: all
+  become: true
   gather_facts: false
   tasks:
-    - name: Install python for Ansible
-      raw: test -e /usr/bin/python || (apt -y update && apt install -y python-minimal)
-      become: true
-      changed_when: false
+    - name: Install some basic packages because Vagrant machines are bare
+      raw: |
+        if [ -x "$(command -v apt-get)" ]
+        then
+          apt-get update && apt-get install -y ca-certificates curl python sudo
+        elif [ -x "$(command -v yum)" ]
+        then
+          yum install -y ca-certificates curl python sudo
+        elif [ -x "$(command -v zypper)" ]
+        then
+          zypper -n --gpg-auto-import-keys refresh && zypper -n install -y ca-certificates ca-certificates-cacert ca-certificates-mozilla curl python python-xml sudo
+        elif [ -x "$(command -v apk)" ]
+        then
+          apk update && apk add ca-certificates curl python sudo
+        elif [ -x "$(command -v pacman)" ]
+        then
+          pacman -Syu --noconfirm ca-certificates curl python2 sudo
+        elif [ -x "$(command -v dnf)" ]
+        then
+          dnf --assumeyes install ca-certificates curl python python-devel python2-dnf sudo
+        elif [ -x "$(command -v emerge)" ]
+        then
+          emerge --ask n ca-certificates curl dev-lang/python-2 gentoolkit sudo
+        fi
 {%- endraw %}
--- a/molecule/provisioner/ansible/playbooks/vagrant/prepare.yml
+++ b/molecule/provisioner/ansible/playbooks/vagrant/prepare.yml
@@ -1,5 +1,4 @@
 ---
-{% raw -%}
 - name: Prepare
   hosts: all
   become: true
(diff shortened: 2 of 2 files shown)


hswong3i:
IMHO, we may relocate it as https://github.com/ansible/molecule/blob/master/molecule/provisioner/ansible/playbooks/vagrant/prepare.yml

molecule/cookiecutter/scenario/driver/vagrant/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/prepare.yml
```diff
@@ -1,11 +0,0 @@
----
```
decentral1se: Why is this deleted? I thought we needed this in the cookiecutter folder?

MarkusTeufelberger: The suggested change is that molecule will use a built-in playbook for the `prepare` step on `vagrant` instead of an explicitly rendered (default) playbook.

molecule/provisioner/ansible/playbooks/vagrant/prepare.yml
```diff
@@ -0,0 +1,30 @@
...
+- name: Prepare
+  hosts: all
+  become: true
+  gather_facts: false
+  tasks:
+    - name: Install some basic packages because Vagrant machines are bare
+      raw: |
+        if [ -x "$(command -v apt-get)" ]
+        then
+          apt-get update && apt-get install -y ca-certificates curl python sudo
```
MarkusTeufelberger: This (and the similar tasks below) installs different packages that are not required by Ansible by default compared to before with very little insight for the user by default which ones are actually being installed.

I'd rather see the LXD provider become more explicit than moving the `prepare` step to something internal to molecule...

MarkusTeufelberger:
By the way, I still need a way to NOT install anything through molecule itself to test roles that prepare hosts to be configured by Ansible (such as this preparation playbook that you propose).

hswong3i:
When developing my Ansible roles (https://github.com/alvistack) with Molecule + LXD, previously I need to manually include my https://github.com/alvistack/ansible-role-bootstrap for almost all other roles in order to have something very basic, e.g. curl sudo python etc.

If not including my bootstrap role, then I will need to manually override every roles Molecule prepare.yml which make it looks so complicated...

IMHO the build in prepare.yml with these packages greatly improve the usability and simplify the complicity (ie 80% use cases); on the other hand, we could still manually override the prepare.yml to vanilla style if looking for clean and pure kickstart (ie 20% use cases).

MarkusTeufelberger:
With LXD you can just build a container image that contains your changes and use that one...?

sudo and curl for example are not required to run Ansible at all. They might be useful tools, but it shouldn't be expected that these are installed or available when writing roles. It would be different if you modify the template for `prepare.yml` (because so far it is just defaulting to using `apt` - because by default molecule will use an Ubuntu VM on vagrant). You are suggesting to install additional packages in a rather hidden way to every VM I launch that neither Ansible nor molecule require.

There are other packages too that most people assume are always present (`bash`, `glibc`, `ssh(d)`...), but this is actually not the case in general. The whole point of tools like molecule is that default installations are rather bare bones in some cases, so people need to be aware in writing their roles that they either need to depend on a bootstrapping role (you can include your role via galaxy for example or just add links to the `prepare.yml` file in the `default` scenario) or write their role(s) in a way that installs all dependencies first before configuring packages.

Again: I'd rather move https://github.com/ansible/molecule/blob/master/molecule/provisioner/ansible/playbooks/lxd/prepare.yml to the cookiecutter template (so far it is the only provider that has `prepare.yml` built in) and I'd be willing to discuss extending the vagrant cookiecutter `prepare.yml` template to something that does a bit more on more platforms, but I don't like the idea of adding more things than absolutely necessary to test machines (e.g. Ansible actually should only need `python-minimal` on Debian based distros, not `python`...), especially not by default and secretly.

hswong3i:
> Again: I'd rather move https://github.com/ansible/molecule/blob/master/molecule/provisioner/ansible/playbooks/lxd/prepare.yml to the cookiecutter template (so far it is the only provider that has prepare.yml built in) and I'd be willing to discuss extending the vagrant cookiecutter prepare.yml template to something that does a bit more on more platforms, but I don't like the idea of adding more things than absolutely necessary to test machines (e.g. Ansible actually should only need python-minimal on Debian based distros, not python...), especially not by default and secretly.

From my point of view, at least if we could provide a sensible "reference" prepare.yml for both use cases, e.g. Docker/LXD/Vagrant, and simplify the difficulties for new comer to understand the "well-known" limitation, keep the changes into cookiecutter template should be a good idea too ;-)

Let's roll this PR back to changes for Vagrant's cookiecutter template, and so another PR for reference prepare.yml for LXD/Docker?

ssbarnea:
I am closing this because it was not updated in very long time.

PR #1882, opened 2019-03-26
Add driver sanity checks (only Docker for now ...)

decentral1se:
Based on point in https://github.com/ansible/molecule/issues/1666#issuecomment-454588682.

Main goals of this change are: implement sanity checks that stop the sequence before anything happens if the driver is missing anything it needs to successfully run the sequence. This will help users get faster feedback when facing the litany of issues that come with using a driver (for the first time, especially ...).

PR #2016, opened 2019-04-28
Disable no_log on creation of resources

MarkusTeufelberger:

#### PR Type

- Feature Pull Request

So far when creating/destroying resources, the output only looked like this:

```
--> Action: 'destroy'
    
    PLAY [Destroy] *****************************************************************
    
    TASK [Destroy molecule instance(s)] ********************************************
    changed: [localhost] => (item=None)
    changed: [localhost] => (item=None)
    changed: [localhost] => (item=None)
    changed: [localhost]
    
    TASK [Populate instance config] ************************************************
    ok: [localhost]
    
    TASK [Dump instance config] ****************************************************
    changed: [localhost]
    
    PLAY RECAP *********************************************************************
    localhost                  : ok=3    changed=2    unreachable=0    failed=0
```

This PR disables `no_log` on resource creation/destruction and adds a label to the loop instead (so the whole config isn't always dumped to stdout).

As far as I understand it, this might still leak some information compared to the current situation when something fails (probably the whole `item` gets logged then), so I'm not so sure if this PR is really that harmless. OTOH it is really frustrating to see only `(item=None)` there and if it fails you don't even know which container/VM is the culprit and you have to hope it fails on the next run too so you can run molecule in debug mode instead.

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -44,6 +44,9 @@
         pull: "{{ item.item.pull | default(omit) }}"
         buildargs: "{{ item.item.buildargs | default(omit) }}"
       with_items: "{{ platforms.results }}"
+      loop_control:
+        label: "molecule_local/{{ item.item.image }}"
+      no_log: false
       when:
         - platforms.changed or docker_images.results | map(attribute='images') | select('equalto', []) | list | count >= 0
         - not item.item.pre_build_image | default(false)
@@ -54,6 +57,9 @@
         docker_host: "{{ item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
         state: present
       with_items: "{{ molecule_yml.platforms | molecule_get_docker_networks }}"
+      loop_control:
+        label: "{{ item }}"
+      no_log: false
 
     - name: Determine the CMD directives
       set_fact:
@@ -93,6 +99,9 @@
         tty: "{{ item.tty | default(omit) }}"
       register: server
       with_items: "{{ molecule_yml.platforms }}"
+      loop_control:
+        label: "{{ item.name }}"
+      no_log: false
       async: 7200
       poll: 0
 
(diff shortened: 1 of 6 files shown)


znerol:
There is the option to override `no_log` by setting the environment variable `MOLECULE_NO_LOG` to `false`. I used the following command to inspect `docker build` failures in the past:

```
MOLECULE_NO_LOG=False molecule converge
```

decentral1se:
Thanks for making an effort with this frustrating issue on user experience.

> As far as I understand it, this might still leak some information

Unfortunately, as seen in and around the debates on https://github.com/ansible/molecule/pull/1828 (and followed on IRC), this is a blocker for any change related to the logging on resource creation/deletion. I sympathise with this frustration and have been trying to remedy it (see also my efforts in #1882 to stop before anything goes wrong) but you'll have to provide guarantees that secrets will not be leaked (maintain current behaviour).


Also see https://github.com/ansible/molecule/pull/2016#issuecomment-487694308.

decentral1se:
> I'd at least like to get the loop_control label in (otherwise this is really spammy)

Let's aim for that then in this PR, no?

> would argue for MOLECULE_NO_LOG=False as a default value (if someone wants to run this in CI with secrets in some weird places, they can set this to True).

Fancy opening a https://github.com/ansible/molecule/blob/master/.github/ISSUE_TEMPLATE/proposal.md? I'm interested in seeing if we can bring this forward.

jctanner:
@MarkusTeufelberger can you or anyone else who's testing this please paste an example stdout from molecule after this patch has been applied?

znerol:
[master-create.txt](https://github.com/ansible/molecule/files/3250747/master-create.txt)
[master-destroy.txt](https://github.com/ansible/molecule/files/3250748/master-destroy.txt)
[pr-create.txt](https://github.com/ansible/molecule/files/3250749/pr-create.txt)
[pr-destroy.txt](https://github.com/ansible/molecule/files/3250751/pr-destroy.txt)

Looks good. Tasks `Build an Ansible compatible image` and `Create molecule instance(s)` as well as `Destroy molecule instance(s)` display useful labels now.

seandst:
Looking at/eyeball-diffing those logs, I think the label values are an improvement, and an idea worth considering applying to other user-facing looped tasks.

It doesn't look like we're passing anything auth-related in the tasks that are now explicitly `no_log: false`, so those are probably (?) fine.

Issue #2023, opened 2019-05-02
Dependencies too constrained


micheelengronne:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible-2.7.10
molecule-2.20.1
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

Tests run on a Docker ubuntu:bionic with python3.6

The command is:

    pip3 install --upgrade -r requirements.txt

the requirements.txt file contains:

```
ansible
ansible-lint
ansible-bender
molecule
yamllint
docker
```

# Desired Behavior

pip dependencies should be resolved without any conflict.

# Actual Behaviour

the install gives:

```
ERROR: molecule 2.20.1 has requirement Jinja2==2.10, but you'll have jinja2 2.10.1 which is incompatible.
ERROR: molecule 2.20.1 has requirement PyYAML==3.13, but you'll have pyyaml 5.1 which is incompatible.
ERROR: molecule 2.20.1 has requirement six==1.11.0, but you'll have six 1.12.0 which is incompatible.
ERROR: molecule 2.20.1 has requirement tabulate==0.8.2, but you'll have tabulate 0.8.3 which is incompatible.
```

molecule constraints its dependencies too much as dependencies with newer minor versions (Semver syntax) conflict.

decentral1se:
See https://github.com/ansible/molecule/issues/1831 (this is probably a duplicate).

micheelengronne:
Yep.

Issue #2022, opened 2019-05-01
Broken unit test: `{'driver': [{'provider': [{'name': ['unallowed value None',`


decentral1se:
See https://travis-ci.com/ansible/molecule/jobs/196973204#L1232. Saw this on latest CI run of the Hetzner driver PR while rebasing. Remined me of https://github.com/ansible/molecule/pull/1714#pullrequestreview-230537350. Not sure of cause ...

Anyone got time for this?

> 
> =================================== FAILURES ===================================
> _ test_driver_provider_name_not_nullable_when_vagrant_driver[_model_driver_provider_name_n...] _
> _config = {'dependency': {'name': 'galaxy'}, 'driver': {'name': 'vagrant', 'provider': {'name': None}}, 'lint': {'name': 'yamllint'}, 'platforms': [{'image': 'centos:latest', 'name': 'instance'}], ...}
>     @pytest.mark.parametrize(
>         '_config',
>         ['_model_driver_provider_name_not_nullable_when_vagrant_section_data'],
>         indirect=True)
>     def test_driver_provider_name_not_nullable_when_vagrant_driver(_config):
>         x = {
>             'driver': [{
>                 'provider': [{
>                     'name': ['unallowed value None', 'null value not allowed']
>                 }]
>             }]
>         }
>     
> >       assert x == schema_v2.validate(_config)
> E       AssertionError: assert {'driver': [{...allowed']}]}]} == {'driver': [{'...allowed']}]}]}
> E         Differing items:
> E         {'driver': [{'provider': [{'name': ['unallowed value None', 'null value not allowed']}]}]} != {'driver': [{'provider': [{'name': ['null value not allowed']}]}]}
> E         Full diff:
> E         - {'driver': [{'provider': [{'name': ['unallowed value None',
> E         ?                                               ----------- ^
> E         + {'driver': [{'provider': [{'name': ['null value not allowed']}]}]}
> E         ?                                      + +++++++++ +++        ^^^^^^
> E         -                                     'null value not allowed']}]}]}

micheelengronne:
The behaviour of this test was already changed twice.

Here https://github.com/ansible/molecule/commit/cd2ea98325bd1a418371eb8a24b0e2a1b5952f05#diff-d3601755c38e1ab64638083914501041

And here https://github.com/ansible/molecule/commit/bd69c6d9af6bc9683fd3eda74f30d9060de878e7#diff-d3601755c38e1ab64638083914501041

[...]

PR #2024, opened 2019-05-02
Less constrained dependencies

micheelengronne:
Dependencies are currently too constrained and conflict with other ansible projects dependencies.

#### PR Type

- Bugfix Pull Request for https://github.com/ansible/molecule/issues/2023

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -80,10 +80,10 @@ install_requires =
     Jinja2 >= 2.10.1
     pexpect >= 4.6.0, < 5
     psutil == 5.4.6; sys_platform!="win32" and sys_platform!="cygwin"
-    PyYAML == 3.13
+    PyYAML >= 3.13
     sh == 1.12.14
-    six == 1.11.0
-    tabulate == 0.8.2
+    six >= 1.11.0
+    tabulate >= 0.8.2
     testinfra == 1.19.0
     tree-format == 0.1.2
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -71,7 +71,7 @@ install_requires =
     ansible-lint >= 4.0.2, < 5
 
     anyconfig == 0.9.7
-    cerberus >= 1.2, < 3
+    cerberus <= 1.3
     click == 6.7
     click-completion == 0.3.1
     colorama == 0.3.9
--- a/setup.cfg
+++ b/setup.cfg
@@ -77,13 +77,13 @@ install_requires =
     colorama == 0.3.9
     cookiecutter == 1.6.0
     python-gilt >= 1.2.1, < 2
-    Jinja2 >= 2.10.1
+    Jinja2 >= 2.10.1, < 3
     pexpect >= 4.6.0, < 5
     psutil == 5.4.6; sys_platform!="win32" and sys_platform!="cygwin"
-    PyYAML >= 3.13
+    PyYAML >= 3.13, < 6
     sh == 1.12.14
-    six >= 1.11.0
-    tabulate >= 0.8.2
+    six >= 1.11.0, < 2
+    tabulate >= 0.8.2, < 1
     testinfra == 1.19.0
     tree-format == 0.1.2
 
--- a/test/unit/model/v2/test_driver_section.py
+++ b/test/unit/model/v2/test_driver_section.py
@@ -134,13 +134,7 @@ def _model_driver_provider_name_not_nullable_when_vagrant_section_data():
     ['_model_driver_provider_name_not_nullable_when_vagrant_section_data'],
     indirect=True)
 def test_driver_provider_name_not_nullable_when_vagrant_driver(_config):
-    x = {
-        'driver': [{
-            'provider': [{
-                'name': ['unallowed value None', 'null value not allowed']
-            }]
-        }]
-    }
+    x = {'driver': [{'provider': [{'name': ['null value not allowed']}]}]}
 
     assert x == schema_v2.validate(_config)
 


micheelengronne:
Agreed, except for pyyaml that goes from major version 3 to 5.

micheelengronne:
I added an upper limit. But it is the widest as it takes only major version into account. Should it be reduced to minor as well ?

decentral1se:
> But it is the widest as it takes only major version into account. 

As far as I know, that's what we've done so far. Thanks!

micheelengronne:
You welcome. As it is currently blocking my stack and my roles development, do you know the next release date where this merge will be ?

decentral1se:
Not sure. https://github.com/ansible/molecule/issues/2022 will block the CI, I think ...

micheelengronne:
Hum, CI failed because of configuration issues in the travis file unrelated to that MR.

Weird.

micheelengronne:
My bad. I misread the tests results. The error is still not related to this MR but is what https://github.com/ansible/molecule/pull/2025 tries to solve.

webknjaz:
@micheelengronne as far as I remember, I kept those specific deps pinned because they don't follow semver.

Have you checked that every each of them is following semver? If not, you need to identify their breaking changes policies and set version boundaries accordingly.

micheelengronne:
I understand your philosophy but `PyYAML` and `Jinja2` with these versions are required by ansible itself on ubuntu:bionic. To keep these 2 pined breaks the installation of molecule with ansible making this project useless (how can we test ansible roles without ansible ?).

`six` with this version is required by ansible-lint on ubuntu:bionic that molecule uses.

`tabulate` with this version is required by ansible-bender on ubuntu:bionic so we can have a separate environment for ansible-bender and molecule as they do not interact.

I am totally with you on ensuring that SemVer is respected but molecule is not the main project (ansible is) so these rules should be enforced at ansible level and for all its subprojects.

From my point of view, molecule should rather prioritizing on compatibility with ansible.

In the current state, molecule is perfectly useless on ubuntu:bionic as we can not have it installed along ansible.

The Semver rules I put in the setup correspond to the current state of ansible required dependencies. I fixed the upper limit to the current major version but if I understand correctly ansible did not so we might need to change the upper limit as well.

micheelengronne:
As a side note, molecule does not work at all on debian:stretch but it is the Debian policy that keeps a completely outdated python3.

decentral1se:
> Have you checked that every each of them is following semver?

Ah, good point. Following from https://github.com/ansible/molecule/issues/2022#issuecomment-488682442, we might upper bound Cerberus on minor versions. To revise my comment above, yes, we need to be sure that projects are following semver if we're doing this (my assumption ...).

webknjaz:
@decentral1se exactly, when I was converting deps to be stored in `setup.cfg` I only set bounds based on the research of their policies.

We can still set upper bounds for non-semver compliant dists but we have to apply different rules (as in not using just a major version) for such limits.

@micheelengronne 
> but molecule is not the main project (ansible is) so these rules should be enforced at ansible level and for all its subprojects.

Nope. We are talking about the direct runtime deps of Molecule, not Ansible's. Ansible deps are declared there and those are transitive deps for Molecule unless specified otherwise.

micheelengronne:
@webknjaz hum, you mean that if I do not declare ansible itself in my requirements.txt, molecule will install a slightly outdated but compatible version ?

webknjaz:
@micheelengronne it's not Molecule, it's pip which is managing versions and it will resolve deps accordingly. In case of Ansible, we have `>= 2.5` set. Ansible has its own dist metadata.
Also, I don't understand why would you bring in `requirements.txt` here. That one is for pinning the env deps, not library deps.

micheelengronne:
because I installed all my pip packages with a requirements.txt file. It is more readable and easier to integrate in multiple systems (CI, doc, other projects...).

Does it do the same for ansible-lint ? Does molecule install it ?

I do not have time to check all possible breaking changes for these deps but if I can have my ansible test environment up and running I would be happy.

webknjaz:
> because I installed all my pip packages with a requirements.txt file. It is more readable and easier to integrate in multiple systems (CI, doc, other projects...).

Just to reiterate: libs are distributed as dists, when you install that into a virtualenv you want to lock that env pinning all of the transitive dependencies. That's fine, but use-cases are different. In the dist metadata of Molecule we declare "hey, Molecule needs these things to run properly". Yet, when building a venv for yourself, as a user you want it to be reproducible, but that's out of scope of a single app.

You should use `requirements.txt`, but that's not about the specific lib. It's about your env.

> Does it do the same for ansible-lint ? Does molecule install it ?

Pip installs things, not Molecule. And there's ansible-lint specified as a dependency so it'll be installed. Which doesn't mean that you shouldn't use `requirements.txt`.

You might want to read this first: https://caremad.io/posts/2013/07/setup-vs-requirement/

micheelengronne:
Yes, my bad. I just twisted words. Of course, I know that pip is in charge of installing and that a dist should not pin.

Hey, isn't it the problem here, btw, molecule dist pining too much :)

So ok. I will go with 2 env. One for molecule to test and one for ansible-bender to create Docker images.

decentral1se:
OK, #2025 is merged now. The latest master build should go green over at https://travis-ci.com/ansible/molecule/builds/110410528. You might want to rebase this and let's get this moving forward. Thanks for the patience.

decentral1se:
What's the status here? It should be rebased to remove the duplicated commit of https://github.com/ansible/molecule/pull/2025? I've restarted the CI build in any case ...

setup.cfg
```diff
@@ -71,19 +71,19 @@ install_requires =
     ansible-lint >= 4.0.2, < 5
 
     anyconfig == 0.9.7
-    cerberus >= 1.2, < 3
+    cerberus <= 1.3
```
webknjaz: Please don't remove the lower limit. It's bad.

test/unit/model/v2/test_driver_section.py
```diff
@@ -134,13 +134,7 @@ def _model_driver_provider_name_not_nullable_when_vagrant_section_data():
...
     indirect=True)
 def test_driver_provider_name_not_nullable_when_vagrant_driver(_config):
-    x = {
-        'driver': [{
-            'provider': [{
-                'name': ['unallowed value None', 'null value not allowed']
-            }]
-        }]
-    }
+    x = {'driver': [{'provider': [{'name': ['null value not allowed']}]}]}
```
webknjaz: This change is unrelated.

micheelengronne: I know. It was another issue that blocked the CI.

micheelengronne:
The CI is still blocked and as @webknjaz stated the SemVer of the dependencies is uncertain.

As I didn't check any of them specifically I think it is premature to merge this patch.

test/unit/model/v2/test_driver_section.py
ssbarnea: Please avoid touching more than one subject in a PR or it is likely that it will never be merged. Just make another PR for the other fix.

Due to the lack of time I am not even doing reviews that cover move than one area as I know we will not merge them.

micheelengronne: It is not related to this PR at all. I investigated this PR https://github.com/ansible/molecule/issues/2022 in response to this comment https://github.com/ansible/molecule/pull/2024#issuecomment-488607398 and created this MR https://github.com/ansible/molecule/pull/2025 for that. But, as I already said here https://github.com/ansible/molecule/pull/2024#issuecomment-488635238 I am well aware this is not related.

ssbarnea:
Closing as it seems outdated. I think that all cores would agree on "Less constrained dependencies" but this should be done on each dependency based on facts. It makes no sense to add ceiling for packages that never broke anything, like six.

PR #2025, opened 2019-05-02
Revert the null value test for Vagrant driver

micheelengronne:
To review with caution. Related to https://github.com/ansible/molecule/issues/2022

#### PR Type

- Bugfix Pull Request

This particular unit test seems broken. I reverted the current behavior for the third time after https://github.com/ansible/molecule/commit/cd2ea98325bd1a418371eb8a24b0e2a1b5952f05#diff-d3601755c38e1ab64638083914501041 and https://github.com/ansible/molecule/commit/bd69c6d9af6bc9683fd3eda74f30d9060de878e7#diff-d3601755c38e1ab64638083914501041

This should be carefully reviewed as it is a personal opinioned patch following https://github.com/ansible/molecule/issues/2022#issuecomment-488613124

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -71,7 +71,7 @@ install_requires =
     ansible-lint >= 4.0.2, < 5
 
     anyconfig == 0.9.7
-    cerberus >= 1.2, < 3
+    cerberus <= 1.3
     click == 6.7
     click-completion == 0.3.1
     colorama == 0.3.9
--- a/test/unit/model/v2/test_driver_section.py
+++ b/test/unit/model/v2/test_driver_section.py
@@ -134,13 +134,7 @@ def _model_driver_provider_name_not_nullable_when_vagrant_section_data():
     ['_model_driver_provider_name_not_nullable_when_vagrant_section_data'],
     indirect=True)
 def test_driver_provider_name_not_nullable_when_vagrant_driver(_config):
-    x = {
-        'driver': [{
-            'provider': [{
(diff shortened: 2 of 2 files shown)


decentral1se:
I think there are two options:

  1) Accept this PR and then bound Cerberus <= 1.3
  2) Bound Cerberus to <=1.2 and then the test should pass

I'm easy with 1 if the minor version bounding is the right approach?

micheelengronne:
As I discovered the molecule code literraly today, I don't know if there is any implications to my MR, hence the caution. So I can't answer to that.

decentral1se:
Sure. Given https://github.com/ansible/molecule/pull/2024#issuecomment-488686654, I think we're good to bound Cerberus to 1.3. It'd be great if you could do that in this PR (or the other). Sorry for all this back and forth now but I think we have the solutions now :)

setup.cfg
```diff
@@ -71,7 +71,7 @@ install_requires =
     ansible-lint >= 4.0.2, < 5
 
     anyconfig == 0.9.7
-    cerberus >= 1.2, < 3
+    cerberus <= 1.3
```
decentral1se: Maybe link to https://github.com/ansible/molecule/issues/2022#issuecomment-488682442 in a comment?

decentral1se:
Since this will fix the broken CI, I'll be merging it when it goes green. 

Please shout before if you have concerns.

PR #2030, opened 2019-05-07
Use selinux shim library

ssbarnea:
Revert the use of site-packages when testing with tox as this proved
to cause other issues and conflicts and use selinux shim module in order
to allow ansible to work correctly from inside the virtualenv.

It should be noted that this is more of a indirect and weak-runtime
dependency due to use of ansible templating with docker.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -57,11 +57,6 @@ commands =
     python -m pytest {posargs}
 whitelist_externals =
     find
-# Enabling sitepackages is needed in order to avoid encountering exceptions
-# caused by missing selinux python bindinds in ansible modules like template.
-# Selinux python bindings are binary and they cannot be installed using pip
-# in virtualenvs. Details: https://github.com/ansible/molecule/issues/1724
-sitepackages = true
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -36,6 +36,7 @@ deps =
     pytest-xdist<1.28.0  # pyup: < 1.28 # blocked by `pytest`
     shade==1.31.0
     yapf>=0.25.0,<0.27  # pyup: < 0.27 # disable updates, conflicts with flake8 as per https://github.com/ansible/molecule/pull/1915
+    selinux  # https://github.com/ansible/molecule/issues/1724
```
webknjaz: This linked issue doesn't explain anything. It's misleading to me. We need some actual comment here.

webknjaz:
Just one question: what will this influence?

webknjaz:
FTR `test_command_init_role_inspec` seems to be consistently failing.

Issue #2012, opened 2019-04-26
yamllint from the molecule virtualenv isn't found


webknjaz:
```console
--> Executing Yamllint on files found in /tmp/mol-test/webknjaz.coll/...
pyenv: yamllint: command not found

The `yamllint' command exists in these Python versions:
  3.7.0
  3.7.1/envs/ansible-opencanary-py3.7.1-pyenv-venv
  3.7.1/envs/molecule_error_example-py3.7.1-pyenv-venv
  3.7.1/envs/molecule-py3.7.1-pyenv-venv
  ansible-opencanary-py3.7.1-pyenv-venv
  molecule_error_example-py3.7.1-pyenv-venv
  molecule-py3.7.1-pyenv-venv

An error occurred during the test sequence action: 'lint'. Cleaning up.
```

Possible solutions:
1) explicitly use the same Python interpreter:
```python
import sys
import sh
getattr(sh, sys.executable).bake(  # instead of `sh.yamllint.bake(`
    '-m', 'yamllint',
    ...
```
2)
```python
import yamllint
```

# Issue Type

- Bug report

# Molecule and Ansible details

master

Molecule installation method (one of):

- source

Ansible installation method (one of):

- pip

# Desired Behavior

Molecule picks up yamllint from the same virtualenv

# Actual Behaviour

Molecule executed from a virtualenv uses `sh` module and fails to find `yamllint` installed right next to it.

webknjaz:
Ref: https://github.com/ansible/molecule/issues/1851#issuecomment-487067137

webknjaz:
Alright... This seems to affect all `sh` invocations.

webknjaz:
After all, this will probably turn into subprocess invocation refactoring...

decentral1se:
Oh dear me, this is a huge breakage ... taking a look, thanks for raising.

decentral1se:
Data point on `sh`:  "Latest commit 858adf0 on 9 Jun 2017"

https://github.com/amoffat/sh

decentral1se:
Wait a second ... `pip freeze ` after a `pip install .` reports that yamllint is not there!?

We never included this in `install_requires`? It was transitive before or?

I opened up https://github.com/ansible/molecule/pull/2033 to start this conversation ...

webknjaz:
@decentral1se it's reproducible if you use an absolute path to the executable in virtualenv without activating that virtualenv. So since `sh` relies on PATH it doesn't work which is ridiculous...

[...]

PR #2033, opened 2019-05-09
Add yamllint to install_requires

decentral1se:
This is required in the run time and is suddenly not present when installing the package as can be seen in the report from https://github.com/ansible/molecule/issues/2012. We now include this in our `install_require` list and manage the bounds directly to ensure we have a working default linter.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -86,6 +86,7 @@ install_requires =
     tabulate == 0.8.2
     testinfra == 1.19.0
     tree-format == 0.1.2
+    yamllint >= 1.15.0, < 2
 
 [options.extras_require]
 docs =
--- a/tox.ini
+++ b/tox.ini
@@ -24,7 +24,6 @@ setenv =
     functional: PYTEST_ADDOPTS=test/functional/ {env:PYTEST_ADDOPTS:}
 deps =
     flake8>=3.6.0,<4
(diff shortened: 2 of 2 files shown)


decentral1se:
See https://github.com/ansible/molecule/issues/2012 for more! Critical bug ...

decentral1se:
WDYT @webknjaz?

setup.cfg
```diff
@@ -86,6 +86,7 @@ install_requires =
     tabulate == 0.8.2
     testinfra == 1.19.0
     tree-format == 0.1.2
+    yamllint >= 1.15.0, < 1.16
```
webknjaz: So it looks like, before this, it was an optional dependency. Do we want to make it mandatory at all times? This will mean that if users opt out of using yamllint they'll get an unnecessary dependency (with transitive deps) installed for no reason.

Also, if we're adding it here we'll need to remove entries from `tox.ini`.

decentral1se: Right. As it is the only one available right now (see https://molecule.readthedocs.io/en/stable/configuration.html#lint), I'm inclined to put it there as default.

webknjaz: Fair enough.

decentral1se:
CI failure unrelated. Please see #2042.

Issue #1849, opened 2019-03-18
Evaluate upgrading to testinfra 2.0.0


seandst:
testinfra 2.0.0 is available. The current requirement is for 1.1.9, which by semver indicates backward-incompatible changes. Those changes, according to testinfra's [changelog](https://github.com/philpep/testinfra/blob/master/CHANGELOG.rst#200), are:

```
- Drop deprecated "testinfra" command, you should use "py.test" instead
- Drop deprecated top level fixtures, access them through the fixture "host" instead.
- Drop support for ansible <= 2.4
```

The change to the testinfra command should not break molecule's integration with testinfra, as molecule uses py.test, as directed by testinfra.
The change removing the top level fixtures may break molecule _users_, but they have been deprecated since testinfra 1.6, released in April of 2017.
The change dropping support for ansible <= 2.4 does not affect molecule (molecule requires ansible >=2.5).

decentral1se:
Thanks for raising @seandst!

> The change removing the top level fixtures may break molecule users, but they have been deprecated since testinfra 1.6, released in April of 2017.

I wonder how we should deal with this with regard to versioning ... does a change log entry suffice!?

philpep:
Hi, I'm the maintainer of testinfra.

Indeed the change which may affect your users for 2.0.0 is the removal of old fixtures ``Command``, ``Package``, but DeprecationWarnings are issued since 1.6 and IIRC since recent pytest releases, I'd say ~ 1 or 2 years ago, these warning are displayed in the output by default.

I think you can even safely pin ``testinfra<3`` here, I have a strong policy regarding API changes using semver X.Y.Z where Z is bugfix, Y new feature without breaking changes (mostly new interfaces and backends to introspect systems) and X is major release dropping deprecated stuff. Probably the version 3 will drop python2 support or introduce a new ansible backend based on https://github.com/ansible/ansible-runner

I think we can do this without breakage, but you known it's python and users might write monkeypaches or rely in deep in the testinfra/ansible api so  ¯\_(ツ)_/¯

In all cases I'd recommend to not pin a specific version because users might want to pin their version and avoid a pkg_resources VersionConflict. Also they probably expect to have a version matching "latest" documentation on readthedocs.

seandst:
> I wonder how we should deal with this with regard to versioning ... does a change log entry suffice!?

I think a change log entry does the job, yeah.

decentral1se:
I've opened up https://github.com/ansible/molecule/pull/2034 against this.

PR #2034, opened 2019-05-09
Add testinfra lower bound of >= 3.0.2

decentral1se:
Closes https://github.com/ansible/molecule/issues/1849.
Closes https://github.com/ansible/molecule/issues/1727.

I've manually tested this and seen no breaking issues :+1:

### DIFF
--- a/test/scenarios/driver/docker/molecule/default/tests/test_default.py
+++ b/test/scenarios/driver/docker/molecule/default/tests/test_default.py
@@ -30,4 +30,4 @@ def test_etc_molecule_ansible_hostname_file(host):
 
 def test_buildarg_env_var(host):
     cmd_out = host.run("echo $envarg")
-    assert cmd_out.stdout == 'this_is_a_test'
+    assert cmd_out.stdout.strip() == 'this_is_a_test'
--- a/test/unit/verifier/test_testinfra.py
+++ b/test/unit/verifier/test_testinfra.py
@@ -311,8 +311,7 @@ def test_execute_bakes(patched_run_command, _patched_testinfra_get_tests,
 
 def test_executes_catches_and_exits_return_code(
         patched_run_command, _patched_testinfra_get_tests, _instance):
-    patched_run_command.side_effect = sh.ErrorReturnCode_1(
-        sh.testinfra, b'', b'')
+    patched_run_command.side_effect = sh.ErrorReturnCode_1(sh.pytest, b'', b'')
     with pytest.raises(SystemExit) as e:
         _instance.execute()
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Unreleased
   `create` and `destroy`. This allows the use of roles in all sequence step playbooks.
 * Removed validation regex for docker registry passwords, all ``string`` values are now valid.
 * Add ``tty`` option to the Docker driver.
+* Specify new lower bound of 3.0.2 for ``testinfra`` which uses the new Ansible test runner.
 
 2.20
 ====
--- a/setup.cfg
+++ b/setup.cfg
@@ -84,7 +84,7 @@ install_requires =
     sh == 1.12.14
     six == 1.11.0
     tabulate == 0.8.2
-    testinfra == 1.19.0
+    testinfra >= 3.0.2, < 4
     tree-format == 0.1.2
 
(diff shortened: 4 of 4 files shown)


decentral1se:
Getting this resolved also helps me move forward https://github.com/ansible/molecule/pull/1811.

seandst:
I've also done some testing with testinfra 3, both with molecule's own tests and also with some of my scenarios that have testinfra tests. Everything worked great :+1:.

decentral1se:
Thanks for also pushing this along @seandst!

BTW, this will fix the always failing master builds now which include an Ansible 2.8 job. See latest failure on https://travis-ci.com/ansible/molecule/jobs/198728955. We should merge this soon to get back on the green line.

decentral1se:
Looking to get this merged soon, anyone marking approved?

test/scenarios/driver/docker/molecule/default/tests/test_default.py
```diff
@@ -30,4 +30,4 @@ def test_etc_molecule_ansible_hostname_file(host):
 
 def test_buildarg_env_var(host):
     cmd_out = host.run("echo $envarg")
-    assert cmd_out.stdout == 'this_is_a_test'
+    assert 'this_is_a_test' in cmd_out.stdout
```
webknjaz: Why `in`?

decentral1se: The test was failing with a missing `\n` in the exact string match. I'd rather not have tests failing with such brittle checks, so I change it to a `in` check. This will be more resilient to any other unimportant changes to string handling and we still get a useful test that checks for what we want.

webknjaz: I'd add that `\n` there then. Or do `str.strip()`. Exact checks are more reliable IMHO.

decentral1se: Done.

decentral1se:
See also the small one in https://github.com/ansible/molecule/pull/2035.

decentral1se:
CI failure unrelated. Please see https://github.com/ansible/molecule/issues/2042.

test/unit/verifier/test_testinfra.py
```diff
@@ -312,7 +312,7 @@ def test_execute_bakes(patched_run_command, _patched_testinfra_get_tests,
 def test_executes_catches_and_exits_return_code(
         patched_run_command, _patched_testinfra_get_tests, _instance):
     patched_run_command.side_effect = sh.ErrorReturnCode_1(
-        sh.testinfra, b'', b'')
+        sh.Command('pytest'), b'', b'')
```
webknjaz: why?

decentral1se: `testinfra` is invoked through `pytest` AFAICT

webknjaz: hm... so what's the reason behind this change anyway?

decentral1se: The test was failing with "I can't find testinfra" from `sh`.

znerol: The `testinfra` command line script was [dropped in 2.0.0](https://github.com/philpep/testinfra/commit/baab30ed1b39e92ce24ccad42a95e3be749c7245#diff-2eeaed663bd0d25b7e608891384b7298). The change here is correct. However, this could also be written as `sh.pytest, b'', b'')` which is more inline with the [sh examples](https://amoffat.github.io/sh/), the `Command()` call is superflous.

decentral1se: OK, will patch to use same style invocation, thanks for the pointer!

decentral1se:
Any final words here @webknjaz? Shall we merge it?

pilou-:
Using one scenario, i tested this matrix without encountering any error: ansible 2.7.10, molecule 2.20.1, testinfra 3.0.3, Python 3.7 (new testinfra works with previous ansible version)

decentral1se:
Thanks for testing this @pilou-.

Bump Bump Bump :shark: :shark: :shark:

hswong3i:
In case someone using Molecule + LXD Driver as me, please refer to https://github.com/philpep/testinfra/pull/445 for testinfra with LXC/LXD support.

PR #2045, opened 2019-05-16
Inspec diagnosis

seandst:
#### PR Type

- Bugfix Pull Request

Two related commits:

## Never truncate exceptions raised by sh in testing
    
    The sh module, by default, truncates the output in string
    representations of exceptions. This change tells sh to stop truncating
    those strings in testing, and also fixes a few instances of direct calls
    to the 'sh' module that were not using the pytest.helpers.run_command
    wrapper.

## Comply with rubocop FrozenStringLiteralComment cop
    
    This was fixed with `rubocop --auto-correct`, similar to running `yapf
    --in-place`. I have insufficient ruby experience (and no interest) to
    agree or disagree with rubocop's opinion.

I did targeted testing on the failing inspec test, but I'm going to let travis do the legwork on making sure my `run_command` changes don't otherwise break functional tests.

closes #2042

Issue #2042, opened 2019-05-14
Fix broken inspec test


decentral1se:
This is blocking PRs from getting reviewed properly and merged. Affected:

* https://github.com/ansible/molecule/pull/2034
* https://github.com/ansible/molecule/pull/2033
* https://github.com/ansible/molecule/pull/2036

I cannot reproduce this locally. I ran:

```
$ tox -e py37-ansible27-functional -- -v -k test_command_init_role_inspec -p no:warning
```

Log is (not useful at all ... another case against `sh`?):

From https://travis-ci.com/ansible/molecule/jobs/199661714#L451.
```console
=================================== FAILURES ===================================
________________________ test_command_init_role_inspec _________________________
temp_dir = local('/tmp/pytest-of-travis/pytest-0/test_command_init_role_inspec0/SVPPY')
    @needs_inspec
    @needs_rubocop
    def test_command_init_role_inspec(temp_dir):
        role_directory = os.path.join(temp_dir.strpath, 'test-init')
        options = {
            'role_name': 'test-init',
            'verifier_name': 'inspec',
        }
        cmd = sh.molecule.bake('init', 'role', **options)
        pytest.helpers.run_command(cmd)
        pytest.helpers.metadata_lint_update(role_directory)
    
        with change_dir_to(role_directory):
>           sh.molecule('test')
/home/travis/build/ansible/molecule/test/functional/docker/test_scenarios.py:104: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/travis/build/ansible/molecule/.tox/ansible27-functional/lib/python3.7/site-packages/sh.py:1427: in __call__
    return RunningCommand(cmd, call_args, stdin, stdout, stderr)
/home/travis/build/ansible/molecule/.tox/ansible27-functional/lib/python3.7/site-packages/sh.py:774: in __init__
    self.wait()
/home/travis/build/ansible/molecule/.tox/ansible27-functional/lib/python3.7/site-packages/sh.py:792: in wait
    self.handle_command_exit_code(exit_code)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[...]

PR #2046, opened 2019-05-16
Place upper bounds on inspec/rubcop

decentral1se:
Refs:
  * https://github.com/ansible/molecule/pull/2045
  * https://github.com/ansible/molecule/issues/2042
  * https://github.com/inspec/inspec/blob/master/CHANGELOG.md
  * https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md

I'm OK with having inspec/rubocop users having to come and ask for new major version tool support for the immediate short term. We can bake in a forward looking major version build into the CI more mid-term.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -608,6 +608,7 @@ install:
   - >
     ! grep 'ERROR: ' ~/.tox-venv-install.log .tox/*/log/*.log
 before_script:
-  - gem install inspec rubocop
+  - gem install inspec -v '~> 3'
+  - gem install rubocop -v '~> 0.69.0'
 script:
   - tox
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Unreleased
   `create` and `destroy`. This allows the use of roles in all sequence step playbooks.
 * Removed validation regex for docker registry passwords, all ``string`` values are now valid.
 * Add ``tty`` option to the Docker driver.
+* Place upper bounds on inspec and rubocop for CI testing.
 
 2.20
 ====


.travis.yml
```diff
@@ -608,6 +608,7 @@ install:
   - >
     ! grep 'ERROR: ' ~/.tox-venv-install.log .tox/*/log/*.log
 before_script:
-  - gem install inspec rubocop
+  - gem install inspec -v '~> 3'
```
decentral1se: `gem` does not support inlining this from my testing. 

If anyone knows how to do it better, I am all ears.

`~>` means, bound before the next major version.

seandst: One suggestion would be to use a Gemfile and bundler, which (imo) is a lot of extra complexity just to get this back down to one line. I think doing it on two separate lines is fine.

seandst: Also, a little bit of interactive ruby to demo how the ~> comparator works for those that may not be familiar with it. In this case, `~> 3` expands to `>= 3, < 4`:

```rb
$ irb
irb(main):001:0> depcheck = Gem::Dependency.new('inspec', '~> 3')
=> <Gem::Dependency type=:runtime name="inspec" requirements="~> 3">
irb(main):002:0> depcheck.match?('inspec', '2.99')
=> false
irb(main):003:0> depcheck.match?('inspec', '3.0')
=> true
irb(main):004:0> depcheck.match?('inspec', '3.99')
=> true
irb(main):005:0> depcheck.match?('inspec', '4.0')
=> false
```

decentral1se:
Can someone with commit bit +1 this as well? Thanks.

decentral1se:
Looks good then @webknjaz?

webknjaz:
I don't know enough about Ruby to make a judgment here. I'll leave it to somebody better knowledgeable.

decentral1se:
@webknjaz, not many are around with the commit bit to get this merged, so that is why I am explicitly asking you to approve this. Given that @seandst has approved this and seems to have more Ruby knowledge than us, let's just go with it. If there are issues, I will follow up prompty, as usual.

PR #2047, opened 2019-05-17
Fix Ansible 2.8 deprecation warnings in provisioner playbook

brianhelba:
**PR Type**
* Bugfix Pull Request

The [`docker_image_facts` Ansible module](https://docs.ansible.com/ansible/latest/modules/docker_image_facts_module.html) causes deprecation warnings in Ansible 2.8 like:
```
[DEPRECATION WARNING]: docker_image_facts is kept for backwards compatibility
but usage is discouraged. The module documentation details page may explain
more about this rationale.. This feature will be removed in a future release.
Deprecation warnings can be disabled by setting deprecation_warnings=False in
ansible.cfg.
```

The preferred alternative, [`docker_image_info`](https://docs.ansible.com/ansible/latest/modules/docker_image_info_module.html) has been available since Ansible 2.1.

This fixes the default `create.yml` playbook for the `molecule.provisioner.ansible.Ansible` provisioner to use `docker_image_info` instead.

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -27,7 +27,7 @@
       register: platforms
 
     - name: Discover local Docker images
-      docker_image_facts:
+      docker_image_info:
         name: "molecule_local/{{ item.item.name }}"
         docker_host: "{{ item.item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
       with_items: "{{ platforms.results }}"


molecule/provisioner/ansible/playbooks/docker/create.yml
```diff
@@ -27,7 +27,7 @@
       register: platforms
 
     - name: Discover local Docker images
-      docker_image_facts:
+      docker_image_info:
```
decentral1se: How do we maintain compat with previous versions of Ansible?

Please see failures on the CI: https://travis-ci.com/ansible/molecule/jobs/200941048#L603.

Do we need some conditional logic here?

brianhelba: I'll investigate and fix.

brianhelba: So I misread the Ansible docs about when `docker_image_info` was introduced (it's 2.8 not 2.1). I'll probably try a conditional, since I don't believe we can set and use a module reference within a variable.

pavelpikta: > So I misread the Ansible docs about when `docker_image_info` was introduced (it's 2.8 not 2.1). I'll probably try a conditional, since I don't believe we can set and use a module reference within a variable.

I think you can try this condition 
```yaml
- name: Discover local Docker images
  docker_image_facts:
    name: "molecule_local/{{ item.item.name }}"
    docker_host: "{{ item.item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
  with_items: "{{ platforms.results }}"
  when: 
    - not item.pre_build_image | default(false)
    - ansible_version.full is version('2.8.0', '<')
  register: docker_images

- name: Discover local Docker images
  docker_image_info:
    name: "molecule_local/{{ item.item.name }}"
    docker_host: "{{ item.item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
  with_items: "{{ platforms.results }}"
  when: 
    - not item.pre_build_image | default(false)
    - ansible_version.full is version('2.8.0', '>=')
  register: docker_images
`

Comment (seandst) [reply molecule/provisioner/ansible/playbooks/docker/create.yml]: > ...I don't believe we can set and use a module reference within a variable.

If I'm understanding this correctly, then I think you can, using a ternery and `action`/`args` in the task body, but it's not the prettiest-looking ansible, and requires both modules to take the same inputs and produce the same outputs. It's also not the prettiest ansible:

```yaml
# Warning! While this is based on working code, I haven't tested this particular snippet.
- name: Determine which docker image info module to use
  set_fact:
    _docker_image_info_module: >-
      {{ ansible_version.full is version_compare('2.8', '>=') |
         ternary('docker_image_info', 'docker_image_facts') }}

- name: Discover local Docker images
  action: "{{ _docker_image_info_module }}"
  args:
    name: "molecule_local/{{ item.item.name }}"
    docker_host: "{{ item.item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
  with_items: "{{ platforms.results }}"
  when: 
    - not item.pre_build_image | default(false)
  register: docker_images
```

ssbarnea: Managed to find a fix for this? I hope we are not going towards branches for each ansible version, it would be a maintenance hell.

ssbarnea:
Outdated, already fixed by other PR.

PR #2050, opened 2019-05-17
Update docker driver test for ansible 2.8

seandst:
In addition to the docker driver test, I also slightly modified the docker driver itself, both changes were
made with the following rationale:

Ansible 2.8 should be the default. The older versions of ansible
should be considered exceptional cases to future-proof the import,
since the new location of the HAS_DOCKER_PY constant will presumably
persist for the forseeable future.

I also removed handling for the case where HAS_DOCKER_PY cannot be
imported, under the assumption that being unable to import this constant
from either its 2.8 or pre-2.8 location is insane for all supported
versions of ansible. My apologies if this assumption is in error.

Please include details of what it is, how to use it, how it's been tested:

Tested with tox envs: lint, format-check, unit, functional (-k docker). ~~Tests were failing with testinfra 1.19, so this needs #2034 to be merged before tests will pass.~~ <-Merged

#### PR Type

- Bugfix Pull Request: Updates docker driver tests for ansible 2.8 compatibility

PR #2051, opened 2019-05-17
CI Fixes for ansible 2.8

### LINKED ISSUE #2049: Get Ansible 2.8 CI testing in place
Now that https://github.com/ansible/molecule/issues/1727 is in, let's deal with some of those CI TODOs to conver Ansible 2.8 proper:

  * https://github.com/ansible/molecule/blob/master/tox.ini#L42
  * https://github.com/ansible/molecule/blob/master/.travis.yml#L59 (Don't allow failure for 2.8)

And then we can push for a pre-release with more certainty ...

seandst:
- Bugfix Pull Request, closes #2049


webknjaz:
This doesn't look like a CI fix, just permission to fail...

seandst:
> This doesn't look like a CI fix, just permission to fail...

Allowing failures against the most recent ansible release is incorrect, and should be fixed. Installing the most recent ansible release with the `--pre` flag is incorrect, and should be fixed. This PR is, therefore, helping to fix CI by removing special cases surrounding ansible 2.8.

seandst:
The travis output is littered with timeouts, making it somewhat difficult to pick out any failures related to ansible 2.8. I'll re-trigger it later, I think.

webknjaz:
Incorrect is a loud word and it is not a bug to be fixed. So the title of this PR is just misleading as it doesn't fix anything, it just unignores things and changes the installation command. Which makes up into two unrelated changes which should be submitted separately.

It cannot be merged in its current state also because it breaks the CI. I've checked the log and it seems the command that test invokes returns 1 which is not a timeout (job lasted ~13 min) + there's a lot of jobs failing, not just one.

seandst:
Unconstructive distinctions about what a "fix" is and judgement calls about text volume aside, could you point me to the non-timeout-related failure that you saw in the log? Every error and fail that saw was caused by a timeout, so I could use a pointer if you've got one.

webknjaz:
> removing cron marker

yes

> add 2.8 also to the regular build

I believe it's already there, except for the marker


Also, we need to remove testing of Ansible 2.5 as per https://docs.ansible.com/ansible/devel/reference_appendices/release_and_maintenance.html#release-status

webknjaz:
> the non-timeout-related failure that you saw in the log?

https://travis-ci.com/ansible/molecule/jobs/201153766#L494

RC=1, I don't see any signs of timeouts (unless `sh` hides that)

seandst:
My check process on that is basically this:
- scroll all the way down to get the total number of fails/errors
- compare the total number of fails/errors with the total number of big purple "timed out" error blocks in the output
- if those numbers differ, find the non-timeout-related cause

Hopefully that helps, but like I said before, there are too many timeout errors causing noise, so I'll retrigger and see if we can get a better sense from travis what negative effect, if any, these changes have on the travis CI runs.

seandst:
> * https://github.com/ansible/molecule/blob/master/.travis.yml#L180 (removing cron marker?)

:+1:

> * https://github.com/ansible/molecule/blob/master/.travis.yml#L437 (add 2.8 also to the regular build?)

The 2.8 builds are already on the list, but I assume they are being blocked by the cron marker. I'm expecting that fixing the cron marker will also fix this.

The logic for the cron marker appears straightforward, the only real issue is that all the repeated yaml snippets start to look the same. I've pushed up the requested changes, and we'll hopefully see the desired travis build matrix as a result.

seandst:
Well sweet, these changes revealed another import error in the ansible 2.8 tests:
```
test/functional/conftest.py:84: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
    @pytest.helpers.register
    def supports_digitalocean():
>       from ansible.modules.cloud.digital_ocean.digital_ocean import HAS_DOPY
E       ModuleNotFoundError: No module named 'ansible.modules.cloud.digital_ocean.digital_ocean'
```

On the face of it, it looks like an approach like the one taken in #2050 would suffice, but the actual deprecation resolution involves using a new module introduced in ansible 2.8 that doesn't use external dependencies. I've opening up #2054 to deal with the problem generally, and #2055 to deal with the ImportError specifically.

decentral1se:
> and #2055 to deal with the ImportError specifically.

Question over in https://github.com/ansible/molecule/pull/2055#issuecomment-494313692. Nice detective work on this one.

seandst:
Rebased to latest master to include #2055, tests retriggered.

decentral1se:
:eyes: :popcorn:

decentral1se:
OK, restarted the build (another instance of https://github.com/ansible/molecule/issues/2057). This is looking good then! I'm going to merge it soon. If anyone has any final thoughts, pleas shout. Once this is in place we can start to think about a pre-release, I hope.

seandst:
Travis is finally happy. :)

gregdek:
A reminder of our community code of conduct:

https://docs.ansible.com/ansible/latest/community/code_of_conduct.html

We will not always agree about technical solutions, but we should always assume positive intent from our fellow contributors, and maintain a constructive working environment. 

Thanks to @decentral1se for his collegial tone here, and for setting the standard that we expect other contributors to follow.

PR #2052, opened 2019-05-17
Fix Deprecations

seandst:
This attempts to clear up most of the deprecation warnings being thrown as a result of molecule code. I spread it out across multiple commits, so if anything is particularly objectionable to reviewers we can pop that commit off and put it up in a separate PR. Most commits are straightforward, but a couple are not:

- The addition of `filterwarnings` to `pytest.ini` is a bit of a special case due to the impressive number of deprecation warnings being raised by that plugin. The plugin maintainers appear unresponsive, so the real solution may be to either fix the plugin or just vendor it here and maintain it ourselves, but for now I would prefer to just stop the noise and figure out the right solution later.
- The changes to the schema tests are *drastic*. This alone potentially warrants splitting that off into its own PR, and is a juicy bikeshed target. :)

Tests: lint, format-check, unit, and limited functional. Ansible 2.8 unit tests will fail due to the issue addressed by #2050.

### DIFF
--- a/molecule/command/base.py
+++ b/molecule/command/base.py
@@ -102,7 +102,7 @@ def execute_cmdline_scenarios(scenario_name, args, command_args):
                 msg = ('An error occurred during the {} sequence action: '
                        "'{}'. Cleaning up.").format(scenario.config.subcommand,
                                                     scenario.config.action)
-                LOG.warn(msg)
+                LOG.warning(msg)
                 execute_subcommand(scenario.config, 'cleanup')
                 execute_subcommand(scenario.config, 'destroy')
                 # always prune ephemeral dir if destroying on failure
--- a/molecule/model/schema_v1.py
+++ b/molecule/model/schema_v1.py
@@ -33,7 +33,7 @@ base_schema = {
             },
             'raw_env_vars': {
                 'type': 'dict',
-                'keyschema': {
+                'keysrules': {
                     'type': 'string',
                     'regex': '^[A-Z0-9_-]+$',
                 },
--- a/molecule/model/schema_v2.py
+++ b/molecule/model/schema_v2.py
@@ -196,7 +196,7 @@ base_schema = {
             },
             'env': {
                 'type': 'dict',
-                'keyschema': {
+                'keysrules': {
                     'type': 'string',
                     'regex': '^[A-Z0-9_-]+$',
                 },
@@ -258,7 +258,7 @@ base_schema = {
             },
             'env': {
                 'type': 'dict',
-                'keyschema': {
+                'keysrules': {
                     'type': 'string',
                     'regex': '^[A-Z0-9_-]+$',
                 },
@@ -334,11 +334,11 @@ base_schema = {
             },
             'env': {
                 'type': 'dict',
-                'keyschema': {
+                'keysrules': {
                     'type': 'string',
                     'regex': '^[A-Z0-9_-]+$',
                 },
-                'valueschema': {
+                'valuesrules': {
                     'nullable': False,
                 },
                 'schema': {
@@ -416,7 +416,7 @@ base_schema = {
                     },
                     'env': {
                         'type': 'dict',
-                        'keyschema': {
+                        'keysrules': {
                             'type': 'string',
                             'regex': '^[A-Z0-9_-]+$',
                         },
@@ -477,7 +477,7 @@ base_schema = {
             },
             'env': {
                 'type': 'dict',
-                'keyschema': {
+                'keysrules': {
                     'type': 'string',
                     'regex': '^[A-Z0-9_-]+$',
                 },
@@ -505,7 +505,7 @@ base_schema = {
                     },
                     'env': {
                         'type': 'dict',
-                        'keyschema': {
+                        'keysrules': {
                             'type': 'string',
                             'regex': '^[A-Z0-9_-]+$',
                         },
@@ -717,7 +717,7 @@ platforms_docker_schema = {
                 },
                 'env': {
                     'type': 'dict',
-                    'keyschema': {
+                    'keysrules': {
                         'type': 'string',
                         'regex': '^[a-zA-Z0-9_-]+$',
                     }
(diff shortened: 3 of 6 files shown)


webknjaz:
Could you please make the title a bit more descriptive? I've got a number of good practices for Git commits which are also applicable to PRs: https://github.com/ansible/molecule/issues/1891#issue-427161772

pytest.ini
```diff
@@ -4,3 +4,7 @@ doctest_optionflags = ALLOW_UNICODE ELLIPSIS
 junit_suite_name = molecule_test_suite
 norecursedirs = dist doc build .tox .eggs
 testpaths = test/
+# filter out deprecation warning spam coming from pytest_verbose_parametrize,
+# since we can't do anything about it until the plugin is updated
```
webknjaz: Make it actionable by adding a TODO/FIXME comment. Also, describe the effect of this change on the system and why someone in future would like to keep it and not delete thinking of refactoring.

Also, add a comment to https://github.com/ansible/molecule/blob/023bef688cc4e78301ec81cf3df42919f9817d08/tox.ini#L33 linking back here so that once we'll update that we'd backtrack the removal of this filtering as well.

pytest.ini
```diff
@@ -4,3 +4,7 @@ doctest_optionflags = ALLOW_UNICODE ELLIPSIS
...
+filterwarnings =
```
webknjaz: Let's also make it stricter:
```suggestion
filterwarnings =
    error
```

We shouldn't be forgiving about other warnings, they should be fixed ASAP.

test/unit/model/v2/test_pre_validate.py
```diff
@@ -18,14 +18,14 @@
...
-
 from molecule.model import schema_v2
 
+# test data dict used to generate tests:
+# key is the test's string id, used in pytest verbose output
+# value is a tuple of (input_molecule_yml_section_data, expected_schema_errors)
+_pre_validate_test_data = {}
 
-@pytest.fixture
-def _model_platforms_docker_section_data():
```
webknjaz: What deprecation is this linked to?

seandst: Sorry about that, I meant to include refs to each specific deprecation being addressed in each commit. I'll get those added next week. For now, though, this was related to the test functions parametrizing `_stream` with the fixture-decorated function itself, not the resolved fixture value, resulting in the fixture function having to be called (e.g. `_stream()`), which is deprecated in pytest 4+: https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly

seandst:
> Misleading title + unrelated changes = 🚫

What changes are unrelated? As mentioned in the description, I'm happy to make any changes requested, but I'm not going to guess at what those changes might be.

webknjaz:
> What changes are unrelated

There's a number of changes to different things which make this patch non-atomic as it seems.

webknjaz:
> I'm not going to guess

Well, when clicking on the PR title I guessed that you're actually deprecating something. Turned out that you do something completely different.

webknjaz:
Also, just don't mix addressing separate deprecations in one PR

seandst:
I'll happily split off the two commits as I mentioned in the description (altering the pytest filterwarnings and refactoring the prevalidation tests). Also, as mentioned above, make sure the commit descriptions are updated to include the specific deprecation being addressed by each commit in every case. I meant to do that initially and it just slipped my mind, sorry about that.

> Also, just don't mix addressing separate deprecations in one PR

I understand the motivation behind this, but there needs to be a balance between not making reviewers have to review cluttered PRs and not making contributors submit a new PR for every small idea. The three remaining commits will remain on this PR as atomic commits for review as separate but related changes.

decentral1se:
>> Also, just don't mix addressing separate deprecations in one PR
> 
> I understand the motivation behind this, but there needs to be a balance between not making reviewers have to review cluttered PRs and not making contributors submit a new PR for every small idea. The three remaining commits will remain on this PR as atomic commits for review as separate but related changes.

Exactly @seandst.

I am +1 on continuing on this PR and getting these warnings issues resolved.

seandst:
PR updated, changes are now limited to straightforward fixes related to deprecation warnings, with the two less-straightforward changes split off into their own PRs.

The fix to the schema tests might end up taking a while longer. Since every schema addition requires (or "should probably require", at least) an update to those tests, I might open up an issue/proposal/somethingelse to discuss how to make working with those tests a little easier before. More/better commenting is certainly needed in that file now to explain what those tests are actually for, and how to add validation scenarios to it.

pytest.ini
seandst: I like the idea on principle, but in practice, this is a sample of what happens in a `tox -e unit` run:
```
_ ERROR collecting test/unit/command/test_lint.py _
test/unit/command/test_lint.py:23: in <module>
    from molecule.command import lint
molecule/command/__init__.py:26: in <module>
    from molecule.command import cleanup  # noqa
molecule/command/cleanup.py:21: in <module>
    import click
.tox/unit/lib/python3.7/site-packages/click/__init__.py:18: in <module>
    from .core import Context, BaseCommand, Command, MultiCommand, Group, \
E     File "/home/semyers/ve/oasis/origin/molecule/.tox/unit/lib/python3.7/site-packages/click/core.py", line 1259
E       """
E        ^
E   SyntaxError: invalid escape sequence \s
```

Turning warning into errors means that unaddressed deprecations in dependencies will cause our tests to fail, which I think is unacceptable. For now, I'm going to limit the scope of change for the upcoming PR to only squashing the pytest_verbose_parametrize warning spam.

seandst:
Ci failures appear to be environmental, including the "normal" timezone time-out, and a somewhat worrisome docker creation failure that I haven't seen before.

pytest.ini
decentral1se: Fair enough so!

decentral1se:
Restarted the build. Thanks for pushing along.

gregdek:
A reminder of our community code of conduct:

https://docs.ansible.com/ansible/latest/community/code_of_conduct.html

We will not always agree about technical solutions, but we should always assume positive intent from our fellow contributors, and maintain a constructive working environment. 

Thanks to @decentral1se for his collegial tone here, and for setting the standard that we expect other contributors to follow.

PR #2063, opened 2019-05-24
Add ansible_args support back to converge command

### LINKED ISSUE #2061: Extra vars are not passed to the playbook anymore with converge.
# Issue Type

- Bug report

```
ansible 2.8.0
  config file = None
  configured module search path = ['/home/olcla/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.6/dist-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.6.7 (default, Oct 22 2018, 11:32:17) [GCC 8.2.0]

molecule (downgraded to 2.20 to fix but problem related to master branch)
```

- pip from git master

- pip

# Desired Behavior

When running `molecule converge -s my_scenario -- -e somevar=value` I expect molecule to pass the extra vars when running playbook.yml

# Actual Behaviour

I was trying to fix deprecations related to ansible 2.8 and had to install molecule from git master to overcome problems with testinfra. When running the above command with the latest molecule master, the extra vars are not passed to the playbook anymore. This is working as expected when downgrading to molecule 2.20.1

seandst:
During a recent command-calling refactor, I missed the handling of the `ansible_args` extra arguments in the calling of the converge command. This restores that functionality to the converge command, which as far as I can tell was the only command affected.

It also adds a test to cover the expected behavior to hopefully ensure it doesn't get missed again.

fixes #2061 

#### PR Type

- Bugfix Pull Request

Unit tested on py{27,37}, ansible{26,28}, functional testing with `-k 'docker and converge'`

### DIFF
--- a/test/unit/command/test_converge.py
+++ b/test/unit/command/test_converge.py
@@ -18,7 +18,10 @@
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
 
+from click.testing import CliRunner
+
 from molecule.command import converge
+from molecule.shell import main
 
 
 # NOTE(retr0h): The use of the `patched_config_validate` fixture, disables
@@ -38,3 +41,19 @@ def test_execute(mocker, patched_logger_info, patched_ansible_converge,
     patched_ansible_converge.assert_called_once_with()
 
     assert config_instance.state.converged
+
+
+def test_ansible_args_passed_to_scenarios_get_configs(mocker):
+    # Scenarios patch is needed to safely invoke CliRunner
+    # in the test environment and block scenario exectution
+    mocker.patch('molecule.scenarios.Scenarios')
+    patched_get_configs = mocker.patch('molecule.command.base.get_configs')
+
+    runner = CliRunner()
+    args = ('converge', '--', '-e', 'testvar=testvalue')
+    ansible_args = args[2:]
+    runner.invoke(main, args, obj={})
+
+    # call index [0][2] is the 3rd positional arguement to get_configs,
+    # which should be the tuple of parsed ansible_args from the CLI
+    assert patched_get_configs.call_args[0][2] == ansible_args
--- a/molecule/command/base.py
+++ b/molecule/command/base.py
@@ -72,7 +72,10 @@ class Base(object):
         self._config.provisioner.manage_inventory()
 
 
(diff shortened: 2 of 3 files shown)


seandst:
@zeitounator If you have a chance to try this out, I'd love to know if this fixes the problem in #2061 for you or not.

zeitounator:
Hi @seandst and thanks for this fast reaction. I can confirm the PR restores the previous behaviour for converge on my side. Cheers.

test/unit/command/test_converge.py
```diff
@@ -38,3 +41,28 @@ def test_execute(mocker, patched_logger_info, patched_ansible_converge,
     patched_ansible_converge.assert_called_once_with()
 
     assert config_instance.state.converged
+
+
+def test_ansible_args(mocker, patched_logger_info, patched_ansible_converge,
```
decentral1se: Perhaps name it `test_ansible_args_correctly_passed` to give more the "why" without the need then for the following comment?

test/unit/command/test_converge.py
```diff
@@ -38,3 +41,28 @@ def test_execute(mocker, patched_logger_info, patched_ansible_converge,
...
+                      patched_config_validate, config_instance):
+    # Use the click test runner and mocks to ensure that ansible_args given on
+    # the converge commandline get correcty passed to the Scenarios constructor
+
+    # Mock out Scenarios to prevent execute_cmdline_scenarios from doing
```
decentral1se: I am for the "why" comments but typically not for the "what" comments because if the code changes then we might not change the comment. The test code seems simple enough to note require such comments? Just a thought, not religious on comment style, promise.

seandst: I err on the side of verbosity, since I never know how much is obvious after spending the time to figure out where to bolt on the patch. If it's clear to you what's going on, then I agree it's fine to reduce the commentary. :)

seandst:
> Oh, yeah, can we get a changelog entry for this?

I actually wrote one, but in the process realized that this bug only ever existed on master. The changelog entry, relative to the previous 2.20 entries, would effectively be "converge _still_ accepts extra ansible command-line args".

test/unit/command/test_converge.py
seandst: Alright, I left a comment in there with a slightly different comment. Nothing asserts against the Scenarios mock, so I think leaving a little note along the lines of "No really, you wanna keep this mock in place..." is a safe choice. Let me know if you think what's there now is better.

PR #2064, opened 2019-05-24
update pyyaml to >=5.1

### LINKED ISSUE #1806: PyYAML update CVE-2017-18342

redshiftzero:
Closes #1806

CVE-2017-18342 flags the previously unsafe default yaml load function in pyyaml. Since version 1.0.6 of this project, it looks like the safe loader (`yaml.safe_load`) was already used everywhere. This PR just updates to a version of pyyaml that does not have the CVE associated with it.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -80,7 +80,7 @@ install_requires =
     Jinja2 >= 2.10.1
     pexpect >= 4.6.0, < 5
     psutil == 5.4.6; sys_platform!="win32" and sys_platform!="cygwin"
-    PyYAML == 3.13
+    PyYAML >= 5.1
     sh == 1.12.14
     six == 1.11.0
     tabulate == 0.8.2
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -80,7 +80,7 @@ install_requires =
     Jinja2 >= 2.10.1
     pexpect >= 4.6.0, < 5
     psutil == 5.4.6; sys_platform!="win32" and sys_platform!="cygwin"
-    PyYAML == 3.13
+    PyYAML >= 5.1
```
decentral1se: Can we also get a `< 6` ? Assuming semantic versioning ... but we should make sure :+1:

decentral1se: Please see https://github.com/yaml/pyyaml/blob/master/CHANGES#L7.

We need to put an upper bound on this, no?

PR #2068, opened 2019-05-26
Avoid sanity checks for AnsiblePlaybook unit tests

decentral1se:
This avoids the need to have the Docker daemon running when running unit tests which should not be necessary and introduces another way for the unit tests to fail.

I've tested this manually like so:

```
$ sudo systemctl stop docker
$  tox -e py37-ansible28-unit -- -n auto -p no:warning 
```

This is necessary for these tests because they call `_instance.execute` directly which is a wrapper function which includes the call to the sanity checks before calling the mocked runner command.

### DIFF
--- a/test/unit/provisioner/test_ansible_playbook.py
+++ b/test/unit/provisioner/test_ansible_playbook.py
@@ -27,7 +27,14 @@ from molecule.provisioner import ansible_playbook
 
 @pytest.fixture
 def _instance(config_instance):
-    return ansible_playbook.AnsiblePlaybook('playbook', config_instance)
+    _instance = ansible_playbook.AnsiblePlaybook('playbook', config_instance)
+
+    # FIXME(decentral1se): temporarily force the sanity checks to not run while
+    # putting the AnsiblePlaybook under test. TBD: reconsider whether invoking
+    # sanity checks within the `execute()` function makes sense.
+    _instance._config.state.change_state('sanity_checked', True)
+
+    return _instance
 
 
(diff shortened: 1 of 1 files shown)


decentral1se:
Also note, this cannot be reproduced on the CI because we always have the docker service up and running. See https://travis-ci.com/ansible/molecule/jobs/203111044#L238.

test/unit/provisioner/test_ansible_playbook.py
```diff
@@ -27,7 +27,9 @@
 
 @pytest.fixture
 def _instance(config_instance):
-    return ansible_playbook.AnsiblePlaybook('playbook', config_instance)
+    _instance = ansible_playbook.AnsiblePlaybook('playbook', config_instance)
+    _instance._config.state.change_state('sanity_checked', True)
```
ssbarnea: My only concern here is that you did not include a comment explaining why this is needed. As the reason is quite well hidden, reading these lines makes anyone wonder why their were added. Using git blame to dig is not always convenient.

decentral1se: Fair point. And I am somehow reconsidering my placement of where I trigger the sanity checks (causing this issue ...) but have not had really time to take a proper look.

decentral1se: OK, something more informative in https://github.com/ansible/molecule/pull/2068/commits/4fed8fc2f3e4b09c4b1f6cee8652d2c0817aab0b.

PR #2072, opened 2019-05-28
Bump pytest* dependencies

decentral1se:
Please see each commit message for context.


ssbarnea:
@decentral1se Really happy to see that you are fixing this, I discovered that after starting to make the same changes myself. The only difference was that my intent was to conert the pinnings to ranges.

decentral1se:
Thanks for keeping the ball rolling with these reviews :clap:

PR #2073, opened 2019-05-28
Move onto latest Tox and remove install hack

decentral1se:
We should get rid of this because it obscures dependency installation errors during the development workflow. The solution is available in the current version of Tox.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-minversion = 3.9.0
+minversion = 3.12.0
 envlist =
     flake8
     yamllint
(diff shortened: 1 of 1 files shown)


decentral1se:
Strange error that I can't reproduce locally ... I've re-ran the builds ...

If it happens again I'll hold off on merging until I can investigate further ...

decentral1se:
Hmmm, strange error. Not going to chase this up now but I think the Tox configuration needs further tweaking to get this sorted out. Anyway, this attempt might be useful to see what is failing for whoever else picks it up.

Issue #2070, opened 2019-05-28
RuntimeError: You must install paramiko package (pip install paramiko) to use the paramiko backend [Molecule verify, Ansible 2.8, Testinfra 3.x, AWS EC2 driver]


jonashackt:
# Issue Type

- Bug report

# Molecule and Ansible details

Prerequisites (install testinfra first to fix https://github.com/ansible/molecule/issues/1727)

```
$ pip install testinfra molecule docker
Installing collected packages: testinfra, molecule, docker, ansible
Successfully installed ansible-2.8.0 docker-4.0.1 molecule-2.20.1 testinfra-3.0.4
```

```
$ ansible --version && molecule --version
ansible 2.8.0
  config file = None
  configured module search path = ['/Users/jonashecht/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.7/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.7.3 (default, Mar 27 2019, 09:23:15) [Clang 10.0.1 (clang-1001.0.46.3)]
molecule, version 2.20.1
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

# Desired Behavior

`molecule verify --scenario-name aws-ec2-ubuntu` should verify a AWS EC2 instance (like it does with Ansible 2.7.x and Testinfra 2.x)

# Actual Behaviour

```
--> Test matrix
    
└── aws-ec2-ubuntu
    └── verify
    
--> Scenario: 'aws-ec2-ubuntu'
--> Action: 'verify'
--> Executing Testinfra tests found in /home/circleci/molecule-ansible-docker-vagrant/docker/molecule/aws-ec2-ubuntu/../tests//...
    ============================= test session starts ==============================
    platform linux -- Python 3.6.1, pytest-4.5.0, py-1.8.0, pluggy-0.11.0 -- /usr/local/bin/python
    rootdir: /home/circleci/molecule-ansible-docker-vagrant/docker/molecule
    plugins: testinfra-3.0.4
collected 3 items                                                              
    
    ../tests/test_docker.py::test_is_docker_installed[ansible:/aws-ec2-ubuntu] FAILED [ 33%]
    ../tests/test_docker.py::test_vagrant_user_is_part_of_group_docker[ansible:/aws-ec2-ubuntu] FAILED [ 66%]
[...]

PR #2076, opened 2019-05-28
Add paramiko dependency (made optional in Ansible 2.8+)

decentral1se:
Closes https://github.com/ansible/molecule/issues/2070.

See https://github.com/ansible/ansible/blob/stable-2.8/changelogs/CHANGELOG-v2.8.rst.
See https://www.paramiko.org/installing.html.

As far as I can tell, in the traceback of #2070, it's because we choose to use the Testinfra paramiko back-end. Maybe we can investigate that choice later on. Testinfra doesn't manage these connection back-ends for us and Ansible use to give us this dependency on previous versions transitively.

Questions:
  * Will this conflict with Ansible < 2.8 dependencies?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -78,6 +78,7 @@ install_requires =
     cookiecutter == 1.6.0
     python-gilt >= 1.2.1, < 2
     Jinja2 >= 2.10.1
+    paramiko >= 2.4.2, < 3
     pexpect >= 4.6.0, < 5
     psutil == 5.4.6; sys_platform!="win32" and sys_platform!="cygwin"
     PyYAML == 3.13


ssbarnea:
I guess that is probably the right approach, even if paramiko seems to be not very well maintained anymore, being full of deprecation warnings. I guess that is the reason why recently https://github.com/ploxiln/paramiko-ng appeared as an alternative for those not willing to wait forever for a new release.

decentral1se:
Yeah, I think it works for Molecule now. Any further comments anyone? 

This fixes a critical bug which I'd like to get out of the way ASAP.

decentral1se:
> because molecule can also be used without paramiko in some cases but I will workflow it.

Yes, we need to investigate on which execution path `paramiko` is required in combination with `testinfra` and then see if we can make it an extra. Perhaps we can default to a different back-end: https://testinfra.readthedocs.io/en/latest/backends.html

PR #2082, opened 2019-05-31
Unpin install requirements to avoid conflicts

ssbarnea:
Converts requirement pins to ranges in order to avoid conflicts. If
needed we will be adding capping to packages that are not compatible.

This fixed errors that we seen on Travis related to conflicts.


ssbarnea:
@themr0c We will add upper limits when we face that issue. So far these minor dependencies are not very active and they all have a good track of not breaking things on updates.

PR #2085, opened 2019-06-05
fix dependency gem for rubocop

Pinaute:
#### PR Type

- Bugfix Pull Request

I use Rubocop linter for the Inspec verifier.
For my gitlab-ci pipeline I use the image `quay.io/ansible/molecule:latest`

The linter rubocop fails with the following error:
https://gitlab.com/Pinaute/sandbox-ansible-pipeline/-/jobs/224766289

To fix the issue, I needed to add two Gem : json and etc
https://gitlab.com/Pinaute/sandbox-ansible-pipeline/-/jobs/225035679

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,8 @@ ENV PIP_INSTALL_ARGS="\
 
 ENV GEM_PACKAGES="\
     rubocop \
+    json \
+    etc \
     "
 
 ENV MOLECULE_EXTRAS="azure,docker,docs,ec2,gce,linode,lxc,openstack,vagrant,windows"


seandst:
Out of curiousity, do you know why the `etc` gem is needed?

Pinaute:
After add gem json i have a similarity error output. So I added the gem etc to solve the problem.
I did the test locally
```
/usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require': cannot load such file -- etc (LoadError)
```
On the rubocop repository it does not specify dependencies. I do not understand why gem are needed in this case.

PR #2088, opened 2019-06-06
WIP: Enhance Vagrant driver configuration

ingobecker:
For all configuration values of the Vagrant driver that are set inside the Vagrantfile using eval you have to supply them as valid ruby code inside your yaml file. That might not be obvious if you just look at the  the documentation and if the value you are going to set is an int or bool you might not even notice that you have to supply them as ruby literals. You just do the right thing intuitively. It works the way you expect it to work. But when it comes to strings, you have to supply them as valid ruby string literal which results in redundant writing of quotes.

For example the config below doesn't work and throws a ruby syntax error:
```
provider_options:
  uri: qemu:///system
```
instead you have to write something like this:
```
provider_options:
  uri: "'qemu:///system'"
```
What makes it even worse is that the behavior is not consistent. Some config option strings are evaled, some are set using the parsed value from the yaml config in a variable assingment. For those values, supplying a string without all the quotes works fine. So to figure out where to use which syntax you have to look at the implementation details.

In this pull request i replaced eval with send. This way you can omit the added quotes in the yaml string and setting variables inside the yaml works more consistently.

### DIFF
--- a/molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
+++ b/molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
@@ -155,7 +155,7 @@ Vagrant.configure('2') do |config|
   c['options'].delete('synced_folder')
 
   c['options'].each { |key, value|
-    eval("config.#{key} = #{value}")
+    config.send "#{key}=", value
   }
 
   ##
@@ -206,7 +206,7 @@ Vagrant.configure('2') do |config|
       # Custom
       provider['options'].each { |key, value|
         if key != 'linked_clone'
-          eval("virtualbox.#{key} = #{value}")
+          virtualbox.send "#{key}=", value
         end
       }
 
@@ -245,7 +245,7 @@ Vagrant.configure('2') do |config|
 
       # Custom
       provider['options'].each { |key, value|
-        eval("vmware.#{key} = #{value}")
+        vmware.send "#{key}=", value
       }
 
       # Raw Configuration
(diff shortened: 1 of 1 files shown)


ssbarnea:
@ingobecker Can you please fix the DCO by adding the correct line to your commit? I would not be able merge it without it. It may be possible for you to manually approve it in the UI, but I am not sure, editing the commit message clearly works.

BTW, nice fix, highly desired!

ingobecker:
It turned out that the fixes from above are not completely solving the issue. I ran into another problem which is a bit more complicated to solve. I would not merge those changes as they are now, because they might lead to other problems which need to be solved. I can give you more detailed informations on that later this evening once i had the chance to get another look at the code.

decentral1se:
Cliff hanger @ingobecker!

ingobecker:
In cases where provider options are set using the `=()` method, the proposed changes work. But not all options use this method. An example from the vagrant-libvirt docs:
```
Vagrant.configure("2") do |config|
  config.vm.provider :libvirt do |libvirt|
    libvirt.cpus = 4
    libvirt.cputopology :sockets => '2', :cores => '2', :threads => '1'
  end
end
```
As you can see, in case of `#cputopology` the value is set using a plain method call instead of asgining the vaule to a property. At this point the code breaks. Mapping provider options available from vagrant-libvirt to a simple yaml dict in molecule doesn't really work out here. The problem is that passing vagrant-libvirt provider options seems to be inconsistent in their api.

For a more complete fix of the issues, the code needs to decide how to set the provider options from the yaml dict to the actual libvirt-vagrant provider object (`#property()` vs `#property=`). To solve this, some more ruby code is needed but i'm not sure, if the vagrantfile template string is the right place for it.

ssbarnea:
@ingobecker please fix conflicts, i will give it another go on macos or linux when done.

ssbarnea:
Closed due to lack of response. Feel free to reopen after rebasing and testing via new role creation.

Issue #1506, opened 2018-10-10
MOLECULE_EPHEMERAL_DIRECTORY is a route inside /tmp directory


morenod:
# Issue Type

- Bug report
- Feature request

# Molecule and Ansible details

```
ansible --version
```

```
molecule --version
```

Molecule installation method (one of):

- source
- pip

Ansible installation method (one of):

- source
- pip
- OS package

# Desired Behavior

Please give some details of the feature being requested or what
should happen if providing a bug report.

# Actual Behaviour (Bug report only)

Please give some details of what is actually happening.
Include a [minimum complete verifiable example](http://stackoverflow.com/help/mcve) with
output of running `molecule --debug`.

I cannot create relative routes for MOLECULE_EPHEMERAL_DIRECTORY because its is forced to be inside /tmp dir

retr0h:
> I cannot create relative routes for MOLECULE_EPHEMERAL_DIRECTORY because its is forced to be inside /tmp dir

Doesn't really describe the problem too well.  Either way, you can do what @wilmardo suggested.  Other than that, you're right you can't change it.

PR #2089, opened 2019-06-06
Refactored temp directory usage

ssbarnea:
To avoid hardcoded temporary directories which can create conflicts with
concurrent use or with multi-user environments, we adopt the use of
the newer tempfile.TemporaryDirectory.

For old python versions we install the backport module which provides
this functionality.

Signed-off-by: Sorin Sbarnea <sorin.sbarnea@gmail.com>
Fixes: #1506

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/molecule/scenario.py
+++ b/molecule/scenario.py
@@ -18,9 +18,15 @@
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
 
+import getpass
 import os
+import platform
 import fnmatch
 import tempfile
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
 
 from molecule import logger
 from molecule import scenarios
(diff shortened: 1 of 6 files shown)


molecule/scenario.py
```diff
@@ -226,8 +232,24 @@ def _setup(self):
             os.makedirs(self.inventory_directory)
 
 
-def ephemeral_directory(path):
+def ephemeral_directory(path=None):
+    """
+    Returns temporary directory to be used by molecule. Morelcule users should
```
themr0c: Typo. This should be: Molecule users.

molecule/scenario.py
```diff
@@ -226,8 +232,24 @@ def _setup(self):
...
-    if d:
-        return os.path.join(tempfile.gettempdir(), d)
-    return os.path.join(tempfile.gettempdir(), path)
+    if not d:
+        # Darwin is the only known platform that returns user-isolated tempdirs
+        # so we do not need to make them unique to each user.
+        if platform.system() == 'Darwin':
+            d = tempfile.gettempdir()
+        else:
+            d = os.path.join(tempfile.gettempdir(), 'mol.' + getpass.getuser())
```
MarkusTeufelberger:
```suggestion
            d = os.path.join(tempfile.gettempdir(), 'molecule.' + getpass.getuser())
```

Might be clearer what's going on or where this folder belongs to.

ssbarnea: I am afraid that composing a longer path could create issues on some platforms. I am aware of various path lengh limitations on all platforms, so I am inclined to avoid extra length. I was close to pick short folder based on a checksum instead of that approach, which makes it harder to identify.

molecule/scenario.py
```diff
@@ -226,8 +232,24 @@ def _setup(self):
...
+    else:
+        d = os.path.abspath(d)
+    if path:
+        d = os.path.abspath(os.path.join(d, path))
+        if not os.path.isdir(d):
+            Path(d).mkdir(parents=True, exist_ok=True)
+    return d
```
abadger: This function is definitely not safe.  ie: malicious users can pre-create paths to gain control over the data that is in here, possibly get molecule to overwrite important files, etc.

This problem might affect the older code as well but one new thing that makes this code less secure is the ```mkdir(parents=True``` which is present here.  If we remove the mkdir from this function, then it is at least, no less secure than the previous code.

ssbarnea:
@sshnaidm do you think you could have a look at this?

molecule/scenario.py
decentral1se: Why this was ignored, I have no idea.

PR #2104, opened 2019-06-15
Remove sharding on functional tests

ssbarnea:
Sharding tests was hacky approach for dealing with Travis build time
limitations, one that caused lots of issues.

In order to simplify the testing and ease the migration toward other
CI system, we are better-off dropping the sharding feature (which also
seems to prevent bumping pytest).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,416 +20,243 @@ services:
 
 language: python
 
-.mixtures:  # is not used by Travis CI, but helps avoid duplication
-  - &if-cron-or-manual-run-or-tagged
-    if: type IN (cron, api) OR tag IS present
-  - &py-27
-    python: "2.7"
-  - &py-36
-    python: "3.6"
-  - &py-37
-    python: "3.7"
-  - &reset-prerequisites
-    addons: {}
-    before_install:
-      - pip install tox-venv tox-tags
-    before_script: []
-    services: []
-  - &lint
-    <<: *py-37
-    <<: *reset-prerequisites
-  - &env-functional
-    TESTS_TYPE: functional
-  - &env-functional-shard_1
-    <<: *env-functional
-    PYTEST_ADDOPTS: >-
-      '-m shard_1_of_3'
-  - &env-functional-shard_2
-    <<: *env-functional
-    PYTEST_ADDOPTS: >-
-      '-m shard_2_of_3'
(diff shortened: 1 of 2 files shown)


themr0c:
Great, but we really can't do that before the need to have builds running on Travis has disappeared, because all builds on Travis will fail by design.

ssbarnea:
That is true and this is why it was marked as a draft. Still, there is something that could potentially make this real: if we manage to optimize the execution time. Currently functional takes around 60min and Travis limit is 50mins. If we cut it to ~40mins we can make it real (we need to have ~10min safety net).

decentral1se:
This level of optimisation won't be possible (or much harder) when https://github.com/ansible/molecule/pull/1811 lands (which should be landed since it introduces the foundation for functional testing drivers). I have no idea on status of the Zuul CI situation but that is what should be pursued. Agree sharding is desirable to get rid of.

decentral1se:
Ok, new idea. 

Since https://github.com/ansible/molecule/pull/2139 and https://github.com/ansible/molecule/pull/2137 show that we can run a whole bunch of Molecule in parallel I propose that we actually move forward with this attempt. We remove sharding altogether.

For the functional test suite, I propose a marker (`-m parallel`, `-m not parallel`) to split the test suite into smaller pieces to be run serially or in parallel. We configure Travis to run the functional test suite twice (with both markers, currently the best-in-class solution ala https://github.com/pytest-dev/pytest-xdist/issues/385). The unit tests will already be run with `-n auto`

This might work! Then I can finally get on with https://github.com/ansible/molecule/pull/1811 and ride into the sunset ...

decentral1se:
Hey, take a look at https://github.com/decentral1se/molecule/commit/cdf902cef7c84ccf42d61ee4515bfb65074b31f4. That's a much smaller diff with less unrelated changes for removing sharding. If we merged #2139 and #2137 then I could work from this sharding diff to refactor the Travis configuration to use `-m parallel/not parallel`.

decentral1se:
FYI, plan has changed now since further experimentation. There are 200+ functional tests which must be run in serial execution mode. So, we can't afford to remove sharding even if we're running some tests in parallel mode (~25 tests only!). So, let's close this off?

ssbarnea:
Closing because is not doable yet.

PR #2107, opened 2019-06-16
Enable profiling during test execution

ssbarnea:
Enables task and role profiling during test execution so we can see
how long different operations took.

This would allow us to focus on. optimizing tasks that take most of the
time and lower the total execution time.

Developer can still define his own callbacks which will take precedence
over the default ones.


themr0c:
Travis fails seemed quite random, I relaunched the build to make sure.

PR #2115, opened 2019-06-20
make cookiecutter yamllint match galaxy

ericsysmin:
Modified the cookiecutter version of .yamllint to match that of Ansible Galaxy to ensure parity between tests.

#### PR Type

- Feature Pull Request

### DIFF
--- a/molecule/cookiecutter/molecule/{{cookiecutter.role_name}}/.yamllint
+++ b/molecule/cookiecutter/molecule/{{cookiecutter.role_name}}/.yamllint
@@ -3,19 +3,31 @@
 extends: default
 
 rules:
-  braces: {max-spaces-inside: 1, level: error}
-  brackets: {max-spaces-inside: 1, level: error}
-  colons: {max-spaces-after: -1, level: error}
-  commas: {max-spaces-after: -1, level: error}
+  braces:
+    max-spaces-inside: 1
+    level: error
+  brackets:
+    max-spaces-inside: 1
+    level: error
+  colons:
+    max-spaces-after: -1
+    level: error
+  commas:
+    max-spaces-after: -1
+    level: error
   comments: disable
   comments-indentation: disable
   document-start: disable
-  empty-lines: {max: 3, level: error}
-  hyphens: {level: error}
+  empty-lines:
+    max: 3
+    level: error
+  hyphens:
+    level: error
   indentation: disable
   key-duplicates: enable
   line-length: disable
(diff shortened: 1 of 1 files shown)


ericsysmin:
This simply makes sure rules exist to match requirements from Ansible Galaxy. I noticed there were inconsistencies.

molecule/cookiecutter/molecule/{{cookiecutter.role_name}}/.yamllint
```diff
@@ -1,3 +1,5 @@
+---
+# Based on ansible-lint config
```
decentral1se: Any ideas how to keep it from going out of date again?

ericsysmin: @decentral1se  No, not really. Unless part of the build process we grab whatever galaxy uses. But I think the new one basically helps catch errors that also appear in Ansible-Lint as some of the YAML checks are the same we'd get in Ansible-Lint

molecule/cookiecutter/molecule/{{cookiecutter.role_name}}/.yamllint
```diff
@@ -7,5 +9,25 @@ rules:
...
+    level: error
+  comments: disable
+  comments-indentation: disable
+  document-start: disable
+  empty-lines:
+    max: 3
+    level: error
+  hyphens:
+    level: error
+  indentation: disable
```
perlun: Late to the party, but @ericsysmin - I strongly question why we would want to disable indentation checking by default? `yamllint` has it active by default so this definitely changes the behavior from what we had previously.

I understand the desire to harmonize with galaxy lint settings though. In this case, it would perhaps be better to leave questionable rules like this commented out with a clear comment indicating that they can be enabled if the use wishes to harmonize with galaxy linter settings.

Likewise with `new-line-at-end-of-file` and `trailing-spaces`; it feels odd to disable these rules.

decentral1se: An issue / PR with change may attract more attention to this.

ssbarnea: No need to reopen this discussion because we are very close of removing the role template completely from molecule and rely on galaxy. Still, I support @perlun idea that yamllint is useful. I only advise to take this to ansible-galaxy and fix it there. I will be glad to support the change.

PR #2124, opened 2019-06-23
Fix docker module deprecations

ssbarnea:
Ansible 2.8 introduces lots of deprecations around docker modules and
sadly there is no backwards compatible way to address them.

Thus we run them with different parameters on 2.8+ than with older versions.

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -50,26 +50,51 @@
         - not item.pre_build_image | default(false)
       register: docker_images
 
-    - name: Build an Ansible compatible image
+    - name: Build an Ansible compatible image (new)
+      when:
+        - ansible_version.full is version_compare('2.8', '>=')
+        - platforms.changed or docker_images.results | map(attribute='images') | select('equalto', []) | list | count >= 0
+        - not item.item.pre_build_image | default(false)
       docker_image:
-        path: "{{ molecule_ephemeral_directory }}"
+        build:
+          path: "{{ molecule_ephemeral_directory }}"
+          dockerfile: "{{ item.invocation.module_args.dest }}"
+          pull: "{{ item.item.pull | default(true) }}"
         name: "molecule_local/{{ item.item.image }}"
         docker_host: "{{ item.item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
         cacert_path: "{{ item.cacert_path | default((lookup('env', 'DOCKER_CERT_PATH') + '/ca.pem') if lookup('env', 'DOCKER_CERT_PATH') else omit) }}"
         cert_path: "{{ item.cert_path | default((lookup('env', 'DOCKER_CERT_PATH') + '/cert.pem')  if lookup('env', 'DOCKER_CERT_PATH') else omit) }}"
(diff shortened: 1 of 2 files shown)


ssbarnea:
@decentral1se I tried the omit approach and it did not work due to the way is implemented. The only other way this could work is to build the `args` dictionary dynamically but again this woudl require many extra tasks to be added (much more verbose than the two tasks approach we had here).

I already asked on #ansible-devel today and apparently there is no other way to address it. Ansible team did not really considered that use case. 

Obviously that we coulds start brancing molecule per ansible release but that would be a maintenance nightmare as nobody has time to backfor fixes to multiple brances.

ssbarnea:
FYI the way omit works is that it generates a special object which ansible knows to ignore.... in most places. Guess what, docker_image module does not support that and it will choke when it receives a dictionary with omits in it.  Even if we can patch it to allow that, nobody is going to backfor the fix to all supported ansible versions, especially as old ones receive only security patches.

decentral1se:
> docker_image module does not support that

Fack.

Yeah, well, I am hesitant on this one. We have to aim for a more maintainable solution. As this is only fixing a warning that is not breaking anything now, I'm not convinced we should get this in.

We need a solution that works for this deprecation and all the rest that are coming.

ssbarnea:
@decentral1se I asked multiple people from Ansible team and the short answer is **nope, this cannot be done**.

Unless we branch molecule we cannot write code without this kind of workarounds, but we can drop the workaround when the minimal supported version would reach 2.8.

Onece could think about using YAML anchors for removing some of the duplication between these two tasks but it would not work well, the result would be more complexity and harder to read or maintain. 

While this has bit of duplication it is very easy to spot and to make changes to both tasks, when at the top is also a good reminder of the branching involved.

Because molecule is a testing framework I do consider the presence of warnings as errors, as they would confuse users of molecule and they cannot do anything to remove them. The users will have their own warnings from their own roles and is up to them to fix them, but if we add our own ones on top of them their logs will look as the xmas tree.

I am even considering finding a way to mark molecule runtime warnings as errors, forcing us to write code that does not trigger runtime warnings.

decentral1se:
> While this has bit of duplication it is very easy to spot and to make changes to both tasks

Fair enough then.

> I am even considering finding a way to mark molecule runtime warnings as errors, forcing us to write code that does not trigger runtime warnings.

That is a good idea. Can we make it configurable? Then run it on the Ansible dev build!

abadger:
Note that this is wrong: 
> docker_image module does not support that and it will choke when it receives a dictionary with omits in it.

Individual modules do not have to support or not support omits.  It is processed earlier than that.

It is possible that you ran into a bug or an intentional limitation with omits (without knowing what you tried, I can't test to see what happened i the code) but I wanted to clarify that this isn't a problem with the docker_image module but in something else that your playbook would have done.

ssbarnea:
@abadger Mainly my first attempt was to use omit on params to make it happy but I hept getting errors from ansible regarding either presence of unknown params or presence of deprecated ones. If you manage to make it work I will be the first to review it (also buying you a beer first time we meet as I hate having two tasks).

My impression is that omit works when used in set_facts but when used directly in the `args:` of docker module, it will not have the same effect.

decentral1se:
If we see issues with our current approach further down the line at least we know that that `omit` approach is not blocked. Good to know.

PR #2125, opened 2019-06-24
Revert "Refactored ephemeral directory use (#2089)"

abadger:
This reverts commit b44fe7174c9e094b5f122a7511161154a5b6505e.

Creating files and directories with predictable names in world writable
directories like tempdir is a security issue.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/molecule/scenario.py
+++ b/molecule/scenario.py
@@ -18,15 +18,9 @@
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
 
-import getpass
 import os
-import platform
 import fnmatch
 import tempfile
-try:
-    from pathlib import Path
-except ImportError:
-    from pathlib2 import Path
 
 from molecule import logger
 from molecule import scenarios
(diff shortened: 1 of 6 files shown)


ssbarnea:
Please raise a bug first that explains the regression introduced by refactoring, otherwise the revert makes no sense.

abadger:
Creating predictably named files and directories is a well known source of CVEs.  For instance:

* https://linuxsecurity.com/news/documentation/safely-creating-temporary-files-in-shell-scripts/
* https://wiki.sei.cmu.edu/confluence/display/c/FIO21-C.+Do+not+create+temporary+files+in+shared+directories

abadger:
This commit should not have been merged in the first place as I noted the security flaw when I was asked to review the Pull request.  Please revert this immediately and rethink how you're going to go about it.

Since the code seems to want a per-user temporary file the secure decision is probably to create the temporary files somewhere in the user's home directory.

decentral1se:
> Blocking this until we have a bug for it.

@ssbarnea, We have the bug. You've been told 4 times now.

We've fixed a bug but we're worse off in terms of security. This should block any further release until we move to creating temporary folders in the home directory. If someone can commit to working on that soon, then we should not revert. If not, then we should revert.

Thanks for pursuing @abadger.

ssbarnea:
I will work on this and I will make sure we do not make a release without addressing this, one way or another.

I am inclined to prefer using operating system temp directories instead of polluting user home directories with temporary stuff.  MacOS does the right thing as the temporary directories are isolated for each user but Linux does not do this. 

The use of system temp directories is important because they are optimized for temporary file storage, likely to be different partitions, ramdisks, cleaned up automatically on reboot or schedule. 

If we do not have a way to obtain this under Linux, I will propose to start using `~/.cache/molecule` based on https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html

ssbarnea:
I created a bug for the raise issue https://github.com/ansible/molecule/issues/2126 -- and I will make a patch to address it.

decentral1se:
Thank you @ssbarnea.

ssbarnea:
Check if https://github.com/ansible/molecule/pull/2127 is fixing the issue.

abadger:
#2127  looks like it will solve the security issue to me.  This can be closed when #2127 is merged.

ssbarnea:
Different fix already merged.

Issue #2126, opened 2019-06-24
molecule should not create predictable temp folders in shared location like /tmp


ssbarnea:
Based on feedback from @abadger, feel free up update it:

> Creating predictably named files and directories is a well known source of CVEs. For instance:
> 
>     https://linuxsecurity.com/news/documentation/safely-creating-temporary-files-in-shell-scripts/
>     https://wiki.sei.cmu.edu/confluence/display/c/FIO21-C.+Do+not+create+temporary+files+in+shared+directories
> 

This bug affects only master as it was introduced by https://github.com/ansible/molecule/commit/b44fe7174c9e094b5f122a7511161154a5b6505e

rockandska:
Just want to open a Feature Request related to this but not for the same concern.
Having molecule temporary files residing in /tmp cause 2 behaviour IMO

- having two users working on the same project cause conflict because it use the same /tmp directory
- having molecule files in /tmp cause to lost the actual state when powering off the computer or have a power outage.

but i just realize, since the molecule instances name doesn't contain an uuid based on the project path yet, that it would result to conflict too if two users launch the same scenario

Having molecule temporary files reside in ~/.cache would be a better place to have those files surviving a reboot and solve this issue too.

Regards,

PR #2127, opened 2019-06-24
Adopt use of XDG cache directories

ssbarnea:
Assures that molecule temp/cache files are created in a safe location
and that they cannot be accessed by other users.

Fixes: #2126

### DIFF
--- a/molecule/scenario.py
+++ b/molecule/scenario.py
@@ -18,11 +18,8 @@
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
 
-import getpass
 import os
-import platform
 import fnmatch
-import tempfile
 try:
     from pathlib import Path
 except ImportError:
@@ -240,16 +237,11 @@ def ephemeral_directory(path=None):
     """
     d = os.getenv('MOLECULE_EPHEMERAL_DIRECTORY')
     if not d:
-        # Darwin is the only known platform that returns user-isolated tempdirs
-        # so we do not need to make them unique to each user.
-        if platform.system() == 'Darwin':
-            d = tempfile.gettempdir()
-        else:
-            d = os.path.join(tempfile.gettempdir(), 'mol.' + getpass.getuser())
-    else:
-        d = os.path.abspath(d)
-    if path:
-        d = os.path.abspath(os.path.join(d, path))
-        if not os.path.isdir(d):
-            Path(d).mkdir(mode=0o700, parents=True, exist_ok=True)
+        d = os.getenv("XDG_CACHE_HOME", os.path.expanduser("~/.cache"))
+    d = os.path.abspath(os.path.join(d, path if path else 'molecule'))
+
+    if not os.path.isdir(d):
+        os.umask(0o077)
+        Path(d).mkdir(mode=0o700, parents=True, exist_ok=True)
+
     return d


decentral1se:
A review would be much appreciated @abadger :+1:

molecule/scenario.py
```diff
@@ -240,16 +237,11 @@ def ephemeral_directory(path=None):
...
-    if path:
-        d = os.path.abspath(os.path.join(d, path))
-        if not os.path.isdir(d):
-            Path(d).mkdir(mode=0o700, parents=True, exist_ok=True)
+        d = os.getenv("XDG_CACHE_HOME", os.path.expanduser("~/.cache"))
+    d = os.path.abspath(os.path.join(d, path if path else 'molecule'))
+
+    if not os.path.isdir(d):
+        os.umask(0o077)
+        Path(d).mkdir(mode=0o700, parents=True, exist_ok=True)
```
decentral1se: Given the original comment in https://github.com/ansible/molecule/pull/2089#discussion_r294474426, we're now OK with this `parents=True` stuff?

ssbarnea: It is ok for me. I think that it would make no sense to question what-if someone already hacked you machine scenario. 

I don't really care what happens if user is using a custom ephemeral directory which points to a non existing location or a location what has previously hacked.

Once the user starts to customize under the hood options it is its responsibility how he uses them.

Now the default behaviour is safe, much safer that what used to be.

abadger: Yeah, sorry I was unclear about the precise reason that parents 9and exists_ok as well) are dangerous.  ```parents=True``` and ```exist_ok=True``` are especially bad in a world-writable directory because they mean that we are unaware whether we have created the directory structure or an attacker has created the directory structure.  When we root the directory structure in the user's home directory, we are okay to use them because we consider that the user's home directory is only writable by the user to begin with.  Applications like ssh, will go a step further and verify permissions on the directory structure to make sure that an attacker cannot write to any of the directories involved.

So this new code looks like it closes the hole.

molecule/scenario.py
```diff
@@ -240,16 +237,11 @@ def ephemeral_directory(path=None):
...
-        else:
-            d = os.path.join(tempfile.gettempdir(), 'mol.' + getpass.getuser())
-    else:
-        d = os.path.abspath(d)
-    if path:
-        d = os.path.abspath(os.path.join(d, path))
-        if not os.path.isdir(d):
-            Path(d).mkdir(mode=0o700, parents=True, exist_ok=True)
+        d = os.getenv("XDG_CACHE_HOME", os.path.expanduser("~/.cache"))
+    d = os.path.abspath(os.path.join(d, path if path else 'molecule'))
```
abadger: Note: With this new code you can just set ```path='molecule'``` in the function's parameter list instead of setting it to ```None``` in the parameter list and then munging it here.

ssbarnea: @abadger I am fully aware of that but I decided not to do it because I want to be able to change its behaviour/location without having to the API. Now the api has default to None, which means `auto`. 

Makes sense?

abadger: Hmmmm....  I think we're getting into a tangent here so I do want to stress, I don't think this discussion blocks this change being merged :-)

If the point of an auto location is to allow you to change it without breaking user expectations at a later time, I think it's reasonable to use None as a sentinel value meaning "auto" but that alone isn't going to be obvious to the user.  I think there's several things that you could do to make that apparent:

1. Document that if path is not given, then an *arbitrary* default value will be chosen.
2. Make the location less convenient to discourage someone from depending on it after perusing the filesystem rather than reading the code's documentation.  I notice that some programs using the ~/.cache directory add a hash or base64 encoded string to the filenames they use.  That would be one possibility here.  If you want the directory to be reusable between runs, using something like this would work: ```"molecule-%s" % uuid.uuid5(uuid.NAMESPACE_URL, 'https://github.com/ansible/molecule').hex[:8]``` (Values are arbitrary as we don't actually need something unique here; just something to discourage users from depending on the name being stable if they come across the name on the filesystem.)

PR #2133, opened 2019-06-25
Use podman_image_info module instead of shell

sshnaidm:
When https://github.com/ansible/ansible/pull/57962 is merged,
let's change shell command to podman_image_info module.

#### PR Type

- Feature Pull Request


sshnaidm:
Well, podman connect module will be only in ansible 2.9 as well. It will be all together.

sshnaidm:
@ssbarnea it was backported to 2.8 via https://github.com/ansible/ansible/pull/58437 so I think we're good.

Issue #1702, opened 2019-01-24
parallelism of pytest tests


themr0c:
# Issue Type

- Bug report

To execute the tests, pytest-xdist is installed as a requirement, but pytest is invoked without the -n option which would permit to have some parallelism during tests.

 See: https://pypi.org/project/pytest-xdist/

In pytest.ini, we could complete the addopts line with some option relevant to travis runner (to be determined).

```
[pytest]
addopts = -n3
```
This option has to be tested: maybe it could influence positively the duration of travis builds.

themr0c:
Travis engines have 2 cores:

https://docs.travis-ci.com/user/reference/overview/#virtualization-environments

webknjaz:
FTR xdist internally identifies Travis and know that it has two cores

webknjaz:
Depends on: #1573
Fix PoC: https://github.com/webknjaz/molecule/commit/87ccd8190064f9bdd8aa1069f8dddad8ed804f47

lwm:
Added https://github.com/ansible/molecule/issues/1573#issuecomment-461816683. Any further ideas on this? I'm wondering if we can use [State](https://github.com/ansible/molecule/blob/master/molecule/state.py#L42) to have all this naming issue resolved internally (some k:v where instance maps to the instance+uuid naming or so) and keep the end-user facing instance names.

webknjaz:
Not sure, it'd be also weird: when users will run `docker ps` they'll see instance names with UUIDs...

webknjaz:
@lwm after rethinking what you wrote I think that you're actually on the right track and using state would probably be beneficial.

decentral1se:
Further data points: https://github.com/ansible/molecule/pull/1715#issuecomment-472048823.

ssbarnea:
I doubt we are facing a pytest-xdist issue here, is mostly the lack of concurrency support in molecule which prevents us from running tests in parallel. I attempted to run functional ones and it mainly felt appart. Still, this does not mean that we shoud not sort this. It just means that it will take a lot of time to make required changes for allowing molecule to run multiple instances in parallel, under the same user on a single machine. 

I do want to be able to run molecule in parallel on multiple roles and multiple scenarios and I already have repositories where this is desirable. At this moment we run sequentially but this increases the runtime considerably as the number of scenarios and roles increases.

ssbarnea:
I created ``feature/parallel`` branch so we can start working towards this goal.

decentral1se:
[...]

PR #2137, opened 2019-06-28
Add experimental ``--parallel`` functionality

decentral1se:
Supercedes the spike in https://github.com/ansible/molecule/pull/2135. Working towards closing https://github.com/ansible/molecule/issues/1702.

I've tried to incorporate comments from last review but as it's all a bit experimental, it was not possible to work them all in. However, I think this is a more elegant solution here. Ready for a review pass. 

Q/A Testing much welcome (see "how to test" in https://github.com/ansible/molecule/pull/2135#issue-292474757).

/cc @ssbarnea @seandst @themr0c

### DIFF
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -18,6 +18,7 @@
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
 
+from uuid import uuid4
 import os
 
 import anyconfig
@@ -109,10 +110,15 @@ class Config(object):
         self.ansible_args = ansible_args
         self.config = self._get_config()
         self._action = None
+        self._run_uuid = str(uuid4())
 
     def after_init(self):
         self.config = self._reget_config()
         self._validate()
+
+    @property
+    def is_parallel(self):
+        return self.command_args.get('parallel', False)
 
     @property
     def debug(self):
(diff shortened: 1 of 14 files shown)


decentral1se:
> a) Make sure user can define environment variable to enable parallel mode.

Done in 295e167a1e1f087235d1aeda0de611a43347b297.

> b) I do not fully understand why the new filter needs any parameters? cannot let the filter implementation take care of adding the uuid when needed?

It's true that it is not ideal. I did this because at run-time, Molecule loads back the configuration from file and the `config.is_parallel` is not available from there. I'll try to pass this in the context of the filter or make it available from the disk saved state.

c) I do have some concerns regarding effects on existing playbooks as people may use ansible fitlering based on hostname, I need to test it to see exactly what happens...

I'll take another run at trying to remove this filter (as noted above). Ideally, this can all be taken care of internally ...

decentral1se:
I've managed to internalise the entire workings of this now with https://github.com/ansible/molecule/pull/2137/commits/e5ac12c196f7db7ddec94e34bb761f48cfd0134c. I use the new `MOLECULE_STATE_FILE` lookup to store the configuration (instead of passing it in manually into the playbook context) and hijack the `from_yaml` loading to re-apply the parallelization changes to the `molecule.yml` at run-time.

(I'll rebase all these fixups into a single commit once the review phase is over ...)

molecule/config.py
```diff
@@ -138,6 +144,10 @@ def action(self, value):
     def project_directory(self):
         return os.getcwd()
 
+    @property
```
decentral1se: Ah, this is already actually covered by the `molecule_directory` property. Will remove.

decentral1se: Oh wait, got it wrong. `molecule_directory` refers to the role molecule folder, not the cache folder. Ignore me ;)

PR #2139, opened 2019-07-01
Attempt to parallelize unit tests

decentral1se:
After https://github.com/ansible/molecule/pull/2137, I think I understand a little further that the unit tests fixtures just "pretend" to build up the usual config/state folder hierarchy manually. So, we override that to add in a random UUID and push `-n auto` into only the unit test environments from the Travis configuration. This might work ...

Another one for https://github.com/ansible/molecule/issues/1702.

### DIFF
--- a/test/unit/conftest.py
+++ b/test/unit/conftest.py
@@ -18,12 +18,17 @@
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
 
+from uuid import uuid4
 import copy
 import functools
 import glob
 import os
 import re
 import shutil
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


test/unit/conftest.py
```diff
@@ -167,9 +172,11 @@ def molecule_scenario_directory_fixture(molecule_directory_fixture):
 
 @pytest.fixture
 def molecule_ephemeral_directory_fixture(molecule_scenario_directory_fixture):
-    path = pytest.helpers.molecule_ephemeral_directory()
+    path = pytest.helpers.molecule_ephemeral_directory(str(uuid4()))
```
ssbarnea: This worries me a little bit because potential piling up of remaining file if you break execution. Can't we make the unique code predictable based on test name and python interpreter used, or something similar.

Feel free to ignore my comment if it too hard to fix at this stage, is more of a question, as we can always fix it later.

decentral1se: Actually, everything is cleaned up. See https://github.com/ansible/molecule/pull/2139/files#diff-93a7df2f35536e469f98fe57544195f7R178. Pytest works like this: yield hands the fixture to the test and then when the test is finished, whatever comes after the yield is the "teardown". So we delete each folder as we make it.

decentral1se:
> This goes against my previous effort to simplify the travis file.
> 
> I see no reason why we would want to define a PYTEST_ADDOPS variable that is different specially for unit tests. Instead just define a global one that would be inhered by all travis jobs.

This is an interim step towards just writing `-n auto` in the `pytest.ini`. We cannot do that now because the functional tests have not been prepared to run in parallel. So, to at least prove that my change is on the way to being correct, I pass `-n auto` only to the unit tests.

I don't see this as complicating the Travis configuration. Arguably it refactors it. It takes a constant and puts it in one place (instead of all over the place) and then references that. When we can paralleliza all the tests, we just remove this: "PYTEST_ADDOPTS: >- '-n auto'" and we're done.

decentral1se:
Done in https://github.com/ansible/molecule/pull/2140.

PR #2140, opened 2019-07-02
Parallelize unit test execution

ssbarnea:
Enable parallel mode when calling unittests. User is still able to disable this behavior by defining his own PYTEST_ADDOPTS value.


ssbarnea:
Hurrah! One step closes to full parallel support.

PR #2144, opened 2019-07-03
Improve pytest support

ssbarnea:
* Fixes issue where `pytest --collect-only` went broken because on CI
  we always call it with specific folders.
* Adds check tox environment which can be used for any further testing
  that we may want that does not fit linting, unit and functional.
* Assures we run check at the same time as lint on Travis
* Check reuses the virtualenv in order to avoid overloading disk with
  extra environments.
* Saves HTML report generated by pytest-html which is handy even for
  local executions.
* Installs pytest-molecule plugin in order to verify that the presence
  of the plugin does not negatively impact pytest execution.


tox.ini
```diff
@@ -30,7 +31,9 @@ deps =
     pytest==4.6.3
     pytest-cov==2.7.1
     pytest-helpers-namespace==2019.1.8
+    pytest-html>=1.21.1
     pytest-mock==1.10.4
+    pytest-molecule>=1.0.1
```
decentral1se: We don't use it.

tox.ini
```diff
@@ -58,6 +61,14 @@ commands =
 whitelist_externals =
     find
 
+# generic environment that should cover anything that is not linter or a unit/functional test
+[testenv:check]
+# reuse existing envdir to avoid increased footprint
+envdir = {toxworkdir}/py27-ansible25-unit
+commands =
+    # verifies that pytest collection works without specific folders mentioned
+    python -m pytest --collect-only --html={envlogdir}/report.html --self-contained-html
```
decentral1se: Testing that collection works and report generation is saved are things that you can do in your personal setup easily. I'd rather not add any further test environments especially something that is for a "generic" need (read: YAGNI).

.travis.yml
```diff
@@ -107,9 +107,9 @@ jobs:
 
   include:
     - <<: *lint
-      name: linting the code
+      name: linting the code and extra checks
       env:
-        TOXENV: lint
+        TOXENV: lint,check
```
decentral1se: Not really into this. The testenv should do one thing. Also in the description of the "check" it said it "should cover anything that is not linter" but then you included it in the linting environment?

tox.ini
```diff
@@ -30,7 +31,9 @@ deps =
     pytest==4.6.3
     pytest-cov==2.7.1
     pytest-helpers-namespace==2019.1.8
+    pytest-html>=1.21.1
```
decentral1se: I wouldn't add this as we don't need it. You can generate reports yourself locally.

tox.ini
decentral1se: > Installs pytest-molecule plugin in order to verify that the presence of the plugin does not negatively impact pytest execution.

You can do that manually. We don't need to add it here?

.travis.yml
ssbarnea: @decentral1se we cannot really afford to add more jobs to Travis. Each job adds a lot of extra delay because they are serialized, only ~3 run in parallel, the other are waiting in queue.

tox.ini
ssbarnea: I have to disagree here with a clear example: we already have `metadata-validation` one which is not always run but which can also break the code.  My intention was to merge these two into a single "check" one.

That pattern does not scale well. The total number of tox environments needs to be kept under control, the same applies to travis jobs, and nobody says that they need to match 1:1.

I am more than willing to hear about which ideas you have.

AFAIK, YAGNI does not apply here, collection was broken and we need to avoid regression, especially as is very easy to break it.

Regarding python-html part, it is not as simple as you could imagine because while I can locally define PYTEST_ADDOPTS with extra params for it, running it from tox it will fail because pytest-html will not be installed into the venv managed by tox. Also, the report will not be stored in tox log specific to each environment because I would not be able to use `{envlogdir}` expansion. So, shortly it would be impossible to use it without altering the source-code. 

Sadly travis does not provide any support for accesing build artifacts, as we would have to upload them ourselves to an external storage. With upcoming zuul situation is much better as it does also knows to collect tox logs dirs by default. If you are curious click on openstack-tox-molecule job from https://review.opendev.org/#/c/667056/ and you can see the report.

tox.ini
ssbarnea: I will remove it.

.travis.yml
decentral1se: Well, how about using this pre-commit then since actually this job does *a lot* of stuff and that is the whole point of pre-commit, in the end, no? We could adapt from https://github.com/szebenyib/pre-commit-pytest?

decentral1se: Or pre-commit serves other more "lint" reasons ... ugh, it's all a big blur to me

tox.ini
decentral1se: We should use a test coverage service instead of pytest-html, no? Why not https://coveralls.io/ or something similar?

tox.ini
decentral1se: https://github.com/ansible/molecule/pull/2144#discussion_r300634637

tox.ini
decentral1se: I am OK with "check" env, we can afford a small fast one, no? Maybe it's just easier to do it this way.

.travis.yml
ssbarnea: While there is always the question about why not running it in lint, I would like to keep the pre-commit run to the linting, Few reasons:
* pre-commit is the only dependency in that environment, as the tool has its own manged venvs
* more check will slowdown lint, and people will start skipping  it before push

PR #2146, opened 2019-07-03
Only test default driver with delegated

decentral1se:
Hear me out first before clicking no.

* "Delegated" is known as the "you're on your own" driver choice and I am A-OK with only testing the ability for it to work with the default driver as a sanity check that it still does what it says it does
* We can significantly reduce the workload of the functional tests in this way so that we can work towards removing the sharding that will simplify the entire functional testing setup.
* We don't have resources to test this. If Zuul CI every does land (no guarantee), we might have more (but my first reason is enough IMHO). It's uncessary workload which is not sustainable for our resources, both in terms of computers and human capacity.

### DIFF
--- a/test/functional/test_command.py
+++ b/test/functional/test_command.py
@@ -44,8 +44,7 @@ def driver_name(request):
 
 
 @pytest.mark.parametrize(
-    'scenario_to_test, driver_name, scenario_name',
-    [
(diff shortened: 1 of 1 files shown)


decentral1se:
> that change makes sense for the moment.

:100:

PR #2147, opened 2019-07-03
WIP: Use markers to run some functional tests with `--parallel`

decentral1se:
Relies on https://github.com/ansible/molecule/pull/2146 and https://github.com/ansible/molecule/pull/2137 to be merged. Adds a new shard which only selects functional tests that can be parallelizable. I've come to realise we cannot removing the sharding since we can only parallelize a small part of the functional tests. This is an experiment to see how this runs on the CI. Up for comments!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,15 +42,19 @@ language: python
   - &env-functional-shard_1
     <<: *env-functional
     PYTEST_ADDOPTS: >-
-      '-m shard_1_of_3'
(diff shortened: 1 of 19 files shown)


decentral1se:
OK, now that #2146 and #2137 are coming down the tubes, I'll close this off and have a fresh PR to show how we can test the `--parallel` stuff with markers in a 4th shard. I am not sure if we'll go with this, but it will be a smaller diff to consider.

Issue #2142, opened 2019-07-03
molecule managed resourced should be identifable by tags


ssbarnea:
Currently molecule does not tag builded images which makes impossible to identify which contrainers or images from a docker host were created by it.

Same issue applies to other providers and we need to assure that we add at least one tag that would allow us to identify molecule managed resources.

Once we do start tagging resources it will be possible to implement a `molecule reset` feature that would cleanup lefovers, from current runs or any other molecule runs.

While implementing this we can safely assume that we cannot rely on any local data persistency, all the information should be contained in each used provider.

PR #2148, opened 2019-07-03
Allows identifying the EC2 instances through the AWS console

gustavomcarmo:
Just replaced the tag **instance** with **Name** for creating the EC2 instances. It allows identifying the EC2 instances through the AWS console.

#### PR Type
- Feature Pull Request

### DIFF
--- a/molecule/cookiecutter/scenario/driver/ec2/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
+++ b/molecule/cookiecutter/scenario/driver/ec2/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
@@ -77,13 +77,14 @@
         instance_type: "{{ item.instance_type }}"
         vpc_subnet_id: "{{ item.vpc_subnet_id }}"
         group: "{{ security_group_name }}"
-        instance_tags:
-          Name: "{{ item.name }}"
+        instance_tags: "{{ item.instance_tags | combine({'instance': item.name})
+          if item.instance_tags is defined
+          else {'instance': item.name} }}"
         wait: true
         assign_public_ip: true
         exact_count: 1
         count_tag:
-          Name: "{{ item.name }}"
+          instance: "{{ item.name }}"
       register: server
       loop: "{{ molecule_yml.platforms }}"
       loop_control:
@@ -104,7 +105,7 @@
     - name: Populate instance config dict
       set_fact:
         instance_conf_dict: {
-          'instance': "{{ item.instances[0].tags.Name }}",
+          'instance': "{{ item.instances[0].tags.instance }}",
           'address': "{{ item.instances[0].public_ip }}",
           'user': "{{ ssh_user }}",
           'port': "{{ ssh_port }}",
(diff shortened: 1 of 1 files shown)


ssbarnea:
@gustavomcarmo Can you please provide more information on that, like: why the instance was not good and why the use of "Name" tag is better. By looking at the code I am unable to figure-out why one would be prefered to another. Does any of them have a special meaning?

Interestengly this seems to be related to https://github.com/ansible/molecule/issues/2142 -- so we may also want to add a tag that marks it as ``managed-by: molecule`` (or owned-by), so we can easily filter resources managed by molecule.

gustavomcarmo:
The *Name* tag is better because it's not possible to identify the EC2 console through the AWS console without. The instances names are shown in the column Name:

aws-console

This console snapshot was taken during the execution of https://github.com/codeyourinfra/codeyourinfra/blob/31e1d815a06e3c1b00accf541394432223166df9/unarchive_from_url_param/molecule/aws/create.yml#L75.

And yes, this PR is related to #2142.

decentral1se:
Should we add a new field to the schema for the user to pass these tags in? Because I can imagine that the next person comes along to add their own tag. They should be configurable?

gustavomcarmo:
I don't think so, unless we can identify some use cases when having configurable tags would be better. This PR has nothing to do with my own preference, I truly believe this tiny change can bring value to those who would like to distinguish the EC2 instances when looking to the AWS console.

decentral1se:
> unless we can identify some use cases when having configurable tags would be better. 

"Because I can imagine that the next person comes along to add their own tag."

https://github.com/ansible/molecule/pull/2148#issuecomment-508443225

gustavomcarmo:
Do you guys prefer the change below?

```yml
        instance_tags: "{{ item.instance_tags | combine({'instance': item.name})
          if item.instance_tags is defined
          else {'instance': item.name} }}"
```

It works as well. And add `instance_tags` to the schema.

I've just tested: https://github.com/codeyourinfra/unarchive_from_url_param/commit/c1128cd0890bf5579987a60fcddc97e41ea8bd53

gustavomcarmo:
@decentral1se I've just pushed the https://github.com/ansible/molecule/pull/2148#issuecomment-511333243 alternative.

molecule/cookiecutter/scenario/driver/ec2/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
```diff
@@ -77,8 +77,9 @@
         instance_type: "{{ item.instance_type }}"
         vpc_subnet_id: "{{ item.vpc_subnet_id }}"
         group: "{{ security_group_name }}"
-        instance_tags:
-          instance: "{{ item.name }}"
+        instance_tags: "{{ item.instance_tags | combine({'instance': item.name})
```
decentral1se: `instance_tags` needs to be added to the schema, no? Check `molecule/model/schema_v2.py` for more.

decentral1se: Not sure exactly what output this gives but it looks fine.

gustavomcarmo: There's no schema validation for ec2 driver yet. Should I implement it too?

decentral1se: Oh god, really? :face-palm:

Let's just take this PR as is then.

decentral1se:
Alright, let's get this one in. Thanks :+1:

PR #2151, opened 2019-07-05
Remove old delegated driver tests

decentral1se:
Follow on from #2146. We don't use these right now.


decentral1se:
No functional changes here, just removing unused files.

PR #2152, opened 2019-07-05
Adopt use of black instead of yapf

ssbarnea:
This change is switching our code-formatter from yapf to black, which
become far more popular, mainly because it is not configurable. This
means that project maintainers no longer have the option to introduce
their own oppionated settings.

Black is now hosted under official Python Foundation GitHub organization
and was already adopted by major projects like pytest and django.

Another serious benefit is speed, black is incomparebly faster than yapf
and can be used as a pre-commit hook.


ssbarnea:
Please review this, is only reformated by black, the faster we do the switch the easier it will be.

decentral1se:
Add a `pyproject.toml` with the following:

```bash
[tool.black]
skip-string-normalization = true
```

ssbarnea:
@decentral1se Addressed your concerns but I will make another change to also change the quotes as I think that one would be useful too (for consistency). Still, it makes no sense to delay adopting your suggestion as the first step.

Regarding existing PRs, we have only 30 now and their authors can easily rebase them and re-run linting, not a real deal. We used to have >100 ones so now we are much better.

decentral1se:
> @decentral1se Addressed your concerns but I will make another change to also change the quotes as I think that one would be useful too (for consistency). Still, it makes no sense to delay adopting your suggestion as the first step.

Huh!? The whole point is to **not** change the quotes to reduce the diff. I have no idea what you mean by "consistency" in this context. My concern is to address consistency (keep using single quote style we already have) while yours is to break consistency and increase the diff size. Please spare me this purism of "not having a config" because Black's "philopsophy" or whatever..

ssbarnea:
@decentral1se please re-review, the current code followed all your comments. I just want to merge it ASAP because otherwise I will have to rebase it after each other PR merged.

Issue #2149, opened 2019-07-04
create playbook doesn't pass item.pull to docker_container so pre-build containers are not updated..


skyscooby:
# Issue Type

- Bug report

# Molecule and Ansible details

Tried setting pull: true in molecule.yml but newer latest image is not actually pullled..
`item.item.pre_build_image = True` for me.. today pull is only used when containers are build locally.

Should the call to this module not pass the pull parameter?
https://github.com/ansible/molecule/blob/master/molecule/provisioner/ansible/playbooks/docker/create.yml#L123

As documented here?
https://molecule.readthedocs.io/en/stable/configuration.html#docker

# Desired Behavior

if pull: true is set for the platform.. it should pull the latest image.

# Actual Behaviour

A fresh latest image is not pulled.

decentral1se:
It's being passed https://github.com/ansible/molecule/blob/master/molecule/provisioner/ansible/playbooks/docker/create.yml#L85 and https://github.com/ansible/molecule/blob/master/molecule/provisioner/ansible/playbooks/docker/create.yml#L62.

skyscooby:
I am using a pre-build (poorly named, should be pre-built) image, rather than building the test images every single test... we have a hundred ansible roles that all are tested with the same base images, there is no resources or time to build this image every single pipeline. But they do get updated occasionally when we move the stable tag for the image we would like the pipelines to pull the newer version of the test conatiner..  the docker_container module also accepts the pull argument .. and in the case a use has defined both pre_build_image = True and pull = True then I believe it should be passed to the docker_container task.

PR #2154, opened 2019-07-05
Pull docker containers at launch as well as build.

skyscooby:
Add pull to docker_container module
Closes: #2149


skyscooby:
DCO added. In terms of tests I added the parameter into the create.yml used by the tests however I'm not sure there is any specific scenario testing around the pull feature. I'll leave that to those intimate with the test suite to decide.

wilmardo:
@skyscooby  Thanks for the PR! To be awesome this needs an update in the `molecule/model/schema_v2.py` and then tests added in `test/unit/model/v2/test_platforms_section.py`

See https://github.com/ansible/molecule/pull/2192 for an example (not a real clean example, if you just look at sysctl it should be clear)

skyscooby:
Thanks @wilmardo, as near as I can tell these two files already have the pull parameter defined. This is not a new parameter but simply attaching the pull request to the launch of the docker container rather than using it only on the build step.

wilmardo:
Ah cool missed that, thanks for checking!

PR #2155, opened 2019-07-06
Add `--parallel` to functional test suite

decentral1se:
We have to add a 4th shard and markers because of the limitations of pytest-xdist. If the tests are passing and there is not so much of a huge difference in time spent on the Travis build, then I think we should move forward with this. We need to have more guarantees that the `--parallel` stuff is working and guarded from regressions.

PS. Let's coordinate merging with #2152.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,15 +42,19 @@ language: python
   - &env-functional-shard_1
     <<: *env-functional
     PYTEST_ADDOPTS: >-
-      '-m shard_1_of_3'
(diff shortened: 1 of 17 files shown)


decentral1se:
Closes https://github.com/ansible/molecule/issues/1702.

decentral1se:
I'm adapting https://github.com/ansible/molecule/pull/1811 to use this 4th shard / pytest marker approach and it appears to be working well. Let's get this merged soon please so I can move ahead with getting the live functional testing done for the Hetzner Cloud driver.

PR #2162, opened 2019-07-08
Make sure unwanted payloads are skipped

decentral1se:
See https://github.com/ansible/molecule/pull/2162#issuecomment-510373989.

### DIFF
--- a/molecule/provisioner/ansible/plugins/filters/molecule_core.py
+++ b/molecule/provisioner/ansible/plugins/filters/molecule_core.py
@@ -68,6 +68,8 @@ def get_docker_networks(data):
 
 
 def _parallelize_config(data):
+    if 'platforms' not in data:
+        return data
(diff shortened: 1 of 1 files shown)


decentral1se:
Unfortunately we can't unit test (import configuration causes issues, see last CI build). It is triggered by the functional tests. The tests in https://github.com/ansible/molecule/pull/1811 do cover this but they're not ready yet. I assume this will break other peoples playbooks, so better get it in first to un-block any releasing. It's a patch for the work in ba351aaf27cbc2e980602f1aafee3e81f947413f.

decentral1se:
> The tests in #1811 do cover this but they're not ready yet. I assume this will break other peoples playbooks, so better get it in first to un-block any releasing. It's a patch for the work in ba351aa.

To clarify, the work in #1811 won't be in now but I have validated this patch manually.

That's what I can do for now.

PR #2163, opened 2019-07-08
Unpin test depedencies and migrate them into test extras

ssbarnea:
This moved test deps into an extra, which will allow us to install
molecule test environment without tox.

By unpining them we will avoid potential version conflicts as the small
risk of detecting breakages. If detected we can easily add capping
until we address the issues.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -125,7 +125,7 @@ test =
     flake8>=3.6.0, < 4
 
     mock>=3.0.5, < 4
-    pytest>=4.6.3, < 6
+    pytest>=4.6.3, < 5
     pytest-cov>=2.7.1, < 3
     pytest-helpers-namespace>=2019.1.8, < 2020
     pytest-mock>=1.10.4, < 2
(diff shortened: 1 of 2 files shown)


ssbarnea:
@decentral1se i do not want to put upper bounds until is proven that they break. I think that the chance of having them breaking something is very small. 

Even if it happens it will be a benefit because developers with active PRs will discover them and make PRs to fix them (otherwise we rely on maintainers to perform the never ending bump-testing).

Note: I am perfectly ok to put upper bonds as soon we discover a problem with one of them but without this even if we assume they use semantic-versioning we are only taking a gamble, most of the project I seen are changing versions without any correlation between chance-of-breaking somethign.

Pip has major releases every other month but usually does not break stuff. Other smaller package could have a minor version bump that cause something.

BTW, If you are really strongly for putting upper bonds, please suggest values to pick.

decentral1se:
It will break and you know it. This is the shifting sands of software.  Look at each dependency, check the changelog and versioning scheme and place an upper bound to match it. Otherwise, just leave the pinning and we can rely on dependabot. That's my 2 cents.

PR #2167, opened 2019-07-10
Unpin psutil

ssbarnea:
Pinning psutil makes installation process more problematic and causing
either conflicts or requiring a development python libraries and the gcc
compiler (psutil being a compiled extension).

As a first step we should unpin it and also aim to have it addressed by
https://github.com/click-contrib/click-completion/issues/25

Related to: https://github.com/ansible/molecule/issues/2168

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -82,7 +82,7 @@ install_requires =
     paramiko >= 2.5.0, < 3
     pathlib2; python_version<"3.2"
     pexpect >= 4.6.0, < 5
-    psutil == 5.4.6; sys_platform!="win32" and sys_platform!="cygwin"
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -82,7 +82,7 @@ install_requires =
     paramiko >= 2.5.0, < 3
     pathlib2; python_version<"3.2"
     pexpect >= 4.6.0, < 5
-    psutil == 5.4.6; sys_platform!="win32" and sys_platform!="cygwin"
+    psutil >= 5.4.6; sys_platform!="win32" and sys_platform!="cygwin"
```
decentral1se: Put an upper bound.

PR #2170, opened 2019-07-10
 fixes #2150: Vagrant driver regression

### LINKED ISSUE #2150: Molecule 2.22rc3: Vagrant driver regression
# Issue Type

- Bug report

Ansible: 2.8.1
Molecule 2.22rc3

Python: 2.7.12
Vagrant: 2.0.2
Virtualbox: 5.2.30r130521
Ubuntu 16.04.6 LTS x86_64

- pip

- pip

# Desired Behavior

Running Molecule test on a sample role with a Vagrant driver and a Virtualbox provider should result in a Virtualbox VM being brought up at some point.

# Actual Behaviour

Debug output snippet:

```
    TASK [Create molecule instance(s)] *********************************************
    task path: /usr/local/lib/python2.7/dist-packages/molecule/provisioner/ansible/playbooks/vagrant/create.yml:8
    <127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: go
    <127.0.0.1> EXEC /bin/sh -c 'echo ~go && sleep 0'
    <127.0.0.1> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /var/go/.ansible/tmp/ansible-tmp-1562189970.7-142759633793038 `" && echo ansible-tmp-1562189970.7-142759633793038="` echo /var/go/.ansible/tmp/ansible-tmp-1562189970.7-142759633793038 `" ) && sleep 0'
    Using module file /usr/local/lib/python2.7/dist-packages/molecule/provisioner/ansible/plugins/libraries/molecule_vagrant.py
    <127.0.0.1> PUT /var/go/.ansible/tmp/ansible-local-10147KOol3i/tmpmVJNT9 TO /var/go/.ansible/tmp/ansible-tmp-1562189970.7-142759633793038/AnsiballZ_molecule_vagrant.py
    <127.0.0.1> EXEC /bin/sh -c 'chmod u+x /var/go/.ansible/tmp/ansible-tmp-1562189970.7-142759633793038/ /var/go/.ansible/tmp/ansible-tmp-1562189970.7-142759633793038/AnsiballZ_molecule_vagrant.py && sleep 0'
    <127.0.0.1> EXEC /bin/sh -c '/usr/bin/python2 /var/go/.ansible/tmp/ansible-tmp-1562189970.7-142759633793038/AnsiballZ_molecule_vagrant.py && sleep 0'
    <127.0.0.1> EXEC /bin/sh -c 'rm -f -r /var/go/.ansible/tmp/ansible-tmp-1562189970.7-142759633793038/ > /dev/null 2>&1 && sleep 0'
    The full traceback is:
    WARNING: The below traceback may *not* be related to the actual failure.
      File "/tmp/ansible_molecule_vagrant_payload_VrLKoY/__main__.py", line 395, in up
        self._vagrant.up(provision=provision)
      File "/usr/local/lib/python2.7/dist-packages/vagrant/__init__.py", line 337, in up
        self._call_vagrant_command(args)
      File "/usr/local/lib/python2.7/dist-packages/vagrant/__init__.py", line 963, in _call_vagrant_command
        stderr=err_fh, env=self.env)
      File "/usr/lib/python2.7/subprocess.py", line 541, in check_call
        raise CalledProcessError(retcode, cmd)

    failed: [localhost] (item={'box': u'ubuntu/xenial64', 'name': u'Ubuntu'}) => {
        "ansible_loop_var": "item",
        "changed": false,
        "invocation": {
            "module_args": {
                "config_options": {},
                "force_stop": false,
                "instance_interfaces": [],
                "instance_name": "Ubuntu",
                "instance_raw_config_args": null,
                "platform_box": "ubuntu/xenial64",
                "platform_box_url": null,
                "platform_box_version": null,
                "provider_cpus": 2,
                "provider_memory": 512,
                "provider_name": "virtualbox",
                "provider_options": {},
                "provider_override_args": null,
                "provider_raw_config_args": null,
                "provision": false,
                "state": "up"
            }
        },
        "item": {
            "box": "ubuntu/xenial64",
            "name": "Ubuntu"
        },
        "msg": "ERROR: See log file '/tmp/molecule/test-role-vagrant/default/vagrant-Ubuntu.err'"
    }
```

Content of sample error file identified in Molecule output:
```
### 2019-07-03 10:08:56 ###
### 2019-07-03 10:08:56 ###
A Vagrant environment or target machine is required to run this
command. Run `vagrant init` to create a new Vagrant environment. Or,
get an ID of a target machine from `vagrant global-status` to run
this command on. A final option is to change to a directory with a
Vagrantfile and to try again.
```

Note that with the same sample role and environment, rolling back to Molecule 2.20.1 via `pip install molecule==2.20.1` fixed the problem and Vagrant boxes were again created successfully.

jonasjoest:
#### PR Type:
Bugfix Pull Request

Fixes #2150 

**Changes:**
Return environment variables if they are already set. This prevents the Ansible-Module `molecule-vagrant` from mistakenly appending another sub-path to the ephemeral directory in it's context.

**Results when retrieving `config.ephemeral_directory` in `molecule-vagrant` Ansible-Module:**
Before (wrong):
`$MOLECULE_EPHEMERAL_DIRECTORY/<CWD>/<SCENARIO_NAME>/`
After (expected):
`$MOLECULE_EPHEMERAL_DIRECTORY`

### DIFF
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -138,7 +138,7 @@ class Config(object):
 
     @property
     def project_directory(self):
-        return os.getcwd()
+        return os.getenv('MOLECULE_PROJECT_DIRECTORY', os.getcwd())
 
     @property
     def cache_directory(self):
--- a/molecule/scenario.py
+++ b/molecule/scenario.py
@@ -145,6 +145,10 @@ class Scenario(object):
 
     @property
     def ephemeral_directory(self):
+        _ephemeral_directory = os.getenv('MOLECULE_EPHEMERAL_DIRECTORY')
+        if _ephemeral_directory:
+            return _ephemeral_directory
+
         project_directory = os.path.basename(self.config.project_directory)
 
         if self.config.is_parallel:


jonasjoest:
Well to be honest in my opinion this is a interface design issue. So it is up to you to decide whether `self._config.scenario.ephemeral_directory` should always result in the commonly used `ephemeral_directory`.

Is it really intended that the vagrant driver uses
`/tmp/mol.[user]/molecule/[project]/[scenario]/molecule/vagrant/default`
instead of
`/tmp/mol.[user]/molecule/[project]/[scenario]` ???


> ```diff
> @@ -471,7 +471,7 @@ def _write_configs(self):
>  
>      def _get_vagrant(self):
>          env = os.environ.copy()
> -        env['VAGRANT_CWD'] = os.environ['MOLECULE_EPHEMERAL_DIRECTORY']
> +        env['VAGRANT_CWD'] = self._config.scenario.ephemeral_directory
>          v = vagrant.Vagrant(out_cm=self.stdout_cm, err_cm=self.stderr_cm, env=env)
>  
>          return v
> ```
If you only change the current directory in vagrant, you are providing just a workaround, but the root problem itself is not addressed at all.



### My understanding for the interface definition is the following:
_Correct me if I'm wrong._
* `config.project_directory` shall always point to the project directory
* `config.scenario.ephemeral_directory` shall always point to the common ephemeral directory, no matter what driver is used (for example `/tmp/mol.[user]/molecule/[project]/[scenario]`)


Both of those requirements are not fullfilled with the current (master) core config implementation:
* `config.project_directory` points to another working directory if a subprocess changes the current working directory (which is the case for the `vagrant` driver since it uses the `sh` library to start `ansible-playbook` in another working directory. This is the reason why the term `vagrant` appears as `basename` for `project_directory`)
* `config.scenario.ephemeral_directory` points to a non-deterministic path that depends on the way the (sub-)process interfering with the config class gets called
* currently, `config.scenario.ephemeral_directory` will be assembled from the following components:
** the env var `MOLECULE_EPHEMERAL_DIRECTORY`  if present **[the root cause of the sub-path problem]**
** the basename of the `project_directory` (which is in fact the current working directory of the (sub-)process, i.e. `vagrant` because the `create.yml` playbook for `vagrant` is located in this folder)
** a few other things like cache directory env vars or scenario name





By the way this issue is not necessarily only vagrant specifc. If any other driver or component, that relies on the base config, would start subprocesses in different `CWDs` than the "real" `project_directory`, both the `project_directory` and `ephemeral_directory` attributes would output the wrong path

decentral1se:
Sorry, haven't had time to look at this but will soon.

jonasjoest:
> * `config.project_directory` shall always point to the project directory
> 
> * `config.scenario.ephemeral_directory` shall always point to the common ephemeral directory, no matter what driver is used (for example `/tmp/mol.[user]/molecule/[project]/[scenario]`)

I can provide some tests that check against those requirements as soon as you confirm them.
But if I have misunderstood the purpose of the related config attributes, that would be fine aswell.

decentral1se:
> I can provide some tests that check against those requirements as soon as you confirm them.

Hey, so - I didn't write this stuff in the first place and have more or less guessed at the same conclusions that you've had. There is no fixed design for these properties as you've also seen because they are generated based on what is passed in.

More broadly ...

I am hesitant to go for a "one true fix" because we don't currently run tests on the drivers due to lack of resources. This is a temporary issue but long-running.

If this issue has been caused by a recent change in direction of the property and it only affects one driver (we have not seen other reports from the release candidates) then I would like to make a fix in the Vagrant implementation only.

What do you think? This is all stop-gap stuff but that's where we're at here in terms of maintenance and capacity to keep the ball rolling on this project.

jonasjoest:
Alright I agree that you don't want to introduce changes that could mess up other drivers. I've tried your solution on my setup. It seems to work (But I haven't tested it in detail).
As already mentioned the ephemeral directory is pretty much messed up with this workaround. But if you are fine with it, I don't care. I think the usual user doesn't need to go there.
Result of your patch:
workaround1

I added another patch as a proposal that ensures that at least the vagrant config and output files are in the correct ephemeral directory. But there are still some "zombie" files like "state.yml" in the nested ephemeral directory.

https://github.com/jonasjoest/molecule/commit/7e67f6cb134fc1a23c5cf254ac7ce7facf9b662a

Result of the new proposal workaround:
workaround2

jonasjoest:
IMHO I'm not really able to tell which of those two patches would stabilize the vagrant driver implementation the most. I just provided you another alternative.

decentral1se:
I haven't had time to look at this. My proposed solution doesn't seem to be the one (thanks for the good detective work). I am leaning back towards your original fix and diagnosis now ... want to apply it again and I'll promise to actually do something beyond talking and manually test it :) What do you think?

jonasjoest:
From my perspective as a "user" it's hard to talk about the "one". Without dedicated test cases it's even harder to suggest what I think. What I can tell is that both patches (yours and my original from the PR) lead to a working vagrant setup. (At least for me on my vagrant-libvirt setup)

What I'm not able to tell is
* Whether it works aswell on other distros, with other vargrant providers and with other special configurations
* The design of project_directory and ephemeral_directory. (Should it be static/deterministic? Then my patch isn't the one neither)

One important side effect of my patch is that user-defined environment variables take effect aswell:
So one can override the path for `ephemeral_directory` or `project_directory` by defining the env vars `MOLECULE_EPHEMERAL_DIRECTORY` and/or `MOLECULE_PROJECT_DIRECTORY` himself. From my perspective I would call that a feature, but you could call that a bug aswell. Depends on the specification what those attributes should return.



Just to complete the report of results, this is what I get when applying my original patch from this PR:
result_pr

decentral1se:
> Looks good to me, but as already mentioned this is all manual testing. I cannot prove that this works on other environments.

That'll do it. I'll also manually test and then we can make a release candidate and see what happens with that. We normally rely on a lot of user Q&A (sadly but that is the current state of the project).

Wanna get your patch in order then?

jonasjoest:
> Wanna get your patch in order then?

Sorry what do you mean by that? Did I miss something?

decentral1se:
Ah, sorry, you already applied it! OK, will test soon!

Issue #2182, opened 2019-07-24
molecule is unable to build containers which need custom /etc/hosts entries


ssbarnea:
Due to missing feature in ansible docker_image module we are unable to pass etc_param to alter it. We already support this for running the container but for building it not, so we may fail to create it.

We cannot fix this until https://github.com/ansible/ansible/issues/59233 is fixed and released.

PR #2183, opened 2019-07-24
Internally use docker module from Ansible 2.9

ssbarnea:
In order to enable molecule to make (internal) use of features or fixes that are not available in older versions of Ansible we temporary fork docker_image module from Ansible 2.9. 

This would only by used internally by molecule during docker create and there are no risks of altering the behavior of tested playbooks as the modules are loaded with different names.
 
Fixes: #2182

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -35,6 +35,9 @@
         _docker_image_info_module: >-
           {{ ansible_version.full is version_compare('2.8', '>=') |
             ternary('docker_image_info', 'docker_image_facts') }}
+        _docker_image_module: >-
+          {{ ansible_version.full is version_compare('2.9', '>=') |
+            ternary('docker_image', 'molecule_docker_image') }}
 
     - name: Discover local Docker images
       action: "{{ _docker_image_info_module }}"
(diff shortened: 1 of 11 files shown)


decentral1se:
https://github.com/ansible/ansible/pull/59540 is merged now?

Merge conflict there unfortunately ...

molecule/provisioner/ansible/playbooks/docker/create.yml
```diff
@@ -74,28 +75,6 @@
         label: "molecule_local/{{ item.item.image }}"
       no_log: false
 
-    - name: Build an Ansible compatible image (old)
```
decentral1se: Breaking change?

molecule/provisioner/ansible/plugins/libraries/README.rst
```diff
@@ -0,0 +1,11 @@
+Ansible Module overrides for Molecule
```
decentral1se: Please do this in the Sphinx documentation as well. 

Will save us time for tickets later on ...

molecule/provisioner/ansible/playbooks/docker/create.yml
ssbarnea: Not really. I was the one that split the call in two old/new for compatibility with various versions of ansible. Because now we call our own copy of that module, we no longer need the conditional, as we only use the newer version (same as one from ansible 2.9).

Tests should prove that it does not break stuff, docker is well tested.

molecule/provisioner/ansible/plugins/libraries/README.rst
ssbarnea: Not sure if I need to document the fact internally we do not use same ansible modules. Do we? Or you refer to another part of the documentation.

decentral1se: Ah, I get you. I thought that this will be possible to change through some experimental end-user intervention. If that is not the case then yes, let's avoid documenting this.

molecule/provisioner/ansible/playbooks/docker/create.yml
decentral1se: Understood.

ssbarnea:
@decentral1se This temporary approach was suggested by ansible cores on `#ansible-devel` and also on https://discuss.python.org/t/how-to-access-a-different-version-of-a-python-library-listed-as-a-dependency/2060/2 

I made it clear that this is only temporary, until two things happen:
* Ansible 2.9 is released (or at least pre-released)
* We switch to use only ansible-runner when we call `playbook.yml`, so we can decouple the version of Ansible used by molecule from the version of ansible that needs to be used by the tested playbooks.

decentral1se:
Those Ansible core people have commit/permissions access here. If they support this, then let them +1 this and away you go. I think adding 1000+ LOC, forking ansible and getting who knows what maintenance issues just for a one line addition to a module is total overkill. That's my position on this.

ssbarnea:
@cloudnuul When you have time have a look at this PR. Some reasons:
* we needed etc_hosts on docker_image for testing rhel7-rui containers
* ansible team refused to backport addition of etc_host param, but merged it to devel, so ansible2.9 will have it
* we needed the new param only for molecule internal code
* mainly this patch is "vendoring" the module from newer ansible.

I do also have another alternative in progress which would only make the vendoring optional, so we can enable it when we want and not by default.

PR #2134, opened 2019-06-26
use password specified in instance config to add host variables in generated inventory files

junaidali:
When using Molecule with AWS Windows instances, connecting over WinRM requires the initial AWS generated password to be available. 

Below are relevant sections for a working test case

molecule.yml

Make sure to change the YOURSUBNETID below

```
---
dependency:
  name: galaxy
driver:
  name: delegated
lint:
  name: yamllint
platforms:
  - name: instance
    image: ami-0781cf90f9cef3437
    instance_type: t2.micro
    vpc_subnet_id: YOURSUBNETID
    groups:
      - windows
provisioner:
  name: ansible
  log: True
  connection_options:
    ansible_connection: winrm
  inventory:
    group_vars:
      windows:
        ansible_winrm_transport: credssp
        ansible_winrm_server_cert_validation: ignore
  lint:
    name: ansible-lint
verifier:
  name: testinfra
  lint:
    name: flake8
  options:
    v: 4
```

create.yml

```
---
- name: Create
  hosts: localhost
  connection: local
  gather_facts: false
  no_log: "{{ not (lookup('env', 'MOLECULE_DEBUG') | bool or molecule_yml.provisioner.log|default(false) | bool) }}"
  vars:
    ssh_user: ec2-user
    winrm_user: Administrator
    ssh_port: 22
    winrm_http_port: 5985
    winrm_https_port: 5986
    winrm_init_userdata_script: init_windows.txt
    security_group_name: ansible-molecule-dev-com
    security_group_description: Security group for testing Molecule
    security_group_rules:
      - proto: tcp
        from_port: "{{ ssh_port }}"
        to_port: "{{ ssh_port }}"
        cidr_ip: '0.0.0.0/0'
      - proto: icmp
        from_port: 8
        to_port: -1
        cidr_ip: '0.0.0.0/0'
      - proto: tcp
        from_port: "{{ winrm_http_port }}"
        to_port: "{{ winrm_http_port }}"
        cidr_ip: '0.0.0.0/0'
      - proto: tcp
        from_port: "{{ winrm_https_port }}"
        to_port: "{{ winrm_https_port }}"
        cidr_ip: '0.0.0.0/0'
    security_group_rules_egress:
      - proto: -1
        from_port: 0
        to_port: 0
        cidr_ip: '0.0.0.0/0'

    keypair_name: molecule-key
    keypair_path: "{{ lookup('env', 'MOLECULE_EPHEMERAL_DIRECTORY') }}/ssh_key"
[...]

PR #2184, opened 2019-07-25
feat(delegated): Adds password and winrm options to instance dict

wilmardo:
#### PR Type

- Feature Pull Request

Follow up of #2134, credits to @junaidali for the initial fix but patched and tested this locally so I had the code already :)

This set of variables seems to cover most of the WinRM connection setups, but of course I am open for discussion if something else needs mapping.

The mutually exclusive isn't a hard exclusive since Ansible would work and tries to use the key first but for documentation purpose I stated it this way to avoid confusion.


wilmardo:
@decentral1se Should I add a changelog entry?

decentral1se:
Yes please. I normally put it together myself when we get near a release ...

wilmardo:
@decentral1se Done!

PR #2196, opened 2019-08-04
[DOC] ssh_connection_options updating documentation

fernandoalex:
- `driver.ssh_connection_options` is expecting a list

#### PR Type

- Docs Pull Request

Using the configuration provided by the documentation molecule fails with:
```
ERROR: Failed to validate.

{'driver': [{'ssh_connection_options': ['must be of list type']}]}
```

### DIFF
--- a/molecule/driver/azure.py
+++ b/molecule/driver/azure.py
@@ -54,7 +54,7 @@ class Azure(base.Base):
         driver:
           name: azure
           ssh_connection_options:
-            -o ControlPath=~/.ansible/cp/%r@%h-%p
+            - '-o ControlPath=~/.ansible/cp/%r@%h-%p'
 
     .. important::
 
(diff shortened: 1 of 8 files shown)


molecule/driver/azure.py
```diff
@@ -54,7 +54,7 @@ class Azure(base.Base):
         driver:
           name: azure
           ssh_connection_options:
-            -o ControlPath=~/.ansible/cp/%r@%h-%p
+            - '-o ControlPath=~/.ansible/cp/%r@%h-%p'
```
ssbarnea: I doubt it will work, but without space between -o and ControlPath it should. Lets see.

What I do not understand is why we did not see a failure on CI related to this?

decentral1se: It's in the documentation string and not executed? Perhaps we should make sure and override a functional test with such a connection setting just to make sure?

PR #2180, opened 2019-07-24
Fixed deprecation warning for pytest.config in test/conftest.py.

atombrella:

PR #2202, opened 2019-08-05
Fixed invalid use of env params on Travis

ssbarnea:
Replaced dictionaries with lists on travis config as the former are
considered invalid syntax.

While invalid travis does parse them but not without side-effects, like
seeing this on ENV section of the Travis jobs:

One of the secure variables in your .travis.yml has an invalid format.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,22 +40,6 @@ language: python
   - &hetzner-cloud-cleanup
     after_script:
       - ./contrib/tokenmgmr.py remove --driver hetznercloud
-  - &env-functional
-    TESTS_TYPE: functional
-  - &env-hetzner-cloud
-    HCLOUD_TOKEN: "$(./contrib/tokenmgmr.py get --driver hetznercloud)"
-  - &env-functional-shard_1
-    <<: *env-functional
-    PYTEST_ADDOPTS: >-
-      '-m shard_1_of_3'
-  - &env-functional-shard_2
-    <<: *env-functional
-    PYTEST_ADDOPTS: >-
-      '-m shard_2_of_3'
-  - &env-functional-shard_3
-    <<: *env-functional
-    PYTEST_ADDOPTS: >-
-      '-m shard_3_of_3'
 
 jobs:
   fast_finish: true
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -40,147 +40,120 @@ language: python
   - &hetzner-cloud-cleanup
     after_script:
       - ./contrib/tokenmgmr.py remove --driver hetznercloud
-  - &env-functional
```
decentral1se: Don't delete this, specify the new list style syntax here.

There is a reason this is here. Changing it later is repeating ourselves everywhere.

ssbarnea: There is a reason why I removed it, we cannot use it any longer. I will explain:

Because `env:` is no longer a dictionary and now is a list, we no longer can use YAML anchors to combine dictionaries. Lists cannot be combined like dictionaries. That is documented limitation of `<<:` operator.

I am for deduplicating code and avoiding repetition. Still, in this case the `deduplicated` code had the same length as the non-deduplicated one (count lines before/after). Mainly we added a new layer which introduced bugs.

If you find another way to reorganize the file in which `env:` blocks are lists, I would be more than happy to take it.

ssbarnea: What is critical in this change is to assure that the triggered jobs use the right argumens as we do not want to introduce typos.

decentral1se:
> One of the secure variables in your .travis.yml has an invalid format.

Where's the error?

.travis.yml
decentral1se: You'd save time if you put this in the description.

The PR just looks like a refactoring without this explanation.

ssbarnea:
> Where's the error?

Take a look at https://github.com/ansible/molecule/pull/2180/checks?check_run_id=182919773 on the ENV column in travis results table.

If you google for that specific error you will find numerous bugs and also PRs that were fixing it, like https://github.com/nodejs/docker-node/pull/949/files 

For example vscode is even failing to lint the file because if this, as [travis.yml schema](http://json.schemastore.org/travis) requires list of strings.
travis-errors

Issue #2204, opened 2019-08-05
Don't be blocked on making a new release by failure on the ansible devel job


decentral1se:
When did this become the case? I have no idea. 

We need to remove this restriction.

ssbarnea:
I am against this proposal. In fact I would be more likely to drop support for oldest ansible in module than not testing our code with ansible-devel.

Despite historical experience, Ansible has not much better test coverage and they do not introduce random changes in devel. That is our chance of preventing a failure, before is released.

In fact my recent PR was to to enable execution of devel all the time, so we do not break molecule releases because we forgot to test on devel.

I personally think that we are better off with a (N-1,N+1) support model than with the `(N-2,N)` one but such decision is not be made by me alone.

decentral1se:
> not testing our code with ansible-devel.

That's not the proposal.

This about continuing to test against ansible devel but not making it a blocking issue for releases.

decentral1se:
> In fact my recent PR was to to enable execution of devel all the time, so we do not break molecule releases because we forgot to test on devel.

See https://github.com/ansible/molecule/pull/2205#issuecomment-518555152.

decentral1se:
Don't care anymore.

PR #2205, opened 2019-08-06
Assures ansibledevel jobs run all the time

ssbarnea:
Decision to skip running ansibledevel bought us some speed boost at
the cost of no longer testing code and introducing costly bugs.

This caused failure to release 2.22rc4 which failed on ansibledevel
jobs, which were not run during normal PRs.

From now on we need to be sure that we test all PRs with the same set of
jobs as we do for releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,55 +43,6 @@ language: python
 
 jobs:
   fast_finish: true
-
-  allow_failures:
-    - <<: *py-37
-      env:
-        - TOXENV=ansibledevel-unit
-
-    - <<: *py-37
-      env:
-        - PYTEST_ADDOPTS='-m shard_1_of_3'
-        - TOXENV=ansibledevel-functional
-    - <<: *py-37
-      env:
-        - PYTEST_ADDOPTS='-m shard_2_of_3'
-        - TOXENV=ansibledevel-functional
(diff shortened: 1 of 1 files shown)


decentral1se:
No. Because we already save in the 2.8 -> 2.9 release cycle of Ansible that we had to wait 1 - 2 months for a fix to come down the line so that we could fix the devel build support here. Furthermore, when we have a working working group, we made a decision that it was untenable to support Ansible devel.

> This caused failure to release 2.22rc4 which failed on ansibledevel jobs, which were not run during normal PRs.

The root cause of this problem is not solved with this change. That is tracked in https://github.com/ansible/molecule/issues/2204.

seandst:
Neither the commit nor the description here on github mention that in addition to enabling the ansible devel builds on all jobs, it also no longer allows failures on those ansible devel jobs. Running the ansible devel jobs makes sense to me, but blocking PRs on their failures does not, for the main reason outlined in @decentral1se's comment:

Failures in molecule's ansible devel builds are not always going to be caused by bugs in molecule, so molecule PRs can not reasonably be blocked on the results of ansible devel build results.

ssbarnea:
After additional talks I agree that running on devel may put too much strain on Travis.

PR #2206, opened 2019-08-06
Adds retries to image building tasks

ssbarnea:
As it was observed in CI, docker_image build can fail due to temporary
network flakiness.

We add 3 retries with 30 delay between them, so we can avoid most
random failures.

Example: https://travis-ci.com/ansible/molecule/jobs/222705245#L582

Error building molecule_local/centos - code: None,
message: Get https://registry-1.docker.io/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)


ssbarnea:
@gabrielecerami @cloudnull Let me know what you think about this approach.

Issue #2207, opened 2019-08-06
anyconfig install failure prevents installation of molecule on rhel8


ssbarnea:
Pasted the error here, I will be back to update with more details as I think is likely to be unrelated to molecule itself:
```
Exception:
Traceback (most recent call last):
  File "/root/rdo/ansible-role-container-registry/.tox/molecule/lib/python3.6/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/root/rdo/ansible-role-container-registry/.tox/molecule/lib/python3.6/site-packages/pip/commands/install.py", line 342, in run
    prefix=options.prefix_path,
  File "/root/rdo/ansible-role-container-registry/.tox/molecule/lib/python3.6/site-packages/pip/req/req_set.py", line 784, in install
    **kwargs
  File "/root/rdo/ansible-role-container-registry/.tox/molecule/lib/python3.6/site-packages/pip/req/req_install.py", line 851, in install
    self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
  File "/root/rdo/ansible-role-container-registry/.tox/molecule/lib/python3.6/site-packages/pip/req/req_install.py", line 1064, in move_wheel_files
    isolated=self.isolated,
  File "/root/rdo/ansible-role-container-registry/.tox/molecule/lib/python3.6/site-packages/pip/wheel.py", line 347, in move_wheel_files
    assert info_dir, "%s .dist-info directory not found" % req
AssertionError: anyconfig>=0.9.7 .dist-info directory not found

```

Already found the culprit: https://github.com/ssato/python-anyconfig/issues/110

PR #2208, opened 2019-08-06
Revert anyconfig version to last known good

ssbarnea:
Also adds note that we should attempt to remove it from our dependencies
due to risks involving its maintenance (or lack of it).

Fixes: #2207

Related to https://github.com/ssato/python-anyconfig/issues/110

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -72,7 +72,8 @@ install_requires =
 
     # TODO(ssbarnea): consider anyconfig removal due to maintenance risks
     # https://github.com/ssato/python-anyconfig/issues/110
-    anyconfig == 0.9.7, !=0.9.8, !=0.9.9
+    # 0.9.8, 0.9.9 are known to be broken
+    anyconfig == 0.9.7
     flake8 >=3.6.0
     cerberus >= 1.3.1
     click >= 6.7
--- a/setup.cfg
+++ b/setup.cfg
@@ -70,7 +70,9 @@ install_requires =
     ansible >= 2.5
     ansible-lint >= 4.0.2, < 5
 
-    anyconfig >= 0.9.7
+    # TODO(ssbarnea): consider anyconfig removal due to maintenance risks
+    # https://github.com/ssato/python-anyconfig/issues/110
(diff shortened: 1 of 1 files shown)


ssbarnea:

ekultails:
LGTM!

ekultails:
I don't think we can officially approve the review since we're not listed as reviewers.

setup.cfg
```diff
@@ -70,7 +70,9 @@ install_requires =
     ansible >= 2.5
     ansible-lint >= 4.0.2, < 5
 
-    anyconfig >= 0.9.7
+    # TODO(ssbarnea): consider anyconfig removal due to maintenance risks
+    # https://github.com/ssato/python-anyconfig/issues/110
+    anyconfig == 0.9.7, !=0.9.8, !=0.9.9
```
seandst: I believe `anyconfig == 0.9.7` already excludes every other version of anyconfig, including 0.9.8 and 0.9.9.

ekultails:
Nevermind, I figured out how to approve/deny reviews (by going to the "Files changed" tab).

@seandst does have a good point that we can simplify the dependency check. @ssbarnea Mind updating that?

setup.cfg
ssbarnea: @seandst Obviously but listing these two excludes we send a clear message: we know these were tested and broken. Is like documenting which ones are known to be broken, so next time someone considers bumping them
they with avoid making the same mistake.

I seen this practice used in the past. I could have written it as a comment but it woud have being longer and
easier to ignore. 

Also, by doing this the pyup bot will be silenced, and it will no longer nag us.

seandst: I assure you that listing those two excludes does not send as clear a message as you suggest. A comment would, however *absolutely* send a clear message about the intent of those excludes, and I'd rather have something in there that's easy to ignore than nothing at all to explain an otherwise nonsensical version requirement.

ssbarnea:
While testing on rhel-8 would require some more advanced tricks the error I got was not specific to rhel itself. I think I could be able to reproduce it with an older version of pip. I am considering writing a new test specially for this. Still no reason to delay merging it before we have a test.

Issue #1908, opened 2019-04-08
Skip dependencies linting


beenje:
# Issue Type

- Feature request

# Molecule and Ansible details

```
ansible --version && molecule --version
ansible 2.7.9
  config file = /Users/benjaminbertrand/.ansible.cfg
  configured module search path = [u'/Users/benjaminbertrand/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/benjaminbertrand/miniconda3/envs/molecule220/lib/python2.7/site-packages/ansible
  executable location = /Users/benjaminbertrand/miniconda3/envs/molecule220/bin/ansible
  python version = 2.7.15 | packaged by conda-forge | (default, Feb 27 2019, 20:44:16) [GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
molecule, version 2.20.0
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

# Desired Behavior

We use one git repository per role.
If a role has some dependencies, running `molecule lint` in a role also checks the other roles that are in the parent directory. That shouldn't be the case. The dependencies should not be checked.

This is an issue because on our gitlab-runners, all projects are cloned in the same directory and linting can fail due to some old roles in the working directory.

I think it's easier to explain with an example:

## How to reproduce the issue

If we take a role_b that depends on role_a:
```
$ cat role_b/meta/main.yml
---
dependencies:
  - role_a

$ cat $ cat role_b/molecule/default/requirements.yml
---
- src: git+https://github.com/beenje/role_a.git
```

If we just clone role_b, everything works fine:
```
$ git clone https://github.com/beenje/role_b.git
$ ls
role_b
$ cd role_b
$ molecule lint
--> Action: 'lint'
--> Executing Yamllint on files found in /Users/benjaminbertrand/Dev/ansible/lint/tmp/role_b/...
Lint completed successfully.
Skipping, no tests found.
--> Executing Ansible Lint on /Users/benjaminbertrand/Dev/ansible/lint/tmp/role_b/molecule/default/playbook.yml...
Lint completed successfully.
```

Now if we clone the role_a (which is a dependency of role_b) in the same directory as where role_b is, `molecule lint` in **role_b** will lint role_a as well, which is not what is expected.
```
$ cd ..
$ git clone https://github.com/beenje/role_a.git
$ ls
role_a	role_b
$ cd role_b
$ molecule lint
--> Action: 'lint'
[...]

PR #2215, opened 2019-08-09
Run dependency before lint on test_sequence

ssbarnea:
Avoids failure to run linting due to missing dependencies, like ansible modules or roles.

Fixes: #1908

### DIFF
--- a/molecule/command/lint.py
+++ b/molecule/command/lint.py
@@ -93,7 +93,7 @@ class Lint(base.Base):
     ),
 )
 def lint(ctx, scenario_name):  # pragma: no cover
-    """ Lint the role. """
+    """ Lint the role (dependency, lint). """
     args = ctx.obj.get('args')
     subcommand = base._get_subcommand(__name__)
     command_args = {'subcommand': subcommand}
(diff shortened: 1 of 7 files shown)


ssbarnea:
@cloudnull @ekultails @gabrielecerami please review this.

molecule/scenario.py
```diff
@@ -197,7 +197,8 @@ def idempotence_sequence(self):
 
     @property
     def lint_sequence(self):
-        return ['lint']
+        # see https://github.com/ansible/molecule/issues/2216
+        return ['dependency', 'lint']
```
gabrielecerami: So now dependency is implicit in lint ...

molecule/command/lint.py
```diff
@@ -93,7 +93,7 @@ def execute(self):
     ),
 )
 def lint(ctx, scenario_name):  # pragma: no cover
-    """ Lint the role. """
+    """ Lint the role (dependency, lint). """
```
gabrielecerami: Not clear here what you mean with this list

molecule/scenario.py
ssbarnea: That is matching behavior with the other commands, run `molecule --help` and you will understand.

PR #2217, opened 2019-08-09
Fix floating IP deletion in openstack driver

haukebruno:
`delete_fip: true` needs to be set in the _destroy_ playbooks rather than in the _create_ ones.

Currently the allocated floating IPs won't be released after instance deletion. Under $circumstances this could lead to quota, resource or billing issues inside an OpenStack project.

#### PR Type

- Bugfix Pull Request


ssbarnea:

ssbarnea:
@haukebruno LGTM but I guess you know that at this moment we do not have integration testing with openstack enabled in CI. 

Can you confirm that you have you succesfully used your patch with molecule?

haukebruno:
> Can you confirm that you have you succesfully used your patch with molecule?

@ssbarnea Yes :) I've tested it a couple times against our Pike/Queens Stack within a molecule scenario.

PR #2219, opened 2019-08-10
Makes molecule drivers plugins

ssbarnea:
Transforms molecule drivers into plugins, which will allow others
to write their own plugins and publish them separately.

Later we will move some plugins out of the code, into their own
repositories, where they can be tested correctly.

Part of #2228

### DIFF
--- a/molecule/api.py
+++ b/molecule/api.py
@@ -1,0 +1,13 @@
+import pkg_resources
+
+
+def molecule_drivers(as_dict=False):
+
+    plugins = {
+        entry_point.name: entry_point.load()
+        for entry_point in pkg_resources.iter_entry_points('molecule_driver')
+    }
+    if as_dict:
+        return plugins
+    else:
+        return list(plugins.keys())
--- a/molecule/command/create.py
+++ b/molecule/command/create.py
@@ -20,8 +20,8 @@
 
 import click
 
-from molecule import config
 from molecule import logger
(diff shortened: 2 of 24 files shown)


ssbarnea:

abadger:
Do we really want to use pkg_resources to implement plugins?  It causes both performance and fragility problems because it has to scan all of the installed python packages in order to function and there are plenty of other libraries to do plugins out there.  I think Fedora Infrastructure uses ```straight.plugin``` https://pypi.org/project/straight.plugin/ (which I've started using for ansible's build scripts) and mitsuhiko (jinja2 author) wrote https://pypi.org/project/pluginbase/ which seemed like a stable, reliable library as well.

ssbarnea:
@abadger Why would I add a dependency on a new module when I can implement this using one of the officially documented ways of writing plugins? See https://packaging.python.org/guides/creating-and-discovering-plugins/

This is what tox, or pytest uses. I really do not want to introduce new dependencies, especially as the projects are far from mainstream. One had last update in 2017, another one is under personal account and has one update in February, and the author in used to push directly to github. I would count both of these projects as high risk.

TBH, I think that the performance concern is a red-herring: others are using it without any performance concerns. Especially for testing tool, which loads the plugins only once, I do not see any reasons to worry -- we could add a `@lru_cache` to it if you want.

I have some concerns around molecule performance, especially its tests but that's another subject.

PR #2193, opened 2019-08-01
Adds base file values into molecule_yml variable

wilmardo:
#### PR Type

- Bugfix Pull Request

Followup of #1437 
Fixes #1423

PR #2223, opened 2019-08-13
Bash style variable expension in env var supports defaults using variables

wilmardo:
Closes: #2250

#### PR Type

- Feature Pull Request

Expand the interpolation so expands default defined variables. Based on the implementation in bash:

```shell
# export TEST=test
# echo ${NONE:-$TEST}
test
# echo ${NONE:-$NONE}


```

When the default variable is not found it returns an empty string.

I use this to postfix the platform names so no conflicts happen, in the CI it uses a predefined Gitlab variable, for local testing it takes the user running molecule.

```yaml
platforms:
  - name: Docker-CE-${CI_COMMIT_REF_SLUG:-$VMWARE_USER}
```

#### Changes:
* Expanded the existing test so it tests both the curly and non curly braced syntax
* Added a testcase for the existing default string interpolation
* Implemented default variable interpolation and added a testcase

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,9 @@ Unreleased
 * Update testinfra to 3.0.6 so we can use ansible verbosity
 * Add ``sysctls`` option to the Docker driver.
 * dependency now runs before lint on default test and lint sequences
+* Bash style variable expension for environment variable defaults added.
+  ``foo: ${UNDEFINED_VAR:-$DEFAULT}`` and ``foo: ${UNDEFINED_VAR-$DEFAULT}``
+  are now supported.
 
 2.20
 ====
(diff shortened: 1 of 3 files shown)


wilmardo:
Same unrelated PR fail as in #2193

wilmardo:
@ssbarnea Thanks for the feedback! I am open for changes and I am keen to what other have to say!

My reasoning is that this behavior lays in the extend of the syntax and is expected for the endusers (Bash is default non strict).

wilmardo:
@decentral1se Any input on this PR?

wilmardo:
@fabianvf Thanks for the opinion! I think the recursive option is a little overkill, at some point you should reconsider your approach ;)

Will fix the conflicts first thing tomorrow :)

PR #2229, opened 2019-08-17
Fix container building

ssbarnea:
* fixed bug with invalid docker tag name for untagged commits
* assured docker runs with network host to have internet access
* removed unneeded sudo which prevented use on MacOS
* assured we run tox -e docker-build on Travis to avoid regressions
* assured we also upload from Travis on deploy

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # This is a multi-stage build which requires Docker 17.05 or higher
-FROM python:3.7-alpine as molecule-builder
+FROM alpine:edge as molecule-builder
 
 WORKDIR /usr/src/molecule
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -56,15 +54,13 @@ ENV GEM_PACKAGES="\
 
 ENV MOLECULE_EXTRAS="azure,docker,docs,ec2,gce,hetznercloud,linode,lxc,openstack,vagrant,windows"
 
+RUN apk add --update --no-cache ${BUILD_DEPS} ${PACKAGES}
+RUN gem install ${GEM_PACKAGES}
+RUN apk del --no-cache ${BUILD_DEPS}
```
fabianvf: We should leave these in a single layer, I think the build deps will still contribute to the final image size because they are being stored in a separate layer.

utils/build-docker.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # Decide to push when all conditions below are met:
+    if os.environ.get('TRAVIS_BUILD_STAGE_NAME', None) == 'deploy':
+        run("docker login quay.io")
+        run("docker push {0}:{1}".format(image_name, version_tag))
+        if os.environ.get('TRAVIS_BRANCH', None) == 'master':
+            run("docker push {0}:{1}".format(image_name, mobile_tag))
+            # if we are a full release on master branch
+            if not Version(version).is_prerelease:
+                run("docker tag {0}:{1} {0}:latest".format(image_name, version_tag))
```
fabianvf: Won't we already have pushed this in line 40 (mobile_tag will be latest)

ssbarnea:
@gundalow this should take care of publishing from travis to quay. You can test it yourself using:
```
TRAVIS_BUILD_STAGE_NAME=deploy TRAVIS_BRANCH=master tox -e build-docker
```
Tag current commit with some release version to see behavior on release (should create latest tag too). It will fail at the end while trying to push to docker. Takes ~10mins due to the multi-staged approach. I do not love its speed but there are other areas needed speed more than this one.

utils/build-docker.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
+from setuptools_scm import get_version
+from packaging.version import Version
+import os
+import sys
+
+
+def run(cmd):
+    print(cmd)
+    r = os.system(cmd)
```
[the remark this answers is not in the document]
ssbarnea: Ok, I will keep in mind. The one that is totally evil is `l` as variable, good that now the linter has a test for it as I was able to spot it as being used by someone.

PR #2233, opened 2019-08-18
Force flake to use py3

bgavrilMS:
`tox -e lint` uses flake8 which defaults to python2, causing an error as the undelying tool (black) is not compatible with py2

#### PR Type

- Bugfix Pull Request


ssbarnea:
@bgavrilMS It seems that you are also affected by the DCO so until we get https://github.com/ansible/molecule/issues/1548 addressed, please rebase with signoff to make the bot happy.

And thank for the fix, not sure how it happened that we missed it.

PR #2238, opened 2019-08-22
fix rubocop and build dependencies

jamesla:
Rubocop now runs without throwing an exception


ssbarnea:
I will close it because it needs rework but feel free to open it again. Ideally you should first raise a bug with a simple reproducing script, so we know exactly what is broken and prevent regression. Please check the newest container, maybe it does not have the rubocop error anymore.

Issue #2228, opened 2019-08-17
Modularize drivers and testers in molecule


ssbarnea:
It seems that nearly everyone agrees that the monolithic approach does not scale and slows down the product development and testing, we will try to make molecule modular so most providers and drivers will be installed as external python packages, developed and tested by their own maintainers.

The goal is to use the same kind plug-gable architecture that proved to work successfully with other python based tools like: [pytest](https://docs.pytest.org/en/latest/plugins.html), [tox](https://tox.readthedocs.io/en/latest/plugins.html) or [flake8](https://flake8.pycqa.org/en/latest/user/using-plugins.html).

This change would require multiple changes that we will do gradually and without breaking the tool behavior. As our providers are currently installed as extras, we will continue to keep them installable as extras. Mainly this transition should be transparent for molecule users.

ssbarnea:
While working at sniping-off azure our of molecule at https://github.com/pycontribs/molecule-azure i realized that we need to **include molecule tests into the package** so we can re-use them while testing a driver.

This will increase a little bit the package size but is a known approach, also documented at http://doc.pytest.org/en/latest/goodpractices.html

Installing tests will not increase dependencies, as these would be installed by the testing extra.

PR #2239, opened 2019-08-22
Move tests inside the package

ssbarnea:
This change will allow use to ship the tests with the package, a move
that would allow use to test molecule plugins reusing testing code
from molecule itself.

This does not mean that test requirements will be installed by default,
they are still installed as an extra.

Partial-Fix: #2228

#### PR Type

- Feature Pull Request


decentral1se:
> This does not mean that test requirements will be installed by default,
they are still installed as an extra.

:+1:

PR #2243, opened 2019-08-23
azure: assure we respect locally configured region

ssbarnea:
Fix bug where test execution would fail if resource group already
existed in another zone.

Also assures that allocated resources are released.

#### PR Type

- Bugfix Pull Request


molecule/test/resources/playbooks/azure/destroy.yml
```diff
@@ -29,8 +29,13 @@
       retries: 300
       with_items: "{{ server.results }}"
 
-    # Mandatory configuration for Molecule to function.
+    - name: Delete virtual network
```
decentral1se: We're only deleting this on this testing scenario? Unsure.

Issue #2227, opened 2019-08-17
quay.io/ansible/molecule DIND container is missing some crucial packages


jamesla:
# Issue Type

- Bug report

The DIND docker image quay.io/ansible/molecule container is missing some crucial packages which are required for pretty much core functionality of molecule.

1. Git  - required for ansible-galaxy to install private ansible roles
2. Rubocop - a linter used by molecule is broken
3. inspec - one of the test frameworks used by molecule
4. ansible 2.8 - only 2.7 is installed and 2.8 brought out a bunch of new modules

This dockerfile used to be kept inside this project back in the retr0h days but I think it's gone now. I am happy to send a PR but where is the dockerfile hosted now?

ssbarnea:
AFAIK, the docker container used by quay.io is the Dockerfile from the root of molecule project and quay is building it itself when a tag is pushed. That made me discover https://github.com/ansible/molecule/issues/2225 which I still have no clue on how to address.

Don't forget to join irc channel, maybe we can fix few things today.

jamesla:
@ssbarnea ok so after a bit more investigation it looks like most of the dockerfile is ok with the exception of Ruby which was broken and have opened a PR here https://github.com/ansible/molecule/pull/2238

However not too sure about the quay.io docker repository and how that works. So I've pushed a temp version to jamesla/molecule in the meantime if anyone else is having problems here.

jamesla:
Ok after a bit more looking around it looks like the build system may be working as there is a build from 3 days ago however it's just never tagging the newest image with the tag latest and pushing that also. This means that if you don't hardcode a tag you get a very old version with the symptoms I posted above.

Who looks after this build system and where is it configured?

decentral1se:
What's the status on this? I see https://github.com/ansible/molecule/pull/2238 was closed but was solving something? Is https://github.com/ansible/molecule/pull/2238#issuecomment-524338726 useful? To answer the build question, we can arrange to contact the admins of this quay setup for a fix but we need to have some idea of the problem first.

jamesla:
There was another PR that went in and fixed the issues. The only remaining issue is that the image tagged as master isn't being pushed (meaning you have to specify the version when pulling the image). Other than that this is all working for me now.

PR #2248, opened 2019-08-26
Fix errors with rubocop not executing

jamesla:
This PR is in reference to issue https://github.com/ansible/molecule/issues/2227

Currently Rubocop doesn't work due to the fact that you need ruby-full rather than the regular ruby apt package and you also need the ruby-dev apt package.

I've also improved the build dependencies to use the meta package alpine-sdk which allows us to let alpine handle all of the build dependencies but size is not a problem here since we remove it before the image creation is finished.

The only way I could add a test for this was to actually run the binary during the image creation process which works but I will admit is an anti-pattern (won't scale) but couldn't think of another way. Open to ideas here.

@ssbarnea You mentioned you had some other concerns. Just let me know what they are and I can address.

Reproduction:

(old)
```
docker run quay.io/ansible/molecule rubocop:2.22 --version
```

docker: Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "exec: \"rubocop:2.22\": executable file not found in $PATH": unknown.
ERRO[0000] error waiting for container: context canceled

(new)
```
docker build . -t test
docker run test rubocop --version
```

0.74.0

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,9 +71,11 @@ LABEL maintainer "Ansible <info@ansible.com>"
 
 ENV PACKAGES="\
     docker \
+    bash \
     git \
     openssh-client \
-    ruby \
+    ruby-full \
+    ruby-dev \
     ansible \
     ansible-lint \
     docker-py \
@@ -115,11 +117,7 @@ ENV PACKAGES="\
     "
 
 ENV BUILD_DEPS="\
-    gcc \
-    libc-dev \
-    make \
-    ruby-dev \
-    ruby-rdoc \
+    alpine-sdk \
     "
 
 ENV PIP_INSTALL_ARGS="\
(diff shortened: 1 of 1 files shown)


ssbarnea:
Please fix conflicts and DCO and I will merge this. Thanks!

Dockerfile
```diff
@@ -138,12 +137,15 @@ ENV MOLECULE_EXTRAS="azure,docker,docs,ec2,gce,hetznercloud,linode,lxc,openstack
...
+    gem install --no-ri --no-doc ${GEM_PACKAGES} \
     apk del --no-cache ${BUILD_DEPS} \
     rm -rf /root/.cache
 COPY --from=molecule-builder \
     /usr/src/molecule/dist \
     /usr/src/molecule/dist
 RUN pip3 install ${PIP_INSTALL_ARGS} "molecule[${MOLECULE_EXTRAS}]"
 
+# Test rubocop binary executes
+RUN rubocop --version
```
decentral1se: Let's test this in the Tox test env and not in the standard Dockerfile?

ssbarnea: I like testing in dockerfile but it needs to avoid creating a new docker layer by reusing one of the already existing RUN commands.

Please add it after the rm from line 143 as `&& rubocop --version`.

jamesla: Ok I've done this and added a tests env var so we can expand upon this without the run command becoming unwieldy.

PR #2252, opened 2019-08-28
Removal of embedded azure driver

ssbarnea:
The same driver will be installable as a plugin named molecule-azure:
https://pypi.org/project/molecule-azure/

Partial: #2228


molecule/api.py
```diff
@@ -9,10 +13,13 @@
...
 
-    plugins = {
-        entry_point.name: entry_point.load()
-        for entry_point in pkg_resources.iter_entry_points('molecule_driver')
-    }
+    plugins = {}
+    for entry_point in pkg_resources.iter_entry_points('molecule_driver'):
+        try:
+            plugins[entry_point.name] = entry_point.load()
+        except Exception as e:
```
cloudnull: I think this might cause some unintended consequences. I think we should limit this exception to "KeyError".

ssbarnea: @cloudnull In fact this piece of code is also present in a separated review and I need it for lots of exceptions, what happens is that pkg_resources can throw a very wide range of exceptions, but redoing the plugin loading will be subject of another change.

cloudnull: understood. thanks for the explanation.

molecule/cookiecutter/molecule/{{cookiecutter.role_name}}/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/molecule.yml
```diff
@@ -53,6 +51,8 @@ platforms:
 {%- elif cookiecutter.driver_name == 'vagrant' %}
   - name: instance
     box: ubuntu/xenial64
+{%- else %}
```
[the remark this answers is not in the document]
ssbarnea: That is on purpose, so we do have an instance mentioned even when the driver nature is unknown to molecule, something that will become the rule when we move them out. I considered that is better to have an instance when mentioned in the template when the driver is another one. This also keeps the code backwards compatible with azure.

PR #2253, opened 2019-08-29
Makes driver loading more resilient

ssbarnea:
Prevents molecule execution failure when a driver fails to load. New code will display the loading failure but continue to load the other drivers.

### DIFF
--- a/molecule/api.py
+++ b/molecule/api.py
@@ -1,4 +1,8 @@
 import pkg_resources
+from molecule import logger
+
+
+LOG = logger.get_logger(__name__)
 
 try:
     from functools import lru_cache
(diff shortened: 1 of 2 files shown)


ssbarnea:
@decentral1se Not at this moment. If it fails to load we will see. FYI I am planning to redo the driver loading as I discovered that current implementation can cause other failures but for the moment this should be a enough. The concept should remain the same: "failure to load a driver should be ignored, mainly because drivers are going to be external / outside our control" and we do not want to break the tool when one driver is broker, one that probably we do not even use.

ssbarnea:
@gabrielecerami In tests we can easily check that `foo` is in reported drivers. Current change is  minimal stuff that we can do to avoid case where user does "pip install molecule-my-broken-driver" and the entire tools becomes impossible to use. In fact this is what happened to me while writing a driver.

decentral1se:
> Where's the test for this :) ?

Yeah, good point. Let's get one of those. You can use the [pytest caplog](https://docs.pytest.org/en/latest/logging.html) fixture to easily make sure that logging errors were outputted as expected.

PR #2254, opened 2019-08-29
Cache loaded drivers

ssbarnea:
Caching loaded drivers has serious performance benefits, to have an idea
running unit tests went from ~4min to 1.5mins.

### DIFF
--- a/molecule/api.py
+++ b/molecule/api.py
@@ -1,6 +1,8 @@
 import pkg_resources
+from molecule.util import memoize
 
 
+@memoize
 def molecule_drivers(as_dict=False):
 
     plugins = {


molecule/api.py
```diff
@@ -1,6 +1,11 @@
+try:
+    from functools import lru_cache
+except ImportError:
+    from backports.functools_lru_cache import lru_cache
 import pkg_resources
 
 
+@lru_cache()
 def molecule_drivers(as_dict=False):
```
decentral1se: Just passing by but wasn't there something like the "memoizer" already built in for caching lookups?

Otherwise, it's a bit meh to have to add a dependency just for this ...

ssbarnea: @decentral1se What can I say,.... I was not aware about existence of `util.memoize`. I will use it as we already have it.

Usually I am against adding depedencies but lru_cache is part of python since 3.3 and the only additional requirement would be the backport for 2.7 --- which I find better than having in-house reimplementation of the same feature. Anyway, I will update it to use memoize.

fabianvf:
Hmm, looking at the build failures it looks like it may not be handling caching with the various arguments well (https://travis-ci.com/ansible/molecule/jobs/229314674#L2467). An easy workaround would be to define two internal functions (ie, `__drivers_as_dict` and `__drivers_as_list`) that are both cached and leave the caching off the top level function. We could also look at fixing the implementation of `memoize` so that it takes kwargs into account (it looks like it does the lookup based on `args` only, presumably because `kwargs` is not a hashable value and would require mutation.

decentral1se:
Looks like #2259 supercedes this?

ssbarnea:
Yes.

PR #2255, opened 2019-08-29
travis: switch to 2 shards instead of 3

ssbarnea:
Using only 2 shards should be fine and will speedup testing by avoiding the queue waiting for new VMs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,45 +51,33 @@ jobs:
 
     - <<: *py-37
       env:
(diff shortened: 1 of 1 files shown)


ssbarnea:
@decentral1se AFAIK this is running the same tests, nothing is removed. The code in conftest does split the test in shards based on the mark used by parsing it: using `shard_x_of_y` is parsed and the number of shards is taken out of "y". Please double check but this is how I read it.

PR #2257, opened 2019-08-29
travis: combine similar tests into less stages

ssbarnea:
* runs unittest for specific python version in a single job
* run linting, docks, check and build-docker in a single job

This change increase the Travis execution speed by lowering the total
resources usage, benefit from caching and especially by avoiding the
queues for new VMs.

This change alone should lower the execution time by at least 10 mins.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,67 +94,25 @@ jobs:
 
   include:
     - <<: *lint
-      name: linting the code
+      name: lint,doc,check,build-docker
       env:
-        - TOXENV=lint
-
-    - <<: *lint
-      name: checking the docs build
-      env:
-        - TOXENV=doc
-
-    - <<: *lint
-      name: check,build-docker
-      env:
-        - TOXENV=check,build-docker
+        - TOXENV=lint,doc,check,build-docker
 
     - <<: *py-37
       env:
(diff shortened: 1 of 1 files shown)


ssbarnea:
@decentral1se I do not find scrolling on ~2800 lines as a real problem, especially as the errors are well colorized when happen. The PR proposer is the most likely person needing to investigate the output.

Moving drivers to plugins is not going to address this part of the CI. Is addressing speed but in a very different area.

We need to make testing as fast as possible, especially as only one change is tested by travis at a time, the others are waiting in the queue. If it takes 2h to test a job, it means that we can run no more than 12 builds per day. Anyone can raise changes, so we can easily build a long queue.

ssbarnea:
@decentral1se The problem you do highlight is caused by https://github.com/tox-dev/tox/issues/578 but I think there is way avoid it if I reorganize the travis file a little bit.

If I call `tox -e foo && tox -e bar & ...`, it will fail-fast without running other environments like it does when you define `TOXENV=foo,bar`. Would you find such an approach acceptable?

Another benefit is that I could also add `PYTEST_ADDOPTS=-n auto` for unit tests in particular, which will also make them run much faster. Using parallel works fine for unittests, only functional are affected by it.

PR #2258, opened 2019-08-29
fix(provisioner): Expand ANSIBLE_ variables with Ansible default locations

wilmardo:
#### PR Type

- Bugfix Pull Request

#### Changes

* Expands the ANSIBLE_ environment variables with the Ansible default locations.

```console
$ ansible-config dump

DEFAULT_FILTER_PLUGIN_PATH(default) = ['/c/Users/wilmaro/.ansible/plugins/filter', '/usr/share/ansible/plugins/filter']
DEFAULT_MODULE_PATH(default) = ['/c/Users/wilmaro/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
DEFAULT_ROLES_PATH(default) = ['/c/Users/wilmaro/.ansible/roles', '/usr/share/ansible/roles', '/etc/ansible/roles']
```

* Moves the internal Molecule modules to a path more like upstream. Both filters and libraries are not like Ansible, filters should be singular. Libraries is singular valid for one playbook or one role but it feels more inline with Ansible to name it modules since library is only used in role or playbook context.

For more information about local modules and plugins see the Ansible documentation:
https://docs.ansible.com/ansible/latest/dev_guide/developing_locally.html#adding-a-module-locally

```console
DEBUG: ANSIBLE ENVIRONMENT
---
ANSIBLE_CONFIG: /c/Users/wilmaro/.cache/molecule/win-cmdb-state-sync/gather-facts/ansible.cfg
ANSIBLE_FILTER_PLUGINS: /c/Users/wilmaro/Development/Intermax/roles/win-cmdb-state-sync/venv/lib/python3.6/site-packages/molecule/provisioner/ansible/plugins/filter:/c/Users/wilmaro/.cache/molecule/win-cmdb-state-sync/gather-facts/plugins/filter:/c/Users/wilmaro/Development/Intermax/roles/win-cmdb-state-sync/plugins/filter:/c/Users/wilmaro/.ansible/plugins/filter:/usr/share/ansible/plugins/filter
ANSIBLE_HOST_KEY_CHECKING: 'False'
ANSIBLE_LIBRARY: /c/Users/wilmaro/Development/Intermax/roles/win-cmdb-state-sync/venv/lib/python3.6/site-packages/molecule/provisioner/ansible/plugins/modules:/c/Users/wilmaro/.cache/molecule/win-cmdb-state-sync/gather-facts/library:/c/Users/wilmaro/Development/Intermax/roles/win-cmdb-state-sync/library:/c/Users/wilmaro/.ansible/plugins/modules:/usr/share/ansible/plugins/modules
ANSIBLE_ROLES_PATH: /c/Users/wilmaro/.cache/molecule/win-cmdb-state-sync/gather-facts/roles:/c/Users/wilmaro/Development/Intermax/roles:/c/Users/wilmaro/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles
ANSIBLE_SSH_CONTROL_PATH_DIR: /dev/shm
ANSIBLE_STDOUT_CALLBACK: yaml
```


molecule/provisioner/ansible.py
```diff
@@ -202,11 +202,11 @@ class Ansible(base.Base):
     ::
 
         ANSIBLE_ROLES_PATH:
-          $ephemeral_directory/roles/:$project_directory/../
```
decentral1se: The library -> modules and filters -> filter path changes are breaking changes or?

wilmardo: Yes but only if a user has overridden these variables in the molecule.yml. Otherwise the new paths are correctly expanded by the code change in ansible.py

wilmardo: But it shouldn't be that much of an issue since the documentation was already wrong. The path was `$ephemeral_directory/libraries` in the code but `$ephemeral_directory/library` in the docs. If have not seen any issues about it. Also the library only contains the internal vagrant module.

The filters plugin could break by the point above.

decentral1se: Uhhhh, right. Well, I'm not sure what to do then. We should perhaps include a more clarifying note in the change log that this could break things for end-users? @ssbarnea what do you think here?

molecule/provisioner/ansible.py
```diff
@@ -202,11 +202,11 @@ class Ansible(base.Base):
...
+          $ephemeral_directory/roles/:$project_directory/../:$HOME/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles
```
ssbarnea: Using $HOME seems like a real bug to me. I am sure that $HOME is not defined on many platforms, do to assume it resolves.

ssbarnea:
First, be sure it passes CI, after this I will have a second look because this change has a lot of potential,.... of breaking stuff. I do understand reasoning behind and support them (to avoid diverging from ansible default behavior), but I need to be sure we do this right or we will all pay the costs of breaking things.

wilmardo:
@ssbarnea Fixed formatting :)

@decentral1se Someone in mind? I could poke on of my colleagues to test this?

decentral1se:
> @decentral1se Someone in mind? I could poke on of my colleagues to test this?

That would be fantastic. If not, well, we can merge and rely on early Q&A testing too.

wilmardo:
Will do tomorrow!

PR #2265, opened 2019-08-31
Informs of the DOCKER_HOST environment variable with the Docker driver.

flowgunso:
Signed-off-by: flow.gunso <flow.gunso@gmail.com>

Add a note to the Docker driver documentation to inform that the DOCKER_HOST environment variable can be used to customize which Docker daemon Molecule uses.

The `docker` binary is still required on the machine that runs Molecule even when using a remote Docker instance, otherwise Molecule fails.

#### PR Type

- Docs Pull Request


molecule/driver/docker.py
```diff
@@ -119,6 +119,11 @@ class Docker(base.Base):
               less secure. For details, please reference `Docker Security
               Configuration`_
 
+    .. note:: With the environment variable ``DOCKER_HOST`` the user can define
+              which `Docker`_ daemon to use. Do note that the `Docker`_ binary
+              must be installed on the machine that run Molecule even if the
```
decentral1se:
```suggestion
              must be installed on the machine that runs Molecule even if the 
```

ssbarnea: It would worth mentioning that user can even use ssh transport protocol to connect to remote docker, something that works well for both docker-cli and docker-py but is not well documented, one of the reasons why I wrote https://docs.sbarnea.com/molecule/using-a-remote-docker some time ago.

The statement is false: user does *not* need to have docker installed on the localhost when using remoting. Docker-py talks directly with the remote host and you can avoid having to install docker locally. Molecule is using only docker-py (via ansible modules)

flowgunso:
Molecule did fail asking for the docker binary, at first. After installing/removing the docker packages, everything works fine.

ssbarnea:
@flowgunso Please fix the linting error and this is good to merge. A welcomed improvement to the docs.

flowgunso:
OK, rebased.

decentral1se:
Uhhhh, very weird CI error:

> File "/home/travis/build/ansible/molecule/molecule/driver/docker.py", line 123 E       SyntaxError: Non-ASCII character '\xc2' in file /home/travis/build/ansible/molecule/molecule/driver/docker.py on line 124, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details

Is this related!?

flowgunso:
Alright, I removed the problematic character but did not add the encoding header line.

PR #2267, opened 2019-08-31
Use pluggy to load plugins

ssbarnea:
Adopts use of pluggy for loading plugins in order to make their loading
more reliable to errors. Using pkg_resources proved to be less reliable
as its entry point iterator can break on any import, spotting to return
more entry points.

Apparently using pluggy is a better bet as it is already used by two
major tools with lots of plugins: pytest and tox.

### DIFF
--- a/molecule/api.py
+++ b/molecule/api.py
@@ -1,26 +1,50 @@
-import pkg_resources
+import pluggy
+from importlib import import_module
 from molecule import logger
-
+import traceback
 
 LOG = logger.get_logger(__name__)
-
 try:
     from functools import lru_cache
 except ImportError:
     from backports.functools_lru_cache import lru_cache
 
 
-@lru_cache()
-def molecule_drivers(as_dict=False):
+CORE_DRIVERS = [
+    "delegated",
+    "digitalocean",
+    "docker",
+    "ec2",
+    "gce",
+    "hetznercloud",
+    "linode",
+    "lxc",
+    "lxd",
+    "openstack",
+    "podman",
+    "vagrant",
+]
 
+
+@lru_cache()
+def molecule_drivers(as_dict=False, config=None):
     plugins = {}
-    for entry_point in pkg_resources.iter_entry_points('molecule_driver'):
+    pm = pluggy.PluginManager("molecule_driver")
+    for driver in CORE_DRIVERS:
+        pm.register(import_module("molecule.driver.%s" % driver), name=driver)
+    try:
+        pm.load_setuptools_entrypoints("molecule_driver")
+    except Exception:
+        # These are not fatal because a broken driver should not make the entire
+        # tool unusable.
+        LOG.error("Failed to load driver entry points %s", traceback.format_exc())
+    for p in pm.get_plugins():
         try:
-            plugins[entry_point.name] = entry_point.load()
+            plugins[pm.get_name(p)] = p.load(config)
         except Exception as e:
-            LOG.error("Failed to load %s driver: %s", entry_point.name, str(e))
+            LOG.error("Failed to load %s driver: %s", pm.get_name(p), str(e))
 
     if as_dict:
         return plugins
     else:
-        return list(plugins.keys())
+        return sorted(list(plugins.keys()))
--- a/molecule/test/unit/test_api.py
+++ b/molecule/test/unit/test_api.py
@@ -1,0 +1,34 @@
+from molecule.driver.base import Base
+from molecule.api import molecule_drivers
+
+
+def looks_iterable(a):
+    try:
+        (x for x in a)
+        return True
+    except TypeError:
+        return False
+
+
+def test_api_molecule_drivers():
+
+    # check if call without parameters returns a list of strings
+    results = molecule_drivers()
+    # print(results)
+    assert looks_iterable(results)
+    # delegated driver is expected to always be present
+    assert 'delegated' in results
+
+
+def test_api_molecule_drivers_as_dict():
+
+    # check that with as_dict it will return a dictionary of driver instances
+    results = molecule_drivers(as_dict=True)
+
+    assert looks_iterable(results)
+
+    # delegated driver is expected to always be present
+    assert 'delegated' in results
+
+    for driver in results.values():
+        assert isinstance(driver, Base)
--- a/setup.cfg
+++ b/setup.cfg
@@ -86,6 +86,7 @@ install_requires =
     paramiko >= 2.5.0, < 3
     pathlib2; python_version<"3.2"
     pexpect >= 4.6.0, < 5
+    pluggy >= 0.7.1, < 1.0
     pre-commit >= 1.17.0, < 2
     psutil >= 5.4.6, < 6; sys_platform!="win32" and sys_platform!="cygwin"
     PyYAML >= 5.1, < 6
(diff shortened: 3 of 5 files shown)


setup.cfg
```diff
@@ -86,6 +86,7 @@ install_requires =
     paramiko >= 2.5.0, < 3
     pathlib2; python_version<"3.2"
     pexpect >= 4.6.0, < 5
+    pluggy >= 0.7.1
```
decentral1se:
```suggestion
    pluggy >= 0.7.1, < 1.0
```

ssbarnea:
@decentral1se We already have tests around molecule_drivers, not very extensive but enough for the moment. My plan is to add more but only once I have the molecule-azure plugin fully working, until then there is a chance that I would want to change implementaiton to suit our needs.

decentral1se:
One. Unit. Test.

molecule/test/unit/test_api.py
```diff
@@ -0,0 +1,34 @@
...
+    except TypeError:
+        return False
+
+
+def test_api_molecule_drivers():
+
+    # check if call without parameters returns a list of strings
+    results = molecule_drivers()
+    # print(results)
+    assert looks_iterable(results)
```
decentral1se: Clearer and more explicit is:

``` python
assert isinstance(results, list)
for result in results:
    assert isinstance(result, str)
assert 'delegated' in results
```

molecule/test/unit/test_api.py
```diff
@@ -0,0 +1,34 @@
...
+    # delegated driver is expected to always be present
+    assert 'delegated' in results
+
+
+def test_api_molecule_drivers_as_dict():
+
+    # check that with as_dict it will return a dictionary of driver instances
+    results = molecule_drivers(as_dict=True)
+
+    assert looks_iterable(results)
```
decentral1se: I would do the same as above and instead make explicit what type you expect.

ssbarnea: If you referred to checking for dict, I would prefer to avoid it because what I want is "hashable", maybe I would even change the signature to make this clear. I may not return a dictionary, but something that behaves like one in the future.

molecule/api.py
```diff
@@ -1,26 +1,50 @@
...
+@lru_cache()
+def molecule_drivers(as_dict=False, config=None):
     plugins = {}
-    for entry_point in pkg_resources.iter_entry_points('molecule_driver'):
+    pm = pluggy.PluginManager("molecule_driver")
+    for driver in CORE_DRIVERS:
+        pm.register(import_module("molecule.driver.%s" % driver), name=driver)
+    try:
+        pm.load_setuptools_entrypoints("molecule_driver")
+    except Exception:
```
cloudnull: can we better scope this exception handling?

molecule/api.py
```diff
@@ -1,26 +1,50 @@
...
+        # These are not fatal because a broken driver should not make the entire
+        # tool unusable.
+        LOG.error("Failed to load driver entry points %s", traceback.format_exc())
+    for p in pm.get_plugins():
         try:
-            plugins[entry_point.name] = entry_point.load()
+            plugins[pm.get_name(p)] = p.load(config)
         except Exception as e:
```
cloudnull: same as above ^

molecule/api.py
```diff
@@ -1,26 +1,50 @@
...
-            LOG.error("Failed to load %s driver: %s", entry_point.name, str(e))
+            LOG.error("Failed to load %s driver: %s", pm.get_name(p), str(e))
 
     if as_dict:
         return plugins
     else:
-        return list(plugins.keys())
+        return sorted(list(plugins.keys()))
```
cloudnull: why does this list need to be sorted? Not that the action has much impact, just a question as to why this is being done for an internal method?

setup.cfg
```diff
@@ -86,6 +86,7 @@ install_requires =
     paramiko >= 2.5.0, < 3
     pathlib2; python_version<"3.2"
     pexpect >= 4.6.0, < 5
+    pluggy >= 0.7.1, < 1.0
```
cloudnull: The development status of the `pluggy` library is set to beta (https://pypi.org/project/pluggy/ - [4 - Beta]). Are we sure we want to run with this given development of this library may break everything?

molecule/api.py
cloudnull: if we do need/want to sort the return we could simply run the following

``` python
sorted(plugins.keys())
```

setup.cfg
decentral1se: If pytest and tox are OK with it, I guess we can run with it too.

molecule/api.py
decentral1se: To clarify, you mean, use a more explicit exception type than `Exception` or?

molecule/api.py
decentral1se: As far as I understand, the `molecule drivers` command will list them? So, this list ends up being shown to the user. I haven't checked the code though. If that is right, sorting makes sense.

decentral1se: As far as I understand, the `molecule drivers` command will list them? So, this list ends up being shown to the user. I haven't checked the code though. If that is right, sorting makes sense.

molecule/test/unit/test_api.py
decentral1se: Shameless copy/pasta:

> my point is: test what you expect the code to return
> ↳if you plan to change it in the future, then change the test
> ↳you might not come back to this test for $reason and then it could make no sense to someone else
> ↳what is "looks iterable" when there is some random failure a few months down the line
> ↳better practice to test against explicit types

molecule/api.py
cloudnull: yes, something more explicit. I'm no fan of `except Exception`, unless its **REALLY** needed.

molecule/api.py
cloudnull: If that's the case, then :+1:

setup.cfg
cloudnull: That may be, but in the name of stability using a beta lib seems risky; maybe its just mislabeled? That said, if we're aware of the risks and the rate of change isn't breaking tox and pytest frequently then, :neutral_face:

molecule/api.py
fabianvf: I feel like it makes sense here, even if we added specific Exception catches we'd probably want the generic one at the end, I can't think of any error that we'd want to let bubble out to the user without wrapping it in the below log message (also as pointed out in the comment exceptions here shouldn't result in a crash, since the driver may not end up being used at all).

molecule/test/unit/test_api.py
ssbarnea: I run out of energy around this patch and that's bad because the driver loading code in master is currently very broken, so broken that any pip conflict would break molecule, not even `molecule --version` would work. Here is the proof https://gist.github.com/ssbarnea/526c2c453d2714e659a9e1f6de1fbf56#file-gistfile1-txt-L61-L85

I more than open to improve the testing and the behavior of the role so anyone is welcomed to update this CR and bring up to mergeable standards. As long it no longer contains `iter_entry_points` I will support it.

decentral1se:
OK, I'm re-opening this with the proposed review changes :+1: 

Also see https://github.com/ansible/molecule/pull/2267#discussion_r321979002.

PR #2271, opened 2019-09-05
ec2: per-platform ansible_connection overrides and win password lookup

troyready:
#### PR Type

- Feature Pull Request

#### Overview

To provision Linux & Windows hosts simultaneously, the same connection_options can't be shared on the provisioner. This PR allows provisioner connection_options to be customized per-platform.

Additionally, if the winrm connection is used on an EC2 instance and a password isn't supplied. the password will be retrieved automatically via boto3 (decryption logic copied from ec2_win_password module)

### DIFF
--- a/molecule/driver/ec2.py
+++ b/molecule/driver/ec2.py
@@ -113,6 +113,8 @@ class EC2(base.Base):
               sudo: False
               ansible_user: Administrator
               # Specify a password to override the automatic lookup
+              # or omit to retrieve automatically
+              # (requires boto3 & cryptography packages)
               # ansible_password: hunter2
               ansible_port: 5986
               ansible_connection: winrm
--- a/molecule/driver/ec2.py
+++ b/molecule/driver/ec2.py
@@ -106,7 +106,7 @@ class EC2(base.Base):
           name: ec2
         platforms:
           - name: instance
-            image: ami-06a0d33fc8d328de0)
+            image: ami-06a0d33fc8d328de0
             instance_type: t3a.medium
             vpc_subnet_id: subnet-1cb17175
             connection_options:
@@ -190,6 +190,11 @@ class EC2(base.Base):
     def ansible_connection_options(self, instance_name):
         try:
             d = self._get_instance_config(instance_name)
+            plat_conn_opts = next(
+                (item for item in self._config.config.get('platforms', [])
+                 if item['name'] == instance_name),
+                {}
+            ).get('connection_options', {})
             conn_opts = util.merge_dicts(
                 {
                     'ansible_user': d['user'],
@@ -199,9 +204,7 @@ class EC2(base.Base):
                     'connection': 'ssh',
                     'ansible_ssh_common_args': ' '.join(self.ssh_connection_options),
                 },
-                # Merge in override 'connection_options' on platform config
-                next((item for item in self._config.config.get('platforms', [])
-                      if item['name'] == instance_name), {}).get('connection_options', {})
+                plat_conn_opts
             )
             if conn_opts.get('ansible_connection') == 'winrm' and (
                     not conn_opts.get('ansible_password')):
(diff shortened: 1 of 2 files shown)


molecule/driver/ec2.py
```diff
@@ -81,6 +98,27 @@ class EC2(base.Base):
...
 
+    Windows EC2 instances can be used as well:
+
+    .. code-block:: yaml
+
+        driver:
+          name: ec2
+        platforms:
+          - name: instance
+            image: ami-06a0d33fc8d328de0)
```
decentral1se:
```suggestion
            image: ami-06a0d33fc8d328de0
```

?

molecule/driver/ec2.py
```diff
@@ -152,15 +190,26 @@ def login_options(self, instance_name):
...
+                {
+                    'ansible_user': d['user'],
+                    'ansible_host': d['address'],
+                    'ansible_port': d['port'],
+                    'ansible_private_key_file': d['identity_file'],
+                    'connection': 'ssh',
+                    'ansible_ssh_common_args': ' '.join(self.ssh_connection_options),
+                },
+                # Merge in override 'connection_options' on platform config
+                next((item for item in self._config.config.get('platforms', [])
```
decentral1se: I'd rather remove the comment and drop the code into a few lines to clarify what is happening here than rely on a comment. Applying `next` without assigning looks unorthodox. Let's not surprise the next one who has to read this.

molecule/driver/ec2.py
```diff
@@ -175,6 +224,20 @@ def _get_instance_config(self, instance_name):
             item for item in instance_config_dict if item['instance'] == instance_name
         )
 
+    def _get_windows_instance_pass(self, instance_id, key_file):
```
decentral1se: Should we be adding `boto3` / `cryptography` as dependencies in the `ec2` extra? I know the ones who don't use Windows instances don't need it though. Perhaps we can document this instead of having the user run into an error like this and then have to install dependencies.

molecule/driver/ec2.py
troyready: copy/paste typo; fixed

molecule/driver/ec2.py
troyready: Good thought; I adjusted it to be a more explicit variable assignment so the flow should be better

molecule/driver/ec2.py
troyready: I added a note about the dependencies in the docstring.

Before that I had started to write up some notes on how they're already _somewhat_ dependencies today. Noting here for posterity & prompting future discussion:

* cryptography is already a a transitive molecule dependency (molecule -> ansible -> cryptography). Since there's no guarantee Ansible won't drop it in the future, the try/catch on import seems like a good compromise/safeguard to make sure no future user gets a stacktrace on it if Ansible changes its deps.

* boto3 (or boto if using legacy modules) are required by the ansible modules that actually create the instances, so it's already an implicit dependency of the ec2 driver. As with the cryptography import, having a try/catch on it should keep the user experience on it predictable (e.g. no early stacktraces if operating in some new half-setup venv)

troyready:
Sure thing - fixed

troyready:
I couldn't agree more about the quotes; I'd be happy to change them back (and have this PR only include relevant changes), but from what I could tell CI was failing because `black` was changing the file

(i.e. it was still failing after the rebase and the only way to get it to exit zero was to run the file through black)

decentral1se:
> I couldn't agree more about the quotes; I'd be happy to change them back (and have this PR only include relevant changes), but from what I could tell CI was failing because black was changing the file

Ah, I see. The Black invocation should be picking up the standard configuration over in https://github.com/ansible/molecule/blob/master/pyproject.toml#L11 which avoids changing quotes. Invoking it through Tox should pick this all up automatically :confused:

PR #2273, opened 2019-09-08
Clean up CODEOWNERS file

decentral1se:
The entries don't make sense anymore because they're not maintaining anymore.

This is probably just annoying for them at this stage.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,20 +1,8 @@
-.github/ @webknjaz @gundalow
-test/conftest.py @webknjaz
+.github/ @gundalow
 
-.gitattributes @webknjaz
-.gitignore @webknjaz
-.git_archival.txt @webknjaz
-.pre-commit-config.yaml @webknjaz @ssbarnea
-.pre-commit-hooks.yaml @webknjaz @ssbarnea
-.readthedocs.yml @webknjaz
+.pre-commit-config.yaml @ssbarnea
+.pre-commit-hooks.yaml @ssbarnea
 .travis.yml @ssbarnea
 
-MANIFEST.in @webknjaz
-Dockerfile @webknjaz
 bindep.txt @ssbarnea
-hooks.yaml @webknjaz
-pytest.ini @webknjaz
-pyproject.toml @webknjaz
-setup.cfg @webknjaz
-setup.py @webknjaz
-tox.ini @webknjaz @ssbarnea
+tox.ini @ssbarnea


ssbarnea:
Your review opens few more important questions related to maintenance.

IMHO I do have expectations like below, which are likely to apply to almost any open-source project:
* a core should be able to perform a review when assigned in less than 30 days, preferably less than a week. If he cannot do it or does not know, he can always himself from assignment.
* core status should not be forever, if you do not do anything with a project for 6mo-1year, I would expect to be "downgraded".
* We need an auto-assignment of reviews to people because most contributors will not know who can help, who is the area expert, who is no longer active,... This can be scripted, or maybe we can even use one of the github steward bots that aim to things like this.
* Molecule is still in "warning" situation when it comes to number of active cores. For example if I leave I have no idea how long it will take for a review to be merged.

Removing core status is a very tricky situation because is high likely to annoy the person being removed. The only positive thing about it is that I see about it is that it can be used as a motivator for "distracted" cores to help with a review or two once in a while.

I kinda diverged on this one as review is only about default reviewers... Here are the problems with new file configuration: i kinda become the only one assigned in the file, how about all the places not covered by a file pattern? I would prefer to see that 2-3 people are automatically assigned to any new review. 

Can we do this? Keep in mind that number of reviews/montg on molecule is quite small and that you can always remove yourself, whatever reason you have (time, knowledge, interest,...)

decentral1se:
> i kinda become the only one assigned in the file, how about all the places not covered by a file pattern? I would prefer to see that 2-3 people are automatically assigned to any new review.

I thought the CODEOWNERS file was just for the case where you definitely want to know if someone is changing a file that you take care of or know is tricky and will need a review. For me, it's like, well, I turn up if I have time and there is no specific file that I should be alerted on. I think I'd rather not be in this file because it is more emails for me.


> a core should be able to perform a review when assigned in less than 30 days, preferably less than a week. If he cannot do it or does not know, he can always himself from assignment.
> 
> core status should not be forever, if you do not do anything with a project for 6mo-1year, I would expect to be "downgraded".

I'd be happy to see these documented.

> Removing core status is a very tricky situation because is high likely to annoy the person being removed. Keep in mind that number of reviews/montg on molecule is quite small and that you can always remove yourself, whatever reason you have (time, knowledge, interest,...)

We can chat about that out-of-band but for what I understand, neither of these points should be the case although thank you for making the point regardless.

Issue #1678, opened 2019-01-18
Migrate to the new linode_v4 module


decentral1se:
# Issue Type
- Feature request

# Issue Description

Since https://github.com/ansible/molecule/pull/1458 is now merged, there are 2 remaining tasks to do.

* Since https://github.com/ansible/ansible/pull/46151 is merged (to arrive in 2.8), we can drop the deprecated linode module being used here. That can't happen until 2.8 is out but should happen before https://github.com/ansible/ansible/issues/51060 happens :) AFAIK, there is already enough functionality in the new module to make this driver work.

* ~~Currently, the functional tests for the linode module are skipped. There is a nice recommendation to use the linode_v4 client fixtures (seen in https://github.com/linode/linode_api4-python/tree/master/test/fixtures) to mock the network connection necessary for running the functional tests. Any test server developed for this work would be of general use to other test drivers who want to mock the network as well.~~
  * Better plan, see: https://github.com/ansible/molecule/issues/1678#issuecomment-493737894.

This work mostly waits on the release of 2.8.

decentral1se:
I've started a discussion on `#ansible-linode` to see if we can get account access for firing up Linode boxes temporarily for the CI testing. Something like we are getting for Hetzner Cloud (see https://github.com/ansible/molecule/pull/1811#issuecomment-487449174).

decentral1se:
FYI, the discussion is underway ... tokens / CI/CD account access is being arranged :rocket:

decentral1se:
I won't be having to follow up on this in the end as I'm passing the torch on the Linode implementation stuff I was doing. Going to close off (please re-open on another ticket if needed, I use my issues as a TODO list).

decentral1se:
OK, I'll try to take a stab at the new module. No functional testing for now but that can come later on when the plugin architecture comes into place. Since we're aiming for a 3.0 release, I won't handle backwards support for the new modules.

PR #2276, opened 2019-09-08
Migrate to new linode_v4 module

decentral1se:
Closes https://github.com/ansible/molecule/issues/1678.

This is a breaking change for 3.0. The old module will not be supported.

cc @gundalow, if I remember correctly, there was some progress on getting Linode sponsorship for CI testing. I have some time again to put into setting this up if the paperwork came through.

(I'm keeping in mind that this will all migrate to a plugin soon but can be done now anyway)


ssbarnea:
@decentral1se keep in mind that there is a "draft" option when you create a WIP review on github. Once you create a non-draft you cannot change it to draft but the other is one click away.

ssbarnea:
Sort conflicts and test it locally. If you confirm I will merge it without having CI for it. We will move the driver out of molecule anyway before doing the 3.0 release so it makes no sense to enable CI here or to delay making changes to the driver.

decentral1se:
OK, I've got this working and tested manually. Let's get it merged then when the CI says :green_heart: 

Functional test suite coming when it is broken out into a plugin :+1:

PR #2280, opened 2019-09-10
Refactor "role:" block in cookiecutter template

### LINKED ISSUE #2279: role template should not use "role:" block
Role template should not use "role:" block because is harder to maintain than the newer task `include_role`.

Note: patches welcomed, this should be a easy to implement first time contribution.

Use UP/DOWN thumbs to vote for this change.

mjlshen:
Signed-off-by: Michael Shen <mishen@med.umich.edu>

#### PR Type

- Bugfix Pull Request

#### Overview

Fixes #2279 
Uses include_role task instead of role block in cookiecutter playbooks


karcaw:
Can I ask why this was done?  is the roles: tag not used anymore?

PR #2281, opened 2019-09-10
refactoring: driver listing

ssbarnea:
- easy access to plugins (drivers or verifiers)
- enables mypy for api methods
- standardize entry point naming `molecule.driver` is prefered to molecule_driver.


molecule/test/unit/test_api.py
```diff
@@ -22,30 +22,20 @@
 from molecule import api
 
 
-def test_api_drivers():
+def test_api_molecule_drivers_as_attributes():
```
decentral1se:
```suggestion
def test_api_drivers_as_attributes():
```

molecule/api.py
```diff
@@ -9,49 +9,47 @@
...
-    "gce",
-    "hetznercloud",
-    "linode",
-    "lxc",
-    "lxd",
-    "openstack",
-    "podman",
-    "vagrant",
-]
+class UserListMap(UserList):
```
decentral1se:
```suggestion
class UserListDict(UserList):
```

Bit more Python centric?

ssbarnea: I will fix it but when I have to touch it next time.

PR #2288, opened 2019-09-13
refactoring: transformed verifiers into plugins

ssbarnea:
This will allow us to add verifiers from outside molecule and to move some of the internal plugins to be external plugins.

### DIFF
--- a/molecule/api.py
+++ b/molecule/api.py
@@ -3,6 +3,7 @@ from importlib import import_module
 from molecule import logger
 from molecule.util import lru_cache
 from molecule.driver.base import Driver  # noqa
+from molecule.verifier.base import Verifier  # noqa
 import traceback
 
 LOG = logger.get_logger(__name__)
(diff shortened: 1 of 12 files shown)


molecule/api.py
```diff
@@ -46,3 +46,22 @@ def drivers(as_dict=False, config=None):
...
+    pm = pluggy.PluginManager("molecule.verifier")
+    try:
+        pm.load_setuptools_entrypoints("molecule.verifier")
+    except Exception:
+        # These are not fatal because a broken verifier should not make the entire
+        # tool unusable.
+        LOG.error("Failed to load verifier entry point %s", traceback.format_exc())
+    for p in pm.get_plugins():
+        try:
+            plugins[pm.get_name(p)] = p(None)
```
decentral1se: `p(None)` looks strange! Is there something we can do about that instead ending up `p()`?

molecule/test/unit/test_config.py
```diff
@@ -473,9 +474,9 @@ def test_molecule_drivers(caplog):
 
 
 def test_molecule_verifiers():
```
decentral1se:
```suggestion
def test_verifiers():
```

ssbarnea: Fixed by extra refactoring.

PR #2289, opened 2019-09-14
Disable deprecation warnings from own dependencies

ssbarnea:
Once we assure we have an upstream bug track we track, we can disable
deprecation warnings from dependencies producing them.

If they do not fix them we should consider finding a way to drop these
dependencies:

* https://github.com/boto/boto3/pull/1603
* https://github.com/boto/botocore/issues/1615
* https://github.com/cookiecutter/cookiecutter/pull/1127 
* https://github.com/pytest-dev/pytest-cov/issues/327
* https://bitbucket.org/astanin/python-tabulate/issues/174/invalid-escape-sequence


ssbarnea:
Also related to https://github.com/astanin/python-tabulate/issues/3

Issue #2181, opened 2019-07-24
presence of futures==3.1.1 can break molecule


ssbarnea:
I discovered that presence of `futures==3.1.1` would break ansible/molecule. Removal fixed the issue. 

I am not yet sure what installed it so I am unsure how we can prevent it, as it not listed as a dependendecy. Still, it could prove useful to share it.
```
$ molecule create                                                                                                                                                         --> Validating schema /Users/ssbarnea/rdo/rdo-infra/testproject/molecule/default/molecule.yml.
Validation completed successfully.
--> Test matrix

└── default
    ├── dependency
    ├── create
    └── prepare

--> Scenario: 'default'
--> Action: 'dependency'
Skipping, missing the requirements file.
--> Scenario: 'default'
--> Action: 'create'
Proceed (y/n)? Traceback (most recent call last):
  File "/Users/ssbarnea/.pyenv/versions/3.6.8/bin/ansible-playbook", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/ssbarnea/os/ansible/bin/ansible-playbook", line 61, in <module>
    import ansible.constants as C
  File "/Users/ssbarnea/os/ansible/lib/ansible/constants.py", line 12, in <module>
    from jinja2 import Template
  File "/Users/ssbarnea/.pyenv/versions/3.6.8/lib/python3.6/site-packages/jinja2/__init__.py", line 82, in <module>
    _patch_async()
  File "/Users/ssbarnea/.pyenv/versions/3.6.8/lib/python3.6/site-packages/jinja2/__init__.py", line 78, in _patch_async
    from jinja2.asyncsupport import patch_all
  File "/Users/ssbarnea/.pyenv/versions/3.6.8/lib/python3.6/site-packages/jinja2/asyncsupport.py", line 13, in <module>
    import asyncio
  File "/Users/ssbarnea/.pyenv/versions/3.6.8/lib/python3.6/asyncio/__init__.py", line 21, in <module>
    from .base_events import *
  File "/Users/ssbarnea/.pyenv/versions/3.6.8/lib/python3.6/asyncio/base_events.py", line 17, in <module>
    import concurrent.futures
  File "/Users/ssbarnea/.pyenv/versions/3.6.8/lib/python3.6/site-packages/concurrent/futures/__init__.py", line 8, in <module>
    from concurrent.futures._base import (FIRST_COMPLETED,
[...]

PR #2291, opened 2019-09-15
Avoid futures SyntaxError

ssbarnea:
Adds run-time check of future version to avoid later failures.
Fixes: #2181

### DIFF
--- a/molecule/shell.py
+++ b/molecule/shell.py
@@ -23,6 +23,8 @@ import os
 import click
 import click_completion
 import colorama
+import pkg_resources
+import sys
 
 import molecule
 from molecule import command
@@ -80,6 +82,16 @@ def main(ctx, debug, base_config, env_file):  # pragma: no cover
     ctx.obj['args']['base_config'] = base_config
     ctx.obj['args']['env_file'] = env_file
 
+
+# runtime environment checks to avoid delayed failures
+if sys.version_info[0] > 2:
+    try:
+        if pkg_resources.get_distribution('futures'):
+            raise SystemExit(
+                "FATAL: futures package found, this package should not be installed in a Python 3 environment, please remove it. See https://github.com/agronholm/pythonfutures/issues/90"
+            )
+    except pkg_resources.DistributionNotFound:
+        pass
 
 main.add_command(command.cleanup.cleanup)
 main.add_command(command.check.check)


molecule/command/init/role.py
```diff
@@ -106,7 +105,7 @@ def execute(self):
 @click.option(
     '--driver-name',
     '-d',
-    type=click.Choice(drivers()),
+    type=click.Choice(api.drivers()),
```
decentral1se: You need to take care here and also use the solution in https://github.com/ansible/molecule/pull/2295 for completion?

molecule/api.py
```diff
@@ -21,6 +21,9 @@ def __getitem__(self, i):
         else:
```
decentral1se: Still has https://github.com/ansible/molecule/pull/2291/commits/e7e4161c5b741b7c491481ef56b326156a8d1f70 from another PR?

molecule/shell.py
```diff
@@ -81,6 +83,16 @@ def main(ctx, debug, base_config, env_file):  # pragma: no cover
     ctx.obj['args']['env_file'] = env_file
 
 
+# runtime environment checks to avoid delayed failures
+if sys.version_info[0] > 2:
+    try:
+        if pkg_resources.get_distribution('futures'):
+            raise SystemExit(
+                "FATAL: futures package found, remove it. See https://github.com/agronholm/pythonfutures/issues/90"
```
decentral1se:
```suggestion
                "FATAL: futures package found, this package should be installed in a Python 3 environment, please remove it. See https://github.com/agronholm/pythonfutures/issues/90"
```

molecule/shell.py
```diff
@@ -81,6 +83,16 @@ def main(ctx, debug, base_config, env_file):  # pragma: no cover
     ctx.obj['args']['env_file'] = env_file
 
 
+# runtime environment checks to avoid delayed failures
+if sys.version_info[0] > 2:
+    try:
+        if pkg_resources.get_distribution('futures'):
+            raise SystemExit(
+                "FATAL: futures package found, this package should be installed in a Python 3 environment, please remove it. See https://github.com/agronholm/pythonfutures/issues/90"
```
decentral1se:
```suggestion
                "FATAL: futures package found, this package should not be installed in a Python 3 environment, please remove it. See https://github.com/agronholm/pythonfutures/issues/90"
```

PR #2292, opened 2019-09-15
refactoring: verifiers should expose their own linters

ssbarnea:
Instead of knowing which linter is used by each verifier, this is now
a property of the verifier itself.

### DIFF
--- a/molecule/verifier/base.py
+++ b/molecule/verifier/base.py
@@ -41,6 +41,7 @@ class Verifier(object):
         :returns: None
         """
         self._config = config
+        self.default_linter = 'ansible-lint'
 
     @abc.abstractproperty
     def name(self):  # pragma: no cover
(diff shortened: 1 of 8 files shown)


molecule/verifier/base.py
```diff
@@ -41,6 +41,7 @@ def __init__(self, config=None):
         :returns: None
         """
         self._config = config
+        self.default_linter = 'ansible-lint'
```
decentral1se: I would leave it as "self.linter" in case the default is changed at some point.

Also, we need to update the `lint()` function on each of these objects?

https://github.com/ansible/molecule/blob/8e2c20797c7d0dadda4a85ef2c8e2fb83a63f321/molecule/verifier/base.py#L107

It is taking the linter from the config but it now lives on the verifier itself?

ssbarnea: That only defining the default (implicit) linter of a verifier, this is not necessarily the one that will be used.

PR #2295, opened 2019-09-15
Fix auto-completion of choices

ssbarnea:
Avoids https://github.com/pallets/click/issues/591 by converting choices
to lists of strings.

### DIFF
--- a/molecule/command/create.py
+++ b/molecule/command/create.py
@@ -105,7 +105,7 @@ class Create(base.Base):
 @click.option(
     '--driver-name',
     '-d',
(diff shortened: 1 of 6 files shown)


molecule/command/create.py
```diff
@@ -105,7 +105,7 @@ def execute(self):
 @click.option(
     '--driver-name',
     '-d',
-    type=click.Choice(drivers()),
+    type=click.Choice([str(s) for s in drivers()]),
```
decentral1se: Almost feel like we should reference the issue here in a comment but I leave it to you.

PR #2298, opened 2019-09-15
verifier: goss removed

ssbarnea:
Removes goss from core as this will be provided by external package molecule-goss.


molecule/verifier/lint/yamllint.py
```diff
@@ -52,7 +52,7 @@ class Yamllint(base.Base):
     .. code-block:: yaml
 
         verifier:
-          name: goss
+          name: ...
```
decentral1se: We should really leave a TODO or something or else we will forget them ...

PR #2300, opened 2019-09-17
refactoring: plugins can report their template dir

ssbarnea:
Fix problem where location of templates was hardcoded.

### DIFF
--- a/molecule/command/init/scenario.py
+++ b/molecule/command/init/scenario.py
@@ -82,7 +82,9 @@ class Scenario(base.Base):
             ).format(scenario_name)
             util.sysexit_with_message(msg)
 
-        driver_template = 'scenario/driver/{driver_name}'.format(**self._command_args)
+        driver_template = api.drivers()[
+            self._command_args['driver_name']
+        ].template_dir()
         if 'driver_template' in self._command_args:
             self._validate_template_dir(self._command_args['driver_template'])
             cli_driver_template = '{driver_template}/{driver_name}'.format(
@@ -98,7 +100,7 @@ class Scenario(base.Base):
         scenario_base_directory = os.path.join(role_directory, role_name)
         templates = [
             driver_template,
-            'scenario/verifier/{verifier_name}'.format(**self._command_args),
+            api.verifiers()[self._command_args['verifier_name']].template_dir(),
         ]
         for template in templates:
             self._process_templates(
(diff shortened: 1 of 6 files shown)


molecule/command/init/scenario.py
```diff
@@ -82,7 +82,9 @@ def execute(self):
             ).format(scenario_name)
             util.sysexit_with_message(msg)
 
-        driver_template = 'scenario/driver/{driver_name}'.format(**self._command_args)
+        driver_template = api.drivers()[
+            self._command_args['driver_name']
```
decentral1se: Uhhhh, this is always taking the driver from the CLI invocation and not from a configuration like `molecule.yml`?

molecule/test/unit/command/init/test_scenario.py
```diff
@@ -105,16 +105,9 @@ def test_execute_with_invalid_driver(
 ):
     _command_args['driver_name'] = 'ec3'
 
-    with pytest.raises(SystemExit) as e:
+    with pytest.raises(KeyError):
         _instance.execute()
 
-    assert 1 == e.value.code
```
decentral1se: We should still be providing something to the user in the case of a `KeyError` instead of a stack trace!?

ssbarnea: @decentral1se You would be surprised but in this case the stack trace is more useful than the previous error message which was totally misleading: referring to missing folder, not inexistent driver.

Try it, that one is very easy to test from cli, before and after: `molecule init role --driver-name foo`.

Usually I am for clear error messages, but a nice error message here would require a wide range of changes. Should be done eventually but not part of this change.

molecule/command/init/scenario.py
ssbarnea: They are not affected, there is a list of template directories, all loaded.

PS. Yes, I am really ware that the construct is very ugly but I did not want to improve code while fixing bugs. We can work to make the code more pythonic later.

molecule/test/unit/command/init/test_scenario.py
decentral1se: OK, sure. However, there is now an increasing list of "things for the future". Please start to mark them in the code! Add a TODO, XXX, whatever or create an issue. If you forget, nobody will remember to do these things until the users have to run into them.

ssbarnea: There where already there, the difference is that now we uncover them.

decentral1se: I see no TODO in this change or added by you. So no, they are not there.

PR #2327, opened 2019-09-24
Fix version parsing error with Ansible prereleases

ssbarnea:
Avoids failing to run run with prereleases by using a PEP440 compliant
version parser.

### DIFF
--- a/molecule/test/functional/conftest.py
+++ b/molecule/test/functional/conftest.py
@@ -21,6 +21,7 @@
 
 import distutils.spawn
 import os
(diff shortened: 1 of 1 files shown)


ssbarnea:
@decentral1se Yeah, master is broken, see https://github.com/ansible/molecule/pull/2326 which failed due to this.

perlun:
(For reference, here is what it looked like in Travis before this change: https://travis-ci.com/ansible/molecule/builds/128917107)

PR #2328, opened 2019-09-25
Dockerfile.j2: add python-apt aptitude packages for Debian-based distros

perlun:
This aims to get rid of warnings like this when installing packages later on, using the [`package`](https://docs.ansible.com/ansible/latest/modules/package_module.html#package-module) ansible module:

```
TASK [ubuntu_tomcat : Install tomcat .deb package] *****************************
[WARNING]: Updating cache and auto-installing missing dependency: python-apt

changed: [ubuntu16.04-8d75d129-86ac-4715-89df-2b4a04733675]
[WARNING]: Could not find aptitude. Using apt-get instead
```

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/molecule/test/resources/playbooks/docker/Dockerfile.j2
+++ b/molecule/test/resources/playbooks/docker/Dockerfile.j2
@@ -9,7 +9,7 @@ FROM {{ item.image }}
 ARG testarg=fail
 ENV envarg=$testarg
 
-RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get install -y python sudo bash ca-certificates iproute2 && apt-get clean; \
+RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get install -y python sudo bash ca-certificates iproute2 python-apt aptitude && apt-get clean; \
     elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install python sudo python-devel python*-dnf bash iproute && dnf clean all; \
     elif [ $(command -v yum) ]; then yum makecache fast && yum install -y python sudo yum-plugin-ovl bash iproute && sed -i 's/plugins=0/plugins=1/g' /etc/yum.conf && yum clean all; \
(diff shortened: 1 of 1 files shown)


MarkusTeufelberger:
This is by design as it is NOT something that is installed by default and should be handled by the role under test instead of being assumed to be installed.

MarkusTeufelberger:
There are tons of dependencies for commonly used modules that are not pre-installed. I don't see why `apt` and `package` should be special cases...

decentral1se:
> This aims to get rid of warnings

MarkusTeufelberger:
The warning is there for a reason and this PR only makes roles tested with containers built from this template less correct.

decentral1se:
> The warning is there for a reason

And these packages were added for a reason.

MarkusTeufelberger:
They belong into a `prepare.yml` file. These warnings will re-appear on actual hosts but not in tests...

decentral1se:
Moleculs is for tests.

MarkusTeufelberger:
And tests are for making sure that things work as expected in production. This warning will show up there. Automatically removing it for every(!) user of molecule by default is making these tests somewhat less useful.

Stuff like this is a prime candidate for a `prepare.yml` playbook. There could be one added by default for the Docker driver too.

ssbarnea:
We should be very careful in our approach of making molecule smart, too smart and it may prevent people from testing correctly. Still, in this case that change is only about molecule tests and we should aim for zero warnings on them. When it comes to testing user roles, the answer is more tricky.

MarkusTeufelberger:
Ah, mea culpa and sorry for the confusion. It wasn't clear to me that this is only changing the template for the molecule tests, not the template that gets rendered for the docker driver (I assumed the tests would test the template that gets rendered for users). The initial description with a `tomcat` installation was also something that I didn't think would apply to a change only for molecule internal tests.

Hint: This would be fixed by implementing #1642

PR #2329, opened 2019-09-25
Removed LXC and LXD providers

ssbarnea:
Based on popularity survey which was also popularized on the mainling
list, these providers are not really used. If there is enough interest
for having them it it is possible to implement them as external plugins.

Reference: https://github.com/pycontribs/pytest-molecule/issues/41


cloudnull:
ping @mnaser @jrosser - is this something that you may use?

snesbittsea:
I use the LXC/LXD driver extensively. I don't use Vagrant because the slow test cycle it imposes and because the docs indicate that the integration with Molecule is pretty much a hack. 

Docker containers don't strike me as a great alternative, especially when working with networking and kernel configuration.

And it strikes me as odd that the Ansible testing framework does not support an Ansible connector.

What would be involved in migrating to the new plugin architecture? Are there any docs? Someone who could point me in the right direction, review my work and answer questions?

decentral1se:
> What would be involved in migrating to the new plugin architecture? Are there any docs? Someone who could point me in the right direction, review my work and answer questions?

Docs are coming. In the meantime you can take a look at the existing plugins (mostly WIP but enough to understand how to do it) over at https://github.com/pycontribs?utf8=%E2%9C%93&q=molecule&type=&language=. We're on IRC on can advise here and would definitely appreciate if you could help this effort.

n0guest:
> I use the LXC/LXD driver extensively. I don't use Vagrant because the slow test cycle it imposes and because the docs indicate that the integration with Molecule is pretty much a hack.

Same goes for us. We're mostly using LXD drivers and less others.

> As it reduces our work load now, that is good. Anyone can easily follow our lead and implement these as a plugin driver and we will support that process. The survey has shown us (however effective it is, that's all we have right now) that there is no usage right now. OTOH, if you see this PR and you need the drivers, please shout now.

I understand, that I'm writing a month late. But it's really an unexpected turn. I'm only hoping that docs are coming soon enough. And we or someone else will write a driver for LXC/LXD as an external plugin.

decentral1se:
Sorry to shock you but nothing to worry about. It is not removed but moving into a plugin. Documentation is my main priority at the moment, so I will come back and chase you! :)

snesbittsea:
Fair enough! In lieu of complete docs, what I think would be helpful to know is:

What services the plugin must provide. 

In general terms, what is the conversion process from the old architecture to the new. To what extent can I "simply" transate the old to the new.

Any obvious gotchas I need to look for.


-steve



---- On Sun, 27 Oct 2019 11:10:13 -0700 decentral1se <notifications@github.com> wrote ----


Sorry to shock you but nothing to worry about. It is not removed but moving into a plugin. Documentation is my main priority at the moment, so I will come back and chase you! :)

n0guest:
> Sorry to shock you but nothing to worry about. It is not removed but moving into a plugin. Documentation is my main priority at the moment, so I will come back and chase you! :)

Oooh ... It's very good news. Thank you for this clarification!

decentral1se:
@snesbittsea, it should behave *exactly* as before, meaning, seamlessly moving to plugin, so no functionality changes at all, and simply live in another repository. I pasted a link to all the plugins we have so far above ^^^, just copy/pasta one and see how it goes!

ssbarnea:
@n0guest Which one LXC or LXD? I am afraid that these were at the bottom of the priority list when it comes to plugins. If they are so important for you maybe you can help with the code migration? 

The best place to start would to look at https://github.com/pycontribs/molecule-azure which is the external plugin which is in the most advanced form at this moment. Some of the others do not even have original code added to their own repositories. If you try to create a new one you could fork it and just replace the plugin code.

n0guest:
> @n0guest Which one LXC or LXD? I am afraid that these were at the bottom of the priority list when it comes to plugins. If they are so important for you maybe you can help with the code migration?

For now we're using LXD driver. Just because "bare" LXC containers are less convenient in comparison with the same containers under LXD management.

> The best place to start would to look at https://github.com/pycontribs/molecule-azure which is the external plugin which is in the most advanced form at this moment. Some of the others do not even have original code added to their own repositories. If you try to create a new one you could fork it and just replace the plugin code.

Yep, I looked quickly at existing repos yesterday. But writing a plugin could be kind a slow in my case (too much other work issues now and not so skilled at Python). But I definitely will try, just because we need it at working state 8)

hswong3i:
@ssbarnea  Could we migrate the legacy LXD drive into https://github.com/ansible-community/molecule-lxd, while remove the support from molecule core on next 3.0 stable release?

In my use case all of my Ansible roles (70+) are default testing with Travis CI + LXD driver + Ubuntu/CentOS/openSUSE (https://github.com/alvistack?q=ansible-role), which simplify lots of works related to systemd on Docker.

ssbarnea:
@hswong3i Can you become its maintainer if I do this? It may take me half a day of work but I will do it if I know there is someone to keep it green. My personal impression is that libvirt driver will be a good replacement in long run.

hswong3i:
@ssbarnea I could give a try for that in Python but I am now a daily Python developer...

On the other hand, my main concern is: which molecule driver could be running on, e.g. Travis CI? Since most public CI don't support VM in VM mode (e.g. VirtualBox, KVM, Xen, etc), therefore Docker and LXD could be the only choice.

Moreover, when testing Ansible role with systemd related, yes we could **hack** that for running with Docker, but generally it works fine with LXD without dirty workaround ;-)

PR #2342, opened 2019-09-30
Use ansible-galaxy to create roles

ssbarnea:
Removed internal cookiecutter template used to create the role and use ansible-galaxy to perform the same task.


jokajak:
Nitpicky but your commit message has a typo in it.

Based on the travis CI it looks like you'll need to update the tests to align with the ansible-galaxy README.md instead of the molecule README.md.

molecule/command/init/role.py
```diff
@@ -72,12 +75,10 @@ def execute(self):
             )
             util.sysexit_with_message(msg)
 
-        template_directory = ''
-        if 'template' in self._command_args.keys():
-            template_directory = self._command_args['template']
-        else:
-            template_directory = 'role'
-        self._process_templates(template_directory, self._command_args, role_directory)
+        r = subprocess.run(["ansible-galaxy", "init", role_name], shell=False)
```
decentral1se: Shouldn't we use `check_output` or? The error message below will not be enough to help users when it fails.

ssbarnea: run() does not hide output, so there is no need to capture and display what was already logged.

decentral1se: In a stack trace? We should handle that with an exception and pass to our logger?

decentral1se:
Fix for failing CI is @ https://github.com/ansible/molecule/issues/2349#issuecomment-538970935. It needs some eyes to decide on a correct solution. Take a look if you have time.

molecule/command/init/role.py
```diff
@@ -43,10 +48,8 @@ class Role(base.Base):
 
     .. option:: molecule init role --role-name foo --template path
 
-        Initialize a new role using a local *cookiecutter* template. This
-        allows the customization of a role while still using the upstream
-        ``molecule`` folder. This is similar to an
-        ``ansible-galaxy init`` skeleton. Please refer to the ``init scenario``
+        Initialize a new role using ansible-galaxy and adds molecule template
+        to it. Please refer to the ``init scenario``
```
decentral1se: "and adds molecule template" -> "and include the default molecule directory"

molecule/command/init/role.py
```diff
@@ -72,12 +75,16 @@ def execute(self):
...
             util.sysexit_with_message(msg)
 
-        template_directory = ''
-        if 'template' in self._command_args.keys():
-            template_directory = self._command_args['template']
-        else:
-            template_directory = 'role'
-        self._process_templates(template_directory, self._command_args, role_directory)
+        try:
+            cmd = ["ansible-galaxy", "init", "-v", "--offline", role_name]
```
decentral1se: * Why use `-v` without checking if the user has passed `--debug`?
* We should catch a [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) instead?
* Should we also `--ignore-certs` to avoid annoying failures?

molecule/cookiecutter/molecule/{{cookiecutter.role_name}}/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/molecule.yml
```diff
@@ -19,6 +19,7 @@ provisioner:
   name: {{ cookiecutter.provisioner_name }}
   lint:
     name: {{ cookiecutter.provisioner_lint_name }}
+    enabled: false
```
decentral1se: ? Also for https://github.com/ansible/ansible/issues/63734?

molecule/test/functional/docker/test_scenarios.py
```diff
@@ -218,8 +218,9 @@ def test_command_test_overrides_driver(
     indirect=['scenario_to_test', 'driver_name', 'scenario_name'],
 )
 def test_command_test_builds_local_molecule_image(
-    scenario_to_test, with_scenario, scenario_name, driver_name
+    scenario_to_test, with_scenario, scenario_name, driver_name, environ
 ):
+    # with mock.patch.dict('os.environ', environ, clear=True):
```
decentral1se: To be removed?

gundalow:
Testing GitHub Actions

molecule/test/functional/docker/test_scenarios.py
ssbarnea: done

molecule/cookiecutter/molecule/{{cookiecutter.role_name}}/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/molecule.yml
ssbarnea: Shroud we include a comment with reasoning? Maybe the team will reconsider that decision  under peer-pressure? Anyway we are going to rewrite the linter logic anyway and probably deprecate these options in near future.

molecule/command/init/role.py
ssbarnea: Even verbose mode is not very verbose on Galaxy. Lets keep it like this because we do not have access at config option. We can make it configurable when we have a proper CfgMgr.

I see no reason to touch --ignore-certs option, I am not keen to hardcode insecure defaults. Better to see a bug raised and we can see what we can do, even if I consider that such option should be controllable using environment or ansible.cfg instead of cli.

molecule/cookiecutter/molecule/{{cookiecutter.role_name}}/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/molecule.yml
decentral1se: Comment would be good.

Issue #2349, opened 2019-10-07
Broken CI: "No match for argument: python"


decentral1se:
https://travis-ci.com/ansible/molecule/jobs/241187632#L1819

> ... 'No match for argument: python\\n', 'There are following alternatives for \"python\": python2, python36\\n', 'No match for argument: python-devel\\n' ...

decentral1se:
Ran into https://github.com/ansible/molecule/issues/2350 while trying to resolve this :(

decentral1se:
Ok, the issue can be resolved with:

```diff
diff --git a/molecule/test/resources/playbooks/docker/Dockerfile.j2 b/molecule/test/resources/playbooks/docker/Dockerfile.j2
index 65279a5e..df4f70d5 100644
--- a/molecule/test/resources/playbooks/docker/Dockerfile.j2
+++ b/molecule/test/resources/playbooks/docker/Dockerfile.j2
@@ -10,7 +10,7 @@ ARG testarg=fail
 ENV envarg=$testarg
 
 RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get install -y python sudo bash ca-certificates iproute2 python-apt aptitude && apt-get clean; \
-    elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install python sudo python-devel python*-dnf bash iproute && dnf clean all; \
+    elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install python2 python36 sudo python2-devel python36-devel python*-dnf bash iproute && dnf clean all; \
     elif [ $(command -v yum) ]; then yum makecache fast && yum install -y python sudo yum-plugin-ovl bash iproute && sed -i 's/plugins=0/plugins=1/g' /etc/yum.conf && yum clean all; \
     elif [ $(command -v zypper) ]; then zypper refresh && zypper install -y python sudo bash python-xml iproute2 && zypper clean -a; \
     elif [ $(command -v apk) ]; then apk update && apk add --no-cache python sudo bash ca-certificates; \
```

Because in testing, we test against the `centos:latest` image, and this new image only provides python2/python36 packages now. If we make this change in the standard cookiecutter Dockerfile.j2, then we are not supporting the latest centos image and installations will break. How to support backwards compatability in this case? 

This may finally force our hand to come up with a more sustainable position on how to deal with the Dockerfile.j2.

decentral1se:
[...]

PR #2351, opened 2019-10-07
Temporarily limit setuptools version

decentral1se:
See https://github.com/ansible/molecule/issues/2350.

In the absence of any answers right now, let's keep the ball rolling with this temporary  fix.

Issue #2350, opened 2019-10-07
CI: 'SpecifierSet' object has no attribute 'split'


decentral1se:
Looks like some sort of pip/setuptools issue related to specifiers (>,<,>=,...)

> (.venv) ➜  molecule (master) ✔ pip show setuptools | grep -i version
> Version: 41.4.0
> (.venv) ➜  molecule (master) ✔ pip show pip | grep -i version 
> Version: 19.2.3

```
(.venv) ➜  molecule (master) ✔ pip install -e .      
Obtaining file:///home/decentral1se/hacking/molecule
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... error
    ERROR: Command errored out with exit status 1:
     command: /home/decentral1se/hacking/molecule/.venv/bin/python /home/decentral1se/hacking/molecule/.venv/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel /tmp/tmp1cijsh20
         cwd: /home/decentral1se/hacking/molecule
    Complete output (28 lines):
    Traceback (most recent call last):
      File "/home/decentral1se/hacking/molecule/.venv/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 207, in <module>
        main()
      File "/home/decentral1se/hacking/molecule/.venv/lib/python3.7/site-packages/pip/(.venv) ➜  molecule (master) ✔ pip install -e .                    
Obtaining file:///home/decentral1se/hacking/molecule
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... error
    ERROR: Command errored out with exit status 1:
     command: /home/decentral1se/hacking/molecule/.venv/bin/python /home/decentral1se/hacking/molecule/.venv/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel /tmp/tmp98lvx227
         cwd: /home/decentral1se/hacking/molecule
    Complete output (28 lines):
    Traceback (most recent call last):
      File "/home/decentral1se/hacking/molecule/.venv/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 207, in <module>
        main()
      File "/home/decentral1se/hacking/molecule/.venv/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 197, in main
[...]

PR #2354, opened 2019-10-07
Focus on CentOS 7 for CI testing

decentral1se:
Closes https://github.com/ansible/molecule/issues/2349.

Of course, we don't resolve the issue that anyone using CentOS 8 docker images will now have their builds broken by Molecule because it cannot match the packages that are pre-installed but we leave that until the bug report rolls in (or someone gets to it). 

This fixes the CI now as we will simply support CentOS 7 images (`:7` tag) in testing.

We can control this from the environment now with `TEST_CENTOS_VERSION`.

### DIFF
--- a/molecule/test/functional/docker/test_scenarios.py
+++ b/molecule/test/functional/docker/test_scenarios.py
@@ -221,7 +221,10 @@ def test_command_test_builds_local_molecule_image(
     scenario_to_test, with_scenario, scenario_name, driver_name
 ):
     try:
-        cmd = sh.docker.bake('rmi', 'molecule_local/centos:latest', '--force')
+        version = os.environ.get('TEST_CENTOS_VERSION', 7)
+        cmd = sh.docker.bake(
+            'rmi', 'molecule_local/centos:{}'.format(version), '--force'
+        )
         pytest.helpers.run_command(cmd)
     except sh.ErrorReturnCode:
         pass
--- a/molecule/test/resources/molecule_docker.yml
+++ b/molecule/test/resources/molecule_docker.yml
@@ -7,7 +7,7 @@ lint:
   name: yamllint
 platforms:
   - name: instance
-    image: centos:latest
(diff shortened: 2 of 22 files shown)


decentral1se:
CI failure is now related to https://travis-ci.com/ansible/molecule/jobs/242987521#L276 which is solved in https://github.com/ansible/molecule/pull/2351 (also see https://github.com/ansible/molecule/issues/2350). Ugh :/

decentral1se:
https://developers.redhat.com/blog/2019/05/07/what-no-python-in-red-hat-enterprise-linux-8/ 

:scream_cat: 

wtf.

molecule/test/resources/playbooks/docker/Dockerfile.j2
```diff
@@ -10,7 +10,7 @@ ARG testarg=fail
 ENV envarg=$testarg
 
 RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get install -y python sudo bash ca-certificates iproute2 python-apt aptitude && apt-get clean; \
-    elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install python sudo python-devel python*-dnf bash iproute && dnf clean all; \
+    elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install python2 python36 sudo python2-devel python36-devel python*-dnf bash iproute && dnf clean all; \
```
ssbarnea: While I can understand the need workaround and force use of centos 7 while we add support for the new one, I do not think we should touch this file in the same change, especially as installing multiple pythons is not to be desired.

decentral1se: Ah, that was supposed to be removed (before, I was trying to support CentOS 8 and it has no `python` package, only `python2` and `python36` etc.).

decentral1se:
Fair enough, I'll try to control from the environment.

decentral1se:
CI is :green_heart: 

image

PR #2357, opened 2019-10-09
Support using ~/path/to/foo ansible environment

pabelanger:
Allow for the ability to use ~ in ansible environment variables. For example:

  provisioner:
    name: ansible
    env:
      ANSIBLE_ROLES_PATH: ~/.ansible/roles

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/molecule/provisioner/ansible.py
+++ b/molecule/provisioner/ansible.py
@@ -407,6 +407,10 @@ class Ansible(base.Base):
 
     @property
     def default_env(self):
+        # Allow user to use ~/path/to/foo in environment variables
+        for key, value in self._config.env.items():
+            self._config.env[key] = os.path.expanduser(value)
+
         env = util.merge_dicts(os.environ.copy(), self._config.env)
         env = util.merge_dicts(
             env,


molecule/provisioner/ansible.py
```diff
@@ -407,6 +407,10 @@ def default_options(self):
 
     @property
     def default_env(self):
+        # Allow user to use ~/path/to/foo in environment variables
+        for key, value in self._config.env.items():
+            self._config.env[key] = os.path.expanduser(value)
+
```
ssbarnea: This worries me a little bit as it would break if genuine `~` is present in other environment variables, ones that do not need/want to be expanded. I am not aware of any existing such cases but I can imagine some could.

Another issue that I see even more problematic is that we have ~20 implementations of `default_env()` across the code base and that is addressing only in one place.

I do not see any reason why this expansion should happen only for provisioner/ansible and not for others. While I can ignore lack of tests for this one I am afraid that the second one is more serious.... also a reason to refactor the code to avoid that duplication and have a single implementation.

decentral1se: > This worries me a little bit as it would break if genuine ~ is present in other environment variables, ones that do not need/want to be expanded.

Which vars do you have that have a `~` in them? We could just add a global `--no-expand` if needed. We should test around this to satisfy your concern.


> Another issue that I see even more problematic is that we have ~20 implementations of default_env() across the code base and that is addressing only in one place.

> I do not see any reason why this expansion should happen only for provisioner/ansible and not for others. While I can ignore lack of tests for this one I am afraid that the second one is more serious.... also a reason to refactor the code to avoid that duplication and have a single implementation.

Those are problems but let's not push this on a contributor who is trying to support their case and not deal with all the things wrong with molecule and variable expansion.

molecule/provisioner/ansible.py
```diff
@@ -407,6 +407,10 @@ def default_options(self):
 
     @property
     def default_env(self):
+        # Allow user to use ~/path/to/foo in environment variables
+        for key, value in self._config.env.items():
```
ssbarnea: This will fail ugly with variables that contain list of paths, like PATH itself. Expand is not able to deal with path lists.

fabianvf: I guess you could do something like split on `:` and apply it to each item, like

```python
self._config.env[key] = ':'.join([os.path.expanduser(entry) for entry in value.split(':')])
```

It would cause an issue if there was an environment variable that contained the pattern `:~`, but wasn't representing a directory, though I'm not sure how likely that is.

ssbarnea: I find it risky and sooner or later someone would hate us is we do it. On the other hand, lets go for a safe approach where we expand only this variable.

Can you please update the review to do this?

ssbarnea:
Closing due to lack of feedback.

PR #2358, opened 2019-10-09
Podman driver shouldn't define /sys/fs/cgroup vol

maxamillion:
Since podman version 0.8.5 it has auto detected when the mounting of
/sys/fs/cgroup is necessary and only mounts a subset of that dir as
needed by the container runtime. In newer versions, explicitly
defining this actually fails with an OCI runtime permission denied
error. Therefore, not only is this not necessary but can cause
failures for users.

Signed-off-by: Adam Miller <admiller@redhat.com>

#### PR Type

- Docs Pull Request

### DIFF
--- a/molecule/driver/podman.py
+++ b/molecule/driver/podman.py
@@ -110,8 +110,6 @@ class Podman(Driver):
         - name: instance
           image: centos:7
           privileged: true
-          volumes:
-            - "/sys/fs/cgroup:/sys/fs/cgroup:rw"
           command: "/usr/sbin/init"
           tty: True
 


maxamillion:
Travis CI failures look unrelated since the failure is in docker functional tests and this is a docs change on something not docker.

decentral1se:
Thanks! RE: CI, we need to merge #2354 first.

perlun:
@maxamillion Thank you for this! :+1: 

@decentral1se This is exactly the problem I mentioned on IRC the other day. Would there be a chance to get this removed from the (published) 2.22 docs at https://molecule.readthedocs.io/en/stable/configuration.html#podman as well? Right now, the docs tend to lure the reader into a configuration that is very likely to fail (on newer podman versions), so it would be great to get this sorted out.

PR #2360, opened 2019-10-11
Make default Dockerfile embedded

ssbarnea:
From now on, creating a Dockerfile for docker/podman containers would
be optional. If the file is not present molecule would use its internal
template.

This ease maintenance of roles using molecule by having one less file
to create and update.

Those relying on custom containers are not affected, as this feature is
still present.

Newly created roles will not have Dockerfiles.


molecule/provisioner/ansible/playbooks/docker/create.yml
```diff
@@ -22,11 +22,27 @@
...
 
+    - name: Check presence of custom Dockerfiles
+      stat:
+        path: "{{ molecule_scenario_directory + '/' + (item.dockerfile | default( 'Dockerfile.j2')) }}"
+      loop: "{{ molecule_yml.platforms }}"
+      register: dockerfile_stats
+
     - name: Create Dockerfiles from image names
+      vars:
+        # TODO(ssbarnea): expose module dir so it can also be used by plugins
```
decentral1se: What about feeding it in from:

https://github.com/ansible/molecule/blob/97a7b60fc47b2b83f3f772aafbf044294be168a6/molecule/provisioner/ansible.py#L593-L601

Does it work?

PR #2362, opened 2019-10-11
Correct python package installation on Dockerfiles

ssbarnea:
This should fix failures with CentOS 8.


decentral1se:
Restarted the failed build (unrelated).

Wanna test it by toggling `TEST_CENTOS_VERSION`?

We could just use `=8` if we can support it like this.

ssbarnea:
> Restarted the failed build (unrelated).
> 
> Wanna test it by toggling `TEST_CENTOS_VERSION` (see `tox.ini`)?
> 
> We could just use `=8` if we can support it like this.

I tested manually. I plan to add a special job that tests the Dockerfile on multiple platforms but it would be outside the scope of this fix.

PR #2367, opened 2019-10-15
Enable release drafter

ssbarnea:
This should automate create of release notes, so our users will be better informed of what went in on each new release.

This github workflow keeps a draft release updated with all merged
changes so when we want to make a new release we already have the full
changelist created.

I initially adopted this workflow on few other projects and I am really
pleased about the results:

* https://github.com/pycontribs/pytest-molecule/releases
* https://github.com/pycontribs/selinux/releases
* https://github.com/pycontribs/jira/releases

#### PR Type

- Feature Pull Request

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -1,0 +1,16 @@
+name: Release Management
+
+on:
+  push:
+    # branches to consider in the event; optional, defaults to all
+    branches:
+      - master
+
+jobs:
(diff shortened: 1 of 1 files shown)


decentral1se:
It's kinda like Github vendor lock in for release change logs.

Can it also produce flat text files too? That'd be nice to also have them too.

But anyway, looks like Github is _the_ place for Molecule.

PR #2371, opened 2019-10-15
Test ansible 2.9 from stable-2.9

ssbarnea:
Temporary use stable-29 branch to test ansible 2.9

Avoids being blocked by currently broken pre-releases. We will move to
standard ranges once a release is made.

This unblocks our CI.


ssbarnea:
Clearly we need to change the way we test because few minutes ago I made this discovery https://github.com/ansible/ansible/issues/63618 -- I guess that 2.9 release may be delayed even more.

PR #2374, opened 2019-10-16
Reconfigured requirements to allow offline testing

ssbarnea:
Avoid using galaxy servers to install test role because this sends us
to github which quite often is refusing to server the files (DDOS protection kicking in).

Ideally we should be able to run all the tests offline.
Caused by https://github.com/ansible/galaxy/issues/2030


ssbarnea:
Closing for the moment but I may resurrect it if needed.

PR #2379, opened 2019-10-18
Remove setup.py hacks

ssbarnea:
This should address the failure to install which seems to still
appears from time to time even with our `pyproject.toml` workaround.

Fixes: #2350


ssbarnea:
The hacks are broken and I do not have the energy to fix it (if even possible). AFAIK is far more important to assure that molecule installs successfully on all supported systems.

I will add installation tests to prevent regressions on packaging. I already did that on pytest-molecule where package installation is tested on multiple systems. Sadly this takes time because it needs to build containers with system deps for these systems.

decentral1se:
I'm not asking you to setup a future proof test suite just be clearer on the changes that are being made. I know we're low on capacity but it would be good to have some opinions from packaging people on this (that should be enough).

PR #2381, opened 2019-10-19
Fix packaging

bristea:
Removes hacks from setup.py and adds extra tests that should
prevent future regressions.

The previous workaround was broken as tox was unable to run twice
on the same environment. The assumption that newer setuptools
was preventing installation only during install was wrong, the
reality was that newer version broke packaging completely.

Fixes: #2379

Signed-off-by: Bogdan Ristea <Bogdan Ristea bogdan.ristea@gmail.com>

#### PR Type

- Bugfix Pull Request


pyproject.toml
```diff
@@ -1,6 +1,5 @@
 [build-system]
 requires = [
-  "setuptools == 41.0.0",  # See https://github.com/ansible/molecule/issues/2350
```
decentral1se: But we require setuptools?

.travis.yml
```diff
@@ -351,4 +351,7 @@ jobs:
 before_install:
   - pip install tox-tags
 script:
+  # keep it like this to first create env and later to run again with tests
+  # in order to avoid regressions to run twice on the same environment.
+  - tox --notest
```
decentral1se: Unclear on this. I'd rather not lose more time on our CI by running tox twice (even if tests are skipped first).

setup.py
```diff
@@ -23,287 +23,6 @@
 
 import setuptools
 
-HAS_DIST_INFO_CMD = False
```
decentral1se: As far as I remember, all these were to assure things can be installed on older setuptools. If we're removing it, we need to document this somewhere that we don't support old versions of whatever. It's hard to tell what this will break TBH.

.travis.yml
ssbarnea: The slowdown is of matter of less than 5 seconds, but the benefits are clear. Is not the first project where I see this done, but only know I fully realised why is needed. Think about it as idempotency test for tox installation ;)

decentral1se: https://github.com/ansible/molecule/pull/2382#pullrequestreview-304255010

pyproject.toml
ssbarnea: This must be transformed into >= in order to revert our previous pinning. It should be enough.

setup.py
ssbarnea: Code should never be added without tests that test for regressions. That is why we ended with a magic setup.py that does so much magic that it breaks installation.

PR #2384, opened 2019-10-20
Reverts setuptools workaround

ssbarnea:
Goes back to `setuptools >= 41.0.0` for build time as missing to mention
it would generate a failure to install from source in editable mode.


ssbarnea:
> I hope adding `pip` isn't going to bite us!

Keep the antiviperin syringe close, just in case.

Issue #2344, opened 2019-10-01
Molecule shows misleading messages when linting is disabled


ansiblejunky:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible 2.8.0
molecule, version 2.20.1
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

- ansible-lint
- yamllint
- flake8

# Desired Behavior

Molecule config file `molecule.yml` has lint section disabled using `enabled: false` and so running `molecule lint` should simply show the following output:
```
> molecule lint
--> Validating schema foo/molecule/default/molecule.yml.
Validation completed successfully.
--> Test matrix

└── default
    └── lint

--> Scenario: 'default'
--> Action: 'lint'
Skipping, lint is disabled.
```

# Actual Behaviour

Molecule outputs misleading messages that show `Executing xxx on files found` which makes the user believe that linting was performed. These messages should not be printed when linting is disabled. The messages are coming from the fact that the `molecule.yml` defines linters for the `verifier` section and `provisioner` section. However, we only ran `molecule lint` so I assume these other sections should never be executed or considered. 

```
> molecule lint
--> Validating schema foo/molecule/default/molecule.yml.
Validation completed successfully.
--> Test matrix

└── default
    └── lint

--> Scenario: 'default'
--> Action: 'lint'
Skipping, lint is disabled.
--> Executing Flake8 on files found in foo/molecule/default/tests/...
Lint completed successfully.
--> Executing Ansible Lint on foo/molecule/default/playbook.yml...
Lint completed successfully.
```

The following is the `molecule.yml`:
```
---
dependency:
  name: galaxy

driver:
  name: vagrant
  provider:
    name: virtualbox

lint:
  name: yamllint
  enabled: False

platforms:
  - name: centos
    box: bento/centos-7.4
    groups:
      - linux

provisioner:
  name: ansible
  lint:
    name: ansible-lint

verifier:
  name: testinfra
  enabled: False
  lint:
    name: flake8
```

ansiblejunky:
It appears the code that performs the lint against each area is:

https://github.com/ansible/molecule/blob/bd7f4767d6a57bf9c1153bcb5c31c5a005fefc40/molecule/command/lint.py#L73-L76

Not sure the `molecule lint` command should span across different phases (lint, verify, provision).

hoanganh25991:
haksjdhfjdsf

rollerd:
[...]

PR #2387, opened 2019-10-20
WIP: Updated linting to run in related test sequence per #2344

rollerd:
Modifying the way that molecule runs linting steps per #2344

- `molecule lint` will now only run the main yamllint against the role
- `molecule verify` and `molecule converge` will now run the linting specified in their respective sections of the molecule.yml file

### DIFF
--- a/molecule/command/converge.py
+++ b/molecule/command/converge.py
@@ -77,6 +77,7 @@ class Converge(base.Base):
         :return: None
         """
         self.print_info()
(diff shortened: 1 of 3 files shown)


decentral1se:
Thanks! Shared some thoughts in https://github.com/ansible/molecule/issues/2344#issuecomment-544603305.

ssbarnea:
Dropped in favour of https://github.com/ansible/molecule/pull/2447

PR #2389, opened 2019-10-21
Added support for internal CIDR

jmpsf:
Right now the ec2 instances created by molecule allow an ingress rule to `0.0.0.0/0` on the ssh port.

As we have a restrictive security policy, we only allow ssh traffic from within the subnet, therefore we need to be able to specify the CIDR that is allowed to ssh into the molecule ec2 instance.

As there can be multiple platforms I've decided to set up the configuration in the `molecule.yml` under the `driver` section. Although I don't know if it's the best place.

```yaml
driver:
  name: ec2
  internal_cidr: 192.168.100.0/24
```

Changes are backward compatible and shouldn't affect current users.

#### PR Type

- Feature Pull Request


jstxu:
Hi @decentral1se, I've checked both cases, either specifying the `internal_cidr` and not specifying it and it worked as expected

jmpsf:
@ssbarnea I've just changed the names of the variables, I don't see how it has broken the tests, could we rerun them?

decentral1se:
Restarted the failed build. It was unrelated.

PR #2390, opened 2019-10-22
ec2 provider: Add support for specifying ssh keypair

jmpsf:
Right now each time a molecule job is run with the ec2 provider, a new key ssh key is created into AWS with the name `molecule_key`.

This means that if several CI jobs are run at the same time, they will have a racing condition, and one of them is going to fail because it can't connect to the created instance.

So in this PR, I add the opportunity to use a specific ssh private key, instead of generating one each job.

As there can be multiple platforms I've decided to set up the configuration in the `molecule.yml` under the `driver` section. 

```yaml
driver:
  name: ec2
  keypair_path: /path/to/ssh/private/key
```

Changes are backward compatible and shouldn't affect current users.

#### PR Type

- Feature Pull Request

### DIFF
--- a/molecule/cookiecutter/scenario/driver/ec2/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
+++ b/molecule/cookiecutter/scenario/driver/ec2/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/create.yml
@@ -27,7 +27,10 @@
         cidr_ip: '0.0.0.0/0'
 
     keypair_name: molecule_key
-    keypair_path: "{{ lookup('env', 'MOLECULE_EPHEMERAL_DIRECTORY') }}/ssh_key"
+    molecule_keypair_path: "{{
+      lookup('env', 'MOLECULE_EPHEMERAL_DIRECTORY') }}/ssh_key"
+    keypair_path: "{{
+      molecule_yml.driver.keypair_path | default(molecule_keypair_path) }}"
   tasks:
     - name: Create security group
       ec2_group:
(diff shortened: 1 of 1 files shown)


jmpsf:
Hi @decentral1se, let me show you why I see a race condition.

https://github.com/ansible/molecule/blob/6c34d80476d1701409724174a3c9774d7a4027d7/molecule/cookiecutter/scenario/driver/ec2/%7B%7Bcookiecutter.molecule_directory%7D%7D/%7B%7Bcookiecutter.scenario_name%7D%7D/create.yml#L39-L42

Tests if a keypair exist locally, beware that `keypair_path` is set at

https://github.com/ansible/molecule/blob/6c34d80476d1701409724174a3c9774d7a4027d7/molecule/cookiecutter/scenario/driver/ec2/%7B%7Bcookiecutter.molecule_directory%7D%7D/%7B%7Bcookiecutter.scenario_name%7D%7D/create.yml#L30

So each role will have a different `keypair_path`.

Then if the ssh key doesn't exist in the `keypair_path` it deletes the AWS keypair called `molecule_key`

https://github.com/ansible/molecule/blob/6c34d80476d1701409724174a3c9774d7a4027d7/molecule/cookiecutter/scenario/driver/ec2/%7B%7Bcookiecutter.molecule_directory%7D%7D/%7B%7Bcookiecutter.scenario_name%7D%7D/create.yml#L44-L48

Then asks AWS to generate a keypair

https://github.com/ansible/molecule/blob/6c34d80476d1701409724174a3c9774d7a4027d7/molecule/cookiecutter/scenario/driver/ec2/%7B%7Bcookiecutter.molecule_directory%7D%7D/%7B%7Bcookiecutter.scenario_name%7D%7D/create.yml#L50-L53

And persists it into the `keypair_path` file

https://github.com/ansible/molecule/blob/6c34d80476d1701409724174a3c9774d7a4027d7/molecule/cookiecutter/scenario/driver/ec2/%7B%7Bcookiecutter.molecule_directory%7D%7D/%7B%7Bcookiecutter.scenario_name%7D%7D/create.yml#L55-L60

So Imagine the first role is being tested, the ssh key is generated by AWS and is now about to launch the instance.

That's when the second role starts to be tested. as it's `keypair_path` is different from role 1, there is no ssh_key present there, so it deletes the AWS `molecule_key` ssh keypair and generates a new one also called `molecule_key`.

Now the role 1 launches the instance with the AWS `molecule_key` ssh keypair.

https://github.com/ansible/molecule/blob/6c34d80476d1701409724174a3c9774d7a4027d7/molecule/cookiecutter/scenario/driver/ec2/%7B%7Bcookiecutter.molecule_directory%7D%7D/%7B%7Bcookiecutter.scenario_name%7D%7D/create.yml#L71-L73

But as it's the ssh keypair of the role 2, once it tries to `prepare` or `converge` we get a `Permission Denied` error.

Cheers

PR #2391, opened 2019-10-23
Remove Hetzner Cloud driver from core

decentral1se:
Will live @ https://github.com/pycontribs/molecule-hetznercloud


till:
@decentral1se How do you use this with head? Now:

```
ERROR: Failed to validate.

{'driver': [{'name': ['unallowed value hetznercloud']}]}
```

Dockerfile:

```
FROM quay.io/ansible/molecule:latest
RUN apk add --update build-base git
RUN pip install --upgrade pip
RUN pip install dnspython requests==2.20.1 docker docker-compose hcloud molecule[hetznercloud]
RUN pip install git+https://github.com/pycontribs/molecule-hetznercloud.git@master#egg=molecule-hetznercloud
```

decentral1se:
Hey @till, sorry you ran into so many issues. We're in a transition period now and I shouldn't have recommended this approach. At current state, you should `pip install -e .` the molecule source and then run the `pip install git+https://github.com/pycontribs/molecule-hetznercloud.git@master#egg=molecule-hetznercloud` command but I've done that now and found another few issues. I better go raise those tickets :sweat_smile:

ssbarnea:
As drivers can request their minimal version of molecule needed, including prerelease, it would be easier to document “pip install molecule-drivername” as the way to install them.

PR #2394, opened 2019-10-23
Include registry in default base test images

ssbarnea:
By including the registry in the base test image we assure that we
can use it with both docker and podman as the former one may not have
a default registry already configured.


ssbarnea:
@sshnaidm ^ this should make it easier for your podman change.

molecule/cookiecutter/molecule/{{cookiecutter.role_name}}/{{cookiecutter.molecule_directory}}/{{cookiecutter.scenario_name}}/molecule.yml
```diff
@@ -17,7 +17,7 @@ platforms:
     size_id: s-1vcpu-1gb
 {%- elif cookiecutter.driver_name == 'docker' %}
   - name: instance
-    image: centos:7
+    image: docker.io/centos:7
```
sshnaidm: Why is here "docker.io/centos:7" and below "docker.io/pycontribs/centos:7" ?

PR #2405, opened 2019-10-25
Adds vpc_id to ec2_group, from subnet facts, and allows to specify SG ip cidr

vinicyusmacedo:
- Added VPC ID to EC2 Group (otherwise, it creates an EC2-Classic Security Group)
Based on #2230
- Added support for other IP CIDRs different than 0.0.0.0/0

### VPC ID
- Create role using EC2 driver: `molecule init role -r test-ec2 -d ec2`
- Include `vpc_subnet_id` to your platform on `molecule.yml`

### IP CIDR
- Create role using EC2 driver: `molecule init role -r test-ec2 -d ec2
- Either include `allowed_ip_cidr` to the platforms (to specify a CIDR) or `auto_public_ip` (to make it use your current public IP). If you include none of these, the default will be 0.0.0.0/0.

PR #2408, opened 2019-10-25
Fixed SG per VPC creation and SSH keypair error on ec2

vinicyusmacedo:
This PR fixes:

- Lack of iterator for subnet facts when creating a security group, introduced by #2405
- Dict breaking when check if a public key existed, introduced by #2389


ssbarnea:
This came too late to be merged as we already removed ec2 from core. You will have to raise it against its new location https://github.com/pycontribs/molecule-ec2/ -- but wait few days so we add the plugin code there.

PR #2417, opened 2019-10-28
Fix wrong template path for drivers

shortmann:
For the new plugin system, molecule has to look for cookiecutter
templates inside the plugin (drivers or verifiers) folder and not in the
main molecule folder.

Signed-off-by: shortmann <kahllund@mac.com>


ssbarnea:
Apparently it needs https://github.com/pycontribs/molecule-azure/pull/12

PR #2428, opened 2019-11-03
docs: fixed broken link to galaxy docs

ssbarnea:

PR #2429, opened 2019-11-03
Avoid adding UUID twice to platform in parallel mode

tadeboro:
Since commit 3b8c8d03, we are placing merged molecule config into the ephemeral directory of the scenario currently being acted upon. This configuration file already contains preprocessed platform names if we run molecule in parallel mode.

Because we forgot to remove the platform name mangling from the filter that is responsible for loading the config file and exposing its content to the playbook tasks, platform names suddenly started having two identical suffixes.

Because those suffixes are relatively long, having two copies hit the upper limit of the hostname length (63 characters). This caused docker to die while trying to create new instances. The exact error message was:

    Error starting container 123...901: 500 Server Error: Internal
    Server Error (\"OCI runtime create failed: container_linux.go:345:
    starting container process caused \"process_linux.go:430: container
    init caused \\\"sethostname: invalid argument\\\"\": unknown\")

If we remove the preprocessing of platform names from the filter function, things go back to normal.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/molecule/provisioner/ansible/plugins/filter/molecule_core.py
+++ b/molecule/provisioner/ansible/plugins/filter/molecule_core.py
@@ -43,9 +43,7 @@ def from_yaml(data):
     i = interpolation.Interpolator(interpolation.TemplateWithDefaults, env)
     interpolated_data = i.interpolate(data)
 
-    loaded_data = util.safe_load(interpolated_data)
-    loaded_data = _parallelize_config(loaded_data)
-    return loaded_data
+    return util.safe_load(interpolated_data)
 
 
 def to_yaml(data):
@@ -65,15 +63,6 @@ def get_docker_networks(data):
                     name = network['name']
                     network_list.append(name)
     return network_list
-
-
-def _parallelize_config(data):
-    if 'platforms' not in data:
-        return data
-    state = util.safe_load_file(os.environ['MOLECULE_STATE_FILE'])
-    if state.get('is_parallel', False):
-        data['platforms'] = util._parallelize_platforms(data, state['run_uuid'])
-    return data
 
 
 class FilterModule(object):


tadeboro:
I left the state intact for now. It seems to me that `is_parallel` and `run_uuid` fields were only used by the function I removed, but it would be nice if someone more familiar with codebase can confirm that this is something that should be removed.

tadeboro:
The CI failure should be fixed by #2428. I will close and reopen this PR once master is green again.

decentral1se:
Hey @tadeboro, nice one for digging into this! Can you confirm that parallel mode continues to work afterwords or you are using without `--parallel`? That was a bit unclear to me. So, by removing this we're sure that the `run_uuid` is already inside the platform instance name? I am wondering in what case that might not be true, do we need instead a conditional statement here? That might also be a hit to performance though (this function runs a lot during run time of Molecule), so ideally, removing is better.

tadeboro:
Let me tell a story of how one developer set out to parallelize Ansible collection's test suite ;)

We use molecule to manage and run our integration tests. Up until now, we were simply executing `molecule test --all` and things were good. But as we were adding more and more scenarios to our test suite, things started to slow down considerably. This is why we are in the process of replacing that simple molecule command with a `Makefile`:

```
modules   := $(wildcard molecule/module_*)
roles     := $(wildcard molecule/role_*)
scenarios := $(modules) $(roles)


.PHONY: all
all: $(scenarios)

.PHONY: modules
modules: $(modules)

.PHONY: roles
roles: $(roles)

.PHONY: $(modules)
$(modules):
	molecule -c config_module.yml test --parallel -s $(notdir $@)

.PHONY: $(roles)
$(roles):
	molecule -c config_role.yml test --parallel -s $(notdir $@)
```

Once we have this ready, we can parallelize out integration tests with a simple `make -j8`. But when I tried to run this command, things failed spectacularly.

After we killed the `no_log` in the create playbook, we were able to determine the cause of the failure: docker did not like our hostname. A few `--debug` runs later we noticed that hostname contained two copies of the `run_uuid`. And now for the big surprise: the person that broke that functionality was us (a few weeks ago, but still us ;). When we dump the merged configuration into the `~/.cache/molecule_parallel/.../molecule.yml`, platform names are already transformed. And since this file is loaded and exposed to the playbook tasks, there is no need to transform it again.

tadeboro:
@mancabizjak @aljazkosir Just to keep you in the loop.

decentral1se:
> And now for the big surprise: the person that broke that functionality was us (a few weeks ago, but still us ;). When we dump the merged configuration into the ~/.cache/molecule_parallel/.../molecule.yml, platform names are already transformed. And since this file is loaded and exposed to the playbook tasks, there is no need to transform it again.

I'm not exactly clear on what you mean here? You're doing something out of the ordinary or you are using some specific functionality of molecule? I am trying to understand whether your fix will break other workflows.

tadeboro:
If I understand things correctly, this is how `molecule` works as far as configuration is concerned:

 1. When we start `molecule`, it merges default configuration, base configuration, and scenario-specific configuration.
 2. If we supply the `--parallel` switch, platform names in the `config["platforms"]` get `run_uuid` suffix appended.
 3. Before each action execution, `molecule` writes dumps a complete configuration into the `molecule.yml` file inside the ephemeral directory.
 4. This file is in the playbooks loaded using the `file` lookup plugin and filtered through `molecule_from_yaml` filter.

Because that last filter is currently adding `run_uuid` suffix to the platform names, names in the playbook end up with double suffix.

Or is there anything that I am missing here?

decentral1se:
> Because that last filter is currently adding run_uuid suffix to the platform names, names in the playbook end up with double suffix.

From what I remember (was a while ago, could be fuzzy / wrong), because we want users to maintain the same UX when running CLI commands like `molecule login --host myinstancename` and not `molecule login --host myinstancename-UUIDNIGHTMARE`, we store the UUID on the state object and then dynamically append it to the platform instance name every time the cached disk files are re-loaded. That's at least what I have in my head.

tadeboro:
Executing `molecule login` command when `--parallel` flag is in use is next to impossible. Because if we want to get `run_uuid`, we need to access the state file, which is stored in the ephemeral directory, which contains `run_uuid` in its name, so we need to access the state file ... Say hello to unterminated recursion ;)

As for the configuration path, I traced it through the sources and it indeed follows the route I described. I can also provide a sample `molecule.yml` file from the ephemeral directory if needed.

decentral1se:
Ah right yep, it's only available on certain sequences anyway (just checked https://molecule.readthedocs.io/en/stable/examples.html#parallel-usage-example). So, can you rebase this PR on master (fix has been merged) and then we can probably go ahead and merge this. Thanks for pushing this!

tadeboro:
I rebased the changes and the CI seems to like them. And thank you for maintaining the tool that makes my life easier.

PR #2436, opened 2019-11-13
lint refresh

ssbarnea:
Needed for avoiding a false positive of:
```
[BLK100] Black would make changes.
```
This was happening because flake8-black adopted new format of black but we were using an older version of black. They need to be kept in sync in order to avoid such errors. Arguably we could even remove flake8-black in the future because we always run black anyway.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,12 @@
 ---
 repos:
   - repo: https://github.com/python/black.git
-    rev: 19.3b0
+    rev: 19.10b0
     hooks:
       - id: black
         language_version: python3
   - repo: https://github.com/pre-commit/pre-commit-hooks.git
-    rev: v2.3.0
+    rev: v2.4.0
     hooks:
       - id: end-of-file-fixer
       - id: trailing-whitespace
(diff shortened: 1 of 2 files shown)


gundalow:
closed/reopen to try and tick Travis-CI back into life

molecule/test/unit/test_config.py
```diff
@@ -376,7 +376,7 @@ def test_set_env_from_file(config_instance):
 
 
 def test_set_env_from_file_returns_original_env_when_env_file_not_found(
-    config_instance
+    config_instance,
```
sshnaidm: And how is that related?

ssbarnea: Black formatting changed, that was made by black, not me.

.pre-commit-config.yaml
```diff
@@ -17,11 +17,11 @@ repos:
       - id: debug-statements
         language_version: python3
   - repo: https://gitlab.com/pycqa/flake8.git
-    rev: 3.7.8
+    rev: 3.7.9
     hooks:
       - id: flake8
         additional_dependencies:
-          - flake8-black
```
ssbarnea: This was tricky because a cached version may have produced different results.

Issue #2435, opened 2019-11-13
anyconfig breaks molecule on centos-8


ssbarnea:
# Issue Type

- Bug report

# Molecule and Ansible details
Initiall seen at https://zuul.opendev.org/t/openstack/build/236553c6a6d84d0788518f449f89bb18/log/job-output.txt

```
File "/home/zuul/test-python/lib/python3.6/site-packages/anyconfig/backends.py", line 44, in <module>
import anyconfig.backend.configobj
```

I am not yet sure which version of anyconfig caused this but because this is the 2nd breaking bug in two weeks and because anyconfig is solely used in a single place in order to us its dictionary-merge strategy, I wonder if it would not be better to drop it as a dependency.

tadeboro:
I vote for replacing the `anyconfig` library with a simple recursive merge function. When I was tracking the configuration building, it just obfuscated the merge process.

ssbarnea:
Feel free to make a PR here.

tadeboro:
I started adding a bit more tests for the `merge_dicts` function in order to get a clear idea on how `anyconfig` merge worked and I already found something *interesting*:

```
@pytest.mark.parametrize('a,b,x', [
    (dict(key={}), dict(key=2), dict(key=2)),
    (dict(key=1), dict(key={}), dict(key={})),
])
def test_merge_dicts(a, b, x):
    assert x == util.merge_dicts(a, b)
```

The merge fails in the first scenario (`TypeError`, `int` is not iterable) but goes through in the second one. And merge being asymmetric does not feel right. Any suggestions on how to proceed here? Should we always fail if only one of the values is `dict`? Should we simply replace the value in this case? Or should we keep the (in my honest opinion useless) existing behavior?

ssbarnea:
This issue seems very similar to Ansible [hash_behaviour](https://docs.ansible.com/ansible/latest/reference_appendices/config.html?highlight=hash_behaviour#default-hash-behaviour) feature so I would be inclined to look at how Ansible does it and to reuse its implementation if possible. We can even document that we are using Ansible logic and avoid extra maintenance on the subject.

tadeboro:
I prepared some code at #2437.

PR #2437, opened 2019-11-14
Remove anyconfig dependency

tadeboro:
The primary goal of this PR is to get rid of the `anyconfig` package dependency. This is what the first commit in this PR does.

In order to make merge function that we created a bit safer to use, we also included the second commit in this PR. In that commit, we made sure `merge_dicts` function operates on a copy of the first input parameter. If that change is not desired, we can remove the second commit from the PR.

Fixes: #2435

### DIFF
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -21,7 +21,6 @@
 from uuid import uuid4
 import os
 
-import anyconfig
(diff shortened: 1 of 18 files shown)


tadeboro:
I fixed the `pre-commit` errors (and properly installed it locally while I was at it).

PR #2443, opened 2019-11-16
Remove GitHub Actions

ssbarnea:
Due to unfortunate surprice GitHub kept under the sleeve when launching
the GitHub Actions we are forced to drop any support for them.

Details: https://github.community/t5/GitHub-Actions/GitHub-Actions-workflows-can-t-be-executed-on-this-repository/m-p/38208#M3260


casz:
Pretty sure there was a simple fix: https://github.community/t5/GitHub-Actions/GitHub-Actions-workflows-can-t-be-executed-on-this-repository/m-p/38466/highlight/true#M3369 🤔

ssbarnea:
That workaround is only for non-paying organizations. Even GitHub support reported that it will not work and even if they decide to change something about it it would likely take weeks or months.

JamesMGreene:
@ssbarnea Some OSS projects qualify for a voucher that will give them a [Teams](https://github.com/pricing) level plan for free, and therefore access to Actions. Probably the best way forward on this would be for an admin of the Ansible organization to [write into Support](https://support.github.com/contact) requesting consideration.

Issue #2462, opened 2019-12-03
Using `--parallel` with Docker driver causes `500 Server Error: Internal Server Error`: container names too long


robertdebock:
# Issue Type

- Bug report

# Molecule and Ansible details

```ansible 2.9.1
  config file = None
  configured module search path = ['/home/robertdb/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.7/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.7.5 (default, Oct 17 2019, 12:16:48) [GCC 9.2.1 20190827 (Red Hat 9.2.1-1)]
```

```
molecule, version 2.22
```

Molecule installation method (one of):
- pip

Ansible installation method (one of):
- pip

# Desired Behavior
Using the parallel (`molecule test --parallel`) option adds a 37 characters UUID to the container name, where the [maximum is 50](https://serverfault.com/questions/923820/docker-ansible-google-compute-engine-debian-packages-a-confusing-story)

I think molecule should truncate docker instance names that are over 50 (or 49?) characters.

# Actual Behaviour

The [build](https://travis-ci.org/robertdebock/ansible-role-bootstrap/jobs/619774227#L324) now fails when using a docker instance name: 'bootstrap-amazonlinux-latest-e75c471e-ff73-4cc8-bda8-848a1aa5a490' (65 characters):

```
315    TASK [Create molecule instance(s)] *********************************************
316    changed: [localhost] => (item=bootstrap-amazonlinux-latest-e75c471e-ff73-4cc8-bda8-848a1aa5a490)
317    
318    TASK [Wait for instance(s) creation to complete] *******************************
319    FAILED - RETRYING: Wait for instance(s) creation to complete (300 retries left).
320    FAILED - RETRYING: Wait for instance(s) creation to complete (299 retries left).
321    FAILED - RETRYING: Wait for instance(s) creation to complete (298 retries left).
322    FAILED - RETRYING: Wait for instance(s) creation to complete (297 retries left).
323    FAILED - RETRYING: Wait for instance(s) creation to complete (296 retries left).
[...]

PR #2463, opened 2019-12-03
Truncate long docker instance names

robertdebock:
[Truncates](https://jinja.palletsprojects.com/en/2.10.x/templates/#truncate) the docker instance name to:
- truncate after `50` characters.
- `True` to cut text
- `''` to indicate the text was truncated. (no character are place.)
- `0` tolerance.

Using these options, the `item.name` is truncated at 50 characters exactly.

Partial-Fixes: #2462

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -142,7 +142,7 @@
 
     - name: Create molecule instance(s)
       docker_container:
-        name: "{{ item.name }}"
+        name: "{{ item.name | truncate(50, True, '', 0) }}"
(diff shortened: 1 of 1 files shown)


ssbarnea:
Happy to see a PR towards making parallel mode work. Still, I see no test being added that assures we do not have an accidental regression.

robertdebock:
CI failed, but on my local Travis account CI succeeded. I guess a single task should be restarted or so.

ssbarnea:
recheck

ssbarnea:
Closing it due lack of maintenance, but happy to reopen it. Likely that we will want a more generic solution that is compatible with running multiple scenarios in parallel.

PR #2468, opened 2019-12-09
Speedup testing by using smaller images

ssbarnea:
Instea of using a `centos:7` image for running tests, we use an `alpine` one which is 4x smaller. This should bring measurable speed improvements on Travis CI which does not cache images.


ssbarnea:
recheck

Issue #2471, opened 2019-12-12
removal of molecule init template command


ssbarnea:
I would like to remove the `molecule init template` feature from molecule as it seems to be very popular. Even its own testing repository was broken and unmantained for a long time.

## Motivation

As we are currently migrating towards using galaxy templating it makes not much use to have this inside molecule itself.

Those really wanting to use cookiecutter templates they can still do it without needing molecule itself.

## Documentation (optional)

This needs to be documented in included in next major release (3.0) as it counds as a breaking change.

`molecule init role` and `molecule init scenario` are **not** affected by this change.

## Anything else?

Please use up/down thumbs to vote for this, eventually adding comments.

Also posted on https://www.reddit.com/r/ansible/comments/eezxpy/proposal_removal_of_molecule_init_template_feature/

PR #2479, opened 2019-12-25
Remove init template command

ssbarnea:
Based on consultation with users (see ticket) it seems that nobody
argued against this change.

Fixes: #2471


ekultails:
Looks like some more of the related tests need to be removed. Example Travis CI failure: https://travis-ci.com/ansible/molecule/jobs/270637079

PR #2480, opened 2019-12-27
tests: use is_subset to validate loaded data

ssbarnea:
Fixes recurring problem with tests which failed for users that use `DOCKER_HOST` to point to a remote docker. This happened because in those cases the loaded configuration dictionary looks different as it has extra parameters added to it.

We failed to catch these on CI because we do not use a remote docker connection.

### DIFF
--- a/molecule/test/unit/driver/test_delegated.py
+++ b/molecule/test/unit/driver/test_delegated.py
@@ -24,6 +24,7 @@ import pytest
 
 from molecule import config
 from molecule.driver import delegated
+from molecule.test.conftest import is_subset
 
 
 @pytest.fixture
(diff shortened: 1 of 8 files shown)


molecule/test/unit/driver/test_delegated.py
```diff
@@ -24,6 +24,7 @@
 
 from molecule import config
 from molecule.driver import delegated
+from ...conftest import is_subset
```
cloudnull: is this intended to be

``` python
from molecule.test.conftest import is_subset
```

ssbarnea: indeed. I wonder why it passed flake8, if I remember well it was able to catch suff like this.

PR #2482, opened 2020-01-04
test: prevent running functional tests when not installed

ssbarnea:
Running functional tests require molecule package to be installed, so
we now test for it and fail fast the testing if needed.

This should simplify development by avoiding confusing errors in
individual tests. Only functional tests are affected so we apply this
requirement only to them.

### DIFF
--- a/molecule/test/functional/conftest.py
+++ b/molecule/test/functional/conftest.py
@@ -40,6 +40,18 @@ LOG = logger.get_logger(__name__)
 IS_TRAVIS = os.getenv('TRAVIS') and os.getenv('CI')
 
 
+@pytest.fixture(scope="session", autouse=True)
+def require_installed_package():
+    import pkg_resources
+
+    try:
+        pkg_resources.require("molecule")
+    except pkg_resources.DistributionNotFound as e:
+        pytest.fail(
+            "Functional tests require molecule package to be installed: {}".format(e)
+        )
+
+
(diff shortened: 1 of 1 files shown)


molecule/test/functional/conftest.py
```diff
@@ -40,6 +40,21 @@
 IS_TRAVIS = os.getenv('TRAVIS') and os.getenv('CI')
 
 
+@pytest.fixture(scope="session", autouse=True)
+def require_installed_module():
+    # import importlib.util
```
sshnaidm: why are these lines commented?

molecule/test/functional/conftest.py
```diff
@@ -40,6 +40,21 @@
...
+    # spec = importlib.util.find_spec("molecule")
+    # __requires__ = ["molecule2"]
+    import pkg_resources
+
+    try:
+        pkg_resources.require("molecule")
```
sshnaidm: why not just "import molecule"?

ssbarnea: Leftovers, removing them.

ssbarnea: `import molecule` would succeed if you run tests from repo-root while some tests will fail too. Doing an import tests if module can be imported which is not the same thing as having a package installed. Hint: query version is one of the things failing.

PR #2483, opened 2020-01-05
test: detect broken podman

ssbarnea:
Improved podman detection so we fail fast if current installation does
not pass `podman info` minimal test. This should save time running
tests on systems what are not properly configured.

Skips will happen only on platforms that are known not to support
podman, but on those supporting it, tests will fail.

Enables `lru_cache` on podman/docker checks in order to improve
testing runtime performance.

I faced https://github.com/containers/libpod/issues/4665 on a RHEL 8 where:
```
podman info
Error: could not get runtime: please update to v2.0.1 or later: outdated conmon version
```

### DIFF
--- a/molecule/test/functional/conftest.py
+++ b/molecule/test/functional/conftest.py
@@ -84,7 +84,7 @@ def skip_test(request, driver_name):
     msg_tmpl = "Skipped '{}' not supported"
     support_checks_map = {
         'docker': supports_docker,
-        'podman': lambda: min_ansible("2.8.6") and platform.system() != 'Darwin',
+        'podman': supports_podman,
         'delegated': lambda: True,
     }
     try:
@@ -253,11 +253,16 @@ def get_docker_executable():
     return distutils.spawn.find_executable('docker')
 
 
+def get_podman_executable():
+    return distutils.spawn.find_executable('podman')
+
+
 def get_virtualbox_executable():
     return distutils.spawn.find_executable('VBoxManage')
 
 
 @pytest.helpers.register
+@util.lru_cache()
 def supports_docker():
     docker = get_docker_executable()
     if docker:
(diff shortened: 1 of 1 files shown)


molecule/test/functional/conftest.py
```diff
@@ -253,11 +253,16 @@ def get_docker_executable():
...
+def get_podman_executable():
+    return distutils.spawn.find_executable('podman')
+
+
 def get_virtualbox_executable():
     return distutils.spawn.find_executable('VBoxManage')
 
 
 @pytest.helpers.register
+@util.lru_cache
```
sshnaidm: is this related to the commit?

molecule/test/functional/conftest.py
```diff
@@ -277,6 +282,30 @@ def supports_docker():
...
+def supports_podman():
+    # Returns true if podman is supported and working
+    # Returns false if podman in not supported
+    # Calls pytest.fail if podman appears to be broken
+    if min_ansible("2.8.6") and platform.system() != 'Darwin':
+        podman = get_podman_executable()
+        if podman:
+            result = util.run([podman, "info"], stdout=PIPE, universal_newlines=True)
+            if result.returncode != 0:
```
sshnaidm: I'd rather use failfast here instead of nested if/if/if but not critical.

molecule/test/functional/conftest.py
```diff
@@ -277,6 +282,30 @@ def supports_docker():
...
+                LOG.error(
+                    "Error %s returned from `podman info`: %s",
+                    result.returncode,
+                    result.stdout,
+                )
+                pytest.fail(
+                    "Cannot run podman tests with a broken podman installation."
+                )
+                return False
+            return True
```
sshnaidm: because of nested "if"s it's not so clear what this "true" is related to, and which conditions are related to multiple "return false"

molecule/test/functional/conftest.py
```diff
@@ -277,6 +282,30 @@ def supports_docker():
...
+@pytest.helpers.register
+@util.lru_cache()
+def supports_podman():
+    # Returns true if podman is supported and working
+    # Returns false if podman in not supported
+    # Calls pytest.fail if podman appears to be broken
+    if min_ansible("2.8.6") and platform.system() != 'Darwin':
+        podman = get_podman_executable()
+        if podman:
+            result = util.run([podman, "info"], stdout=PIPE, universal_newlines=True)
```
sshnaidm: I'd rather use "podman ps", it's really checking if podman is functioning on the host

PR #2497, opened 2020-01-22
Add the ability to configure docker_networks


evrardjp:
Currently, on creation, the docker_network module only receives the
name of the network to create. It is therefore impossible to configure
advanced details like IPAM.

Without this patch, a molecule.yaml containing multiple platforms with
multiple networks with static IPs will basically fail, as it is not possible
to determine the network ranges, because we don't pass IPAM details at the
network creation time.

It is a problem, as everyone having docker_networks should basically have
their own create/destroy playbooks to overcome this.

This fixes the behaviour by adding a new key into platforms,
named docker_networks. A new key was necessary because "networks" is used
inside the docker_container creation, which takes different arguments
than docker_network.

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -89,13 +89,16 @@
 
     - name: Create docker network(s)
       docker_network:
-        name: "{{ item }}"
+        name: "{{ item.name }}"
         docker_host: "{{ item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
         cacert_path: "{{ item.cacert_path | default((lookup('env', 'DOCKER_CERT_PATH') + '/ca.pem') if lookup('env', 'DOCKER_CERT_PATH') else omit) }}"
         cert_path: "{{ item.cert_path | default((lookup('env', 'DOCKER_CERT_PATH') + '/cert.pem')  if lookup('env', 'DOCKER_CERT_PATH') else omit) }}"
         key_path: "{{ item.key_path | default((lookup('env', 'DOCKER_CERT_PATH') + '/key.pem') if lookup('env', 'DOCKER_CERT_PATH') else omit) }}"
         tls_verify: "{{ item.tls_verify | default(lookup('env', 'DOCKER_TLS_VERIFY')) or false }}"
         state: present
+        appends: "{{ item.appends | default(omit) }}"
+        enable_ipv6: "{{ item.enable_ipv6 | default(omit) }}"
+        ipam_config: "{{ item.ipam_config | default(omit) }}"
       with_items: "{{ molecule_yml.platforms | molecule_get_docker_networks }}"
       loop_control:
         label: "{{ item }}"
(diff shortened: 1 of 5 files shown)


evrardjp:
Hello folks.

I am not sure where I should start on the addition of tests -- I see no unittest for the ansible plugin for get_docker_networks, nor do I see a functional test to evaluate if networks are properly created (I suspect this would be done by calling testinfra itself?). Is there anyone that could guide me into what needs to be done?

I am just doing this on my free time, so any pointer would be welcomed :)

evrardjp:
My bad. This is indeed WIP. I will ask questions on the irc chan to query for further help when I will be back from vacation.

rafaelpirolla:
It would be interesting to be able to create more than one network per container. And since you could use the same network in more than one container (more complex test case) maybe it would be better to define the networks in the driver session?
```
driver:
  name: docker
  networks_create:
    - name: one
      ipam_config:
        - subnet: 172.10.0.0/16
          gateway: 172.10.0.1
    - name: two
      ipam_config:
        - subnet: 172.20.0.0/16
          gateway: 172.20.0.1
```
What do you think?

ssbarnea:
For advanced use cases, I would rather switch to ansible as provisioneer and DIY using a "create.yml" playbook.

rafaelpirolla:
The changes involved to fully support this use case are almost the same...
docker.py
```
def get_docker_networks_create(data):
    return data['networks_create']

def get_docker_networks(data):
    network_list = []
    for platform in data:
        if "networks" in platform:
            return platform['networks']
    return network_list

class FilterModule(object):
    """ Core Molecule filter plugins. """

    def filters(self):
        return {
            'molecule_get_docker_networks': get_docker_networks,
            'molecule_get_docker_networks_create': get_docker_networks_create,
        }
```

create.yml
```
    - name: Create docker network(s)
      docker_network:
        name: "{{ item.name }}"
        docker_host: "{{ item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
        cacert_path: "{{ item.cacert_path | default((lookup('env', 'DOCKER_CERT_PATH') + '/ca.pem') if lookup('env', 'DOCKER_CERT_PATH') else omit) }}"
        cert_path: "{{ item.cert_path | default((lookup('env', 'DOCKER_CERT_PATH') + '/cert.pem')  if lookup('env', 'DOCKER_CERT_PATH') else omit) }}"
        key_path: "{{ item.key_path | default((lookup('env', 'DOCKER_CERT_PATH') + '/key.pem') if lookup('env', 'DOCKER_CERT_PATH') else omit) }}"
        tls_verify: "{{ item.tls_verify | default(lookup('env', 'DOCKER_TLS_VERIFY')) or false }}"
        ipam_config: "{{ item.ipam_config | default(omit) }}"
        state: present
      with_items: "{{ molecule_yml.driver | molecule_get_docker_networks_create }}"
      loop_control:
        label: "{{ item }}"
      no_log: false
...
- name: Create molecule instance(s)
networks: "{{ item.networks | default(omit) }}"
```

destroy.yml
```
    - name: Delete docker network(s)
      docker_network:
        name: "{{ item.name }}"
        state: absent
      with_items: "{{ molecule_yml.platforms | molecule_get_docker_networks }}"
```

ssbarnea:
I am not against implementing this feature, especially if it does not introduce backwards incompatible changes. Be sure you rebase, I am not going to review it with conflicts.

PR #2499, opened 2020-01-24
Support the podman --systemd option (instead of --privileged)

westurner:
Support the Podman ``--systemd=True`` option for running systemd in containers without ``--privileged``.

This has apparently been possible since https://developers.redhat.com/blog/2019/04/24/how-to-run-systemd-in-a-container/

In ``role/molecule/default/molecule.yml``:

```yaml
        driver:
          name: podman
        platforms:
        - name: instance
          image: centos:7
          systemd: True
          command: "/usr/sbin/init"
          tty: True
        - name: instance 2
          image: fedora:31
          systemd: True
          command: "/sbin/init"
```

- [x] Please include details of what it is, how to use it, how it's been tested
- [x] Please include an entry in the CHANGELOG.md if the change will impact users.

#### PR Type

- Feature Pull Request

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,8 @@ Changelog
 * Use pluggy to load plugins
 * Windows & Linux EC2 instances can be tested simultaneously
 * Windows EC2 instances can be provisioned using the automatically-generated password
+* Support the Podman ``--systemd=True`` option for running systemd in
+  containers without ``--privileged``.
 
 2.22
 ====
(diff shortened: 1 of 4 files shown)


ssbarnea:
recheck

sshnaidm:
Firstly, the `--systemd` option is not replacement for `--privileged` option, they are completely different. So the commit message is confusing.
Secondly the`--sytemd` option is `True` [by default](https://github.com/containers/libpod/blame/master/docs/source/markdown/podman-run.1.md#L781-L783) So the commit message is misleading again. 
The only case for this commit is to set it to `False`, and I don't think there is such use-case for molecule.

PR #2501, opened 2020-01-24
docker: networks_cli_compatible default to true

ssbarnea:
Avoids below deprecation warning from code inside molecule by setting
the networks_cli_compatible to true which is also future proof, unless
user mentions another value.

> [DEPRECATION WARNING]: Please note that docker_container handles networks
> slightly different than docker CLI. If you specify networks, the default
> network will still be attached as the first network. (You can specify
> purge_networks to remove all networks not explicitly listed.) This behavior
> will change in Ansible 2.12. You can change the behavior now by setting the new
>  `networks_cli_compatible` option to `yes`, and remove this warning by setting
> it to `no`. This feature will be removed in version 2.12. Deprecation warnings
> can be disabled by setting deprecation_warnings=False in ansible.cfg.
>

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -139,6 +139,7 @@
         ulimits: "{{ item.ulimits | default(omit) }}"
         networks: "{{ item.networks | default(omit) }}"
         network_mode: "{{ item.network_mode | default(omit) }}"
+        networks_cli_compatible: "{{ item.networks_cli_compatible | default(true) }}"
         purge_networks: "{{ item.purge_networks | default(omit) }}"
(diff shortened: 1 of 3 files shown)


pabelanger:
This failed to merge, because of extra branch protection.  Today, zuul.a.c doesn't enfore this in the gate pipeline. Step 1, is to remove code reviewer requirement, step 2, update zuul to get branch protection settings and only enqueue once all are reported

ssbarnea:
That explains it. Sadly I cannot reconfigure that repository. @gundalow ?

pabelanger:
I've removed it for now, however next week we should sync on branch protection. We can simply it, and work to improve zuul. There is a good reason today why we don't enforce code review, on the pipeline, this is because a single user cannot approve their own PR.

PR #2506, opened 2020-01-25
Reduce the duration of functional testing

ssbarnea:
The longest-running test was ~350s and now ~130s.

Tested locally via:
```
pytest -sx -k "test_command_test[driver/docker-docker-multi-node]"
```


ssbarnea:
recheck

Issue #2293, opened 2019-09-15
Externalize linting from molecule v3


ssbarnea:
# Proposal: lint-outsourcing

*Author*: Sorin Sbarnea <@ssbarnea> IRC: zbr

*Date*: 2019-09-15

- Status: Accepted after consultation on [reddit](https://www.reddit.com/r/ansible/comments/e1wq19/proposal_on_ansible_molecule_v3_major_linting/).
- Proposal type: core design
- Estimated time to implement: 2 weeks

As would like to get some community feedback regarding a major change I would like to make for v3 version of Ansible Molecule. While doing other refactoring the need to rework linting become more pressing. The whole idea is to make `molecule lint` easier to use and more flexible at the same time.

Another reason for this task is that linting should be performed at repository level and not at scenario level. One repository can have lots of roles and scenarios and linting at scenario level does not make much sense, creating a lot of duplicated configuration.

# Molecule v3 :: Externalize linting

Linting is important but the reality is that each project may have different needs and preferences. Projects are likely to want to use more than linter, maybe to disable some checks.

We were close to completely removing linting support from molecule as it was adding a lot of complexity and coupling between different components. IRC channel has at least one question related to linting each week, the common ones where to configure it, how to disable it, why is running at the wrong step.

Having only one lint command which runs one configurable shell tasks is the way to go. This would allow users to use any linter they want, **without having to modify molecule code to add support for a new linter**.

We did our best to avoid execution failures when you start using v3 so this is what it will happen:

## Migrating linter config from v2 to v3

If you removed or disabled linting in `molecule.yml`, there is nothing for you to change.

An error would be raised if obsolete lint config items are found. In order to make it easier to migrate you can see this complex setup before and after migration.

```yaml
    # old v2 format:
    lint:
      name: yamllint
      enabled: true
    provisioner:
      lint:
        name: ansible-lint
      options: ...
      env: ...
    verifier:
      lint:
        name: flake8
```

If you want to make molecule avoid any linting remove all linting related options from the config.
If you still want to perform linting, look at example below:

```yaml
    # new v3 format, calling multiple linters:
    lint: |   
        yamllint .
        ansible-lint
        flake8
```

## Behaviour to implemented

* If no `lint` section exists assumes no linting
* If old linting options are found, error will be displayed pointing to this ticket, which contain example on how to use new syntax
* Update documentation
* Update templates

## Feedback needed

[...]

PR #2513, opened 2020-01-26
Refactored molecule linting

ssbarnea:
Changes configuration of linting to a model that does not require
writing plugins in order to support other linters. 

Mainly the new setup executes a shell command as part of the lint sequence step.

When no command is mentioned, linting will be skipped.

* only one place to configure linting in molecule
* user decides what commands to run (more flexible)
* molecule no longer have knowledge of specific linters

More details regarding for this breaking change can be found on linked ticket.

Fixes: #2293


molecule/test/functional/docker/test_scenarios.py
```diff
@@ -262,7 +261,7 @@ def test_host_group_vars(scenario_to_test, with_scenario, scenario_name):
     options = {'all': True}
     cmd = sh.molecule.bake('test', **options)
     out = pytest.helpers.run_command(cmd, log=False)
-    out = util.strip_ansi_escape(out.stdout.decode('utf-8'))
+    out = util.strip_ansi_escape(out.stdout)
```
sshnaidm: how is that related?

docs/configuration.rst
```diff
@@ -83,18 +83,49 @@ Podman
...
+Molecule was able to use up to three linters and while it was aimed to flexible
+about them, it ended up creating more confusions to the users. We decided to
+maximize flexibility by just calling an external shell command.
+
+.. code-block:: yaml
+
+    lint: |
+        yamllint .
+        ansible-lint
+        flake8
```
sshnaidm: Is "env" supported? Worth to mention what are possible options here.
Also from which directory it's executed.

zuul.d/layout.yaml
```diff
@@ -6,7 +6,7 @@
     vars:
       tox_envlist: py27-ansible28-unit
       tox_environment:
-        PYTEST_REQPASS: 559
+        PYTEST_REQPASS: 472
```
sshnaidm: this is very weird thing to do

docs/getting-started.rst
```diff
@@ -103,9 +103,8 @@ keys represent the high level components that Molecule provides. These are:
 * The :ref:`driver` provider. Molecule uses `Docker`_ by default. Molecule uses
   the driver to delegate the task of creating instances.
 
-* The :ref:`linters` provider. Molecule uses :std:doc:`Yamllint
-  <yamllint:index>` by default to ensure that best practices are encouraged
-  when writing YAML.
+* The :ref:`root_lint` command. Molecule can call external commands to ensure
```
sshnaidm: > The :ref:`root_lint` command

Above it's just `lint` command

molecule/command/lint.py
```diff
@@ -71,18 +73,21 @@ def execute(self):
...
+        if cmd == 'yamllint':
+            LOG.warning(
+                "Deprecated linter command found, migrate to newer syntax using <url>"
+            )
+            cmd = 'yamllint .'
+
+        try:
+            # import pdb; pdb.set_trace()
+            LOG.info("Executing: %s" % cmd)
+            util.run(cmd, shell=True, universal_newlines=True, check=True)
```
sshnaidm: maybe worth to support `env` here

molecule/command/lint.py
```diff
@@ -71,18 +73,21 @@ def execute(self):
...
+        # v3 migration code:
+        cmd = self._config.lint
+        if cmd == 'yamllint':
+            LOG.warning(
+                "Deprecated linter command found, migrate to newer syntax using <url>"
+            )
+            cmd = 'yamllint .'
+
+        try:
+            # import pdb; pdb.set_trace()
```
sshnaidm: leftovers?

molecule/command/lint.py
ssbarnea: Not in this PR. In fact this will reuse env from parent process. I want to avoid extra logic arond env because if needed it means that user would not be able to run the linter w/o molecule.

zuul.d/layout.yaml
ssbarnea: I know but it worked fine so far. Even got a new PR from a new user which didn't ask anything about it, he just updated the number because he was adding a new test.

I bet you will see the real benefit as soon we will have something changes on Zuul nodes that triggers some tests to skip. We still have to re-enable podman ones, which are skipped now.

docs/getting-started.rst
ssbarnea: Thanks, fixed.

molecule/command/lint.py
```diff
@@ -23,13 +23,19 @@
...
 
 
 class Lint(base.Base):
     """
-    Lint Command Class.
+    Lint command executes external linters.
+
+    You need to remember to install those linters. For convenience, there is a
+    package extra that installs the most common ones, use it like
+    ``pip install "molecule[extra]"``.
```
dmsimard: Should this be ``pip install molecule[lint]`` rather than ``pip install molecule[extra]`` ?

ssbarnea: Indeed.

PR #2514, opened 2020-01-26
Update internal ansible.cfg defaults

ssbarnea:
Adds several new defaults to internal ansible.cfg so:

- future proof interpreter detection enabled (avoid warnings)
- enable ssh pipelining (performance)
- sets forks to 500 (performance)
- display failures to stderr instead stdout (compliance)

Users can still override these values in molecule.yml files.


molecule/provisioner/ansible.py
```diff
@@ -383,13 +383,17 @@ def default_config_options(self):
         return {
             'defaults': {
                 'ansible_managed': 'Ansible managed: Do NOT edit this file manually!',
+                'display_failed_stderr': True,
+                'forks': 500,
```
geerlingguy: 500 seems high to me; though not sure what prompts this change. I usually set it to 50 as a pretty sane default (especially for CI where the systems might not have the capacity I have on my own workstation).

ssbarnea: @geerlingguy Thanks for looking into these. I was also wondering if we should put a bigger timeout like 30s.

I kinda have mixed feelings about it as I do not want to delay failures too much but I do not want to get failures due to slow external services (cloud provisioning being one of them).

My main concern is to make CI testing as fast as possible, so if you know other tricks that may help, they would be welcomed.

Note: mitogen is out of question till https://github.com/dw/mitogen/issues/597

pabelanger: pipelining below, should speed things up. Maybe revert this and see the impact?  With zuul today, we only have 1vcpu / 1gb RAM. We could bump molecule testing to more, but ideally keeping this smaller helps with total capacity in testing.

PR #2519, opened 2020-01-31
Make linters an optional dependency

ssbarnea:
Makes installation of linters optional in molecule, allowing it to be installed without installing them. If you are using them you will need to either add them as dependencies or use the newly
introduced 'lint' extra when you install molecule.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -41,7 +41,7 @@ Keep in mind that on selinux supporting systems, if you install into a virtual
 environment, you may face :gh:`issue <ansible/ansible/issues/34340>` even
 if selinux is not enabled or is configured to be permissive.
 
-It is your reponsability to assure that soft dependencies of Ansible are
+It is your responsibility to assure that soft dependencies of Ansible are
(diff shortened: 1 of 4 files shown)


docs/installation.rst
```diff
@@ -41,7 +41,7 @@ Keep in mind that on selinux supporting systems, if you install into a virtual
 environment, you may face :gh:`issue <ansible/ansible/issues/34340>` even
 if selinux is not enabled or is configured to be permissive.
 
-It is your reponsability to assure that soft dependencies of Ansible are
+It is your responsability to assure that soft dependencies of Ansible are
```
geerlingguy: 'responsibility'

ssbarnea: @geerlingguy I find both requires but the core issue here around how that they should be used in parallel and be relatively independent tools. I faced too many bugs myself (and support tickets) related to linting such as "lint works at repo level but molecule fails with another linting error", or how to I disable linting in molecule: guess what that may be a stuggle even for its maintainers! ;)

PR #2524, opened 2020-02-02
Shell dependency: use proper lexical parsing

sio:
Shell dependency manager used simple splitting on whitespace for parsing command lines. That would lead to incorrect parsing and execution of more complex commands like `echo "hello world"`.

This PR introduces proper lexical parsing with shlex module from Python standard library. Unit tests and changelog entry are included.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,7 @@ Changelog
 * Use pluggy to load plugins
 * Windows & Linux EC2 instances can be tested simultaneously
 * Windows EC2 instances can be provisioned using the automatically-generated password
+* Shell dependency manager now uses proper lexical parsing of complex command lines
 
 2.22
 ====
--- a/molecule/dependency/shell.py
+++ b/molecule/dependency/shell.py
@@ -20,6 +20,7 @@
 """Shell Dependency Module."""
 
 import os
+import shlex
 
 import sh
 
@@ -97,7 +98,7 @@ class Shell(base.Base):
 
         :return: None
         """
-        command_list = self.command.split(' ')
+        command_list = shlex.split(self.command)
         command, args = command_list[0], command_list[1:]
 
         self._sh_command = getattr(sh, command)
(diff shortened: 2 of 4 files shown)


ssbarnea:
@sio AFAIK we are already using `sh` library in other places in molecule and that had support for parsing shell. Not sure why we should re-implement it instead of assuring we re-use it from sh? Can you investigate this path?

sio:
@ssbarnea, I've checked both [documentation] and [source code] for **sh** - and I don't see any parsing capabilities. **sh** expects lists of strings for command arguments, not unparsed strings. This is mentioned in their FAQ: [Why do my arguments need to be separate strings?](https://amoffat.github.io/sh/sections/faq.html?highlight=parsing#why-do-my-arguments-need-to-be-separate-strings)

Molecule shell dependency manager, on the other hand, [expects](https://molecule.readthedocs.io/en/stable/configuration.html#shell) an unparsed command line string from configuration. That means there must be some parsing logic, and there was previously - just not a very good one. This PR is literally an one-line change:

```diff
-        command_list = self.command.split(' ')
+        command_list = shlex.split(self.command)
```

Molecule currently fails to execute commands as simple as `echo "hello world"`, and I've encountered similar error in-the-wild. That's why I came here to propose a fix.

[documentation]: https://amoffat.github.io/sh/index.html
[source code]: https://github.com/amoffat/sh

ssbarnea:
@sio I hope you are running molecule from master for development, so I can rely on early feedback from you regarding recent changes.

sio:
@ssbarnea, I was running the stable version from PyPI until I've encountered this bug. Currently the problematic config works fine with molecule from `sio:shlex`, but I didn't include it into CI runs yet.  Latest `master` does not feel any different from 2.22 on a limited subset of configs I've tried.

I'm currently considering switching CI to use molecule from a specific git commit, but only until these changes make their way into the release.

sio:
Thank you, @ssbarnea! This review process was a pleasure ^_^

PR #2526, opened 2020-02-02
Made init command required opts into args

ssbarnea:
This fix previous CLI bug which added some role name and scenario name as options instead of arguments, when they in fact were not optional.

Old: molecule init role -r foo
New molecule init role foo

Old: molecule init scenario -s bar
New: molecule init scenario bar


ssbarnea:
recheck

ssbarnea:
recheck

Issue #2013, opened 2019-04-26
Make Ansible default verifier


webknjaz:
# Proposal: Make Ansible default verifier

*Author*: Sviatoslav Sydorenko <@webknjaz> IRC: webknjaz

*Date*: 2019-04-26

- Status: New
- Proposal type: core design, consistent UX
- Estimated time to implement: 1h

## Motivation

The UX is inconsistent with the Ansible Engine and Galaxy which promise the ability to keep using YAML. "native" way of doing testing should come first which will improve the experience for new users.

### Problems

What problems exist that this proposal will solve?
- Inconsistent UX across Ansible-related projects

## Solution proposal

- set Ansible to be a default verifier while keeping other options available.

## Dependencies (optional)

- Now that #1714 is merged, this is finally possible!

## Testing (optional)

N/A

## Documentation (optional)

Just a `:versionchanged:` pointer.

## Anything else?

At Ansible Core and Galaxy we try to ensure the consistent user experience so we all agreed that defaults should be "native" to be "closer" to the user expectations and simplicity promises.
This will also facilitate the further introduction of Ansible Collections creation, testing and publishing workflows.

MarkusTeufelberger:
I see some value in having a tool that is NOT Ansible to verify something that was modified by Ansible.
Also writing good assertions in Ansible is surprisingly hard and brittle.
Lastly the Ansible verifier is very new, I'm not so sure if there are not still some kinks in there (e.g. imho it should fail the test if any job in the verification reports `changed`).

None of these reasons are good enough to warrant a veto, I'd just be more comfortable with that change for e.g. Molecule around version 1.23, not directly the next release.

webknjaz:
@MarkusTeufelberger your argument makes sense. The thing is that we don't want to recommend learning a programming language additionally to what Ansible users already know because this makes things inconsistent and more complicated. Modules/plugins authors will develop content in Python though and having playbooks to verify those will also bring in a proper integration testing into the scenario.

MarkusTeufelberger:
Modules I co-developed are extensively integration tested (https://github.com/ansible/ansible/tree/devel/test/integration), but in those the assertions and state generation are usually quite close together.

[...]

PR #2534, opened 2020-02-04
Makes ansible default verifier

ssbarnea:
Replace testinfra with ansible as the default verifier.

Now testinfra is an optional dependency which is instalable as an extra. This requires us to bump major version because users may need to add this extra to keep their code compatible.

This does *not* mean we deprecate testinfra, we only transition to a modular architecture.

Fixes: #2013
Relates: #2228


ssbarnea:
recheck

PR #2536, opened 2020-02-06
Makes pytest failures more verbose

ssbarnea:
Fixed issue where test failures did not allow us to see what did not match due
to the default abbreviation. This makes output more verbose for errors, so
we can see more information.


ssbarnea:
recheck

Issue #2543, opened 2020-02-07
3.0a6 & 3.0a7: Verifier testinfra does not run


unglaublicherdude:
# Issue Type

- Bug report

# Molecule and Ansible details

I actually run my tests within the docker images taged as 3.0a6 and 3.0a7

```
ansible --version && molecule --version
```

Molecule installation method (one of):

- docker

Ansible installation method (one of):

- docker

Detail any linters or test runners used:

Nothing special. I just go interactively into the official container.

```
$ docker run -it --entrypoint=/bin/sh quay.io/ansible/molecule:3.0a6
$$ molecule init role --verifier testinfra rolename
$$ cd rolename
$$ molecule test --all
[...]
--> Scenario: 'install_docker_with_platform_container'
--> Action: 'verify'
Skipping, verify playbook not configured.
[...]
```

# Desired Behavior

It should have run the tests the initializer placed in the tests directory.

# Actual Behaviour

It skips the tests.

unglaublicherdude:
I want to add, that the python lib testinfra is also missing in this image. So even if you "trick" molecule by adding an empty playbook, the testinfra tests will fail because of the missing lib.

https://github.com/ansible-community/molecule/commit/3d28ab3e4de9f221ff5afb186fbf273a73799b46#diff-380c6a8ebbbce17d55d50ef17d3cf906

PR #2545, opened 2020-02-07
Assure verify step is not skipped with testinfra

ssbarnea:
Fixed regression in previous pre-release cased by the switch from having ansible the default verifier instead of testinfra. Ansible verifier behavior is to skip verify step when playbook is missing, but with testinfra skip should not happen because the tests are not a playbook.

Fixes: #2543


ssbarnea:
recheck

ssbarnea:
recheck

PR #2548, opened 2020-02-10
Test test_command_init_scenario with delegated

ssbarnea:
Instead of using podman or docker to test this long scenario, switch to use
only delegated for testing it.


ssbarnea:
recheck

pabelanger:
recheck

ssbarnea:
recheck

PR #2549, opened 2020-02-10
Change init commands to avoid override by default

ssbarnea:
Changed default value of override to False in order to prevent overriding
already existing files when expanding templates.

Fixes bug where running `molecule init scenario foo` would override
your own .yamllint file.


ssbarnea:
Currently blocked by https://github.com/cookiecutter/cookiecutter/pull/1076

PR #2551, opened 2020-02-13
Revert "Rename playbook.yml to converge.yml"

ssbarnea:
This reverts commit 9dda2e69ac17f7277ac6870b363a6d3b02c54d2a.

That change has other serious implications and caused unexpected bugs. We will reconsider it for the next major release, as v3.0 already has a too long list of changes.


ssbarnea:
recheck

ssbarnea:
> We have a lot of changes that replaced playbook.yml with converge.yml. Let's not break again the whole world. Please make it both available.
> Like it should have been done for transition.

This is what I was planning to do but this change had serious implications to other speudo-steps which rely on playbook, like check or idempotence. The switching of default verifier also interacted with this because some playbooks are optional with one verifier but not with another.

My first attempt was to fix it, but the number of required changes started to get out of control and I decided that maybe is better to lower the scope.

Now that the gated failed, I will give it another go, maybe I can avoid the revert.

PR #2559, opened 2020-02-16
Corrects playbook.yml deprecation message

ssbarnea:
Previous implementation wrongly swapped arguments on deprecation
message, which confused users about rename direction.


ssbarnea:
recheck

PR #2563, opened 2020-02-20
Update instruction with correct flag

zdtsw:
Minor fix for https://molecule.readthedocs.io/en/latest/getting-started.html with incorrect information when init a new role dir
Old: $ molecule init role my-new-role
Should be: $ molecule init role --role-name my-new-role

- Docs Pull Request

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -34,7 +34,7 @@ To generate a new role with Molecule, simply run:
 
 .. code-block:: bash
 
-    $ molecule init role my-new-role
+    $ molecule init role --role-name my-new-role
 
 You should then see a ``my-new-role`` folder in your current directory.
 


ssbarnea:
This is because you read the documentation from latest but used outdated versions 2.22. You either switch to the old documentation or better use the newer versions.

zdtsw:
As I can see from your release versions, 2.22 is the "latest version", 3.X are all marked as pre-release
I installed molecule from pip, 2.22 is the  latest one I can get, following documentation.
In my opinion,  the best way to not confuse someone newly start using molecule is to have >2.22 official released.

ssbarnea:
@zdtsw install latest using `pip install --pre molecule`, or wait a couple of hours and the 3.0.0 release will appear, it needs to finish last CI/CD pipelines.

Most of the time the differences in documentation are not observable but in that particular case we delayed a release for too long.

Issue #2569, opened 2020-02-20
Invoking a scenario fails if parallel scenarios exist


greg-hellings:
# Issue Type

- Bug report

# Molecule and Ansible details

```
$ ansible --version && molecule --version
ansible 2.9.4
  config file = /home/ghelling/.ansible.cfg
  configured module search path = ['/home/ghelling/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/ghelling/venv/tools3/lib64/python3.7/site-packages/ansible
  executable location = /home/ghelling/venv/tools3/bin/ansible
  python version = 3.7.6 (default, Jan 30 2020, 09:44:41) [GCC 9.2.1 20190827 (Red Hat 9.2.1-1)]
molecule 3.0.0
   ansible==2.9.4 python==3.7
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

# Desired Behavior

I have two scenarios in parallel directories under molecule, as well as a shared folder to hold shared code for both of them:

```
$ tree -L 1 molecule
molecule
├── docker
├── openstack
└── shared
```

I would like to be able to run just the `docker` scenario and test how the migration to 3.0 is coming along

```
$ molecule lint -s docker
<successful lint run>
```

# Actual Behaviour

Because the OpenStack driver is not yet out/installed (even though the "docker" scenario is specifically being selected):

```
$ molecule lint -s docker
---
driver:
  name: openstack
platforms:
  - flavor: ci.m1.medium
    image: CentOS-7-x86_64-GenericCloud-released-latest
    name: test-dhcpd

ERROR: Failed to pre-validate.

{'driver': [{'name': ['unallowed value openstack']}]}
```

ssbarnea:
That is because you do not have the openstack driver installed and all scenarios are loaded regardless what command you try to run.

v3 made some changes that may make possible to avoid this, but that feature is not yet implemented. If you install molecule-openstack, you should be fine.

greg-hellings:
I realize that's a workaround, but the lazy-load feature in molecule would be wonderful. I run scenarios across the breadth of my collection in CI with tox (I wrote a plugin for it to auto-generate environments for Ansible collections with molecule) and minimizing the virtualenv dependencies would be ideal.

If the lazy-load is a planned feature just waiting on person-hours, I'm happy to look into it as a contribution.

PR #2571, opened 2020-02-21
Avoid loading all scenarios if not needed

greg-hellings:
Partial-Fix: #2569


greg-hellings:
I'm sure this is a trash way to do it, but it was quick, dirty, and works. I wanted to get the conversation started on proper implementation while the issue was still fresh in my mind.

PR #2572, opened 2020-02-22
Extend driver api

ssbarnea:
- allow driver to expose ansible modules
- allow driver to expose embedded playbooks


hswong3i:
@ssbarnea I see so this is what we are now missing for `molecule-vagrant` custom module and embed playbooks ;-)

Issue #2560, opened 2020-02-17
 Migrating to molecule v3 checklist


ssbarnea:
# How to migrate to molecule v3 (3.0a9 or newer)
Only **delegated**, **docker** and **podman** providers are ready, so if you use others, delay till they are updated or preferably give a hand with that. 

- remove `scenario.name` from `molecule.yaml`, from now on, only folder name determines the scenario name
- remove `lint:` sections from molecule.yaml or [replace them with newer syntax from documentation](https://molecule.readthedocs.io/en/latest/configuration.html#lint)
- if you use `testinfra`, add it as a pip dependency, is no longer installed by default
- if you use `linters`, you need to add them as pip dependencies
- (optional) rename `playbook.yml` to `converge.yml` to avoid a deprecation message

## Known regressions with v3

* TBH

# When not to migrate to v3

If you use one of:

* non-core providers like  [Azure](https://github.com/ansible-community/molecule-azure), [EC2](https://github.com/ansible-community/molecule-ec2), [DigitalOcean](https://github.com/ansible-community/molecule-digitalocean), [GCE](https://github.com/ansible-community/molecule-gce), [HetznerCloud,](https://github.com/ansible-community/molecule-hetznercloud) [Linode](https://github.com/ansible-community/molecule-linode), [LXD](https://github.com/ansible-community/molecule-lxd), [OpenStack](https://github.com/ansible-community/molecule-openstack) or [Vagrant](https://github.com/ansible-community/molecule-vagrant), you probably do want to delay migration until they are updated. Keep in mind that as well as molecule, they are fully community supported, so if nobody steps in to help, they will be affected by bit rot and become unusable.

* you are using [goss](https://github.com/ansible-community/molecule-goss) verifier which was removed. It can still be re-implemented as a plugin, if anyone is willing to help.

zeitounator:
@ssbarnea. Shouldn't playbook.yml be renamed to converge.yml? I guess it's a typo...

ssbarnea:
@zeitounator Thanks, fixed. That is the proof that most people TLDR.

geerlingguy:
Can you add a little more detail to the `lint:` section? E.g. maybe a link to the relevant documentation page(s)?

hswong3i:
@ssbarnea as per https://github.com/ansible-community/molecule-vagrant/issues/2, could I give a hand for molecule-vagrant?

SirUli:
[...]

PR #2574, opened 2020-02-23
Warn users they should use `set -e` for lint

zeitounator:
Playing serveral linters without `set -e` will not catch
the errors in first ones if the latest suceeds.

This is following discussions in #2560

#### PR Type

- Docs Pull Request


ssbarnea:
I picked another one first, nolonger needed.

PR #2578, opened 2020-02-24
Suggest using python3 and pip3 in example docs

brianhelba:
Some common distros do not alias "pip" and "python" to Python 3 versions. This will encourage users following the examples to install and run Molecule with Python 3.

This does not change the actual support status of Python 2.

#### PR Type
- Docs Pull Request

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,10 +54,10 @@ RUN apk add --update --no-cache --repository http://dl-3.alpinelinux.org/alpine/
 ENV MOLECULE_EXTRAS="docker,docs,windows,lint"
 
 # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=917006
-RUN pip3 install -U wheel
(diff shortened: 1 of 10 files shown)


ssbarnea:
I am clearly in favour of python3 first approach as I plan to drop py2 support this yet. I am still not sure if we do have the pip3 everywhere where we have python3.

manics:
Just a warning: I've seen reports elsewhere of virtualenvs which do not have the `python3`/`pip3` aliases. In this situation a user could inadvertently run a global `pip3` instead of the virtualenv `pip`.

ssbarnea:
I am very aware of virtualenv, cli inconsistency but that can easily be avoided running `python3 -m virtualenv` or `python3 -m venv`. Also pip works as a module.

brianhelba:
@ssbarnea Thanks for the review! I just added a new commit to address your feedback and rebased the branch to resolve a merge conflict.

PR #2586, opened 2020-02-25
Opensuse image does not contains sysvinit by default

rosineygp:
- If package not installed, problems to test services will happens even
command: /sbin/init has been configured

I have been tried to run services for a while, everything works in centos and oraclelinux but fails in opensuce, using docker driver.

#### PR Type

- Bugfix Pull Request


rosineygp:
> I am not sure if we really want this by default as systemd with containers is a mined-field.
> 
> I think that this may be a case where you need to build your own container.
> 
> I will not close this right away, maybe I am missing some good reason for including it.

If not apply this... the following modules not working in test for opensuse:
- service
- systemd

In documents has no infomantion how to solve it. only this notation for centos:

```yaml
....
platforms:
  - name: instance
    image: centos:7
    command: /sbin/init
    tmpfs:
      - /run
      - /tmp
    volumes:
....
```
Everythings works in centos, because it have `/sbin/init` by default.

PR #2591, opened 2020-02-28
Support Both `entrypoint` and `command` with Simplified Default

### LINKED ISSUE #2593: Proposal: Support Both entrypoint and command with Simplified Default
# Proposal: Support Both `entrypoint` and `command` with Simplified Default

*Author*: Wong Hoi Sing Edison <@hswong3i>

*Date*: 2020-02-29

- Status: New
- Proposal type: core design, schema changes
- Estimated time to implement: **DONE** (see #2591)

## Motivation

For docker and podman driver, by default we only handle the `command` overwrite with `sleep infinity`.

### Problems

In case base image have both `ENTRYPOINT` and `CMD` specified, our `sleep infinity` may not functioning correctly.

## Solution proposal

- Add `entrypoint` for `ENTRYPOINT` support, with default `''`
  - `''` means enforce rollback to docker's build-in default `ENTRYPOINT` with `/bin/sh`
  - **NOT** setting it as `/bin/sh` else it will be replaced into `exec /bin/sh` which not works
  - See https://docs.docker.com/engine/reference/run/#entrypoint-default-command-to-execute-at-runtime#entrypoint-default-command-to-execute-at-runtime
- Update `command` with new default `bash -c "sleep infinity"`
- Remove `override_command` for simplify logic

For simple test case which we are NOT care about the base image original `ENTRYPOINT` and `CMD`, with this patch we will default it with `''` and `bash -c "sleep infinity"` therefore generally working for Molecule:

```
platforms:
  - name: instance
    image: centos:7
    entrypoint: ''
    command: 'bash -c "sleep infinity"'
```

In case we ARE caring about the base image origin `ENTRYPOINT` and `CMD` (e.g. [mariadb:10.4](https://github.com/docker-library/mariadb/blob/master/10.4/Dockerfile)), just simply set them as `~` in order to resume the original behavior:

```
platforms:
  - name: instance
    image: mariadb:10.4
    entrypoint: ~
    command: ~
```

When combine with systemd based container we could specify `ENTRYPOINT` and `CMD` as `init` and `bash -c "sleep infinity"`, therefore much similar as real world production use cases:

```
platforms:
  - name: instance
    image: centos:7
    privileged: true
    volumes:
      - "/sys/fs/cgroup:/sys/fs/cgroup:rw"
    entrypoint: init
    command: 'bash -c "sleep infinity"'
    tty: True
    environment:
      container: docker
```

## Dependencies (optional)

N/A

## Testing (optional)

Corresponding test cases updated (see #2591; moreover, CI passed)

## Documentation (optional)

Corresponding documentation updated (see #2591)

## Anything else?

Ultimately, if we `docker commit` the result container as a static image with original base image's `ENTRYPOINT` and `CMD` during `verify.yml`, we could therefore push this brewed image hub.docker.com as an alternative of [ansible-bender](https://github.com/ansible-community/ansible-bender) project with standard Ansible supporting.

hswong3i:
Fixes #2593

#### PR Type

- Feature Pull Request


hswong3i:
@ssbarnea @TomasTomecek may I have your comment if this approach also works for building docker image with molecule by `docker commit` the result running container during `verify.yml`, therefore overcome the difficulties that using [ansible-bender](https://github.com/ansible-community/ansible-bender) with non-docker-based and non-ansible-standard-based solution (e.g. no `import_playbook` nor `group_vars` support)?

hswong3i:
@ssbarnea sorry for my incorrect procedure, #2593 created as per requested ;-)

hswong3i:
`entrypoint` should handle in this way:
- `''` (default): reset `ENTRYPONT` as docker default `/bin/sh -c` (see https://docs.docker.com/engine/reference/run/#entrypoint-default-command-to-execute-at-runtime#entrypoint-default-command-to-execute-at-runtime)
- `~`: omit `entrypoint` during `docker_image`, preserve base image's `ENTRYPOINT` or fail as default as docker's `/bin/sh -c`
- `init`: supply to `docker_image` as per specified, useful for systemd-based image

Similar for `command`:
- `bash -c "sleep infinity"` (default): sleep for running molecule tests
- `~`: omit `command` during `docker_image`, preserve base image's `CMD`
- `echo "hello world"`: supply to `docker_image` as per specified

TomasTomecek:
> @ssbarnea @TomasTomecek may I have your comment if this approach also works for building docker image with molecule by `docker commit` the result running container during `verify.yml`, therefore overcome the difficulties that using [ansible-bender](https://github.com/ansible-community/ansible-bender) with non-docker-based (e.g. not Travis CI friendly with podman and buildah) and non-ansible-standard-based solution (e.g. no `import_playbook` nor `group_vars` support)?

I'm sorry I can't comment here, not familiar with molecule that much.

molecule/driver/docker.py
```diff
@@ -111,11 +111,21 @@ class Docker(Driver):
             buildargs:
                 http_proxy: http://proxy.example.com:8080/
 
-    If specifying the `CMD`_ directive in your ``Dockerfile.j2`` or consuming a
-    built image which declares a ``CMD`` directive, then you must set
-    ``override_command: False``. Otherwise, Molecule takes care to honour the
-    value of the ``command`` key or uses the default of ``bash -c "while true;
-    do sleep 10000; done"`` to run the container until it is provisioned.
+    Molecule takes care to honour the value of the ``entrypoint`` and
```
geerlingguy: Just a quick question; do we use British or US English in docs? ('honour' vs 'honor')

molecule/driver/docker.py
```diff
@@ -111,11 +111,21 @@ class Docker(Driver):
...
+    ``command`` keys, or defalt with ``''`` and ``bash -c "sleep infinity"``
+    to run the container until it is provisioned.
+
+    When attempting to utilize base image original ``ENTRYPOINT`` and ``CMD``
+    behavior, make sure to unset the ``entrypoint`` and ``command``. An
```
geerlingguy: Since the `entrypoint` control is a new feature, why not default it to `~` so users can choose to override it if desired, otherwise, leave it alone?

molecule/driver/docker.py
hswong3i: I just try to rewrite the comment based on original sentence structure; IMHO, vote for US English ;-)

molecule/driver/docker.py
hswong3i: Depends on how we define the default behavior: molecule style sleep infinity for running test, or out of control depends on base image default?

Here I vote for molecule default with sleep infinity, because this should be most common molecule use cases: run and test our Ansible logic, but not testing the base image original behavior.

Therefore the ‘’ (ie /bin/sh -c) and /bin/bash -c “sleep infinity” is the only combo that could make any base image running with our expect default behavior in a manageable style.

ssbarnea:
As I do not see us going towards a clear agreement on what should be default, I will say that best approach would be not to change current behavior, so unless updated I will close it.

Others are invited to send feedback on this, it find quite hard to refuse contribution, especially those tha had a lot of invested effort.

hswong3i:
@ssbarnea may I have a clarify on which combination looks good as not to change current behavior? I will update the code as per recommendation.

ssbarnea:
@hswong3i Thanks you. Mainly my concern is that this is the most popular driver and a change an behaviour would create havoc among current users.

PR #2592, opened 2020-02-28
Define reusable molecule jobs

ssbarnea:
These jobs will be reused by molecule plugins, avoiding the need to define them there.


ssbarnea:
recheck

PR #2608, opened 2020-03-12
Remove use of distutils

ssbarnea:
- Use shutil.which() which is a replacement for deprecated
find_executable.
- Use packaging version parsing

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,3 +41,5 @@ repos:
         args: []
         entry: mypy molecule/
         pass_filenames: false
+        additional_dependencies:
+          - packaging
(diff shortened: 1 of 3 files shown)


ssbarnea:
Closing to mark I am not working on this. Apparently building the snap with multipass on macos or centos-8 seems to be a permanent source of frustration. Probably I will give it another try in a couple of months.

Saviq:
Hi @ssbarnea, the last output from Multipass is:

```
2020-04-08 08:31:19.656890 | ubuntu-bionic | Please ensure multipassd is running and '/var/snap/multipass/common/multipass_socket' is accessible
```

Can you add a check for that? `ls -l` that path, and also `journalctl -u snap.multipass.*`?

Saviq:
@ssbarnea: `ERROR! Syntax Error while loading YAML.`

Saviq:
@ssbarnea we need the journal output… it looks like Multipass failed to start:
```
ls: cannot access '/var/snap/multipass/common/multipass_socket': No such file or directory
```

Saviq:
@ssbarnea this suggests the user running it is not in the `sudo` group:
```
Please check that you have read/write permissions to '/var/snap/multipass/common/multipass_socket'
```

PR #2612, opened 2020-03-17
Increase timeout on molecule-tox-packaging

ssbarnea:
Packaging can take a lot of time when it tries to verify packaging on lots of containerized platforms.


ssbarnea:
recheck

PR #2615, opened 2020-03-20
Add support for memory and kernel_memory module parameters

guerzon:
Signed-off-by: Lester Guerzon <lesterguerzon@pm.me>

This pull request adds support for the `memory` and `kernel_memory` parameters of `docker_container`, allowing users to specify memory limits for platforms using the docker driver.

#### PR Type

- Feature Pull Request

### DIFF
--- a/molecule/provisioner/ansible/playbooks/docker/create.yml
+++ b/molecule/provisioner/ansible/playbooks/docker/create.yml
@@ -121,7 +121,6 @@
         hostname: "{{ item.hostname | default(item.name) }}"
         image: "{{ item.pre_build_image | default(false) | ternary('', 'molecule_local/') }}{{ item.image }}"
         pull: "{{ item.pull | default(omit) }}"
-        kernel_memory: "{{ item.kernel_memory | default(omit) }}"
         memory: "{{ item.memory | default(omit) }}"
         state: started
         recreate: false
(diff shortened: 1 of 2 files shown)


molecule/provisioner/ansible/playbooks/docker/create.yml
```diff
@@ -121,6 +121,8 @@
         hostname: "{{ item.hostname | default(item.name) }}"
         image: "{{ item.pre_build_image | default(false) | ternary('', 'molecule_local/') }}{{ item.image }}"
         pull: "{{ item.pull | default(omit) }}"
+        kernel_memory: "{{ item.kernel_memory | default(omit) }}"
```
geerlingguy: https://docs.docker.com/engine/reference/run/#kernel-memory-constraints#kernel-memory-constraints for reference. I've never had to touch kernel_memory, but having memory configurable is nice for some environments/tests.

guerzon: I only added `kernel_memory` as a "good to have"; i only really need `memory`. My use-case does not require `kernel_memory` so I will not have a chance to test it. Hence, removing it in this MR.

Some relevant docker warning:
`Docker warning: You specified a kernel memory limit on a kernel older than 4.0. Kernel memory limits are experimental on older kernels, it won't work as expected and
can cause your system to be unstable.`

ssbarnea:
recheck

ssbarnea:
recheck

Issue #2621, opened 2020-03-24
molecule init scenario with delegated creates broken create and destroy playbooks


ssbarnea:
It seems that this basic use-case got broken during the migration to v3:
```
molecule init role foo -d delegated
cd foo

--> Found config file /Users/ssbarnea/.config/molecule/config.yml
--> Test matrix

└── default
    ├── dependency
    ├── create
    └── prepare

--> Scenario: 'default'
--> Action: 'dependency'
Skipping, missing the requirements file.
--> Scenario: 'default'
--> Action: 'create'

    PLAY [Create] ******************************************************************

    TASK [Populate instance config dict] *******************************************
    Tuesday 24 March 2020  13:28:10 +0000 (0:00:00.020)       0:00:00.020 *********
fatal: [localhost]: FAILED! => {"msg": "'server' is undefined"}

    PLAY RECAP *********************************************************************
    localhost                  : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0

    Tuesday 24 March 2020  13:28:10 +0000 (0:00:00.289)       0:00:00.310 *********
    ===============================================================================
    Populate instance config dict ------------------------------------------- 0.29s

ERROR:
FAIL: 2
```

Same happens with destroy command.

The situation is worsened by the fact that even if you remove the create and destroy playbooks, you endup with an error like:
```
ERROR! the playbook: /Users/ssbarnea/c/opendev/ffff/molecule/default/destroy.yml could not be found
```

## Desired behavior
* presence of create and destroy playbooks should be optional when using delegated driver
* when using init we should always assure that if we produce create or destroy playbooks, they are not broken.

PR #2623, opened 2020-03-24
Avoid errors with delegated create and destroy templates

ssbarnea:
* Assure delegated templates can succesfully run
* Make create and destroy playbooks optionals, if they miss only
  a warning will be triggered.

Fix: #2621


ssbarnea:
recheck

PR #2625, opened 2020-03-24
Fix Getting Started docs as default verifier is now Ansible not testinfra

matthewcosgrove:
Fix Getting Started docs as default verifier is now Ansible not testinfra

#### PR Type

- Docs Pull Request

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -65,7 +65,7 @@ directories:
 .. code-block:: bash
 
     $ ls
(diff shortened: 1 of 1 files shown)


matthewcosgrove:
Updated and squashed the commits

Ran `tox -e lint` which I found in [https://github.com/ansible-community/molecule/blob/master/docs/testing.rst](https://github.com/ansible-community/molecule/blob/master/docs/testing.rst)

ssbarnea:
recheck

PR #2631, opened 2020-03-29
Paralelize host_vars

tadeboro:
When Molecule is executed with the --parallel switch, it transforms the platform names in order to avoid name collisions on the service that is hosting the tests.

Currently, Molecule only parallelizes platform names and does not touch the inventory section at all. This makes it impossible to specify per-platform variables in the molecule.yml file.

For example, let us assume that we have this molecule.yml file:

    platforms:
      - name: my_platform
        # other stuff here

    provisioner:
      inventory:
        host_vars:
          my_platform:
            my_var: my_value

If we run such a scenario, the playbook tasks have access to the my_var variable. But as soon as we add the --parallel flag, this variable is no longer accessible because the host names from the platforms section and the provisioner sections do not match.

Changes in this PR ensure that when we parallelize platform names, we also parallelize any host_vars entries that match the original name. All other host_vars entries (the ones that do not match any of the platform names) are left alone and are not transformed.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/molecule/platforms.py
+++ b/molecule/platforms.py
@@ -75,9 +75,7 @@ class Platforms(object):
         :return: None
         """
         if parallelize_platforms:
-            config.config["platforms"] = util._parallelize_platforms(
-                config.config, config._run_uuid
-            )
+            util.parallelize_platform_names(config.config, config._run_uuid)
         self._config = config
 
     @property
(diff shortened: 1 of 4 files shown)


ssbarnea:
Closing as it seems abandoned. I also think that this is likely the wrong direction for fixing bugs with the experimental parallel option. Molecule should take care of isolating instances but not the inventory hostnames. A "foo" host from scnario A should likely not be same as one from scenario B, even if at runtime we should continue to see the same platforms.

PR #2634, opened 2020-03-31
added example for gitlab-ci using virtenv and tags, requested in issu…

belfast77:
Adding an example to the CI doc for gitlab-ci.

This example uses virtualenv and tags.

Tags are used to be sure the job is picked up by a runner capable of executing it.

This example has been tested with a Centos7 Runner against GitLab Community Edition 12.6.2

This is in answer to the request https://github.com/ansible-community/molecule/issues/2493

#### PR Type
- Docs Pull Request

### DIFF
--- a/docs/ci.rst
+++ b/docs/ci.rst
@@ -114,6 +114,43 @@ Here is an example using Docker in Docker
       script:
         - cd roles/testrole && molecule test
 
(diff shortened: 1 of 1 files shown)


belfast77:
That seems to have worked, apologies for not squashing commits before.

ssbarnea:
There is no need to squash commits because this is what Zull will do anyway when merging.

PR #2636, opened 2020-03-31
Add podman to bindep

TristanCacqueray:
#### PR Type

- Bugfix Pull Request


ssbarnea:
No longer needed as we install it with zuul.

PR #2639, opened 2020-04-02
Implemented reset for docker driver

ssbarnea:
From now on reset command will clean all docker containers and networks
that are managed my molecule, label owner=molecule

We also removed network creation from docker driver testing in order
to avoid concurrency issues.

### DIFF
--- a/molecule/command/reset.py
+++ b/molecule/command/reset.py
@@ -23,6 +23,7 @@ import click
 
 from molecule import logger
 from molecule.command import base
+from molecule.api import drivers
 
 
 LOG = logger.get_logger(__name__)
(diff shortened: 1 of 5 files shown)


ssbarnea:
> Didn't test, but everything looks good at a glance. I like this funcationality, because every once in a while, I end up in a situation where I have to manually blow away a bunch of molecule stuff if things get out of sync when testing multiple scenarios and multiple projects.

You described very well the reason why I ended up writing this today. I did not have time to implement the same functionality for other drivers, but we can always do this in follow-ups.

ssbarnea:
@cloudnull You forgot to submit that note? Github fails to show me any comment on code  🤔

PR #2640, opened 2020-04-06
Correct typo in collection argument roles-file -> role-file

greg-hellings:

#### PR Type

- Bugfix Pull Request

This fixes a typo in #2609 that causes installs to fail when "role-file" is specified.


ssbarnea:
recheck

PR #2642, opened 2020-04-07
Removed --driver-template option

ssbarnea:
Removed `--driver-template` option from molecule init commands as it did add more complexity without a tangible outcome. Those still interested about this feature can use cookiecutter directly.


ssbarnea:
recheck

PR #2645, opened 2020-04-08
Add memory-swap to create docker driver

ElfoLiNk:
Add memory-swap parameter to docker driver

https://docs.docker.com/config/containers/resource_constraints/#--memory-swap-details

If --memory and --memory-swap are set to the same value, this prevents containers from using any swap. This is because --memory-swap is the amount of combined memory and swap that can be used, while --memory is only the amount of physical memory that can be used.

- Feature Pull Request


ssbarnea:
Passing parameters to the docker module seems to be a never ending source of PR, maybe is time to refactor implementation and find a way to pass all values defined in molecule file to the module, avoiding the need to update molecule code every time these modules are updated.

PR #2655, opened 2020-04-16
cross-platform development environment in docker

kevingentile:
The following changes implement a cross-platform development environment using Docker. Developers can build and spin up the image defined in `Dockerfile.dev`, mount the molecule source-code and develop inside the container.

### Usage

```
cd molecule
docker-compose up -d
# select molecule container in vscode remote explorer OR
docker-compose exec dev /bin/bash
# OR
docker exec -it molecule_dev_1 /bin/bash
```

See:
 * [Container development in Visual Studio Code](https://code.visualstudio.com/docs/remote/containers).

#### PR Type
- Feature Pull Request

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -3,7 +3,8 @@ services:
   dev:
     build:
       context: .
-      dockerfile: tools/Dockerfile.dev
     volumes:
       - "./:/etc/molecule"
+      - "./tools/entrypoint.sh:/usr/local/bin/entrypoint.sh"
     privileged: true
+    entrypoint: entrypoint.sh
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,9 @@
+version: "3.7"
+services:
+  dev:
+    build:
+      context: .
+      dockerfile: tools/Dockerfile.dev
+    volumes:
+      - "./:/etc/molecule"
+    privileged: true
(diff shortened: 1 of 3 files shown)


ssbarnea:
We already have a Dockerfile for building molecule in the root. Why you are not able to re-use that one?

Please use global .gitignore for user-specific files (aka editor specific ones). While I use vscode myself and like it, allowing it to be added would open a pandora box of other excludes. For example OpenStack project has a strict policy regarding what can get into gitignores, for the same reasons: to avoid noise. I kinda like that policy.

kevingentile:
> We already have a Dockerfile for building molecule in the root. Why you are not able to re-use that one?

I'm playing with this some more now. The current image doesn't start the Docker daemon so we would still require an alternate entrypoint script to do so. Additionally, we will still want to mount the source-code and do a `pip install -e .` .

I see 3 approaches to the divergence from the Dockerfile in the root:

1. modify the existing Dockerfile to include an alternate entrypoint script that can be targeted in the docker-compose file
1. include a separate development Dockerfile with the alternate entrypoint extending public latest image (not possible until #2656)
1. mount and target an alternate entrypoint in the docker-compose configuration 

> Please use global .gitignore for user-specific files (aka editor specific ones). While I use vscode myself and like it, allowing it to be added would open a pandora box of other excludes. For example OpenStack project has a strict policy regarding what can get into gitignores, for the same reasons: to avoid noise. I kinda like that policy.

Thank you, this is a feature I was unaware of and will certainly help me moving forward. I've removed the entry.

kevingentile:
With the latest revision the user can execute the dev container through:

docker-compose run --rm dev /bin/bash

ssbarnea:
Closing as it seems not to pass CI/CD, lacks updates and it does not appear to be clear what it solves. We can reopen if needed.

PR #2659, opened 2020-04-20
Avoid using pre-release linters

ssbarnea:
We should not use pre-release dependencies.


ssbarnea:
Obsoleted by https://github.com/ansible-community/molecule/pull/2663

Issue #2657, opened 2020-04-17
Missing cookiecutter template option for init


kevingentile:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version
ansible 2.9.6
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.2 (default, Feb 29 2020, 17:03:31) [GCC 9.2.0]
molecule 3.0.4.dev4+g21ce2f4d
   ansible==2.9.6 python==3.8
```

Molecule installation method (one of):

- source

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

# Desired Behavior

I'm attempting to create my own cookiecutter template to configure new roles. Ansible Galaxy provides users with a [role skeleton](https://docs.ansible.com/ansible/2.7/reference_appendices/galaxy.html#using-a-custom-role-skeleton). I am now attempting to further configure the molecule generated files.
 
I expected to be able to pass my custom cookiecutter configuration to molecule either through a flag or an environment variable.

# Actual Behaviour
The documentation for `molecule init scenario` [mentions a local cookiecutter template](https://github.com/ansible-community/molecule/blob/1563796089710b3e0505acb3ec8f57ba80834c45/molecule/command/init/scenario.py#L57). The scenario command [does a lookup for "driver_template" as a command argument](https://github.com/ansible-community/molecule/blob/1563796089710b3e0505acb3ec8f57ba80834c45/molecule/command/init/scenario.py#L92) but this value doesn't appear to ever be set in the project. Molecule then defaults to the local ([project owned](https://github.com/ansible-community/molecule/tree/master/molecule/cookiecutter)) cookiecutter template.

This appears to have been addressed in #2117. Am I missing something?

nikParasyr:
Hi,

I think the functionality of this was removed with: https://github.com/ansible-community/molecule/pull/2642

I think that the idea is that if you have a cookiecutter template you can directly use it to create it your role layout, instead of using ` molecule init role/scenario`

kevingentile:
@nikParasyr thanks for the link. 

I see how introducing cookicutter would add additional complexity to the project. 
However, given that it _is_ still a part of the project, would it make more sense to remove this feature with the remaining cookiecutter integration(s)?

nikParasyr:
[...]

PR #2665, opened 2020-04-21
WIP: add cookiecutter template option

kevingentile:
Addresses #2657

#### PR Type

- Feature Pull Request

Allow users to define a cookiecutter template for configuring a molecule scenario

```
molecule init -t ~/mycookiecutter role ~/testrole
cd ~/testrole
molecule init -t ~/mycookiecutter scenario customscenario
``` 

Here we assume the cookiecutter directory structure reflects that of the [default](https://github.com/ansible-community/molecule/tree/master/molecule/cookiecutter) from the "molecule" directory down with changes only to `molecule.yml`. Do we want to make this available as an environment variable?

### DIFF
--- a/molecule/command/init/init.py
+++ b/molecule/command/init/init.py
@@ -19,6 +19,8 @@
 #  DEALINGS IN THE SOFTWARE.
 """Base class used by init command."""
 
+import click
+
 from molecule.command import base
 from molecule import logger
 from molecule.command.init import role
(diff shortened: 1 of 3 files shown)


ssbarnea:
There are no plans to reinstate the custom template functionality in molecule, mainly because you can always create your own cookiecutter templates that do not rely on molecule itself.

PR #2666, opened 2020-04-23
Allow drivers to overwrite template

nikParasyr:
Related to https://github.com/ansible-community/molecule-openstack/issues/3

Drivers can overwrite the cookiecutter template of molecule to define driver specific configuration. This allows "external" drivers to specify options that are driver specific, independently of molecule. 
For example the openstack needs to define flavor, image per vm that is booted ( ec2 same for instance type).

There is the alternative of populating the molecule template with external driver specific but i think that it's not the best way of addressing this.  

I've tested this with openstack/ec2/docker/podman drivers. Docker and podman templates are not affected, openstack/ec2 drivers either, but they will be able to change them in the future. 

#### PR Type
- Feature Pull Request

### DIFF
--- a/molecule/command/init/role.py
+++ b/molecule/command/init/role.py
@@ -86,11 +86,11 @@ class Role(base.Base):
             api.drivers()[self._command_args["driver_name"]].template_dir(),
             api.verifiers()[self._command_args["verifier_name"]].template_dir(),
         ]
+        self._process_templates("molecule", self._command_args, role_directory)
         for template in templates:
             self._process_templates(
                 template, self._command_args, scenario_base_directory
             )
-        self._process_templates("molecule", self._command_args, role_directory)
 
         role_directory = os.path.join(role_directory, role_name)
         msg = "Initialized role in {} successfully.".format(role_directory)
--- a/molecule/command/init/scenario.py
+++ b/molecule/command/init/scenario.py
@@ -106,11 +106,11 @@ class Scenario(base.Base):
             driver_template,
             api.verifiers()[self._command_args["verifier_name"]].template_dir(),
         ]
+        self._process_templates("molecule", self._command_args, role_directory)
         for template in templates:
             self._process_templates(
                 template, self._command_args, scenario_base_directory
             )
-        self._process_templates("molecule", self._command_args, role_directory)
 
         role_directory = os.path.join(role_directory, role_name)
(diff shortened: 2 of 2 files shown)


decentral1se:
Hey cool, I need this too. Thanks for working on it.

> I've tested this 

Can you explain how? I'm not quite clear how I can also test it.

nikParasyr:
> I've tested this

Maybe that was a bit of an over-statement. What I did was:
- For docker/podman/openstack/ec2 drivers check that this change does not affect the role/scenario initiliazation 
- For openstack/ec2 checked that i can add a molecule.yml on the cookiecutter templates that overwrites the one defined by molecule

If you think we should add unit/functional tests, I can work on it. I might need some help though.

decentral1se:
> For openstack/ec2 checked that i can add a molecule.yml on the cookiecutter templates that overwrites the one defined by molecule

Ah! I get it now, thanks for clarifying.

> If you think we should add unit/functional tests, I can work on it. I might need some help though.

If you can, that'd be great. There is only one maintainer here and if there isn't a test added then some hours could be lost down the line... I guess adding integration tests isn't the horrendous nightmare it used to be :)

/cc @ssbarnea

decentral1se:
Alternatively, I will be using this functionality and we could merge and hash out any fixes through usage :cowboy_hat_face:

ssbarnea:
Because we use Zuul, with very little efforth, I should be able to trigger jobs that tests some of the plugins. Maybe I should do this with vagrant and libvirt ones which do not require a cloud to run.

For the cloud ones it will take more time because I never ended up setting credentials for them in zuul.

nikParasyr:
> Alternatively, I will be using this functionality and we could merge and hash out any fixes through usage 

Openstack is another driver that will.

> Because we use Zuul, with very little efforth, I should be able to trigger jobs that tests some of the plugins.

That would be really nice. I had troubles how to test this only within molecule. This will solve it. (an alternative would be to add a dummy driver on molecule just to be able to test functionality like this)

> For the cloud ones it will take more time because I never ended up setting credentials for them in zuul.

In the scope of this PR, I don't think you actually need to deploy. I would say it is on each external driver that chooses to use this functionality to ensure that molecule.yml (and other files that it overwrites) are valid, with tests etc on the external-driver site

PR #2677, opened 2020-04-28
Add upper bound for sh dependency

decentral1se:
Fixes: https://github.com/ansible-community/molecule/issues/2676.


decentral1se:
I've updated the bound to account for the new `sh` release.

The urgency of a required release is not really an issue now btw.

robinmulder:
Until there is a proper fix, this should probably be:

```
sh >= 1.12.14, < 1.13
```

decentral1se:
Let's go with https://github.com/ansible-community/molecule/pull/2679.

PR #2680, opened 2020-04-29
Adds external plugins to the bundle dockerfile

ssbarnea:
Adds known drivers to molecule Dockerfile image and adds a minimal verification.


decentral1se:
This awaits a next release, right? I see https://quay.io/repository/ansible/molecule?tab=tags has 3.0.4.

PR #2682, opened 2020-04-29
Fix environment handling for sh.py >= 1.13

apatard:
Since sh.py 1.13, a new check on the environment variables has been added
with commit https://github.com/amoffat/sh/commit/5b274778fb4be647bbb95f9bec14153f186bcc1f.

This leads to failures in the testsuite since the code is passing
os.environ, which is not a dict().

It's also breaking MOLECULE_ENV_FILE handling with errors like (from the testsuite):

E           TypeError: Invalid special arguments:
E
E             'env': value None of env key 'MOLECULE_ENV_FILE' must be a str

so:
- use dict() to convert os.environ to a dict in _rebake_command(),
- use str() when defining MOLECULE_ENV_FILE to make sure self.env_file is a string.

To avoid any other surprise with 1.14 or newer, I've changed the sh.py requirement
to < 1.14. The requirement is not 1.13 but 1.13.1 to get the fix for sh.py bug #520
(https://github.com/amoffat/sh/issues/520).

#### PR Type

- Bugfix Pull Request


ssbarnea:
recheck

PR #2685, opened 2020-05-02
Treat warnings as errors during testing

ssbarnea:
Harden Molecule own testing by treating any runtime warnings as errors.


ericsysmin:
So, at this point, we're safe to assume and mandate molecule supports Python 3.6+ only, and are no longer testing on Python <3.6?

ssbarnea:
@ericsysmin That is correct, in fact these are only leftovers from the main removal.

ssbarnea:
recheck

PR #2689, opened 2020-05-06
docs: Dockerfile no longer created by default

geerlingguy:
#### PR Type

- Bugfix Pull Request

The getting started guide listed a `Dockerfile.j2` file as being scaffolded when using `molecule init role`, but that is no longer the case, after https://github.com/ansible-community/molecule/commit/a74b2f3e24ffa3f61fde3253470856c2e13aaabc was added, and there was no longer a need for the lengthy process of setting up the custom local Docker image.

This PR just removes that from the getting-started section of the docs since I was a little nervous that my local scaffold didn't have a Dockerfile.j2.

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -63,11 +63,7 @@ directories:
 .. code-block:: bash
 
     $ ls
-    Dockerfile.j2  INSTALL.rst  molecule.yml  converge.yml  verify.yml
-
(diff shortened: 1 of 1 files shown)


ericsysmin:
@geerlingguy maybe we should propose some documentation examples of a Dockerfile.j2 that supports non-systemd and systemd containers? At least having something there for reference is better than nothing. We've all noticed that assuming a docker container has the required packages for Ansible etc to work is sometimes faulty. Like the assumption, a docker container has python 3.x installed.

Issue #1879, opened 2019-03-26
[docker_network] Allow more options


rockandska:
# Issue Type

- Feature request

# Desired Behavior

For now it seems impossible to use `docker_network` options other than `name` ([create.yml#L51-L56](https://github.com/ansible/molecule/blob/master/molecule/provisioner/ansible/playbooks/docker/create.yml#L51-L56))

Would be handy to have the other options available.

For example, i want to test the availability of a VIP (deployed with haproxy / keepalived roles) and I need to be sure that the VIP i test will not be taken by a container, so to avoid this, i fix the network/ip range used by the network dedicated to the tests.
The only workaround i've found for now is to update the network in "prepare.yml" and restart the attached containers but will be better to have all configuration in one place :)

```
- name: Update docker network
  hosts: localhost
  gather_facts: false
  become: false
  tasks:
    - name: "Update docker network: my_docker_network"
      docker_network:
        name: "{{ item }}"
        docker_host: "{{ item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
        ipam_options:
          subnet: 172.28.0.0/16
          iprange: 172.28.5.0/24
          gateway: 172.28.5.254
        force: true
      with_items: "{{ molecule_yml.platforms | molecule_get_docker_networks | unique }}"
      when: item == 'my_docker_network'

    - name: "Restart containers"
      docker_container:
        name: "{{ item.name }}"
        docker_host: "{{ item.docker_host | default(lookup('env', 'DOCKER_HOST') or 'unix://var/run/docker.sock') }}"
        state: started
        restart: true
        networks:
          - name: my_docker_network
        purge_networks: false
      with_items: "{{ molecule_yml.platforms }}"
```

Regards,

decentral1se:
Fair enough! Just after seeing https://github.com/ansible/molecule/issues/1878 and now thinking that we should really take a moment to extend these tasks with the full module options so as to avoid these missing functionality ...

PR #2696, opened 2020-05-11
Allow configuration of docker_network in platforms

actatux:
Map a list of docker_networks to [Ansible's docker_network module](https://docs.ansible.com/ansible/latest/modules/docker_network_module.html). This
change allows to define additional parameters, while the previous code
only allowed to define the name of the Docker network.

Keep backward compatibility if a network name is defined for the
platform, without a docker_network definition.

Example:
```
platforms:
  - name: instance1
    image: docker.io/pycontribs/centos:8
    docker_networks:
      - name: foo
        ipam_config:
          - subnet: "10.11.0.0/16"
            gateway: "10.11.0.254"
    networks:
      - name: foo
        ipv4_address: "10.11.0.1"
  - name: instance2
    image: docker.io/pycontribs/centos:8
    networks:
      - name: foo
        ipv4_address: "10.11.0.2"
  - name: instance3
    image: docker.io/pycontribs/centos:8
    networks:
      - name: bar
```
This will create 2 Docker networks `foo` and `bar`.
Docker network `foo` is created with a subnet 10.11.0.0/16 so we can define IP addresses for instance 1 and 2 and attach these interfaces to this network.
Docker network `bar` is created as previously.

Fixes: #1879

### DIFF
--- a/molecule/driver/docker.py
+++ b/molecule/driver/docker.py
@@ -41,7 +41,11 @@ class Docker(Driver):
     Molecule leverages Ansible's `docker_container`_ module, by mapping
     variables from ``molecule.yml`` into ``create.yml`` and ``destroy.yml``.
 
-    .. _`docker_container`: https://docs.ansible.com/ansible/latest/docker_container_module.html
+    Molecule leverages Ansible's `docker_network`_ module, by mapping variable
+    ``docker_networks`` into ``create.yml`` and ``destroy.yml``.
+
+    .. _`docker_container`: https://docs.ansible.com/ansible/latest/modules/docker_container_module.html
+    .. _`docker_network`: https://docs.ansible.com/ansible/latest/modules/docker_network_module.html
     .. _`Docker Security Configuration`: https://docs.docker.com/engine/reference/run/#security-configuration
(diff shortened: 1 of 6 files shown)


actatux:
Related to #1879.

actatux:
I only consider `labels` and `state` as internal defaults.
`docker_host`, `cacert_path`, `cert_path`, `key_path` and `tls_verify` were defaulting to [docker_network](https://docs.ansible.com/ansible/latest/modules/docker_network_module.html)'s default values, so I removed these.

actatux:
Note the warning with this change:
```
    TASK [Delete docker network(s)] ************************************************
[WARNING]: Using a variable for a task's 'args' is unsafe in some situations
(see https://docs.ansible.com/ansible/devel/reference_appendices/faq.html#argsplat-unsafe)
    changed: [localhost] => (item=ice1-1)
    changed: [localhost] => (item=bar)
    changed: [localhost] => (item=ice2-1)

```

actatux:
Hello @ssbarnea - does the new code answer your change request?

ssbarnea:
recheck

PR #2701, opened 2020-05-13
Clarify Docker bind-mount

bradshjg:
Change language clarifying how Molecule can start sibling containers. In my experience, Docker-in-Docker generally refers to the build available at https://hub.docker.com/_/docker used for Docker development. This _may_ be a bit of a nit though 😄.

#### PR Type

- Docs Pull Request


ssbarnea:
recheck

PR #2716, opened 2020-06-07
Support root containers for podman

### LINKED ISSUE #2713: Add a config option to run podman as root
# Issue Type

- Feature request

```
ansible 2.9.9
molecule 3.0.4
```

- pip

- OS package

# Desired Behavior

Podman allows running containers in a rootless mode, although this introduces some limitations.
For example this makes it difficult to test ansible roles which rely on a working systemd instance running in the container.

It would therefore be nice if it would be possible to specify whether podman should be executed as root or as the current user in the molecule config.

sshnaidm:
By adding "rootless: false" support root containers for podman
driver.
#### PR Type

- Feature Pull Request

Fixes #2713

### DIFF
--- a/molecule/model/schema_v3.py
+++ b/molecule/model/schema_v3.py
@@ -358,6 +358,7 @@ platforms_podman_schema = {
                 "cgroup_manager": {"type": "string"},
                 "storage_opt": {"type": "string"},
                 "storage_driver": {"type": "string"},
+                "rootless": {"type": "boolean"},
             },
         },
     }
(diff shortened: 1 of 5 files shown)


ssbarnea:
recheck

molecule/model/schema_v3.py
```diff
@@ -355,6 +355,7 @@ def pre_validate_base_schema(env, keep_string):
                 "network": {"type": "string"},
                 "cert_path": {"type": "string"},
                 "tls_verify": {"type": "boolean"},
+                'rootless': {'type': 'boolean'},
```
ssbarnea, line 358: I bet this will be refused by black and converted to double quotes.

PR #2722, opened 2020-06-11
Fixes around ansible-devel testing

ssbarnea:
Makes additional changes in order to test the new packaging while is developed
by @abadger

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,21 @@
 # This is a multi-stage build which requires Docker 17.05 or higher
-FROM alpine:edge as molecule-builder
+FROM docker.io/alpine:edge as molecule-builder
 
 WORKDIR /usr/src/molecule
 
-ENV PACKAGES="\
-gcc \
-git \
-libffi-dev \
-make \
-musl-dev \
-openssl-dev \
+# testing needed for: py3-arrow py3-tabulate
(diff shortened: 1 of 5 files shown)


felixfontein:
You should make clear whether you are talking about ansible-base's `devel` branch, or about ansible (as in ACD) - which doesn't have anything called "devel" AFAIK.

cloudnull:
> You should make clear whether you are talking about ansible-base's `devel` branch, or about ansible (as in ACD) - which doesn't have anything called "devel" AFAIK.

While I approved the change, I too would like more details in the commit message regarding this change.

gundalow:
Github.com/ansible/ansible = `ansible-base`
No such thing as ACD, it's called the (new) `ansible` package

abadger:
Note that this will only work for a couple weeks.  I'll be removing the fedora people files once ansible-base 2.10 beta 1 and ansible-2.10 aloha 1 are available on pypi

ssbarnea:
> Note that this will only work for a couple weeks. I'll be removing the fedora people files once ansible-base 2.10 beta 1 and ansible-2.10 aloha 1 are available on pypi

We will update it asap we have content on pypi. A termporary working hack is better then a broken CI/CD pipeline. I am afraid to keep the jobs disabled for a long as we may endup with code merged that break on 2.10

ssbarnea:
recheck

PR #2728, opened 2020-06-18
Remove hardcoded test files for docker and podman

ssbarnea:
Removes test files that where used during testing and which duplicated
embedded molecule ones. Keeping them in sync seems something we failed to do and is better to avoid duplicating them.


ssbarnea:
recheck

PR #2730, opened 2020-06-21
Adopt sphinx_ansible_theme

ssbarnea:
- use [sphinx_ansible_theme](https://github.com/ansible-community/sphinx_ansible_theme)


ssbarnea:
@webknjaz Please take a look at the outcome of this. We can use molecule as early adopter for the shared theme and once we are done with it we can do the same for the linter.

If you spot a bug in the theme, raise it directly to its repository. Thanks.

PR #2732, opened 2020-06-23
Set correct default directory for tests

jyoder-deloitte:
The default test directory created by verifiers is `molecule/default/tests` but the base `Verifier` class defaulted to `molecule/default/molecule`.  The `testinfra.py` verifier was overriding the `directory` property to correct the bug in the base class, but other spun-off verifiers aren't doing that (e.g. `molecule-goss` and `molecule-inspec`).  This change makes all the non-Ansible verifiers work again.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/molecule/test/unit/verifier/test_ansible.py
+++ b/molecule/test/unit/verifier/test_ansible.py
@@ -57,7 +57,7 @@ def test_enabled_property(_instance):
 
 def test_directory_property(_instance):
     parts = _instance.directory.split(os.path.sep)
-    # default verifier is ansible, which keeps tests in molecule folder.
+    # The ansible verifier expects directory to end in molecule
     assert "molecule" == parts[-1]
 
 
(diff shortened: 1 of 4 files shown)


ssbarnea:
I am not sure what to think about this because I kinda prefer to see the default value to be what we want for the default verifier which is Ansible.

I incline a bit against it, but I will allow others to express their preference before making a decision.

jyoder-deloitte:
@ssbarnea I couldn't actually find any refence to `directory` in the Ansible verifier.  I only left the check on the off chance I missed something.  The Ansible verifier uses the `verify.yml` playbook.  I actually didn't need to change the property in `ansible.py` at all to make the verifier work, I just had to change it for the test.

jyoder-deloitte:
After more checking the `verifier.directory` attribute is definitely not used by the Ansible verifier.  Only the `scenario.directory` attribute to find the `verify.yml` playbook. Only the non-Ansible verifiers use the `directory` property. I've removed the unnecessary test and the override of the attribute in `ansible.py`.

PR #2737, opened 2020-06-30
Add mol command alias

ssbarnea:
Makes it easier to run molecule by also installing a `mol` script. `mol` is the international abbreviation for molecule and it seems not to be used by any other known tool.

Using the abbreviated name can be proved useful not only for terminal usage but also when naming jobs in CI/CD pipelines, as their names could easily become very long.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,6 +46,14 @@ Molecule supports only the latest two major versions of Ansible (N/N-1),
 meaning that if the latest version is 2.9.x, we will also test our code with
 2.8.x.
 
+Once installed, the command line can be called using any of the methods below:
+
+.. code-block:: bash
+
+    molecule ...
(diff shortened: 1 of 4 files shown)


mperry2:
I recommend reverting this commit. We should not pollute the path with duplicate executables. This is exactly what aliases exist for, and users can easily create an alias if they wish to use `mol` to call `molecule`.

tadeboro:
This PR has been part of Molecule for almost a year and a half now and removing the alias would constitute a backward-incompatible API break. I do agree that the `mol` alias is not really needed (I did not even know it existed and I use Molecule a lot), but removing it is not as simple as reverting the commit. We should at least print the deprecation warning for some time before removing it. But this is probably quite a low-priority task since there are other more urgent issues in the tool that we need to solve.

PR #2745, opened 2020-07-10
Correct ANSIBLE_COLLECTIONS_PATH

ssbarnea:
Respect conversion to singular made by
https://github.com/ansible/ansible/pull/70007

### DIFF
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -21,6 +21,7 @@
 
 from uuid import uuid4
 import os
+import pkg_resources
 
 from ansible.module_utils.parsing.convert_bool import boolean
 
@@ -101,6 +102,22 @@ class Config(object, metaclass=NewInitCaller):
 
     def write(self):
         util.write_file(self.config_file, util.safe_dump(self.config))
+
+    @property
+    def ansible_version(self):
+        """Return current version of ansible."""
+        return pkg_resources.parse_version(
+            pkg_resources.get_distribution("ansible").version
+        )
+
+    @property
+    def ansible_collections_path(self):
+        """Return collection path variable for current version of Ansible."""
+        # https://github.com/ansible/ansible/pull/70007
+        if self.ansible_version >= pkg_resources.parse_version("2.10.0.dev0"):
+            return "ANSIBLE_COLLECTIONS_PATH"
+        else:
+            return "ANSIBLE_COLLECTIONS_PATHS"
 
     @property
     def config_file(self):
--- a/molecule/dependency/ansible_galaxy/collections.py
+++ b/molecule/dependency/ansible_galaxy/collections.py
@@ -36,7 +36,7 @@ class Collections(AnsibleGalaxyBase):
     def default_env(self):
         general = super(Collections, self).default_env
         return util.merge_dicts(
-            general, {"ANSIBLE_COLLECTIONS_PATHS": self.install_path}
+            general, {self._config.ansible_collections_path: self.install_path}
         )
 
     @property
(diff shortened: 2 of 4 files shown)


molecule/config.py
```diff
@@ -102,6 +103,20 @@ def after_init(self):
...
+    @property
+    def ansible_version(self):
+        """Return current version of ansible."""
+        return pkg_resources.get_distribution("ansible").version
+
+    @property
+    def ansible_collections_path(self):
+        """Return collection path variable for current version of Ansible."""
+        # https://github.com/ansible/ansible/pull/70007
+        if self.ansible_version >= "2.10":
```
felixfontein, line 115: That is not how you should compare version numbers. Better use `LooseVersion` (`from distutils.version import LooseVersion`) to compare version numbers.

ssbarnea: Fixed using more modern version from packaging instead of outdated distutils.

felixfontein: If you already have `packaging` as a requirement, that's better. If you don't, you need to add it as a requirement.

felixfontein: You don't seem to have it in your dependencies (https://github.com/ansible-community/molecule/blob/master/pyproject.toml#L3-L7), so you need to add it (https://pypi.org/project/packaging/).

setup.cfg
```diff
@@ -76,6 +76,7 @@ install_requires =
     cookiecutter >= 1.6.0, != 1.7.1
     python-gilt >= 1.2.1, < 2
     Jinja2 >= 2.10.1
+    packaging
```
sshnaidm, line 79: better to use `from distutils.version import LooseVersion`
no need to add dependencies
It works in `podman_container` like:
```python
if minv and LooseVersion(minv) > LooseVersion(self.podman_version):
   ...
```
https://github.com/sshnaidm/ansible-podman-collections/blob/8bfe7402f02dea320d116278739180b5c5c85a55/plugins/modules/podman_container.py#L892-L893

ssbarnea: In fact I went to pkg_resources as is newer and already used.

molecule/config.py
```diff
@@ -102,6 +103,22 @@ def after_init(self):
...
+        """Return current version of ansible."""
+        return pkg_resources.parse_version(
+            pkg_resources.get_distribution("ansible").version
+        )
+
+    @property
+    def ansible_collections_path(self):
+        """Return collection path variable for current version of Ansible."""
+        # https://github.com/ansible/ansible/pull/70007
+        if self.ansible_version >= pkg_resources.parse_version("2.10"):
```
felixfontein, line 117:
```suggestion
        if self.ansible_version >= pkg_resources.parse_version("2.10.0.dev"):
```
Otherwise `stable-2.10` currently doesn't pass this test (with version `2.10.0.dev0`).

geerlingguy:
Something in here _seems_ like it might be breaking my test repo on builds with ACD: https://github.com/geerlingguy/ansible-content-testing/issues/17

felixfontein:
@geerlingguy maybe you have `ANSIBLE_COLLECTIONS_PATHS` set, and molecule now (with this change) only sets `ANSIBLE_COLLECTIONS_PATH`, and Ansible uses paths from both (or only the former, if both are set)?

geerlingguy:
@felixfontein - Testing that assumption now. I have `PATHS`, and also tried `PATH` but neither worked individually. I'll try setting both and see if that makes any difference.

Also I opened a follow-up issue to try to get some sanity into this variable so we won't have to deal with this in 2.9 vs 2.10 for a couple years at least... https://github.com/ansible/ansible/issues/70750

PR #2748, opened 2020-07-13
Clear the APT package index files in default test images

brianhelba:
On fresh production systems, APT typically is missing its package index files, which requires an `apt-get update` or use of Ansible's `apt.update_cache` to populate.

Molecule's default test image should reflect production deployment environments to the extent possible. By providing a test image with an empty package cache, it ensures that roles under test must consider this case.


ssbarnea:
@brianhelba Thanks. Really good point. I hope to see more patches from you. Maybe you also want to become a core?

PR #2767, opened 2020-07-30
Add -f alias for --force on prepare command

greg-hellings:

#### PR Type

- Feature Pull Request

"molecule prepare --force" can now be accessed via "molecule prepare -f" as well


ldarmbuster:
Great addition!

Issue #2733, opened 2020-06-23
Help output is truncated


mperry2:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible 2.9.9
  config file = /export/home/mperry/.ansible.cfg
  configured module search path = ['/export/home/mperry/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /export/home/mperry/.local/lib/python3.7/site-packages/ansible
  executable location = /export/home/mperry/.local/bin/ansible
  python version = 3.7.3 (default, Dec 20 2019, 18:57:59) [GCC 8.3.0]
molecule 3.0.4
   ansible==2.9.9 python==3.7
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

- n/a

# Desired Behavior

The help output should have complete sentences and describe the function of each command. The output should continue on to a new line if necessary similar to the help output of the `-c` and `-e` options listed in the "Options" section of the help.

# Actual Behaviour

Some of the help output for the molecule commands is truncated with an ellipsis. Many of these are incomplete sentences, so the help output for those commands is not useful.

Example output is reproduced below where you can see the check, cleanup, converge, idempotence, prepare, and test commands have incomplete output.

```
$ molecule --help
Usage: molecule [OPTIONS] COMMAND [ARGS]...

  Molecule aids in the development and testing of Ansible roles.

  Enable autocomplete issue:

    eval "$(_MOLECULE_COMPLETE=source molecule)"

Options:
  --debug / --no-debug    Enable or disable debug mode. Default is disabled.
  -c, --base-config TEXT  Path to a base config.  If provided Molecule will
                          load this config first, and deep merge each
                          scenario's molecule.yml on top. By default is
                          looking for config in current VCS repository and if
                          not found it will look on user home.
                          (/export/home/mperry/.config/molecule/config.yml)

  -e, --env-file TEXT     The file to read variables from when rendering
                          molecule.yml. (.env.yml)

  --version               Show the version and exit.
  --help                  Show this message and exit.

Commands:
  check        Use the provisioner to perform a Dry-Run (destroy,...
  cleanup      Use the provisioner to cleanup any changes made to external...
  converge     Use the provisioner to configure instances (dependency,...
  create       Use the provisioner to start the instances.
  dependency   Manage the role's dependencies.
  destroy      Use the provisioner to destroy the instances.
  drivers      List drivers.
[...]

PR #2768, opened 2020-07-30
Avoid abridged click output

ssbarnea:
Fixes: #2733
Related: https://github.com/pallets/click/issues/486

### DIFF
--- a/molecule/command/base.py
+++ b/molecule/command/base.py
@@ -216,6 +216,7 @@ def click_group_ex():
     # yellow : special commands, like full test sequence, or login
     return click.group(
         cls=HelpColorsGroup,
+        context_settings=dict(max_content_width=9999),
         help_headers_color="yellow",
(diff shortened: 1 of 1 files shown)


molecule/command/base.py
```diff
@@ -216,6 +216,7 @@ def click_group_ex():
     # yellow : special commands, like full test sequence, or login
     return click.group(
         cls=HelpColorsGroup,
+        context_settings=dict(max_content_width=9999),
```
greg-hellings, line 219: Would it be smart to put a comment in the code explaining why you've done this? Even if it's just `# Check commit msg on this one`?

ssbarnea: Usually blame can be used but in this case adding a commend makes sense, especially as the number seems weird. I will let the user engage with Pallets people around why logic values like None, 0 or -1 do not work..

PR #2769, opened 2020-07-30
WIP: Switch supported Ansible versions to 2.9+

ssbarnea:
- drops 2.8 jobs
- adds 2.10 jobs
- mark as major in order to include in 3.1.0

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -63,7 +63,7 @@ setup_requires =
 
 # These are required in actual runtime:
 install_requires =
-    ansible >= 2.8  # keep it N/N-1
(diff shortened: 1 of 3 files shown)


greg-hellings:
Is this also an appropriate time to add Python 3.8 testing? What is the support matrix for Python versions?

ssbarnea:
@greg-hellings py38 is not tested because we did not had nodes with recent fedora or ubuntu ready (not sure if this was sorted). Is not because we do not want to test, is only because Zuul was not able provide a py38 to us. AFAIK, it works as that is what I use on macos.

Issue #2757, opened 2020-07-20
Can't install Molecule with latest Ansible devel releases (2.11)


geerlingguy:
# Issue Type

- Bug report

# Molecule and Ansible details

```
$ ansible --version
ansible-base-2.11.0.dev0

$ molecule --version
3.0.6
```

Molecule installation method (one of):

```
pip install molecule
```

Ansible installation method (one of):

```
pip uninstall -y ansible
pip install https://github.com/ansible/ansible/archive/devel.tar.gz --disable-pip-version-check
```

# Desired Behavior

Molecule should be able to run with Ansible installed from devel.

# Actual Behaviour

Molecule seems to be failing with:

```
pkg_resources.DistributionNotFound: The 'ansible' distribution was not found and is required by the application
```

The problem here is that if I install Ansible at devel first, then install Molecule via pip, Ansible 2.9.10 is installed on top of the ansible-base devel install, and that breaks Ansible too.

geerlingguy:
Related issue: https://github.com/ansible-collections/community.kubernetes/issues/167

geerlingguy:
It looks like even with ansible base 2.10 stable, I can't get Molecule to detect that ansible is installed:

Install with: `pip install https://github.com/ansible/ansible/archive/stable-2.10.tar.gz --disable-pip-version-check`

Get error when running `molecule`:

```
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.8/x64/bin/molecule", line 5, in <module>
    from molecule.__main__ import main
  File "/opt/hostedtoolcache/Python/3.7.8/x64/lib/python3.7/site-packages/molecule/__main__.py", line 22, in <module>
    from molecule.shell import main
  File "/opt/hostedtoolcache/Python/3.7.8/x64/lib/python3.7/site-packages/molecule/shell.py", line 90, in <module>
    prog_name="molecule", version=molecule.__version__, message=_version_string()
  File "/opt/hostedtoolcache/Python/3.7.8/x64/lib/python3.7/site-packages/molecule/shell.py", line 55, in _version_string
    pkg_resources.get_distribution("ansible").version,
  File "/opt/hostedtoolcache/Python/3.7.8/x64/lib/python3.7/site-packages/pkg_resources/__init__.py", line 482, in get_distribution
    dist = get_provider(dist)
[...]

PR #2777, opened 2020-08-09
Check ansible-base version before ansible version 

btravouillon:
Fixes `pkg_resources.DistributionNotFound: The 'ansible' distribution was not found and is required by the application` reported in #2757 when only ansible-base is installed.

This will report the version of ansible if < 2.10, the version of ansible-base if >= 2.10.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -97,9 +97,12 @@ class Config(object, metaclass=NewInitCaller):
     @property
     def ansible_version(self):
         """Return current version of ansible."""
-        return pkg_resources.parse_version(
-            pkg_resources.get_distribution("ansible").version
-        )
+        try:
+            ansible_version = pkg_resources.get_distribution("ansible-base").version
+        except Exception:
+            ansible_version = pkg_resources.get_distribution("ansible").version
+
+        return pkg_resources.parse_version(ansible_version)
 
     @property
     def ansible_collections_path(self):
--- a/molecule/shell.py
+++ b/molecule/shell.py
@@ -49,13 +49,15 @@ def _version_string() -> str:
     v = pkg_resources.parse_version(molecule.__version__)
     color = "bright_yellow" if v.is_prerelease else "green"  # type: ignore
     msg = "molecule %s\n" % _colorize(molecule.__version__, color)
+
+    try:
+        ansible_version = pkg_resources.get_distribution("ansible-base").version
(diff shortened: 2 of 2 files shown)


btravouillon:
In the Ansible 2.10 era, I'm not sure if molecule still needs to require ansible. I guess ansible-core is enough.

btravouillon:
Something like below?

```diff
diff --git a/molecule/config.py b/molecule/config.py
index e55b7f7c..77042fcb 100644
--- a/molecule/config.py
+++ b/molecule/config.py
@@ -97,9 +97,12 @@ def write(self):
     @property
     def ansible_version(self):
         """Return current version of ansible."""
-        return pkg_resources.parse_version(
-            pkg_resources.get_distribution("ansible").version
-        )
+        if pkg_resources.get_distribution("ansible-base"):
+            ansible_version = pkg_resources.get_distribution("ansible-base").version
+        else:
+            ansible_version = pkg_resources.get_distribution("ansible").version
+
+        return pkg_resources.parse_version(ansible_version)
 
     @property
     def ansible_collections_path(self):
diff --git a/molecule/shell.py b/molecule/shell.py
index 7cc9113f..129ff085 100644
--- a/molecule/shell.py
+++ b/molecule/shell.py
@@ -49,10 +49,16 @@ def _version_string() -> str:
     v = pkg_resources.parse_version(molecule.__version__)
     color = "bright_yellow" if v.is_prerelease else "green"  # type: ignore
     msg = "molecule %s\n" % _colorize(molecule.__version__, color)
+
+    if pkg_resources.get_distribution("ansible-base"):
+        ansible_version = pkg_resources.get_distribution("ansible-base").version
+    else:
+        ansible_version = pkg_resources.get_distribution("ansible").version
+
     msg += _colorize(
         "   ansible==%s python==%s.%s"
         % (
-            pkg_resources.get_distribution("ansible").version,
+            ansible_version,
             sys.version_info[0],
             sys.version_info[1],
         ),
```

ssbarnea:
Use an implementation that does not call get_distribution twice as this is an expensive call to make. I suspect you may need a try/catch approach. Maybe for v in ["a-base","a"]: x = get_dist(); if x return x?

You are welcomed to pick any implementation as long you do not call get_distribution twice for the same package as this will double the execution time.

PR #2789, opened 2020-08-19
Replace Zuul CI with Github Actions

ssbarnea:
Considering the current reliability of Ansible Zuul CI insance, we need an alternative so we avoind being blocked from testing or releasing patches.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -1,0 +1,127 @@
+name: tox
+
+on:
+  create:  # is used for publishing to PyPI and TestPyPI
+    tags:  # any tag regardless of its name, no branches
+  push:  # only publishes pushes to the main branch to TestPyPI
+    branches:  # any integration branch but not tag
+      - "master"
+    tags-ignore:
+      - "**"
+  pull_request:
+  schedule:
+    - cron: 1 0 * * *  # Run daily at 0:01 UTC
+
+jobs:
+  build:
+    name: ${{ matrix.tox_env }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - tox_env: lint
+          - tox_env: docs
+          - tox_env: py36
+            PREFIX: PYTEST_REQPASS=450
+          - tox_env: py37
+            PREFIX: PYTEST_REQPASS=450
+          - tox_env: py38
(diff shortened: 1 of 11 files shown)


.github/workflows/tox.yml
```diff
@@ -0,0 +1,142 @@
...
+on:
+  create:  # is used for publishing to PyPI and TestPyPI
+    tags:  # any tag regardless of its name, no branches
+  push:  # only publishes pushes to the main branch to TestPyPI
+    branches:  # any branch but not tag
+      - >-
+        **
+    tags-ignore:
+      - >-
+        **
```
greg-hellings, line 12: I wonder if this structure will pick up force-pushes of tags, or only creations.

Also, you could do

```yaml
branches:
  - "**"
tags-ignore:
  - "**"
```

If you wanted it to be more compact.

.github/workflows/tox.yml
```diff
@@ -0,0 +1,142 @@
...
+      fail-fast: false
+      matrix:
+        name: [
+          "linting",
+          "packaging",
+          "py36",
+          "py37",
+          "py38",
+          "devel",
+        ]
```
greg-hellings, line 32: Why JSON format here?

.github/workflows/tox.yml
```diff
@@ -0,0 +1,142 @@
...
+jobs:
+  build:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        name: [
+          "linting",
```
greg-hellings, line 26: Should be "lint"

greg-hellings: In fact, this whole structure maybe should be pulled dynamically by a previous build step that generates the list from calling "tox -l"?

tox.ini
```diff
@@ -19,8 +19,16 @@ isolated_build = True
...
+# do no put * in passenv as it may break builds do to reduced isolation
+passenv =
+    CI
+    DOCKER_*
+    PODMAN_*
+    GITHUB_*
+    TERM
+    HOME
 setenv =
+    ANSIBLE_CONFIG={toxinidir}/dev/null
```
greg-hellings, line 31: This should be `{toxinidir}/.ansible.cfg` instead of /dev/null.

Issue #2788, opened 2020-08-19
Tox py38-ansible28 missing pytest


rgarrigue:
# Issue Type

- Bug report

# Molecule and Ansible details

```
master branch / 33f1ec99
```

Molecule installation method (one of):

- source

Ansible installation method (one of):

- tox

Detail any linters or test runners used:

- Xubuntu 20.04 over Dell Precision M6700

# Desired Behavior

I want tox to install it however it's required for the test to run properly.

# Actual Behaviour

```
...
py38-ansible28-functional develop-inst-noop: /home/remy/personal/molecule
py38-ansible28-functional installed: ansible==2.8.14,appdirs==1.4.3,CacheControl==0.12.6,certifi==2019.11.28,cffi==1.14.2,chardet==3.0.4,colorama==0.4.3,contextlib2==0.6.0,cryptography==3.0,distlib==0.3.0,distro==1.4.0,html5lib==1.0.1,idna==2.8,ipaddr==2.2.0,Jinja2==2.11.2,lockfile==0.12.2,MarkupSafe==1.1.1,molecule==3.0.8.dev2,msgpack==0.6.2,packaging==20.3,pep517==0.8.2,pkg-resources==0.0.0,progress==1.5,pycparser==2.20,pyparsing==2.4.6,pytoml==0.1.21,PyYAML==5.3.1,requests==2.22.0,retrying==1.3.3,selinux==0.2.1,six==1.14.0,urllib3==1.25.8,webencodings==0.5.1
py38-ansible28-functional run-test-pre: PYTHONHASHSEED='19837123'
py38-ansible28-functional run-test: commands[0] | pip check
No broken requirements found.
py38-ansible28-functional run-test: commands[1] | sh -c 'PYTEST_ADDOPTS= python -m pytest -p no:cov --collect-only 2>&1 >/home/remy/personal/molecule/.tox/py38-ansible28-functional/log/collect.log'
/home/remy/personal/molecule/.tox/py38-ansible28-functional/bin/python: No module named pytest
ERROR: InvocationError for command /usr/bin/sh -c 'PYTEST_ADDOPTS= python -m pytest -p no:cov --collect-only 2>&1 >/home/remy/personal/molecule/.tox/py38-ansible28-functional/log/collect.log' (exited with code 1)
py38-ansible29-unit create: /home/remy/personal/molecule/.tox/py38-ansible29-unit
py38-ansible29-unit installdeps: ansible>=2.9,<2.10, selinux
[...]

PR #2790, opened 2020-08-19
Fix missing pytest dependency

rgarrigue:
#### PR Type

- Bugfix Pull Request #2788


ssbarnea:
That is not needed, have you tried to run `tox -r ..." to recreate the environments. pytest is listed as part of "test" extra and will be installed. I am ready to bet this is caused by outdated virtualenv  on your machine.

rgarrigue:
The PR was created a bit too fast, I still have the original issue, missed it on checking.

I'll give a shot to tox -r, though the tox venv is brand new, I just cloned molecule originaly to make https://github.com/ansible-community/molecule/pull/2791 PR

PR #2793, opened 2020-08-20
Fix problems with CI

greg-hellings:
Collect values rather than maintain them locally
Cleanup unnecessary JSON syntax

* Should address most of the issues currently with enabling Github Actions on the repo


greg-hellings:
This PR would land against your POC Github Actions branch, not against master.

ssbarnea (approved):

PR #2800, opened 2020-08-25
Update interpolation.py

joej164:

#### PR Type
- Docs Pull Request

The name of the file that needs to be in the root is `.env.yml`.  Updating this so others don't get tripped up.


ssbarnea (approved):

ssbarnea:
@larmbuster When you created the fix, you did not allow edit from maintainers and this change cannot pass CI as is using outdated code. I created a branch myself and updated.

joej164:
thank you, i thought i had checked the box to allow updates, but i guess I didn't.  Thanks for fixing the issue.

PR #2802, opened 2020-08-29
Fixed linters

ssbarnea:
Among normal bumping of linters temporary pins down pydocstyle due to recent regression.

Related: https://github.com/PyCQA/pydocstyle/pull/506


ssbarnea:
> Do you really want to pin pydocstyle to a less than, or just exclude the known broken version?

It did not merge so I have no guarantee that it will be merged before the next tag is added. If it was merged I would have picked for exclusion. Anyway, I am not worried much about this because next bumping of linter would remind me to replace the downpinning.

PR #2811, opened 2020-09-01
Remove docker driver from core

ssbarnea:


geerlingguy:
@ssbarnea  This seems to be a rather huge change in functionality; is there anywhere this was discussed (the reasoning behind the change)? If this changes the default if unspecified, I know most of my own molecule configs will be broken and require changes.

ssbarnea:
Removal of drivers from core was initially planned for 3.0 but we postponed docker and podman due to lack of time.

Remember that this is not dropping support for them, is only part of moving them to individual projects with their own CI/CD pipelines. Read https://github.com/ansible-community/molecule/milestone/19 -- I will not release 3.1 before these drivers are functional, without changes for the users.

I suspect it will take some time to finish this move. If needed, I will create a `stable/3.0` branch in order to make patches there.

geerlingguy:
@ssbarnea - The main thing that would need to be changed is you would _have_ to install molecule with `pip3 install molecule[docker]` now, right? And make sure your `molecule.yml` defines `driver.name: docker`?

PhilippSelenium:
This should really be more prominent in the documentation. Took me quite some time to find that my builds were broken because I was missing:
```
pip3 install molecule[docker]
```

PR #2814, opened 2020-09-01
Removed support for gilt

ssbarnea:
As [gilt project](https://github.com/retr0h/gilt) is not actively maintained, we remove support for installing dependencies with it.


geerlingguy:
I've honestly never heart of Gilt.

ssbarnea:
> I've honestly never heart of Gilt.

Gilt was created by @retr0h, molecule author a long time ago, likely before galaxy was a thing. I doubt there is any real usage in the wild.

Even if there are some users still using it, they will have enough time to migrate to either galaxy or shell installers.

decentral1se (commented):
:+1:

Issue #2820, opened 2020-09-09
Update GH actions example


sestrella:
**PR Type**
- Docs Pull Request

The "commit" event described in the documentation doesn't seem to be valid:

![image](https://user-images.githubusercontent.com/2049686/92658124-a9cfde80-f2bb-11ea-8e74-51fa2c7a6d2c.png)

"pull_request" on the other hand works fine:

![image](https://user-images.githubusercontent.com/2049686/92658106-9c1a5900-f2bb-11ea-83db-c429ce6a9c58.png)

PR #2821, opened 2020-09-09
Correct GitHub Actions example

sestrella:
Fixes: #2820


1efty (approved):

1efty:
Change is harmless but couldn't find anything in action docs that says otherwise. https://docs.github.com/en/actions/reference/events-that-trigger-workflows

ssbarnea (approved):

PR #2842, opened 2020-09-21
Added hint about current directory for "initialization"

wouterj:
##### SUMMARY

It took me quite some time and the help of this https://github.com/ansible-community/molecule/issues/1133 issue (where the author also needed quite some time) to figure out that I was running this command from the wrong directory (namely the parent of the role's directory).

I think it would help if the docs were a bit more clear about this :)

##### ISSUE TYPE
- Docs Pull Request

+label: docsite_pr

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -40,7 +40,8 @@ You should then see a ``my-new-role`` folder in your current directory.
 
     For future reference, if you want to initialize Molecule within an
     existing role, you would use the ``molecule init scenario -r
-    my-role-name`` command.
+    my-role-name`` command from within the role's directory (e.g.
+    ``my-role-name/``).
 
 Molecule Scenarios
 ------------------


larmbuster (commented):
I don't think this change is necessary. I think it is generally known that the role name is the path to the role's directory structure. For example, `molecule init scenario -r ~/ansible/roles/my-role-name` is also an acceptable path. Let's get some more comments on this first though.

wouterj:
Hmm, I'm quite new to ansible and it's community, so I'm a weak on "common" or best practices.

To illustrate my situation:  
In our directory structure, we have a mono repo with some roles saved in `roles/...`, similar to this: https://docs.ansible.com/ansible/latest/user_guide/playbooks_best_practices.html#directory-layout I've used molecule to set-up new roles using `molecule init role my-new-role` from inside `roles/`. So when I wanted to add molecule to an existing role, I was trying the same and I tried to execute `molecule init scenario -r my-existing-role` from within `roles/`, which caused the error "Please choose the proper role name.". As there weren't many details on "proper", I took a while before I discovered I had to execute this command from within the role directory (so within `roles/my-existing-role`)

ssbarnea (approved):

MarkusTeufelberger:
As an aside, role names should not contain the `-` character, so it might be better to give `my_existing_role` as an example. Ansible-lint will complain about the role name otherwise.

Issue #2845, opened 2020-09-24
order of tests files for testinfra


atmaniak:
The os.walk is not sorted so the test file are always played in a different order.

Is it possible to have an option to force test order in testinfra ?

PR #2846, opened 2020-09-25
Sort tests files for testinfra by filename

atmaniak:
With sorted, tests files for testinfra are now executed in the same order, since os.walk return a list in an arbitrary maner.

- Bugfix #2845

### DIFF
--- a/molecule/verifier/testinfra.py
+++ b/molecule/verifier/testinfra.py
@@ -200,7 +200,7 @@ class Testinfra(Verifier):
 
         :return: list
         """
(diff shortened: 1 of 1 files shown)


greg-hellings:
I'm looking into the py36/37 errors separately. Those don't look like they're part of your PR. But the lint errors definitely are. Can you please run "tox -e lint" locally and then push the resulting changes. It's failing the "black" hook, which should auto-reformat the offending files if you run it locally, commit, and push the changed files.

ssbarnea (approved):

PR #2847, opened 2020-09-25
Avoid importlib-metadata conflict with tox

greg-hellings:
importlib-metadata does not support installing versions after 2.0 on
Python versions prior to py38.
Also fix the path for ANSIBLE_CONFIG file

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -38,7 +38,7 @@ passenv =
     SSH_AUTH_SOCK
     TERM
 setenv =
-    ANSIBLE_CONFIG={toxinidir}/dev/null
+    ANSIBLE_CONFIG={toxinidir}/.ansible.cfg
     ANSIBLE_CALLABLE_WHITELIST={env:ANSIBLE_CALLABLE_WHITELIST:timer,profile_roles}
     ANSIBLE_DISPLAY_FAILED_STDERR=1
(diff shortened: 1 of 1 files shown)


greg-hellings:
Getting tox/build-containers to build properly is being a pain. build-containers tests all the main plugins, including molecule-azure, which currently fails because ansible[azure] is no longer a thing.

tox.ini
```diff
@@ -38,7 +38,7 @@ passenv =
     SSH_AUTH_SOCK
     TERM
 setenv =
-    ANSIBLE_CONFIG={toxinidir}/dev/null
+    ANSIBLE_CONFIG={toxinidir}/.ansible.cfg
```
ssbarnea, line 41: This this really needed, why? Seems unrelated.

greg-hellings: If ANSIBLE_CONFIG points to a non-existent path, then the value is ignored and Ansible falls back to any that exist on the local system. This causes local runs of tests to fail, at least on my system, because of some of my ~/.ansible.cfg settings.

True, this isn't strictly necessary, but it was necessary to pass in my environment.

ssbarnea (approved):

PR #2849, opened 2020-09-28
Fixed CustomLogger super call

mattp-:
In cases where another logger uses getLoggerClass to subclass and then
setLoggerClass's itself,  molecules logger passes the wrong value to
super() causing a recursionError.

#### PR Type
- Bugfix Pull Request

### DIFF
--- a/molecule/test/unit/test_logger.py
+++ b/molecule/test/unit/test_logger.py
@@ -20,6 +20,7 @@
 
 from __future__ import print_function
 
+import logging
 import sys
 
 import colorama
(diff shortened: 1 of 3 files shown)


ssbarnea:
@mattp- please rebase, we fixed the unrelated CI failures.

molecule/test/unit/test_logger.py
```diff
@@ -177,3 +179,16 @@ def test_markup_detection_tty_no(mocker):
...
+    class FooLogger(logging.getLoggerClass()):
+        log = logging.getLogger('FooLogger')
+        pass
+
+    logging.setLoggerClass(FooLogger)
+
+    try:
+        log = FooLogger('foo')
+    except RecursionError as exc:
+        raise pytest.fail("CustomLogger caused recursion error: {0}".format(exc))
```
greg-hellings, line 194: Why are you catching this, and not just letting the error pass out to pytest automatically?

mattp-: not sure, I guess that also works :) ill change

greg-hellings:
You accidentally made an extra test and it passes. Somewhere in the test code you'll see definitions for how many tests there are supposed to be. You'll need to bump that up to 433 to account for your additional test.

mattp-:
@greg-hellings should be good now I believe, pending that last build-containers task. thanks!

greg-hellings (approved):
lgtm

ssbarnea (approved):

PR #2852, opened 2020-09-30
Fixed build-containers

ssbarnea:
Since Ansible 2.10 was released build-containers job was failing
due to inability to upgrade Ansible. Now we avoid installing it from
apk and we will avoid any conflicts.


greg-hellings (approved):
This mixes well with #2847 where I fixed the other tests. I was toying with this same change in that PR, but wasn't sure if it still hit the point of this test.

PR #2857, opened 2020-10-03
Refactored logging and terminal support (WIP, feedback welcomed)

ssbarnea:
- major change as it changes the way molecule display its own output
  and the one from ansible. No huge changes expected but user should
  be aware.
- replace use of colorama with rich
- refactored internal logging to make use of rich logging


decentral1se (commented):
> replace use of colorama with rich

Why?

Other questions:
- You removed a lot of tests but didn't replace them? That might make this unstable.
- Can we avoid defining the terminal theme in our code? Is there now default set?
- Is that virtualenv packaging change related or? Hard to know from my side.
- Is this PR still WIP since there are a lot of comments? Some tests are commented out that are also potentially unrelated to this change? That was also unclear as to why they were included.

Just taking a pass but those are the things that jumped to mind.

ssbarnea:
I already found some bugs so it is somewhere between WIP and ready for review.

The primary goal at this stage is to reach a level where we like the result and we are not aware of any regressions.

Note: I do not plan to merge this change without getting at least 2x passing reviews, is just too risky.

I am aware of commented/removed tests, I plan to restore at least some of them, but only after we reach the desired output/behavior.

decentral1se:
That's all fine, feel free to re-ping. I still don't understand why you are moving to `rich` from `colorama`.

ssbarnea:
@decentral1se Rich uses colorama itself, mainly what we do is to switch to a higher level "API". Main advantage is that rich is able to highlight printed/logged code automatically, making it much easier to read (works quite nice even without telling it what to do). I will post some screenshots soon.

Likely that I will split this change in multiple PRs once it seems fine. FYI, I switched ansible-lint to it few months ago, take at look at screenshots from https://github.com/ansible/ansible-lint/pull/978 and https://github.com/ansible/ansible-lint/pull/971

For example we could get rid of tabulate but I wanted to avoid increasing the scope.

decentral1se:
Aha, I did see some absolutely wonderful new colorful logging coming from ansible-lint side latest version. Keep doing your magic then! If you want to split or keep it all here, that's fine, just split into separate commits...will try to keep up with this.

PR #2858, opened 2020-10-03
Documentation: fixing --parallel flag position and typos

Arnoways:
Hello,

I recently had the need to run Molecule tests in parallel mode using another tool (https://github.com/ovh/venom to be specific), and found some errors/typos in the docs. 
I also added the line about running the molecule command with the --parallel flag to be more explicit on how to avoid concurrency issues (mine was over apt lock :smile: ).

#### PR Type
- Docs Pull Request


ssbarnea:
That is good addition. Still I would rather move the argument to root instead of changing documentations. That is the kind of option that should be shared among all commands (ignore for those that do not need it).

ssbarnea (approved):

PR #2869, opened 2020-10-08
Removed molecule ansible filters

ssbarnea:
Fixes: #2872


collection/playbooks/validate.yml
```diff
@@ -0,0 +1,14 @@
...
+    - ssbarnea.molecule
+  tasks:
+
+    - name: Check if molecule_header filter can be used (1)
+      debug:
+        msg: "{{ '' | ssbarnea.molecule.molecule_header }}"
+
+    - name: Check if molecule_header filter can be used (2)
+      debug:
+        msg: "{{ '' | molecule_header }}"
```
ssbarnea, line 14: @bcoca Apparently it chokes here with:
```
fatal: [localhost]: FAILED! => {"msg": "template error while templating string: no filter named 'molecule_header'. String: {{ '' | molecule_header }}"}
```

Not sure how to convince Ansible to find the filter using short name.

.github/workflows/tox.yml
```diff
@@ -111,7 +111,9 @@ jobs:
           xargs git tag --delete
       - name: Build dists
         run: python -m tox
-      - name: Publish to test.pypi.org
+      - name: Publish collection
+        run: python -m tox -e packaging
+        - name: Publish to test.pypi.org
```
greg-hellings, line 116: This indentation seems off

collection/playbooks/validate.yml
greg-hellings: Docs specifically say you can only use short name for modules. Anything else always requires FQCN

PR #2903, opened 2020-10-19
Reflect changes in the molecule core in the documentation

fourstepper:
This change is made to reflect the change in disconnection of the molecule drivers from the core in the docs

#### PR Type

- Docs Pull Request


docs/installation.rst
```diff
@@ -90,7 +90,15 @@ Install Molecule:
 
 .. code-block:: bash
 
-    $ python3 -m pip install --user "molecule[lint]"
+    $ python3 -m pip install --user "molecule[lint] molecule-docker"
+
+Molecule uses docker by default, which is maintained separately as
```
ssbarnea, line 95: That is no longer true: molecule does not use docker by default in any way. The only driver in core is delegated.

The command line you suggested is correct but also similar to `molecule[docker,lint]`, which is shorter.

fourstepper: Thanks for the insight. I will edit it

fourstepper: This should now be factual. I have also deleted the link the the Configuration page as I don't think it hold enough relevant information regarding the topic (for now)

ssbarnea (approved):

larmbuster (approved):

PR #2914, opened 2020-10-23
Disable Ansible interpreter warnings

ssbarnea:
Switches Ansible warnings around interpreter being picked to silent in order to avoid noise during testing.

### DIFF
--- a/lib/molecule/provisioner/ansible.py
+++ b/lib/molecule/provisioner/ansible.py
@@ -386,7 +386,7 @@ class Ansible(base.Base):
                 "retry_files_enabled": False,
                 "host_key_checking": False,
                 "nocows": 1,
-                "interpreter_python": "auto",
+                "interpreter_python": "auto_silent",
             },
             "ssh_connection": {
                 "scp_if_ssh": True,
--- a/lib/molecule/test/unit/provisioner/test_ansible.py
+++ b/lib/molecule/test/unit/provisioner/test_ansible.py
@@ -100,7 +100,8 @@ def test_default_config_options_property(_instance):
             "display_failed_stderr": True,
             "forks": 50,
             "host_key_checking": False,
-            "interpreter_python": "auto",
+            # https://docs.ansible.com/ansible/devel/reference_appendices/interpreter_discovery.html
+            "interpreter_python": "auto_silent",
             "nocows": 1,
             "retry_files_enabled": False,
         },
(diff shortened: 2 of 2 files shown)


geerlingguy:
I still like to have warnings present, because while like 80% of them are non-actionable, annoying, and IMO should not be warnings, there are 20% or so that help me to proactively change things that need changing for future compatibility.

This should be easy to configure for the end user but by default I think they should be enabled.

greg-hellings:
I'm in agreement with @geerlingguy. Many of the warnings are overzealous - like the python discovery ones - but I rely on it to find deprecations early and prevent failures preemptively. Every time I move to a new version of Ansible and see new warnings pop up it's a great reminder to run through my roles and update them.

ssbarnea:
@geerlingguy @greg-hellings Feel free to to use the format review option and request changes if you disagree with the proposal. I do find opposition beneficial to the project.

I am also against disabling Ansible warnings in general and this is why I complained on Ansible issue tracker that they have only one nuclear disable deprecation warnings.

Still, **this change is about a very particular warning, the Python interpreter being used**. I almost always seen these warnings as useless, as the user does not really care about which interpreter is used as long ansible succeeds. Because molecule is used during **testing**, it means that core will run with newer interpreter before reaching production. This makes the warning useless, because user will already discover an issue before he goes to production.

This ansible warning is useful for those that do not have tests, as it warns them that in next version of Ansible you may see a behavior change. But when you test your code (molecule!), you will always run the tests with the newer version of Ansible before running the same code in production. This means that displaying this warning in molecule context is just noise.

I would rather see a warning when there is something actionable (a way to address it ), and not when something will happen in the future but you cannot do anything about it now. These warning in particular reminded me of the story about the boy that cried wolf... How i see it: warning presence makes people resistant to them and they will soon ignore all of them. Less warnings but all actionable, makes it much easier for people to not ignore them.

greg-hellings (approved):
If we can really nuke just this one pointless warning, then I'm in favor of this change.

ssbarnea:
> If we can really nuke just this one pointless warning, then I'm in favor of this change.

Yes, it is only that particular one. I will still let it wait for more feedback, before merging it (or not).

geerlingguy:
I'm okay with it, in that case.

larmbuster:
I agree with silencing this particular warning.

Issue #2678, opened 2020-04-28
Remove sh package as a molecule dependency


ssbarnea:
As we previously had some problems with `sh` library and today [a new release also broke us](https://github.com/amoffat/sh/issues/520), it is time to remove it as a dependency.

If someone makes a PR to remove sh and replace it with either an internal implementation or another library that is better maintained (test coverage and multiple maintainers), I will be very happy to fast-track it.

Related: https://github.com/ansible-community/molecule/issues/2676

decentral1se:
Regression was handled in a timely manner, however.

> https://github.com/amoffat/sh/issues/520#issuecomment-620637323

Yajo:
https://plumbum.readthedocs.io/ rocks :guitar:

ssbarnea:
Plumbum seems nice but I did not test it yet. Also seems to be another project backed by a single developer, which worries me. I am more inclined to add some DIY code than ending up in the same place one year later. Still, no action will be taken unless we have a strong reason.

greg-hellings:
I have had good luck with plumbum in the past, and it has been pretty solidly stable. But for the amount of calling that Molecule does, we can do this ourselves with popen.

ssbarnea:
I started working on this during the weekend, implemention would count as major change which will impact plugins, so it will only go in in `3.2.0`.

PR #2917, opened 2020-10-27
Remove dependency on sh library

ssbarnea:
* Main change is replace use of `sh` library with `subprocess-tee` while adopting a loose-coupling approach, so we do not rely on implementation details of the subprocess library.
* This will be part of 3.2.0 major release as it will likely break at least some plugins that are using `molecule.util.run_command` or the `bake()`. 
* It will initially be tagged as 3.2.0a0 in order to give time for plugin projects to adapt their test suites.
* This enables us to capture and process the output of the executed commands while still displaying it in real time, still no extra processing is done in this change.

Fixed: #2678
Related: https://github.com/amoffat/sh/issues/450 https://github.com/amoffat/sh/issues/512


larmbuster:
I like the simplicity of `subprocess-tee`.

decentral1se:
:clap: :clap: :clap:

PR #2920, opened 2020-10-27
Remove dependency on tree-format

ssbarnea:
Replaces our limited use of tree-format with a simple YAML dump for printing the matrix steps.

This changes the output a little bit, the new one looks like:


felixfontein:
Why do you request a review if you merge it shortly after that anyway?

ssbarnea:
@felixfontein I did not see reviews in a long time. If you see anything to fix I will do a follow-up on it.

PR #2921, opened 2020-10-27
Move pexpect from a runtime dependency to test

ssbarnea:
We do not really need pexpect as a runtime dependency as it was using only inside login. Using just subprocess.run() seems to work fine, so we can lower our dependencies.

### DIFF
--- a/lib/molecule/command/login.py
+++ b/lib/molecule/command/login.py
@@ -19,15 +19,10 @@
 #  DEALINGS IN THE SOFTWARE.
 """Login Command Module."""
 
-import fcntl
 import os
-import signal
-import struct
-import sys
-import termios
+from subprocess import run
 
 import click
-import pexpect
 
 from molecule import logger, scenarios, util
 from molecule.command import base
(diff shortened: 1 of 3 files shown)


lib/molecule/command/login.py
```diff
@@ -147,19 +142,8 @@ def _get_login(self, hostname):  # pragma: no cover
         login_options["lines"] = lines
         login_cmd = self._config.driver.login_cmd_template.format(**login_options)
 
-        dimensions = (int(lines), int(columns))
         cmd = "/usr/bin/env {}".format(login_cmd)
-        self._pt = pexpect.spawn(cmd, dimensions=dimensions)
```
ssbarnea, line 152: @retr0h Can you please comment on why the pexpect implementation was added in the first place for the interactive login? I am asking because I would not want to introduce a serious regression just because I do not know the reasons behind original implementation.

I did some basic tests and using subprocess.run() appeared to be working fine for me.

greg-hellings: Possible reason from the docs: "The run() function was added in Python 3.5"

If we previously supported < 3.5, then this call wouldn't have been possible.

greg-hellings (approved):
Obviously pending input from @retr0h, but it looks good to me.

PR #2926, opened 2020-10-29
Harden smoketest to avoid ansible becoming a python dependency

ssbarnea:
Molecule drivers should never made ansible package as a dependency
as molecule itself is decoupled from Ansible python module.

This removes drivers that are prevent this decoupling.

Related: https://github.com/ansible-community/molecule-vagrant/pull/78


greg-hellings:
I like it.

greg-hellings (approved):

larmbuster (approved):

PR #2927, opened 2020-10-29
Set ansible_ssh_pass with ansible_password because of a testinfra issue

lmesz-bitrise:
We are using delegated driver to connect to a recently created host on which ssh keys are not set yet, when we had tried to execute molecule test on that it stopped and asked for password in the verify phase even though the password was set in the create.yml.
For verification we use testinfra, under the hood it uses pytest with ansible backend.
After some investigation figured out the reason is that ansible_ssh_pass is not set and based on the doc it is mandatory if key is not set. This [issue](https://github.com/philpep/testinfra/issues/580) has been reported to testinfra, but until it is not fixed this PR implements a small workaround.

### DIFF
--- a/lib/molecule/driver/delegated.py
+++ b/lib/molecule/driver/delegated.py
@@ -198,6 +198,8 @@ class Delegated(Driver):
                     )
                 if d.get("password", None):
                     conn_dict["ansible_password"] = d.get("password")
+                    # Based on testinfra documentation, ansible password must be passed via ansible_ssh_pass
+                    # issue to fix this can be found https://github.com/philpep/testinfra/issues/580
                     conn_dict["ansible_ssh_pass"] = d.get("password")
                 if d.get("winrm_transport", None):
                     conn_dict["ansible_winrm_transport"] = d.get("winrm_transport")
--- a/lib/molecule/driver/delegated.py
+++ b/lib/molecule/driver/delegated.py
@@ -198,6 +198,7 @@ class Delegated(Driver):
                     )
                 if d.get("password", None):
                     conn_dict["ansible_password"] = d.get("password")
+                    conn_dict["ansible_ssh_pass"] = d.get("password")
                 if d.get("winrm_transport", None):
                     conn_dict["ansible_winrm_transport"] = d.get("winrm_transport")
                 if d.get("winrm_cert_pem", None):
--- a/lib/molecule/driver/delegated.py
+++ b/lib/molecule/driver/delegated.py
@@ -199,7 +199,7 @@ class Delegated(Driver):
                 if d.get("password", None):
                     conn_dict["ansible_password"] = d.get("password")
                     # Based on testinfra documentation, ansible password must be passed via ansible_ssh_pass
-                    # issue to fix this can be found https://github.com/philpep/testinfra/issues/580
+                    # issue to fix this can be found https://github.com/pytest-dev/pytest-testinfra/issues/580
                     conn_dict["ansible_ssh_pass"] = d.get("password")
                 if d.get("winrm_transport", None):
                     conn_dict["ansible_winrm_transport"] = d.get("winrm_transport")


ssbarnea (commented):
Can you please provide some links that clarify the difference between `ansible_ssh_pass` and `ansible_password`.

If we fix it we better know what we are doing, maybe even including a clarifying comment in the source code.

larmbuster:
I can't find `ansible_ssh_pass` in the newer docs, but I believe it's the old version of `ansible_password`. I did find `ansible_ssh_pass` in the [docs for version 2.3](https://docs.ansible.com/ansible/2.3/intro_inventory.html) and it seems to be a SSH specific password variable.

greg-hellings:
I believe that `ansible_ssh_pass` is deprecated in favor of `ansible_password`, since there are alternatively connection types that might be supported.

lmesz-bitrise:
Sorry I was not clear with it. The parameter is not for ansible but for testinfra pytest backend. Here is the doc that mention the parameter. https://testinfra.readthedocs.io/en/latest/backends.html#ansible

lmesz-bitrise:
> Can you please provide some links that clarify the difference between `ansible_ssh_pass` and `ansible_password`.

If we fix it we better know what we are doing, maybe even including a clarifying comment in the source code.

> * Do we need both?

If we only use that inventory for pytest than we don't need `ansible_password` `ansible_ssh_pass` is enough

> * Do we have a case where one may be different than another?

Unfortunately, I don't know but if it is only for pytest then no.

> * Are you aware that you can mention inventory variables directly in molecule file, bypassing any need to process them with molecule.

Ohh I wasn't aware of it, will check it.

Sorry I was not clear with the change. The parameter is not for ansible but for testinfra verifier. For test it uses pytest and pytest to establish the connection uses ansible backend. The documentation can be found [here](https://testinfra.readthedocs.io/en/latest/backends.html#ansible)
In case of password and not key we need to use that variable since the connection for ansible is hardcoded here https://github.com/ansible-community/molecule/blob/master/lib/molecule/driver/base.py#L75

ssbarnea:
> > Can you please provide some links that clarify the difference between `ansible_ssh_pass` and `ansible_password`.
> 
> If we fix it we better know what we are doing, maybe even including a clarifying comment in the source code.
> 
> > * Do we need both?
> 
> If we only use that inventory for pytest than we don't need `ansible_password` `ansible_ssh_pass` is enough
> 
> > * Do we have a case where one may be different than another?
> 
> Unfortunately, I don't know but if it is only for pytest then no.
> 
> > * Are you aware that you can mention inventory variables directly in molecule file, bypassing any need to process them with molecule.
> 
> Ohh I wasn't aware of it, will check it. How can I do this?
> 
> Sorry I was not clear with the change. The parameter is not for ansible but for testinfra verifier. For test it uses pytest and pytest to establish the connection uses ansible backend. The documentation can be found [here](https://testinfra.readthedocs.io/en/latest/backends.html#ansible)
> In case of password and not key we need to use that variable since the connection for ansible is hardcoded here [https://github.com/ansible-community/molecule/blob/master/lib/molecule/driver/base.py#L75](https://github.com/ansible-community/molecule/blob/master/lib/molecule/driver/base.py?rgh-link-date=2020-10-30T07%3A42%3A25Z#L75)

When in doubt, try to ask on `#ansible-devel`, other devs are more than willing to help with questions like this. Give them link to this PR. I suspect is an old version of the one you tried and in this case the correct fix is different.

lmesz-bitrise:
> > > Can you please provide some links that clarify the difference between `ansible_ssh_pass` and `ansible_password`.
> > 
> > 
> > If we fix it we better know what we are doing, maybe even including a clarifying comment in the source code.
> > > * Do we need both?
> > 
> > 
> > If we only use that inventory for pytest than we don't need `ansible_password` `ansible_ssh_pass` is enough
> > > * Do we have a case where one may be different than another?
> > 
> > 
> > Unfortunately, I don't know but if it is only for pytest then no.
> > > * Are you aware that you can mention inventory variables directly in molecule file, bypassing any need to process them with molecule.
> > 
> > 
> > Ohh I wasn't aware of it, will check it. How can I do this?
> > Sorry I was not clear with the change. The parameter is not for ansible but for testinfra verifier. For test it uses pytest and pytest to establish the connection uses ansible backend. The documentation can be found [here](https://testinfra.readthedocs.io/en/latest/backends.html#ansible)
> > In case of password and not key we need to use that variable since the connection for ansible is hardcoded here [https://github.com/ansible-community/molecule/blob/master/lib/molecule/driver/base.py#L75](https://github.com/ansible-community/molecule/blob/master/lib/molecule/driver/base.py?rgh-link-date=2020-10-30T07%3A42%3A25Z#L75)
> 
> When in doubt, try to ask on `#ansible-devel`, other devs are more than willing to help with questions like this. Give them link to this PR. I suspect is an old version of the one you tried and in this case the correct fix is different.

Alright trying to connect. Old version of which component? I mean I doubt a little cause the documentation exactly says that to use that variable.
_If ssh identity file is not provided via –ssh-identity-file flag, testinfra will try to use ansible_ssh_private_key_file, ansible_private_key_file and, finally, ansible_user with ansible_ssh_pass variables, both should be specified._

ssbarnea:
@tadeboro Gave a very useful link: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/ssh_connection.html#parameter-password -- which indicates that there are 3 variables around password, not two.

I suspect that the bug may be a testinfra related bug? I see you already ruled out an outdated Ansible issue.

lmesz-bitrise:
> @tadeboro Gave a very useful link: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/ssh_connection.html#parameter-password -- which indicates that there are 3 variables around password, not two.

Do you suggest setting all three variables?

> 
> I suspect that the bug may be a testinfra related bug? I see you already ruled out an outdated Ansible issue.

Sorry I don't understand that part. What kind of bug?

tadeboro:
As far as Ansible is concerned, setting ` ansible_password`  or ` ansible_ssh_pass`  is exactly the same thing. So I guess something strange happens during the verification step. Unfortunatelly, there is no output attached to this PR, so it is hard to know what exactly went wrong.

lmesz-bitrise:
Nothing strange happened. Based on testinfra code it is intended.
As I posted earlier from the [doc](https://testinfra.readthedocs.io/en/latest/backends.html#ansible) it says _...and, finally, ansible_user with ansible_ssh_pass variables, both should be specified_.
I took a look and it is because they explicitly grab that parameter [relevant code](https://github.com/philpep/testinfra/blob/master/testinfra/utils/ansible_runner.py#L81)

lmesz-bitrise:
@ssbarnea you were right, your suggestion also works:
```
---
dependency:
  name: galaxy
driver:
  name: delegated
platforms:
  - name: instance
provisioner:
  name: ansible
  lint:
    name: ansible-lint
  inventory:
    host_vars:
      instance:
        ansible_ssh_pass: vagrant
verifier:
  name: testinfra
  additional_files_or_dirs:
    - ../../../../node/tests/*

tadeboro:
So, this seems like a testinfra bug, not a molecule bug, and should be fixed there. But for the sake of end-user convenience, I guess this workaround should be added to Molecule also. I would just add a comment over the added line that describes the why behind the change. Something like "testinfra expects older alias for SSH password here" just to make sure we know why this line of code is there in a year or two.

And I guess this same information should go into commit message also because right now, it is next to impossible to determine why this change is needed (the commit message says nothing about testinfra at all).

ssbarnea:
> So, this seems like a testinfra bug, not a molecule bug, and should be fixed there. But for the sake of end-user convenience, I guess this workaround should be added to Molecule also. I would just add a comment over the added line that describes the why behind the change. Something like "testinfra expects older alias for SSH password here" just to make sure we know why this line of code is there in a year or two.
> 
> And I guess this same information should go into commit message also because right now, it is next to impossible to determine why this change is needed (the commit message says nothing about testinfra at all).

I have maintainer rights on testinfra so I could help use merging a fix there. Here are few action points:

* Create upstream (testinfra) bug with details
* Add source code comment this PR that points to the testinfra bug, so next time we look at the code we know why it was added and we will also know when we can remove the workaround
* Update PR title/description with new information, remove unnecessary details.

Once done, I will approve the workaround.

lmesz-bitrise:
> > So, this seems like a testinfra bug, not a molecule bug, and should be fixed there. But for the sake of end-user convenience, I guess this workaround should be added to Molecule also. I would just add a comment over the added line that describes the why behind the change. Something like "testinfra expects older alias for SSH password here" just to make sure we know why this line of code is there in a year or two.
> > And I guess this same information should go into commit message also because right now, it is next to impossible to determine why this change is needed (the commit message says nothing about testinfra at all).
> 
> I have maintainer rights on testinfra so I could help use merging a fix there. Here are few action points:
> 
> * Create upstream (testinfra) bug with details

I created a testinfra [bug](https://github.com/philpep/testinfra/issues/580), I hope added all the necessary information.

> * Add source code comment this PR that points to the testinfra bug, so next time we look at the code we know why it was added and we will also know when we can remove the workaround

Added with the bug in it.

> * Update PR title/description with new information, remove unnecessary details.

Updated accordingly.

> 
> Once done, I will approve the workaround.

Thank you for the help with the whole PR.

ssbarnea (approved):

PR #2933, opened 2020-10-30
Add verbosity argument

ssbarnea:
Allow easy control over Ansible verbosity by allowing -v[vvv] molecule arguments.

Previously Ansible verbosity was controlled by directly defining ANSIBLE_VERBOSITY variable or inside molecule.yml files, which was not very convenient for command line users. This change enables molecule user to specify `-v` the same way as it does with Ansible itself.

This flag is currently used to control only Ansible verbosity but in the future we may want to control verbosity of molecule itself and also other called subprocesses.


greg-hellings (approved):
Vehemently approved!

PR #2935, opened 2020-10-31
Remove molecule container building from core

ssbarnea:
Container building, testing and publishing will not be taken care of by https://github.com/ansible-community/molecule-dist project.

Main reason for this move is the fact that container building is distribution issue with specific
issues. Apparently the build process itself seems to be break quite easy for reasons outside the project (usually building some dependencies).

Decoupling of container building follows a distribution model that is also used by Ansible itself. The new project will also be able to produce other kind of packaging like debs, rpms, snaps,...


apatard (approved):
I'm not sure what's the use of it, but if it's there, I guess some people are needing that. **But**
I agree that this should be handled elsewhere. Splitting is more flexible. For instance, it will allow people who can fix any problem with it to release a fixed version without waiting for a new molecule release.

PR #2949, opened 2020-11-04
Check for successful `ansible --version` call

Matoking:
Check that the `ansible --version` command executes successfully or raise a subprocess.CalledProcessError. Otherwise , Ansible might print a message starting with "the full traceback was:" to stdout, eventually leading to the following cryptic exception:

```
packaging.version.InvalidVersion: Invalid version 'full'
```

With the check in place, the following exception is printed instead:

```
subprocess.CalledProcessError: Command ['ansible', '--version']' returned non-zero exit status 250.
```

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/lib/molecule/config.py
+++ b/lib/molecule/config.py
@@ -461,10 +461,18 @@ def ansible_version(version: str = None) -> Version:
                     stdout=subprocess.PIPE,
                     stderr=subprocess.PIPE,
                     universal_newlines=True,
+                    check=True,
                 )
                 .stdout.splitlines()[0]
                 .split()[1]
             )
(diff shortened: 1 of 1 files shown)


ssbarnea (changes requested):
You also need to update exception catch to include the new exception or it will endup with an ugly stack trace instead of the desired LOG.fatal execution path.

ssbarnea (approved):
Looks ok but I cannot merge due to the linting issue which is unrelated. I will fix it later and merge your patch. 

And thank you!

ssbarnea:
Obsoleted byt https://github.com/ansible-community/molecule/pull/2956

PR #2957, opened 2020-11-08
fix CI docs of GitHub Actions

Yuhta28:
PR Type
Docs Pull Request
Hello.
I fix the former a pull  request(https://github.com/ansible-community/molecule/pull/2947)
I am not used to Git Hub, and I would try a pull request again. 
I remove sample code of Git Hub Actions doc.
Instead I put a link to https://github.com/ansible-community/molecule-docker/blob/master/.github/workflows/tox.yml#L67-L68 based on the advice.
Thank you.

### DIFF
--- a/docs/ci.rst
+++ b/docs/ci.rst
@@ -38,7 +38,11 @@ and run ``molecule test`` in ubuntu.
             python-version: ${{ matrix.python-version }}
         - name: Install dependencies
           run: |
-            sudo apt install docker
+            sudo apt install apt-transport-https ca-certificates curl gnupg-agent software-properties-common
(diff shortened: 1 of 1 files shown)


ssbarnea (changes requested):
I think you would better rewrite the example to make it use the already built container with molecule: https://quay.io/repository/ansible/toolset

ssbarnea:
Closing as I do not see it adding any value and the phrasing seems problematic. I am not against updating the docs on this side but at least we need to send the user to an working example. I would recommend adding this example to https://github.com/ansible-community/toolset where molecule would be run from inside the builded container.

Issue #2961, opened 2020-11-10
Switch inventory for testinfra from inventory_file to inventory_directory


amarao:
# Issue Type
- Feature request

# Molecule and Ansible details
```
ansible 2.9.11
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/amarao/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/amarao/venvs/ansible29/lib/python3.8/site-packages/ansible
  executable location = /home/amarao/venvs/ansible29/bin/ansible
  python version = 3.8.6 (default, Sep 25 2020, 09:36:53) [GCC 10.2.0]
molecule 3.2.0a0
    ansible:2.9.11 python:3.8
    delegated:3.2.0a0 from molecule
    libvirt:0.0.3 from molecule_libvirt
    openstack:0.1 from molecule_openstack
```

installation method
- pip

# Desired Behavior
Right now when testinfra is called, it's called with `self._config.provisioner.inventory_file`. This cause ansible-inventory (inside testinfra) to read inventory file. If inventory has links, `group_vars` and `host_vars` works fine, but `hosts` link is ignored. This creates a lot of problem with accessing variables from other hosts.

Consider this:
host_vars (or group_vars) for the instance:

```yaml
hypervisor: '{{ groups.hypervisor[0] }}'
hypervisor_ip: '{{ hostvars[hypervisor].ansible_host }}'
hypervisor_user: '{{ hostvars[hypervisor].ansible_user }}'
ansible_ssh_extra_args: '-J {{ hypervisor_user }}@{{ hypervisor_ip }}'
```
and this inventory:
```yaml
provisioner:
  name: ansible
  inventory:
    links:
      hosts: ${TEST_BENCH_INVENTORY:-../../inventories/test_bench.yaml}
      group_vars: group_vars/
      host_vars: host_vars/
```
when converge playbook is called, it has access to other hosts inside `inventory_directory` (so I can uses `hostvars[hypervisor]`), but when testinfra is called, it can't see `hosts` link inside inventory directory, because right now testinfra is called with `self._config.provisioner.inventory_file`.

#Proposal

Change is to `self._config.provisioner.inventory_directory`.

# The problem
Right now testinfra is called without specifying hosts to test. If this change is made, testinfra would try to 'test' excessive hosts (like hypervisor in my case). This can be offset by passing explicit list of hosts to test (`--hosts` option) or by using special group for testing purposes.

[...]

PR #2962, opened 2020-11-10
Switch from inventory file to directory for testinfra call

amarao:
#### PR Type
- Feature Pull Request

It should allow to use hostvars of non-managed hosts inside testinfra call (including delegation). See Issue #2961 for details.

### DIFF
--- a/lib/molecule/driver/base.py
+++ b/lib/molecule/driver/base.py
@@ -73,7 +73,7 @@ class Driver(object):
         """
         return {
             "connection": "ansible",
-            "ansible-inventory": self._config.provisioner.inventory_file,
+            "ansible-inventory": self._config.provisioner.inventory_directory,
         }
 
     @abc.abstractproperty
--- a/lib/molecule/test/unit/driver/test_delegated.py
+++ b/lib/molecule/test/unit/driver/test_delegated.py
@@ -63,7 +63,7 @@ def test_config_private_member(_instance):
 def test_testinfra_options_property(_instance):
     assert {
         "connection": "ansible",
-        "ansible-inventory": _instance._config.provisioner.inventory_file,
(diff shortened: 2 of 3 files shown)


ssbarnea (approved):

ssbarnea:
@amarao Does this break users that do not have a directory? Ansible allows both files or directory approach and if we implement support for folders while breaking the flat-inventory users we are in bit trouble, especially because the number of users using simple inventories is greater.

cloudnull:
:+1: to what @ssbarnea said, I'd like know this doesn't break existing use-cases.

cloudnull (changes requested):
can we configure the provisioner with both file and directory support?

amarao:
@ssbarnea, When molecule constructs inventory, it writes `ansible_inventory.yaml`.  Current behavior: the path to this inventory fi is then passed to testinfra: (`self._config.provisioner.inventory_file`). Proposed behavior: write the same file, but pass the directory (`self._config.provisioner.inventory_directory,`) with `ansible_inventory.yaml`  inside as inventory.

As far as I understand ansible, there is no difference between inventory file and an inventory directory with the same file inside. Ansible  reads all yaml files from the inventory directory, so 'inventory file' is used in both cases. The difference is only if there are additional files. (This is the case with 'linked' hosts and group/hostvars directories). Group vars and Host vars are not affected, because host_group_vars plugin is searching them in both cases (file and directory).

If user used only constructed inventory (no links), then Ansible will read directory with a single file, no issues.
If user used links, than Ansible (under testinfra) start to see those links. (This is the reason for patch - I want to access data from `hosts` link).

The single issue I found is that tests may start to run against linked host, but it's easily solved:
```
verifier:
  name: testinfra
  options:
    force-ansible: true
    hosts: vms
```
(I have 'hypervisor' linked host and vms are provisioned and tested).

PR #2966, opened 2020-11-15
Prevent Ansible args from leaking to create and destroy

### LINKED ISSUE #2963: ansible_args passed to all steps
# Issue Type

- Bug report

```
molecule 3.2.0a0
    ansible:2.9.7 python:3.8
    delegated:3.2.0a0 from molecule
    podman:0.3.0 from molecule_podman
    vagrant:0.5 from molecule_vagrant
```

- pip

- OS package

# Desired Behavior

I have a particular use case where I want to test a full playbook and it's roles. However, the playbook has to be called with specified tags. I would call the playbook in production like this:
```
ansible-playbook site.yml -t install
```
Therefore, I configured my `molecule.yml` like this:
```
---
dependency:
  name: galaxy
driver:
  name: vagrant
platforms:
  - name: opensuse15
    box: generic/opensuse15
  - name: debian10
    box: generic/debian10
provisioner:
  name: ansible
  ansible_args:
   - --tags=install
  playbooks:
    converge: ../../site.yml
verifier:
  name: ansible
```
I would expect molecule to pass these parameters to the converge step only. 

# Actual Behaviour
Molecule passes the tags also to create and destroy, which makes my machines never start/stop.

A good create command looks like this
```
COMMAND: ansible-playbook --diff --inventory /home/t/.cache/molecule/playbook/default/inventory --skip-tags molecule-notest,notest /home/t/.local/lib/python3.8/site-packages/molecule_vagrant/playbooks/create.yml
```
With `ansible_args` it looks like that: 
```
COMMAND: ansible-playbook --diff --inventory /home/t/.cache/molecule/playbook/default/inventory --skip-tags molecule-notest,notest --become --tags=install /home/t/.local/lib/python3.8/site-packages/molecule_default/playbooks/create.yml
```
The create playbook doesn't do anything with just this tag.

I have reprodiced this with both vagrant and podman driver.

tadeboro:
When we were removing the dependency on the sh Python package from the codebase, we inadvertently changed the behavior of the Ansible arguments that user supplies either on the command line on in the provisioner part of the scenario configuration (commit 1cefe5c21e8ef338a5769d65e4e0e3eb6b2256fc).

Changes in this commit restore the previous state when Ansible args were not passed to the create and destroy playbook executions.

Fixes #2963 

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/lib/molecule/provisioner/ansible_playbook.py
+++ b/lib/molecule/provisioner/ansible_playbook.py
@@ -67,14 +67,20 @@ class AnsiblePlaybook(object):
             if options.get("become"):
                 del options["become"]
 
-        ansible_args = list(self._config.provisioner.ansible_args) + list(
-            self._config.ansible_args
-        )
-
-        # if ansible_args:
-        #     if self._config.action not in ["create", "destroy"]:
-        #         # inserts ansible_args at index 1
-        #         self._ansible_command.cmd.extend(ansible_args)
+        # We do not pass user-specified Ansible arguments to the create and
+        # destroy invocations because playbooks involved in those two
+        # operations are not always provided by end users. And in those cases,
+        # custom Ansible arguments can break the creation and destruction
+        # processes.
+        #
+        # If users need to modify the creation of deletion, they can supply
+        # custom playbooks and specify them in the scenario configuration.
+        if self._config.action not in ["create", "destroy"]:
+            ansible_args = list(self._config.provisioner.ansible_args) + list(
+                self._config.ansible_args
+            )
+        else:
+            ansible_args = []
 
         self._ansible_command = util.BakedCommand(
             cmd=[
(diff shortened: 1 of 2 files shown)


ssbarnea (approved):

ssbarnea:
I knew it! I deliberately simplified the code because I did not see why these should not be passed and because it did not break any tests.... so blame me for introducing that bug.

I approved the fix but I will let you decide if you want to merge it like this or if you can also add two bits to it:
* a small comment before the if explaining why we do not want args for create/destroy - why they are special
* a test that would fail if someone removes this code (if ansible args do slip in, again)

tadeboro:
>     * a small comment before the if explaining why we do not want args for create/destroy - why they are special

I will add this comment to the code.

>     * a test that would fail if someone removes this code (if ansible args do slip in, again)

The test is already there (https://github.com/ansible-community/molecule/blob/94624bcd5f119f5553b7ab0edd3855307bae01f4/lib/molecule/test/unit/provisioner/test_ansible_playbook.py#L109), but it failed to detect the change because we updated it by mistake. I already fixed the test, so we should be safe in the future (until we mess up the tests again, that is ;)).

ssbarnea (approved):

PR #2851, opened 2020-09-29
Enable folding sections when run under Travis

cognifloyd:
Add output when molecule is running in travis-ci to make the logs easier to jump through.
Travis log folding is explained [here](https://github.com/travis-ci/travis-ci/issues/2285).

#### PR Type
- Feature Pull Request

PR #2967, opened 2020-11-16
Add log folds/groups/sections for CI services

cognifloyd:
Add output when molecule is running in some CI services (Travis, Github Actions, GitLab Pipelines) to make the logs easier to jump through.

This is a follow up to #2851

#### PR Type

- Feature Pull Request

#### CI Docs:

- Travis log folding is explained [here](https://github.com/travis-ci/travis-ci/issues/2285), [here](https://github.com/travis-ci/docs-travis-ci-com/issues/949), and [here](https://github.com/travis-ci/travis-rubies/blob/9f7962a881c55d32da7c76baefc58b89e3941d91/build.sh#L38-L44) (CI vars documented [here](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables)).
- Github Action output groups are explained [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#grouping-log-lines) (CI vars documented [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables)).
- GitLab Pipelines collapsible sections are explained [here](https://docs.gitlab.com/13.5/ee/ci/pipelines/#custom-collapsible-sections) (CI vars documented [here](https://docs.gitlab.com/ee/ci/variables/predefined_variables.html)).

##### Travis CI:

- [x] implement based on Travis CI experience as documented in various places.
- [x] test/validate molecule with Travis CI. Tested with [this job](https://travis-ci.com/github/cognifloyd/community.mongodb/builds/187516426) on Travis (_nb molecule works just fine in this job, but the job itself failed because the tests haven't been fully ported to molecule 3_).
- [x] take a screenshot with Travis CI.
TravisCI Screenshot

##### Github Actions:
- [x] implement based on Github Actions docs.
- [x] test/validate molecule with Github Actions. Tested with [this job](https://github.com/cognifloyd/ansible-role-ssh-chroot-jail/runs/1407017062?check_suite_focus=true) on Github Actions.
- [x] take a screenshot with Github Actions.
Github Actions Screenshot

##### GitLab Pipelines:
- [x] implement based on GitLab Pipelines docs.
- [x] test/validate molecule with GitLab Pipelines. Tested with [this job](https://gitlab.com/cognifloyd/molecule_test/-/jobs/852538147) on GitLab Pipelines
- [x] take a screenshot with GitLab Pipelines.
GitLabs Pipelines Screenshot
_nb: The title in GitLab has to be all one color, or it ends up with funky output._


cognifloyd:
Tested on all 3 platforms. I think this is ready to merge.

cognifloyd:
Tested on all 3 platforms. I think this is ready to merge.

cognifloyd:
Just noticed a straggler TODO (already done) so I rewrote history and removed it.

ssbarnea:
@cognifloyd I cannot state how happy I am to see this implemented as I wanted to see it myself for years. Sadly my iterm does not yet support this feature and [very few terminals](https://stackoverflow.com/questions/52812618/ansi-escape-sequence-for-collapsing-folding-text-maybe-hierarchically) seems to support it.

i am not surprised it became popular on CI systems as doing it in HTML is quite easy.

I will be happy to list this as no1 feature of 3.2 release. Still, I would like to request few changes to prepare it for the future: mainly name the feature collapsable_sections, foldable_sections or similar and avoid naming it and the classes around "CI".

While the current implementation works only with few particular CI systems, it is not something unique to CI and sooner or later will be enabled for terminals too. Make it generic and we enable it when we detect a context that supports it, and do nothing when is not supported.

cognifloyd:
Following the chain of links:
https://gitlab.freedesktop.org/terminal-wg/specifications/-/issues/3 is a discussion around adding folding to terminal emulator spec (seems unlikely to be accepted based on the primarily negative response)
https://gitlab.freedesktop.org/terminal-wg/specifications/-/issues/4 takes a step back from the presentation (folding) to add semantic markers that could be used by a terminal emulator to add log folding

More description of the [propsal in #4](https://gitlab.freedesktop.org/terminal-wg/specifications/-/issues/4) is at http://per.bothner.com/blog/2019/shell-integration-proposal/

So there might be escape sequences at some point that can denote the start/end of output. Terminal emulators might add support for turning that into folding or collapsible sections. Support in CI environments would also need them to support that standardized sequence.

So, maybe I'll call this `mark_sections`. And I'll divide this into two utility functions:
- `get_section_marker_style() -> Optional[str]` returns the name of the section_marker_style. For now this would be one of `travis_ci_fold`, `github_action_group`, `gitlab_ci_section`, or `None`.
- `dispatch_section_markers()` very similar to the current implementation.

Or, instead of doing a function, we can actually expand this out into a more extensible command pattern.

```python
from abc import ABC, abstractmethod

class SectionMarker(ABC):
    @classmethod
    @abstractmethod
    def can_handle():
        """"Investigate environment to see if section markers are supported."""

    @abstractmethod
    def start_marker(section_id: str, title: str = None):
        """"Print the start marker. title may include markers and emoji."""

    @abstractmethod
    def end_marker(section_id: str):
        """"Print the end marker."""
```

This is a lot heavier weight in terms of objects. Do we need to go this far?

cognifloyd:
I switched the target branch to devel - I'm guessing this is for 3.2 and master is for 3.1?
I'll work on refactoring these in a bit.

ssbarnea:
@cognifloyd nope. the only branch receiving patches is the default branch (master). All the others are temporary and usually deleted after merge and reused later.

ssbarnea:
@cognifloyd Look at https://github.com/ansible-community/molecule/pull/2975/files which I created to facilitate your change.

PR #2970, opened 2020-11-17
Use src instead of lib for modules

ssbarnea:
Replace use of lib/ folder with src/ which is preferable as it causes less problems. This folder is also recognized and used by multiple tools without extra configurations.

Pros for src:
* recommended by pytest team
* auto-recognized by vscode (not further configuration needed). See: https://github.com/microsoft/pylance-release/issues/236
* is not part of default `.gitignore` repository template on github
* `lib` usually with `lib64` are used for binaries, acting as temp build dir, may confuse developers coming from other languages

Cons for src:
* Ansible project itself uses `lib/`


tadeboro (approved):
Look OK to me.

apatard (approved):
as said on irc, as long as we're sure it won't change in few months, fine for me.

PR #2975, opened 2020-11-17
Refactor execute to use wrapper for logging

ssbarnea:
Avoid each implementation of Command to call logging and use a wrapped to log executions of each subclass.

### DIFF
--- a/src/molecule/command/base.py
+++ b/src/molecule/command/base.py
@@ -37,6 +37,18 @@ MOLECULE_GLOB = os.environ.get("MOLECULE_GLOB", "molecule/*/molecule.yml")
 MOLECULE_DEFAULT_SCENARIO_NAME = "default"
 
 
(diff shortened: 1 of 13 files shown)


ssbarnea:
@cognifloyd They do not need any calls because any class inheriting from command.Base will call the print_info() from the base class. That is what this change is adding as we no longer need to call it in each implementation. `execute()` is wrapped dynamically with the logging.

"Running ..." was present with list, login, 

There are some unrelated problems with "matrix" and "check", "test" and "converge" do not produce their own sections because they are sequences, but each command within them is displayed and that was the goal.

I am not saying that the "sequences" should not print sometime but because section support in terminal is likely not nested, we would break the rendering if we would display nested ones.

We clearly need to find a way to display the sequences in a different section when we do `--help`, so we can decouple them from the basic commands.

PR #2976, opened 2020-11-17
Display each command as a collapsable section under CI

cognifloyd:
Follow up to #2967 based on #2975 

Please see #2967 for screenshots.

#### PR Type

- Feature Pull Request

### DIFF
--- a/src/molecule/console.py
+++ b/src/molecule/console.py
@@ -14,6 +14,7 @@ theme = Theme(
         "danger": "bold red",
         "scenario": "green",
         "action": "green",
+        "section_title": "bold cyan",
         "logging.level.notset": Style(dim=True),
         "logging.level.debug": Style(color="white", dim=True),
         "logging.level.info": Style(color="blue"),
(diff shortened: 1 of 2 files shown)


cognifloyd:
That's a first stab at implementing Sectioners. `get_sectioner()` is a bit ugly and is not test friendly. I'll have to refactor that later. I've spent too much time on this right now.

cognifloyd:
Reworked according to comments in #2984.
I still need to re-add tests.

cognifloyd:
Before I was wrapping `execute_subcommand()` so modifying the execute_subcommand test seemed logical, to make sure that for each of the wrapper functions, the functionality didn't change.

Doing so is not quite as obvious now.

I suppose we could decorate a mock function using `logger.get_section_loggers()` parametrized with each combo of envvars (and calling `logger.get_section_loggers.cache_clear()` between each run), and then ensure that that mock function returns correctly in each case.

Anyway, those are my thoughts, I'll come back to finish this later.

ssbarnea (approved):

PR #2978, opened 2020-11-18
Removed custom logger

ssbarnea:
Molecule no longer uses a custom logger class and we make use of the standard python logger class.

This change replaces `logger.success()` and `logger.out()` with `logger.info()` calls.

### DIFF
--- a/src/molecule/test/unit/test_config.py
+++ b/src/molecule/test/unit/test_config.py
@@ -292,9 +292,7 @@ def test_preflight_exists_when_validation_fails(
     patched_logger_critical.assert_called_once_with(msg)
 
 
-def test_validate(
-    mocker, config_instance, patched_logger_debug, patched_logger_success
-):
+def test_validate(mocker, config_instance, patched_logger_debug):
     m = mocker.patch("molecule.model.schema_v3.validate")
     m.return_value = None
 
--- a/src/molecule/verifier/testinfra.py
+++ b/src/molecule/verifier/testinfra.py
@@ -161,8 +161,6 @@ class Testinfra(Verifier):
             cmd=["pytest", *util.dict2args(options), *self._tests, *args],
             cwd=self._config.scenario.directory,
             env=self.env,
-            stdout=LOG.out,
-            stderr=LOG.error,
         )
         # print(self._testinfra_command.cmd)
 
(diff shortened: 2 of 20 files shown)


src/molecule/test/unit/test_config.py
```diff
@@ -292,9 +292,7 @@ def test_preflight_exists_when_validation_fails(
     patched_logger_critical.assert_called_once_with(msg)
 
 
-def test_validate(
-    mocker, config_instance, patched_logger_debug, patched_logger_success
-):
+def test_validate(mocker, config_instance, patched_logger_debug, patched_logger_info):
```
tadeboro, line 295: Do we really need the `patched_logger_info` fixture here? It is not used anywhere in the test (and after a quick look, `LOG.info` is also not called anywhere in the code path we are testing).

src/molecule/verifier/testinfra.py
```diff
@@ -161,7 +161,6 @@ def bake(self):
             cmd=["pytest", *util.dict2args(options), *self._tests, *args],
             cwd=self._config.scenario.directory,
             env=self.env,
-            stdout=LOG.out,
```
tadeboro, line 164: Why did we leave the `stderr` param in this case while we removed it from other bake methods?

tadeboro (commented):
I found no major issues in the code. I do have two minor questions, but that is it. Once we resolve them, things are good as far as I am concerned.

On an unrelated note: I also noticed that the `BakedCommand` is a data class during the review. Which all fine and dandy, but those are only available in python >= 3.7, and Molecule advertises compatibility with 3.6. Not sure why tests did not catch this. Unfortunately, I have no time to investigate right now,

src/molecule/verifier/testinfra.py
ssbarnea: I removed stderr also, just missed it.

Background: All the bakes are vestigial from when we used sh library and remained just to lower the size of the change that removed sh. Now that we use subprocess_tee.run to execute subprocesses we no longer need to do tricks with stdout/stderr.

I am not yet in a hurry to cleanup these bakes yet. I want to spread the refactoring across several major releases, so we lower the risks of making really bad release.

src/molecule/test/unit/test_config.py
ssbarnea: fixed

ssbarnea:
@tadeboro There is a bug that we missed to add dataclasses for py36 but our testing failed to spot it because another dependency did install it. I will add a patch just to be sure.

PR #2984, opened 2020-11-18
Enable multiple section wrappers

ssbarnea:
Moves the recently added section_logger to the logger module and enable us to use multiple wrappers. This should make
implementation of https://github.com/ansible-community/molecule/pull/2976 much easier.

### DIFF
--- a/src/molecule/logger.py
+++ b/src/molecule/logger.py
@@ -22,11 +22,13 @@
 import logging
 import sys
 from functools import lru_cache
+from typing import Callable, Iterable
 
 from enrich.console import Console
 from enrich.logging import RichHandler
 
 from molecule.console import should_do_markup, theme
+from molecule.text import underscore
 
 SUCCESS = 100
 OUT = 101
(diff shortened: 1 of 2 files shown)


ssbarnea:
@cognifloyd Should this be enough to make the section adding easy and clean?

src/molecule/logger.py
```diff
@@ -74,6 +76,29 @@ def get_logger(name=None) -> logging.Logger:
...
+        rt = func(*args, **kwargs)
+        # section close code goes here
+        return rt
+
+    return wrapper
+
+
+def get_section_loggers() -> Iterable[Callable]:
+    """Return a list of section wrappers to be added."""
+    return [section_logger]
```
cognifloyd, line 99: are you imagining a separate wrapper function for each of the output types?
```python
return [github_actions_grouper, gitlab_sectioner, travis_ci_folder, section_logger]
```
Then the environment var checks would happen during each of those?
or something more like what I had before:
```python
return [ci_section_wrapper, section_wrapper]
```

cognifloyd (commented):
Where do you want the CI-specific formatting to live? Under console.py? Under logger.py? In a utility file somewhere else?

Do you like the class-style format better? Or are you cool with the function wrapper?

Happy to adjust the PR, I'm just not sure what your vision is for a more generic impl that allows for adding terminal-specific escapes later.

Are you thinking that the CI-specific wrappers should live in enrich somewhere?

src/molecule/logger.py
ssbarnea: Add them to the logger module. I was considering console initially but due to some circular import issues I decided to go for logger. logger can import console but console does not, this being the main reason.

Yep, I did fancy adding these to enrich but I am afraid that it may not be so easy. If I am correct, there wrappers may require some parameters like section name.

For the moment just extend get_section_loggers to make it return additional wrappens when needed, add them inside the same file and also add `lru_cache()` wrapper on it for performance reasons, as we do not want it to run each time it wraps a class.

PR #2985, opened 2020-11-19
Remove mock as a dependency

clickthisnick:
unittest.mock is built into python3 so mock should not be needed.
This package is only for python3

`pytest-mock` (mocker fixture)
Also will use unittest.mock:
https://github.com/pytest-dev/pytest-mock/blob/9e1464bb87d551b1a242104b68bad8d8a1429316/src/pytest_mock/plugin.py#L22

#### PR Type
- Feature Pull Request

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -110,7 +110,6 @@ test =
     # related tools/plugins but w/o ansible, which can be installed separated.
     ansi2html >= 1.6.0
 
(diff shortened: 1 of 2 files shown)


mattp- (approved):
lgtm

ssbarnea (approved):

PR #2986, opened 2020-11-20
docs/getting-started.rst: Add missing argument to init role

apatard:
The default driver is 'delegated' nowadays and this document is
about using the 'docker' driver, so "--driver-name docker" has to
be appended to the "molecule init role" command line.

Ref: https://github.com/ansible-community/molecule/discussions/2916

#### PR Type

- Docs Pull Request

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -32,7 +32,7 @@ To generate a new role with Molecule, simply run:
 
 .. code-block:: bash
 
-    $ molecule init role my-new-role
+    $ molecule init role my-new-role --driver-name docker
 
 You should then see a ``my-new-role`` folder in your current directory.
 


ssbarnea (changes requested):
I am inclined to remove the docker references from the guide instead? This would make the guide much easier to follow. Once the user is able to run with default delegated he could look for another guide on how to do it with another driver?

apatard:
I thought about doing that too but I took the quickest way to fix the issue and made this PR.
I agree that moving this doc to a different part and provide a new "Getting Started" with the (default) delegated driver is the way to go but it will take a little bit more time. In the mean time, the current doc will still be broken.

ssbarnea (approved):

PR #2988, opened 2020-11-20
Fix connection option handling for delegated driver

apatard:
The documentation and the code tends to suggest that in order to configure the ansible connection plugin, the option to use is 'connection' but the test handling this option is testing 'ansible_connection', so fix that.

### DIFF
--- a/src/molecule/driver/delegated.py
+++ b/src/molecule/driver/delegated.py
@@ -185,7 +185,7 @@ class Delegated(Driver):
                 conn_dict["ansible_user"] = d.get("user")
                 conn_dict["ansible_host"] = d.get("address")
                 conn_dict["ansible_port"] = d.get("port")
-                if d.get("ansible_connection", None):
+                if d.get("connection", None):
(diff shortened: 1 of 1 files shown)


ssbarnea (changes requested):
This is exactly the opposite, we should always follow Ansible and adapt to its changes, so we avoid confusing users. Based on https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html we need to assure we use ansible_connection and not just connection.

OK to update the docs, ok even if it breaks some users config, we will mark it as major and explain it that we follow ansible.

apatard:
hm. not sure to follow. The change is only for the case of managed instances and here, the options are already not following ansible. See https://github.com/ansible-community/molecule/blob/master/src/molecule/driver/delegated.py#L185. If the idea is to get ansible variables by adding 'ansible_' as prefix, then my change is fine (and the identity option has possibly to be renamed). If not, then all options has to be renamed.
Am I missing something ?

ssbarnea (approved):

PR #2993, opened 2020-11-23
src/molecule/console.py: Don't always set output wrapping

apatard:
The output soft wrapping with python enrich is enabled for all
terminals. This means that lines are wrapped to 80 when molecule
is run from tox or when a module ends importing this console module.
In the later case, this will wrap its output to 80 columns and
possibly break it, since the output won't be valid JSON as found in
https://github.com/ansible-community/molecule-vagrant/issues/84.

As a possible solution, enable the soft wrapping on the same condition
as for the force_terminal terminal.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/src/molecule/console.py
+++ b/src/molecule/console.py
@@ -72,7 +72,11 @@ def should_do_markup() -> bool:
 
 
 console = Console(
-    force_terminal=should_do_markup(), theme=theme, record=True, redirect=True
+    force_terminal=should_do_markup(),
+    theme=theme,
+    record=True,
+    redirect=True,
+    soft_wrap=should_do_markup(),
 )
 # Define ANSIBLE_FORCE_COLOR if markup is enabled and another value is not
 # already given. This assures that Ansible subprocesses are still colored,


ssbarnea (changes requested):
I think the option name confused you because soft_wrap=True means do not wrap. soft_wrap=False means wrap at terminal width, default behavior on rich.

Default behavior of enrich.console.Console is to soft_wrap=True (not wrapping). Is one of the reasons I created enrich.

If is not working like this maybe there is a bug we need to fix, but we should not need to ever manually set soft_wrap=False.

apatard:
ok, so a bug in enrich ?
fwiw, it's easy to reproduce it with: https://gist.github.com/apatard/d8ba985b9f48ba908057a726549dbbc2
It's producing:
```
python3 ./myexec.py 
11111111111111111111111111111111111111111111111111111111111111111111111111111111
11111111111111111111111
```
This leads to tox output looking like:
```
ANSIBLE_CALLBACK_WHITELIST=profile_tasks,timer ANSIBLE_COLLECTIONS_PATH=/home/va
grant/.cache/molecule/test-init/test-scenario/collections:/home/vagrant/.ansible
/collections:/usr/share/ansible/collections:/etc/ansible/collections
ANSIBLE_CONFIG=/home/vagrant/.cache/molecule/test-init/test-scenario/ansible.cfg
```
and also breaks molecule-vagrant

apatard:
Fixed with enrich 1.2.3, so this PR is useless. I'm closing it.

PR #2996, opened 2020-11-26
Assure provisioner logging is enabled by default

ssbarnea:
This fixes confuses where creating instances fail without details.
This change will not affect users that already mentioned their
preferences in config or as environment variable.

Related: https://github.com/ansible-community/molecule-podman/issues/22


apatard (approved):
Depending on no default value is quite annoying and with default to false would make MOLECULE_NO_LOG useless. So, fine for me.

Issue #2997, opened 2020-11-28
Molecule should auto install tested collection during dependency step


ssbarnea:
# Issue Type

- Feature request

# Desired Behavior

When the user creates scenarios inside a collection they also need to install the collection itself before running molecule or the tests would likely fail to access roles and modules from within the collection.

Until now users either manually added ansible-galaxy install the local collection with something like:
```
    sh -c "rm -rf dist/* .cache/collections/*"
    # build collection
    ansible-galaxy collection build -v -f --output-path dist/
    # install collection in custom location
    sh -c "ansible-galaxy collection install -f dist/*.tar.gz -p .cache/collections"
```

Clearly this is not something convenient and we want to have a seamless experience where molecule scenarios can be run right away without extra user interventions. The tool can easily detect the presence of a collection repository by cheking presence of the `galaxy.yml` file.

It should be noted that Molecule itself already knowns to isolated ansible collections and roles paths to scenario specific folders, so installing these should not pollute user folders.

At this moment dependency already knows to install dependencies but does nothing about local collections, the one being subject to the test.

Sadly Ansible does not provide any way to perform editable installs for collections, so we will need to install it each time forced.

ssbarnea:
Molecule already does this, closing.

provenvelocity:
FYI the solution of using ansible-lint does not work with the latest release of ansible. I converted my requirements to version pinning to force working on the latest and not get behind... famous last words.  

The run with the requirements fails.
ansible
docker
molecule
molecule-docker
molecule-podman
molecule-vagrant
python-vagrant
pytest
pytest-testinfra
yamllint
flake8

But running with this does work:
ansible==3.4.0
docker==5.0.0
molecule==3.3.0
molecule-docker==0.2.4
molecule-podman==0.3.0
molecule-vagrant==0.6.3
pytest==6.2.4
pytest-testinfra==6.3.0
python-vagrant==0.5.15
yamllint==1.26.1
flake8==3.9.2

provenvelocity:
@ssbarnea

PR #2998, opened 2020-11-28
Automatically install current collection

ssbarnea:
If run inside a collection repository dependency step will first
build and install the current collection. This avoids failure to
run test scenarios unless user already installed the collection.

Fixes: #2997

### DIFF
--- a/src/molecule/dependency/base.py
+++ b/src/molecule/dependency/base.py
@@ -22,7 +22,10 @@
 import abc
 import logging
 import os
+import re
 import time
+from pathlib import Path
+from subprocess import CalledProcessError
 
 from molecule import constants, util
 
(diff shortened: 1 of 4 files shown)


src/molecule/dependency/base.py
```diff
@@ -80,13 +82,32 @@ def execute_with_retries(self):
...
 
-    @abc.abstractmethod
-    def execute(self):  # pragma: no cover
+    def execute(self):
         """
         Execute ``cmd`` and returns None.
 
         :return: None
         """
+        if os.path.isfile("galaxy.yml"):
```
tadeboro, line 91: This check is too naive to be effective. For example, we have our molecule scenarios stored under `tests/integration/molecule`, which means that we run Molecule from the `tests/integration` collection subfolder where galaxy.yml file will never be found.

src/molecule/dependency/base.py
```diff
@@ -80,13 +82,32 @@ def execute_with_retries(self):
...
+            LOG.info("Collection detected at %s", os.getcwd())
+
+            dist = Path(self._config.scenario.ephemeral_directory) / "dist"
+            dist.mkdir(parents=True, exist_ok=True)
+            result = util.run_command(
+                f"ansible-galaxy collection build -v -f --output-path {dist}",
```
tadeboro, line 97: Again, this build command is too naive because it assumes that we are running molecule from the collection root. And there is another problem: users might need to some something else before they build the collection artifact. Because `ansible-galaxy collection build` command is really primitive, there is usually quite a few steps we need to perform before we get to building (or we risk packaging together things like virtual environments, coverage reports - and the are huge BTW, etc.).

tadeboro (changes requested):
As already stated on IRC, I have a hard time justifying this functionality in the Molecule. I would much rather document how to set up a collection development environment and then let molecule run ansible from within the collection where it can automatically use the content without installing it.

src/molecule/dependency/base.py
ssbarnea: I will address this because molecule already looks for repository root location as is used in config file lookup, so we should be able to re-use repo-root path and avoid depending on cwd.

greg-hellings:
This will address an issue I am currently encountering where trying to get Molecule to pick up an existing collection is difficult but also causes any other collections on my development machine to be overwritten by the ansibile-galaxy command.

greg-hellings (approved):
This would be a wonderful improvement to my current development cycle.

src/molecule/dependency/base.py
```diff
@@ -80,13 +83,41 @@ def execute_with_retries(self):
...
-    @abc.abstractmethod
-    def execute(self):  # pragma: no cover
+    def execute(self):
         """
         Execute ``cmd`` and returns None.
 
         :return: None
         """
+        workdir = util.find_vcs_root(self._config.project_directory)
+        collection_marker = os.path.join(workdir, "galaxy.yml")
```
greg-hellings, line 93: What if `galaxy.yml` doesn't live at the root of the current VCS?

MarkusTeufelberger:
I think there first needs to be some consensus (and documentation) on where the `molecule` folder should end up in a collection repository.

ssbarnea:
Closing in favour of https://github.com/ansible-community/molecule/pull/3077 which is reusing ansible-lint prerun code.

Issue #2999, opened 2020-11-28
Debug output on all molecule commands


mdpdesign:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible 2.10.3
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/mdp/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/mdp/ansible-roles/molecule-venv/lib/python3.6/site-packages/ansible
  executable location = /home/mdp/ansible-roles/molecule-venv/bin/ansible
  python version = 3.6.9 (default, Oct  8 2020, 12:12:24) [GCC 8.4.0]
molecule 3.2.1.dev1 using python 3.6
    ansible:2.10.3
    delegated:3.2.1.dev1 from molecule
    vagrant:0.5.dev1 from molecule_vagrant
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

# Desired Behavior

Molecule should output at most the INFO log messages when running commands without `--debug`

# Actual Behaviour

For all commands like list, create, verify molecule outputs some DEBUG log along side the INFO without explicitly using `--debug` switch. IMHO debug output should be only showed when using `--debug` switch - unless this is intended?

For command: `molecule list`

```
DEBUG    Validating schema /home/mdp/ansible-roles/jenkins_swarm/molecule/default/molecule.yml.
DEBUG    Validating schema /home/mdp/ansible-roles/jenkins_swarm/molecule/service/molecule.yml.
DEBUG    Validating schema /home/mdp/ansible-roles/jenkins_swarm/molecule/vmware/molecule.yml.
INFO     Running default > list
INFO     Running service > list
INFO     Running vmware > list
                    ╷             ╷                  ╷               ╷         ╷
  Instance Name     │ Driver Name │ Provisioner Name │ Scenario Name │ Created │ Converged
╶───────────────────┼─────────────┼──────────────────┼───────────────┼─────────┼───────────╴
  win10             │ vagrant     │ ansible          │ default       │ false   │ false
  win12             │ vagrant     │ ansible          │ default       │ false   │ false
  win10-svc         │ vagrant     │ ansible          │ service       │ false   │ false
  win12-svc         │ vagrant     │ ansible          │ service       │ false   │ false
  jsw-win10-schtask │ delegated   │ ansible          │ vmware        │ true    │ true
[...]

PR #3000, opened 2020-11-28
Refactor logging

tadeboro:
Molecule reinvented the logging for some reason and did not use the standard Python logging facilities as they are meant to be used. This PR brings some sanity to the logging part of the code.

See commit messages for more detailed change descriptions.

#### PR Type

- Bugfix Pull Request

Fixes: #2999

### DIFF
--- a/src/molecule/api.py
+++ b/src/molecule/api.py
@@ -1,16 +1,16 @@
 """Molecule API Module."""
 
+import logging
 import traceback
 from collections import UserList
 
 import pluggy
 
-from molecule import logger
 from molecule.driver.base import Driver  # noqa
 from molecule.util import lru_cache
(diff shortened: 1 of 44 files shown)


ssbarnea (changes requested):
I see a regression: "Found config file" which is printed before does no longer happen with this patch.

ssbarnea (changes requested):
Can you please fix the conflicts and ping me when this seams ready for review? (no known undesired side-effects). I will have to test it manually to assure we do not mess the logging even more with it but I need your confirmation first.

tadeboro:
I rebased the PR and tested it again locally. The only visible change of behavior from the master branch now should be the suppression of debug messages when running Molecule with no verbosity or debug flags.

ssbarnea (approved):

PR #3016, opened 2020-12-20
POC: Use a  pinned extra for certified dependencies

ssbarnea:
Save a list of pinned dependencies inside an extra named "pinned" in
order to allow users to install an exact version that we knew it
working at the build time.

I should enable user to produce more controlled installations and avoid cases where a dependency released after our package would break it, as this would not be allowed when using the pinned extra: 
```
pip install 'molecule[pinned]'
```

Known missing bits:
- setup.cfg extra is not updated based on requirements.txt
- we are missing a pre-commit hook for checking that the extra does
  not get out of sync.
- not included in any CI worflow


decentral1se (commented):
Well that's a nice idea.

From a maintenance point of view, this feels like duplication of effort for what is more or less offered by Docker.

PR #3029, opened 2021-01-07
Support multiple base configs

lebauce:
Allow specifing the --base-config option multiple times.
The `base-config` option is really useful to avoid duplication but, in a complicated setup,
we often end up having multiple `base-config` files with a lot of duplication between these files.

#### PR Type

- Feature Pull Request


decentral1se (commented):
Implementation looks legit. 

I don't even want to know why you need multiple base configs :running_man:

src/molecule/shell.py
```diff
@@ -88,14 +88,16 @@ def print_version(ctx, param, value):
 @click.option(
     "--base-config",
     "-c",
-    default=LOCAL_CONFIG,
+    multiple=True,
+    default="LOCAL_CONFIG",
```
tadeboro, line 92: Is there a reason for changing the contents of the default value? I would expect that the default value should be `[LOCAL_CONFIG]` now, but to be honest, I do not know enough about click to be 100% sure.

tadeboro (changes requested):
We also have a quite complex setup and would probably benefit from this change.

That being said, there is one area that definitely needs improvement in this PR: documenting each base config file's priority. Right now, help text only states that we can use multiple base config files, but it does specify what happens if two of them contain the same configuration options with different values. Reading through the implementation, the last one wins, but I do not know enough about click to be certain if argument parsing preserves the order of options.

I added one comment about the default value that needs to be resolved before merging.

src/molecule/shell.py
lebauce: There are absolutely zero reason to change the content, it's a typo. Sorry about that

ssbarnea (approved):

tadeboro (approved):
Thank you, @lebauce, for your contribution!

PR #3030, opened 2021-01-07
Fix base config default when not using vcs

lebauce:
When not using a VCS, LOCAL_CONFIG can be None

#### PR Type

- Bugfix Pull Request


tadeboro (approved):
Thank you for your contribution, @lebauce.

PR #3033, opened 2021-01-12
Allow including YAML files

lebauce:
When using multiple scenarios with many platforms, we either need to duplicate the entries in every scenario or provide a base config. The latter does not really provide fine-grained granularity. Being able to include YAML files allows to share only some specific sections between scenarios.

#### PR Type

- Feature Pull Request

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -306,6 +306,9 @@ project.
 Sharing Across Scenarios
 ========================
 
+Sharing whole files
+-------------------
+
 Playbooks and tests can be shared across scenarios.
 
 .. code-block:: bash
@@ -354,6 +357,30 @@ In this second example the actions `create`, `destroy`,
         prepare: ../resources/playbooks/prepare.yml
 
 .. _parallel-usage-example:
+
+Sharing YAML sections
+---------------------
+
+Portions of YAML files can be isolated and included in
+other files. For instance, you can create a file ``provisioner.yml``:
+
+.. code-block:: yaml
+
+    name: ansible
+    playbooks:
+      create: ../resources/playbooks/create.yml
+      destroy: ../resources/playbooks/destroy.yml
+      converge: ../resources/playbooks/converge.yml
+      prepare: ../resources/playbooks/prepare.yml
+
+The file can therefore be used in a scenario using:
+
+.. code-block:: yaml
+
+    provisioner: !include provisioner.yml
+
+Molecule will look for files relatively to the directory
+of the file that is currently being loaded.
 
 Running Molecule processes in parallel mode
 ===========================================
--- a/src/molecule/config.py
+++ b/src/molecule/config.py
@@ -23,7 +23,8 @@ import copy
 import functools
 import logging
 import os
-from typing import Callable, MutableMapping, TypeVar
+import pathlib
+from typing import Callable, MutableMapping, Optional, TypeVar
 from uuid import uuid4
 
 from packaging.version import Version
@@ -294,7 +295,8 @@ class Config(object, metaclass=NewInitCaller):
         for base_config in base_configs:
             with util.open_file(base_config) as stream:
                 s = stream.read()
-                self._preflight(s)
+                include_path = pathlib.Path(os.path.dirname(base_config))
+                self._preflight(s, include_path=include_path)
                 interpolated_config = self._interpolate(s, env, keep_string)
                 defaults = util.merge_dicts(
                     defaults, util.safe_load(interpolated_config)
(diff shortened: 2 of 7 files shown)


src/molecule/config.py
```diff
@@ -416,9 +418,11 @@ def _get_defaults(self) -> MutableMapping:
             },
         }
 
-    def _preflight(self, data: MutableMapping):
+    def _preflight(self, data: MutableMapping, include_path: str = ""):
```
decentral1se, line 421: Can that be a `pathlib.Path`? Not sure.

src/molecule/model/schema_v3.py
```diff
@@ -367,9 +367,9 @@ def _validate_molecule_env_var(self, molecule_env_var, field, value):
                 self._error(field, msg)
 
 
-def pre_validate(stream, env: MutableMapping, keep_string: str):
+def pre_validate(stream, env: MutableMapping, keep_string: str, include_path=""):
```
decentral1se, line 370: Missing type declaration like you included elsewhere?

src/molecule/util.py
```diff
@@ -224,29 +255,50 @@ def safe_dump(data: Any, explicit_start=True) -> str:
     )
 
 
-def safe_load(string) -> Dict:
+def new_safe_loader_with_include(import_path: str):
```
decentral1se, line 258: When is "new"? Today yes but tomorrow not. I would just go with `safe_loader_with_include`

decentral1se (commented):
Overall I think this is a good change! Thanks. I am a bit wary to see re-implementation of a feature that YAML supports but since we do generation we need to do it ourselves? Maybe I don't understand it well and you can explain better. Is there anyway to avoid this? Perhaps not. In this case, then I think the implementation looks good. Left some minor comments.

Some documentation would be good.

src/molecule/model/schema_v3.py
lebauce: Fixed

src/molecule/util.py
lebauce: True :-) Fixed

src/molecule/config.py
lebauce: Changed to pathlib.Path

lebauce:
@decentral1se Thank you for your review. I'm not sure to understand what you mean by "re-implementation of a feature that YAML supports". AFAIK, including other files is not supported by YAML.

I tried to address your comments and added some doc.

decentral1se:
> AFAIK, including other files is not supported by YAML.

Ahhhh, OK, I thought it did that.

Good then, looking good, thanks for thrashing out the comments.

Let's see what the rest say.

docs/examples.rst
```diff
@@ -355,6 +358,30 @@ In this second example the actions `create`, `destroy`,
...
+.. code-block:: yaml
+
+    name: ansible
+    playbooks:
+      create: ../resources/playbooks/create.yml
+      destroy: ../resources/playbooks/destroy.yml
+      converge: ../resources/playbooks/converge.yml
+      prepare: ../resources/playbooks/prepare.yml
+
+The file and therefore be used in a scenario using:
```
larmbuster, line 376: Should the "and" be a "can" instead?

lebauce: Indeed. Fixed, thanks

decentral1se (approved):

larmbuster (approved):

ssbarnea:
With all due respect, I am inclined to refuse this feature as it would introduce a serious amount of complexity and prevent us from doing other changes to molecule scenarios. At least now they are contained YAML files. Adding include opens a pandora box.

I would rather wait a little bit more and see if there are other ways to improve the user experience without this complexity.

decentral1se:
> prevent us from doing other changes to molecule scenarios

Which changes?

> Adding include opens a pandora box

Having a limited form of this feature which is experiimental (weve done this before, remember parallel builds?) and not publicly documented seems to me a fine way to test this out! Some limited form of inclusion can be worth it, I think.

ssbarnea (changes requested):
This will prevent use of simple (JSON/YAML) schema based validation for the file format and make almost impossible to integrate it with editors.

Molecule already supports layering config by loading the shared config, which should be enough for the moment (limited to only one shared config per repo).

I would be more inclined to support an approach that would allow to mention alternative file to be inherited from than adding a custom YAML extension.

I do want to keep using a simple YAML parsing, one without extensions.

That is why I would find a feature like `extends: filename` a more appropriate approach, especially as we can consider `.config/molecule/config.yml` to be the implicit value when the key is absent.

lebauce:
@ssbarnea I think that adding YAML files at the parsing phase is somehow less complex that loading YAML files using variable interpolation (like for playbooks for instance). Allowing to specify external files (https://molecule.readthedocs.io/en/latest/examples.html#sharing-across-scenarios) requires adding specific code for every section.

lebauce:
@ssbarnea Should I close the PR ?

PR #3077, opened 2021-03-19
Make molecule reuse prerun from ansible-lint

ssbarnea:
As ansible-lint is able to detect repository layout and run prepare ansible from running it, we rely on its logic in order to assure ansible is able to run properly.

This means that dependencies will be installed into .cache folder and that the role import path will be updated to include it.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -6,6 +6,27 @@ Configuration
 
 .. _variable substitution:
 
+
+Prerun
+------
+
+In order to help Ansible find used modules and roles, molecule will perform
+a prerun set of actions. These involve installing dependencies from
+``requirements.yml`` specified at project level, install a standalone role
+or a collection. The destination is ``project_dir/.cache`` and the code itself
+is reused from ansible-lint, which has to do the same actions.
+
+This assures that when you include a role inside molecule playbooks, Ansible
+will be able to find that role, and that the include is exactly the same as
+the one you are expecting to use in production.
+
+If for some reason the prerun action does not suits your needs, you can still
+disable it by adding `prerun: false` inside the configuration file.
+
+Keep in mind that you can add this value to ``.config/molecule/config.yml``
+file in order to avoid adding it to each scenario.
+
+
 Variable Substitution
 ---------------------
 
--- a/src/molecule/command/base.py
+++ b/src/molecule/command/base.py
@@ -28,6 +28,7 @@ import shutil
 from typing import Any, Callable
 
 import click
+from ansiblelint.prerun import prepare_environment
 from click_help_colors import HelpColorsCommand, HelpColorsGroup
 
 import molecule.scenarios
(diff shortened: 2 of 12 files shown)


src/molecule/command/base.py
```diff
@@ -90,6 +91,8 @@ def execute_cmdline_scenarios(scenario_name, args, command_args, ansible_args=()
     :returns: None
 
     """
+    prepare_environment()
```
apatard, line 94: I've not checked but will this properly fill ansible-lint options ? If someone is having:
```
skip_list:
- 'role-name'
```
Will this be properly handled ?

ssbarnea: Yep. I do have a WIP change on my fork of your repo, I used it as a model. That is how I had to make two more patches for the linter :P

src/molecule/data/validate-dockerfile.yml
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env ansible-playbook
 ---
 - hosts: localhost
+  collections:
+    - community.docker
```
apatard, line 5: btw, I may be missing something but shouldn't all these docker-related changes be in a different PR ?

ssbarnea: It is linked, once I added ansible-lint as a direct dependency and moved ansibe-base, I ended up with dockerfile build failing due to missing docker_image module.

Still, I will try to do the changes around that in a preparatory change, so we make this PR smaller. There is still more work to be done on this PR.

tadeboro (approved):
I tested changes on our collections and it works just fine. Having to install ansible-lint just for setting up the test run is a bit of a bummer, but you cannot always win I guess.

docs/configuration.rst
```diff
@@ -6,6 +6,27 @@ Configuration
 
 .. _variable substitution:
 
+
+Prerun
+------
+
+In order to help Ansible find used modules and roles, molecule will perform
+a prerun set of actions. These involve installing dependencies from
+``requiresments.yml`` specified at project level, install a standalone role
```
apatard, line 15: typo on ``requirements.yml`` ?

docs/configuration.rst
```diff
@@ -6,6 +6,27 @@ Configuration
...
+or a collection. The destination is ``project_dir/.cache`` and the code itself
+is reused from ansible-lint, which has to do the same actions.
+
+This assures that when you include a role inside molecule playbooks, Ansible
+will be able to find that role, and that the include is exactly the same as
+the one you are expecting to use in production.
+
+If for some reason the prerun action does not suits your needs, you can still
+disabled it by adding `prerun: false` inside the configuraiton file.
```
apatard, line 24: configuration

apatard: 'still disable' ?

apatard:
I've done a quick test. While it doesnt solve my bug, this change make molecule behaves in the same way as ansible-lint,
which is nice. The ``prerun`` option is working as expected too.
I found some typos in the doc. Except for the earliers comments, I'm happy with this

PR #3080, opened 2021-03-23
Enable codespell

ssbarnea:
Also includes a bunch of typo fixes.


apatard (approved):
I'm surprised it was not already the case. Sound a great idea.

ssbarnea:
> I'm surprised it was not already the case. Sound a great idea.

The linter already had this enabled, but not molecule.

Issue #3045, opened 2021-02-03
molecule does not find `module_utils` of tested ansible role


ltalirz:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version
```

ansible 2.10.6
molecule 3.2.3

[See here for CI logs](https://github.com/dokku/ansible-dokku/pull/98/checks?check_run_id=1822645321#step:5:147)

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

# Desired Behavior

When testing an [ansible role that includes a `module_utils` folder](https://github.com/ltalirz/ansible-dokku/tree/issue_88_clone_version), a python module inside this folder (e.g. `dokku_app_utils.py`) should be importable from other ansible modules via:
```python
from ansible.module_utils import dokku_app_utils
```

# Actual Behaviour

[This role](https://github.com/ltalirz/ansible-dokku/tree/issue_88_clone_version) uses a `module_utils` folder to share code between ansible modules it provides.
Running the linked branch directly using ansible 2.10.6 on a server works fine.

When testing the role using molecule, I run into the following error (see  [CI logs](https://github.com/dokku/ansible-dokku/pull/98/checks?check_run_id=1822645321#step:7:684)):
```
fatal: [instance]: FAILED! => {"msg": "Could not find imported module support code for ansible.modules.dokku_clone.  Looked for (['ansible.module_utils.dokku_app_utils.dokku_app_present', 'ansible.module_utils.dokku_app_utils'])"}
```

For some reason that I haven't figured out, the `module_utils` folder is found when running the role directly via ansible, but not found when the role runs via molecule.

I've tried 
 - making the name of the module unique
 - adding `__init__.py`s

but that didn't help (and did not seem likely, since the role runs fine locally).

Any ideas?

P.S. Nice work on the issue template selector! It's a great way to communicate all the available communication channels.

ltalirz:
Oh wait - I just realized that the issue must be specific to the `verify` stage. The python module in question is actually used  (successfully) during `converge`.

ltalirz:
Issue persists with ansible 3.3.0 and molecule[docker] 3.3.0.

tadeboro:
I took a look at this issue and as far as I can tell, Molecule is actually doing too much here, masking the real problem.

When I execute the `molecule --debug verify` command, I see this in the output (I manually split the things across multiple lines for convenience; this is actually a single line):

```
[...]

PR #3102, opened 2021-05-06
Add default for `ANSIBLE_MODULE_UTILS` variable

ltalirz:
Hi, this is a try at fixing #3045  - i.e. `molecule` not finding ansible modules defined in a role's `module_utils` folder.
See also the [ansible docs](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-module-utils-path).

I am not yet entirely sure whether this fixes the issue - in the role where I encountered this issue I have tried setting the `ANSIBLE_MODULE_UTILS` environment variable manually in the [molecule.yml](https://github.com/dokku/ansible-dokku/pull/98/commits/e4306e6b449a966961bbe703f5367782547855a5) and that did not seem to resolve it.

Nevertheless it seems to me that this should be the right path.
Please let me know whether you agree, and what kind of tests would need to be added here.

#### PR Type

- Bugfix Pull Request
- Feature Pull Request


tadeboro:
https://github.com/ansible-community/molecule/issues/3045#issuecomment-835518824 contains the proper direction for fixing this.

PR #3105, opened 2021-05-09
Make sure we detect dependency installation failure

### LINKED ISSUE #3070: Add support for checking exit codes on shell dependencies
# Issue Type

- Feature request

```
ansible 2.10.5
  config file = /Users/jhg03a/<redacted>/ansible.cfg
  configured module search path = ['/Users/jhg03a/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.9.2 (default, Feb 24 2021, 13:26:09) [Clang 12.0.0 (clang-1200.0.32.29)]
molecule 3.2.3 using python 3.9
    ansible:2.10.5
    delegated:3.2.3 from molecule
    docker:0.2.4 from molecule_docker
```

- pip

- pip

# Desired Behavior

Currently it appears that the dependency shell module doesn't take into account the exit code from the command.  If something goes wrong in the dependency phase, it's highly likely the rest of the run is going to fail or have inconsistent results.

### Example:
```yaml
dependency:
  name: shell
  command: 'false'
```

tadeboro:
Right now, we just ignore the failure because we do not check the return code of the children's process. This commit just makes sure we are notified when the dependency command fails.

Fixes #3070

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/src/molecule/dependency/base.py
+++ b/src/molecule/dependency/base.py
@@ -23,8 +23,9 @@ import abc
 import logging
 import os
 import time
+from subprocess import CalledProcessError
 
-from molecule import constants, util
+from molecule import util
 
 LOG = logging.getLogger(__name__)
 
@@ -53,11 +54,11 @@ class Base(object):
 
         try:
             # print(555, self._sh_command)
-            util.run_command(self._sh_command, debug=self._config.debug)
+            util.run_command(self._sh_command, debug=self._config.debug, check=True)
             msg = "Dependency completed successfully."
             LOG.info(msg)
             return
-        except Exception:
+        except CalledProcessError:
             pass
 
         for counter in range(1, (self.RETRY + 1)):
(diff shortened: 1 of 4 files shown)


ssbarnea (approved):

fourstepper:
In GitLab CI, it is quite easy to pull dependencies from gitlab over HTTPs without a secret by setting the git config like this: `git config --global url."https://gitlab-ci-token:${CI_JOB_TOKEN}@gitlab.com/".insteadOf https://gitlab.com/`

As I have this set, this PR is a regression for me, as molecule just fails on dependency resolution, when before I could just enter away at it and it would pick up roles downloaded through a different requirements file (over SSH)

Is it possible to change the behaviour to either not make this default or make it optional to check successful dependency download? Perhaps with an environment variable?

ssbarnea:
I not think failure to install deps should be ignored but if you propose a PR that sorts your use case without affecting other users I will likely accept it.

tadeboro:
Silently ignoring the fact that dependencies were not installed is not something one should do by default. I would suggest that you parameterize your molecule configuration file using an environment variable that controls whether dependency installation should be attempted or not. Something like this should work:

```
dependency:
  name: galaxy
  enabled: ${INSTALL_DEPS:-true}
```

Then we can set the `INSTALL_DEPS` variable to `false` if we would like to skip the dependency installation.

fourstepper:
You are right. By default we shouldn't ignore dependency installation failures. This is a good enough solution. Thank you

Issue #2028, opened 2019-05-06
molecule zsh completion support is broken


codequokka:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version

ansible 2.7.10
  config file = None
  configured module search path = ['/Users/nbchk/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/nbchk/.ghq/github.com/codequokka/ansible-role-pypro3/.venv/lib/python3.7/site-packages/ansible
  executable location = /Users/nbchk/.ghq/github.com/codequokka/ansible-role-pypro3/.venv/bin/ansible
  python version = 3.7.3 (default, Mar 27 2019, 09:23:15) [Clang 10.0.1 (clang-1001.0.46.3)]
molecule, version 2.20.1
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

- None

# Desired Behavior
As fish and bash completion support, zsh also should have same one.

- fish
```
bash-3.2$ eval "$(_MOLECULE_COMPLETE=source molecule)"                                                                                                      bash-3.2$  . /Users/nbchk/.ghq/github.com/codequokka/ansible-role-pypro3/.venv/bin/activate
bash-3.2$ molecule l
lint   list   login
```

- bash
```
❯ eval (env _MOLECULE_COMPLETE=source-fish molecule)
❯ molecule list
lint  (Lint the role.)  list  (Lists status of instances.)
```

# Actual Behaviour
Zsh has no completion in pressing tab.

```                                                       
% eval "$(_MOLECULE_COMPLETE=source_zsh molecule)"
% molecule l

```

gundalow:
@codequokka Thanks for the suggestion, I've changed this to `enhancement`, rather than bug.

In Ansible this was done by using `argparse` https://github.com/ansible/ansible/pull/50610

decentral1se:
:+1: :+1: :+1:

codequokka:
@gundalow Thank you for replying.
Molecule uses click and click-completion packages to parse command line args.
These packages are supposed to generate completion script for bash, zsh and fish.
So, I don't know why it doesn't generate completion script only for zsh.
Do you know the reason why?

Lirt:
It works in my case, however it doesn't complete arguments for commands (this is intended I guess). 

```bash
# ZSH version
zsh --version
[...]

PR #3118, opened 2021-05-17
Fix shell completion (click 8)

ssbarnea:
Fixes: #2028


src/molecule/shell.py
```diff
@@ -115,9 +112,10 @@ def main(ctx, debug, verbose, base_config, env_file):  # pragma: no cover
     """
     Molecule aids in the development and testing of Ansible roles.
 
-    Enable autocomplete issue:
+    To enable autocomplete for a supported shell execute command below after
+    replacing SHELL with either bash, bash or fish:
```
tadeboro, line 116:
```suggestion
    replacing SHELL with either bash, zsh, or fish:
```

tadeboro (changes requested):

tadeboro (approved):

PR #3136, opened 2021-06-03
Added missing option name to `molecule init`

trashbat:
The `molecule init` command doesn't work as written.

#### PR Type

- Docs Pull Request


ssbarnea:
This is NOT a bug, you are using a outdated version of molecule. Newer versions do not have the --role-name argument, this being a positional argument.

trashbat:
Apologies, I didn't realise I had an older version installed in my path.

PR #3137, opened 2021-06-03
Change ANSIBLE_LIBRARY PATH processing

cloudnull:
When the environment variable ANSIBLE_LIBRARY is used, molecule is
currently extending the path with the contents of the found variable,
however, this will cause problems as the system driven path will
conflict with a user defined path. To correct this issue the path
processing will now insert the user provided variable into the path.

Signed-off-by: Kevin Carter <kecarter@redhat.com>

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/src/molecule/provisioner/ansible.py
+++ b/src/molecule/provisioner/ansible.py
@@ -915,7 +915,13 @@ class Ansible(base.Base):
 
         Adds modules directory from molecule and its plugins.
         """
-        paths = [util.abs_path(os.path.join(self._get_plugin_directory(), "modules"))]
+        paths = list()
+        if os.environ.get("ANSIBLE_LIBRARY"):
+            paths = list(map(util.abs_path, os.environ["ANSIBLE_LIBRARY"].split(":")))
+
+        paths.append(
+            util.abs_path(os.path.join(self._get_plugin_directory(), "modules"))
+        )
 
         for d in drivers():
             p = d.modules_dir()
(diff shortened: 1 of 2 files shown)


src/molecule/provisioner/ansible.py
```diff
@@ -940,7 +940,7 @@ def _get_modules_directories(self):
         )
 
         if os.environ.get("ANSIBLE_LIBRARY"):
-            paths.extend(map(util.abs_path, os.environ["ANSIBLE_LIBRARY"].split(":")))
+            paths.insert(0, map(util.abs_path, os.environ["ANSIBLE_LIBRARY"].split(":")))
```
mwhahaha, line 943: I think it's better to move this up to L918 with something like...

        paths = []
        if os.environ.get("ANSIBLE_LIBRARY"):
            paths.extend(map(util.abs_path, os.environ["ANSIBLE_LIBRARY"].split(":")))
        paths.extend([util.abs_path(os.path.join(self._get_plugin_directory(), "modules"))])

cloudnull: done.

cloudnull:
This change will not pass the docs checks until https://github.com/ansible-community/molecule/pull/3138 merges.

tadeboro (approved):

tadeboro:
@cloudnull Would you be so kind as to rebase your work on top of the main branch and then force push it? This should turn the last red test into a green one (In the log at https://readthedocs.org/projects/molecule/builds/13925451/, I still see the old extension being used). Thank you in advance!

cloudnull:
> @cloudnull Would you be so kind as to rebase your work on top of the main branch and then force push it? This should turn the last red test into a green one (In the log at https://readthedocs.org/projects/molecule/builds/13925451/, I still see the old extension being used). Thank you in advance!

Done

PR #3162, opened 2021-06-23
Avoid None among plugin directories

ssbarnea:


src/molecule/provisioner/ansible.py
```diff
@@ -782,7 +783,7 @@ def manage_inventory(self):
         else:
             self._link_or_update_vars()
 
-    def abs_path(self, path):
+    def abs_path(self, path: str) -> Optional[str]:
```
tadeboro, line 786: As far as I can tell, this method will always return string. `utils.abs_path` will return `None` only if it receives a falsy argument, but this will not happen here.

src/molecule/provisioner/ansible.py
```diff
@@ -945,7 +946,7 @@ def _get_modules_directories(self):
             ]
         )
 
-        return paths
+        return [path for path in paths if path is not None]
```
tadeboro, line 949: If I understand this method correctly, the only source of `None`s in the `ANSIBLE_LIBRARY` environment variable (`path/1::path/2` will result in `["/abs/path/1", None, "/abs/path/2"]`). So maybe we should eliminate bad values there?

tadeboro (changes requested):

src/molecule/provisioner/ansible.py
ssbarnea: The only reason why I updated the signature was to keep mypy happy but I am considering removing the method in the future basically because it allowed current bug from being introduced, or at least making it not allow None.

src/molecule/provisioner/ansible.py
ssbarnea: That is exactly what we do here, assure we do not return None in that list.

tadeboro: But we are doing it in the wrong place. This should be done in line 920 where `None`s can be introduced.

tadeboro (changes requested):

src/molecule/provisioner/ansible.py
ssbarnea: By looking at the body of the function I feel far more confident to remove None values at the end. There are few other areas which could return None, especially as `util.abs_path` can return None.

I do not remember exactly the error that motivated me to make this PR, but my impression is that the None came from `_get_plugin_directory` and not from `ANSIBLE_LIBRARY`.

As that used to raise an ugly runtime exception, I prefer to sanitize the values before returning the list.

PR #3166, opened 2021-06-25
``molecule test`` add support of ansible args

abikouo:
When running molecule test ansible args are now supported:

``molecule test -s test_scenario --driver-name test_driver -- -v --tags test --diff``

#### PR Type
- Feature Pull Request

### DIFF
--- a/src/molecule/command/test.py
+++ b/src/molecule/command/test.py
@@ -137,7 +137,9 @@ class Test(base.Base):
     help="Enable or disable parallel mode. Default is disabled.",
 )
 @click.argument("ansible_args", nargs=-1, type=click.UNPROCESSED)
-def test(ctx, scenario_name, driver_name, __all, destroy, parallel, ansible_args):  # pragma: no cover
+def test(
+    ctx, scenario_name, driver_name, __all, destroy, parallel, ansible_args
+):  # pragma: no cover
     """Test (dependency, lint, cleanup, destroy, syntax, create, prepare, converge, idempotence, side_effect, verify, cleanup, destroy)."""
     args = ctx.obj.get("args")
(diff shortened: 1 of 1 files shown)


src/molecule/command/test.py
```diff
@@ -55,6 +55,13 @@ class Test(base.Base):
 
         Target all scenarios.
 
+    .. option:: molecule test -- -vvv --tags foo,bar
+
+        Providing additional command line arguments to the `ansible-playbook`
+        command.  Use this option with care, as there is no sanitation or
+        validation of input.  Options passed on the CLI override options
+        provided in provisioner's `options` section of `molecule.yml`.
```
tadeboro, line 63: This is not true since those options are combined: https://github.com/ansible-community/molecule/blob/60b68140fb5c650c47019f5db238c0864dbd43ed/src/molecule/provisioner/ansible_playbook.py#L75-L76

tadeboro (changes requested):
The docs are a bit misleading, but once that is fixed, I think this is ready to go.

src/molecule/command/test.py
abikouo: I just made a copy of this description from here:
https://github.com/ansible-community/molecule/blob/60b68140fb5c650c47019f5db238c0864dbd43ed/src/molecule/command/converge.py#L53-L54

tadeboro: That description is also wrong and should be fixed, but not in this PR since it has nothing to do with adding a new option to the test subcommand.

tadeboro (changes requested):
As I said before, black is complaining. Before we can merge this, that issue must be resolved.

PR #3168, opened 2021-06-25
Fix pip documentation link

tadeboro:
Pip's documentation has been rearranged a bit, including the usage document that the Molecule documentation pointed to. This commit updated the link so that it now points to the quickstart document.

#### PR Type

- Docs Pull Request

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -34,7 +34,7 @@ Ubuntu 16.x
 Pip
 ===
 
-:std:doc:`pip <pip:usage>` is the only supported installation method.
+:std:doc:`pip <pip:getting-started>` is the only supported installation method.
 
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):

ssbarnea (changes requested):
I think correct is `getting-started`? the one you proposed made me do a redirect.

tadeboro:
> I think correct is `getting-started`? the one you proposed made me do a redirect.

Yep, that would be the right one. I used grep on the `python -msphinx.ext.intersphinx https://pip.pypa.io/en/latest/objects.inv` output and failed to notice that the reference I selected is a redirect.

ssbarnea (approved):

Issue #3171, opened 2021-07-02
Broken functional tests for init role and init scenario


ssbarnea:
Few hours ago I discovered that we are only running `unit/` tests on CI/CD and that some of the functional tests we had got broken at some point.

Fixing these would not be easy so I will attempt a gradual fixing, with first priority being to ensure we run all tests, and marking the 4 broken ones accordingly:
```
FAILED src/molecule/test/functional/test_command.py::test_command_init_role[delegated] - assert 1 == 0
FAILED src/molecule/test/functional/test_command.py::test_command_init_scenario[delegated] - assert 1 == 0
FAILED src/molecule/test/functional/test_command.py::test_command_dependency[galaxy] - AssertionError: assert 2 == 0
FAILED src/molecule/test/functional/test_command.py::test_command_dependency[shell]
```

The first two tests are failing because converge fails on the newly created scenarios, for reasons outside our control - like lots of linting errors originating from galaxy templates.

The last two are failing because they utility collection `community.molecule` is installed inside the ephemeral molecule folder and this folder is no longer used by the called commands. That is a regression that was triggered by the introduction of prerun phase, which automatically installs deps into a different location and defines ansible collection include path. When this is defined, it prevents molecule from adding its own ephemeral folder to the include paths. 

While I am not yet sure, I believe it we should likely stop messing with role and collection paths inside molecule itself, basically stop using and configuring `ephemeral/collections` and `ephemeral/roles`. 

I do suspect that the second bug is the same as #3135

zhan9san:
> The last two are failing because they utility collection community.molecule is installed inside the ephemeral molecule folder and this folder is no longer used by the called commands. That is a regression that was triggered by the introduction of prerun phase, which automatically installs deps into a different location and defines ansible collection include path. When this is defined, it prevents molecule from adding its own ephemeral folder to the include paths.

The last two can be fixed by both #3514 and https://github.com/ansible/ansible-compat/pull/135

ssbarnea:
We already have a mechanism for installing required collections at runtime for plugins, we should use the same here.

PR #3173, opened 2021-07-02
Remove collection and role path adulterating

ssbarnea:
This major change removes all the core related to modification of ansible include paths for collections and roles.

This also replaces the dependency on ansible-lint with ansible-compat thin library, one that aims to provide exactly the bootstrapping functionality both projects need.

Once this is merged, molecule itself will have no special knowledge of these folders and will make its behavior less likely to be different than just Ansible execution.

Molecule will rely on ansible-compat runtime functionality for injecting temporary roles and collections include paths and installing the requirements.

Who is unlikely to impacted:
* you have `requirements.yml` at repository root or scenario folder
* you used shell tasks running `ansible-galaxy role|collection install`, but did not specify a custom target path

Who is likely to be impacted:
* You manually modified ansible collections paths or ansible roles paths inside molecule.yml file
* You used custom path locations when installing roles or collections

Related: #3171


ssbarnea:
@tadeboro @felixfontein Do you think that we should remove any code in molecule that modifies the collection and role include paths?

I am asking this because I think we no longer need it as the prerun functionality can take care of these. The reported bugs with dependency are related to conflict between what molecule did and the newer cache.

.pre-commit-config.yaml
```diff
@@ -50,11 +50,11 @@ repos:
     hooks:
       - id: mypy
         # empty args needed in order to match mypy cli behavior
-        args: []
+        args: ['--no-incremental']
         entry: mypy src/
         pass_filenames: false
         additional_dependencies:
-          - ansible-lint>=5.0.10
+          - ansible-compat>=0.2.0
```
felixfontein, line 57: I don't think we should depend on non-stable releases, and once there is a stable release, we should use semver to make sure there are no breaking changes.
```suggestion
          - ansible-compat>=1.0.0,<2.0.0
```

setup.cfg
```diff
@@ -66,7 +66,7 @@ setup_requires =
 
 # These are required in actual runtime:
 install_requires =
-    ansible-lint >= 5.0.12  # only for the prerun functionality
+    ansible-compat >= 0.2.0
```
felixfontein, line 69: Same as above
```suggestion
    ansible-compat >= 1.0.0, < 2.0.0
```

felixfontein:
I'm not using molecule so I can't really say much about this. I personally prefer less magic, so the idea sounds good to me.

PR #3178, opened 2021-07-06
Run functional tests in serial

ssbarnea:
This change allows us to run tests in parallel using xdist while keeping the functional tests that are known to fail serial.

Related: https://github.com/pytest-dev/pytest-xdist/issues/385

### DIFF
--- a/src/molecule/test/conftest.py
+++ b/src/molecule/test/conftest.py
@@ -24,6 +24,7 @@ import random
 import string
 
 import pytest
+from filelock import FileLock
 
 from molecule import config
 from molecule.scenario import ephemeral_directory
@@ -138,3 +139,14 @@ def reset_pytest_vars(monkeypatch):
     for var_name in tuple(os.environ):
         if var_name.startswith("PYTEST_"):
             monkeypatch.delenv(var_name, raising=False)
+
+
+@pytest.fixture(autouse=True)
+def block_on_serial_mark(request):
+    # https://github.com/pytest-dev/pytest-xdist/issues/385
+    os.makedirs(".tox", exist_ok=True)
+    if request.node.get_closest_marker("serial"):
+        with FileLock(".tox/semaphore.lock"):
+            yield
+    else:
+        yield
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -75,6 +75,7 @@ repos:
           - tenacity
           - types-PyYAML
           - types-dataclasses
+          - types-filelock
           - types-setuptools
   - repo: https://github.com/PyCQA/pylint
     rev: v2.11.1
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,5 +15,34 @@ skip-string-normalization = false
 profile = "black"
 known_first_party = "molecule"
 
+[tool.pytest.ini_options]
+markers = [
+    "serial: Run this test serially via filelock.",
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -23,3 +23,4 @@ pip-wheel-metadata
 docs/docstree
 .docker
 molecule/**
+semaphore.lock
```
webknjaz, line 26: Why not put this under `.tox`? Or even use a proper pytest location?

src/molecule/test/conftest.py
```diff
@@ -144,3 +145,13 @@ def reset_pytest_vars(monkeypatch):
     for var_name in tuple(os.environ):
         if var_name.startswith("PYTEST_"):
             monkeypatch.delenv(var_name, raising=False)
+
+
+@pytest.fixture(autouse=True)
+def block_on_serial_mark(request):
+    # https://github.com/pytest-dev/pytest-xdist/issues/385
+    if request.node.get_closest_marker("serial"):
+        with FileLock("semaphore.lock"):
```
webknjaz, line 154: Instead of hardcoding this lock file in a project root, plz use a `tmp_path` fixture (maybe do this in a wrapper fixture with a session scope).

ssbarnea:
We cannot use a random lock location because it will fail to identify other instances running. It if quite common to have a test suite open and running (maybe stopped inside debugger) on on window and try to run other tests. Considering that IDEs like vscode could also run tests I would be quite careful not not miss that case.

Using `.tox` folder is incorrect as tests may be running outside tox itself, or even before tox run first time, in which case .tox folder would be missing.

But at the same time I do understand why you dislike the idea of creating semaphore files all around the code base, is not cool.

Maybe someone has a better idea? I am not in a rush to merge this change.

webknjaz:
I think `.tox` is fine. Even if tox is not used, the location is already gitignored. OS packages will probably get the same gitignore anyway. Do you have a real world example of such a failure?
In some projects I also have a `test-results` directory committed to Git, for test results and coverage (Circle CI required pointing to a directory and I picked up this pattern after having to deal with it). Such a dir could also be used if necessary.

.gitignore
ssbarnea: done

src/molecule/test/conftest.py
ssbarnea: We now use .tox folder and that is unique (opposed to tmp_path).

src/molecule/test/conftest.py
```diff
@@ -138,3 +139,13 @@ def reset_pytest_vars(monkeypatch):
     for var_name in tuple(os.environ):
         if var_name.startswith("PYTEST_"):
             monkeypatch.delenv(var_name, raising=False)
+
+
+@pytest.fixture(autouse=True)
+def block_on_serial_mark(request):
+    # https://github.com/pytest-dev/pytest-xdist/issues/385
+    if request.node.get_closest_marker("serial"):
+        with FileLock(".tox/semaphore.lock"):
```
tadeboro, line 148: Do we have guarantees that the `.tox` directory will always be there when we run the tests? Or, will this create the missing directory if it is missing?

tadeboro (approved):
I have no major complaints, just one small comment about the tox folder. If the folder is always there or if we do not support running tests without tox, feel free to ignore my comment and merge as is.

src/molecule/test/conftest.py
ssbarnea: Fixed.

PR #3180, opened 2021-07-07
Use portable cache from ansible-compat

ssbarnea:
This simplifies use of caching across multiple versions of Python.
Implementation is inspired from the the approach taken by https://github.com/apache/airflow/ project.

Related: https://github.com/python/mypy/issues/1362

### DIFF
--- a/src/molecule/api.py
+++ b/src/molecule/api.py
@@ -5,9 +5,9 @@ import traceback
 from collections import UserList
 
 import pluggy
+from ansible_compat.ports import cache
 
 from molecule.driver.base import Driver  # noqa
-from molecule.util import lru_cache
 from molecule.verifier.base import Verifier  # noqa
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


src/molecule/api.py
```diff
@@ -56,7 +56,7 @@ def drivers(config=None) -> UserListMap:
     return plugins
 
 
-@lru_cache()
+# @cache
```
webknjaz, line 59: why comment it out?

src/molecule/compat.py
```diff
@@ -0,0 +1,18 @@
+"""Compatibility layer for various Python versions."""
```
webknjaz, line 1: Could you plz `git mv src/molecule/{,_}compat.py`?

tadeboro (changes requested):
The cache commit seems OK, but this PR also contains setuptools cleanup which should not be part of the current batch of changes.

ssbarnea:
> The cache commit seems OK, but this PR also contains setuptools cleanup which should not be part of the current batch of changes.

Sorted. Got combined by mistake while working on different changes.

PR #3185, opened 2021-07-11
feat: add MOLECULE_PROJECT_BASE_DIRECTORY environment variable

xabinapal:
Simple PR to add a new `MOLECULE_` environment variable that contains the directory basename (not the absolute path) of the project directory.

This can be useful, for example, to create instances with unique dynamic names per project/collection/role... when the driver does not support slashes in the instance name or the resulting instance name would be too long to be readable.

I have not added neither `MOLECULE_SCENARIO_BASE_DIRECTORY` (as it would have the same value as `MOLECULE_SCENARIO_NAME`) nor `MOLECULE_EPHEMERAL_BASE_DIRECTORY` (as except when it is customized by the user, it also contains the scenario name).

#### PR Type

- Feature Pull Request

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -47,6 +47,8 @@ MOLECULE_EPHEMERAL_DIRECTORY      Path to generated directory, usually
                                   ``~/.cache/molecule/<scenario-name>``
 MOLECULE_SCENARIO_DIRECTORY       Path to scenario directory
 MOLECULE_PROJECT_DIRECTORY        Path to your project directory
+MOLECULE_PROJECT_BASE_DIRECTORY   Base project directory name, i.e. the latest
+                                  directory name in the project path.
 MOLECULE_INSTANCE_CONFIG          ?
 MOLECULE_DEPENDENCY_NAME          Dependency type name, usually 'galaxy'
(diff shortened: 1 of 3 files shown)


ssbarnea (changes requested):
This variable really confuses me, based on its name I assume is parent of MOLECULE_PROJECT_DIRECTORY and not the basename.

I am more inclined to believe that `MOLECULE_PROJECT_DIRECTORY_BASENAME` would be more clear?

The second part is why we cannot make use of basename and avoid adding a new variable?

I would really want to clarify the business-case for adding one extra variable as we already have lots of them and removing them is a very tricky business.

ssbarnea:
I will close this because it lacks a clear reasoning for it. Users can still make use of ansible functionality to extract the basename from already existing variable: `MOLECULE_PROJECT_DIRECTORY`.

Any variable we add increases the maintenance and testing cost, that being the reason I oppose this change. Any such addition should sort a problem for which there is no easy workaround. In that case we do have a workaround.

I hope you understand the reasoning behind this decision. That is not set in stone and I suppose that we will add a feature that helps naming managed resources soon as I am aware of that need (hostnames, cloud resource groups or tags, each with its own limitations).

PR #3190, opened 2021-07-15
Replace ansible-lint dependency with ansible-compat

ssbarnea:
Related: https://github.com/ansible-community/ansible-compat/issues/4


ssbarnea:
@tadeboro Can you please look at this? I think is now i good shape and I would not want to miss that from the upcoming [v4.0.0](https://github.com/ansible-community/molecule/releases/tag/untagged-ae7183272bcfb24a0998)

PR #3199, opened 2021-07-19
Made prune more resilient to filesystem errors

ssbarnea:
Fixed observed problem with prune() where it would have failed due
to exceptions being raised during pruning.

This was observed occurring randomly while running tests with xdist.

Related: #3178

### DIFF
--- a/src/molecule/scenario.py
+++ b/src/molecule/scenario.py
@@ -128,13 +128,19 @@ class Scenario(object):
         ] + self.config.driver.safe_files
         files = util.os_walk(self.ephemeral_directory, "*")
         for f in files:
-            if not any(sf for sf in safe_files if fnmatch.fnmatch(f, sf)):
-                os.remove(f)
+            try:
+                if not any(sf for sf in safe_files if fnmatch.fnmatch(f, sf)):
+                    os.remove(f)
+            except FileNotFoundError:
+                pass
 
         # Remove empty directories.
         for dirpath, dirs, files in os.walk(self.ephemeral_directory, topdown=False):
             if not dirs and not files:
-                os.removedirs(dirpath)
+                try:
+                    os.removedirs(dirpath)
+                except OSError as exc:
+                    LOG.info("Failed to remove directory: %s", exc)
 
     @property
     def name(self):


tadeboro:
Do we really want to silently handle such situations? Since it is not safe to run multiple molecule commands at once for a single project, errors we want to handle here can only happen when users abuse Molecule.

ssbarnea:
@tadeboro We are not really silent about it. We ignore only a missing file in first case making the removal idempotent (at the end file is still missing, we do not care if it did vanish or we were the ones removing it).

For the second case, we do emit info message. The reason for this approach is exactly for giving some feedback.

tadeboro:
> @tadeboro We are not really silent about it. We ignore only a missing file in first case making the removal idempotent (at the end file is still missing, we do not care if it did vanish or we were the ones removing it).

No, we are silent covering up a race condition here because the file was gone while the prune operation was in progress. And there are only two realistic sources of that: the user either ran more than one Molecule instance in the project or the cache folder was removed. And none of those scenarios should be silently ignored.

> For the second case, we do emit info message. The reason for this approach is exactly for giving some feedback. If you want we can make them warnings, I do not mind.

The same things hold for this situation.

If we really want to make things safe, we should probably protect the stateful resources with a lock and make sure only only Molecule instance is accessing it at any time.

ssbarnea:
I plan to address the parallel issue and lock support via ansible-compat by adding support in it for that, especially as it has a method for running any ansible tasks. That should make the implementation quite clean.

src/molecule/scenario.py
```diff
@@ -128,13 +128,19 @@ def prune(self):
         ] + self.config.driver.safe_files
         files = util.os_walk(self.ephemeral_directory, "*")
         for f in files:
-            if not any(sf for sf in safe_files if fnmatch.fnmatch(f, sf)):
-                os.remove(f)
+            try:
+                if not any(sf for sf in safe_files if fnmatch.fnmatch(f, sf)):
+                    os.remove(f)
```
webknjaz, line 133: Only wrap this line with try-except. We shouldn't expect `fnmatch()` to cause `FileNotFoundError`.

webknjaz:
I agree with Tadej, this looks like a band-aid. Why not make tests use proper separate tmp dirs and introduce a lock file for preventing race conditions?

src/molecule/scenario.py
```diff
@@ -128,13 +128,19 @@ def prune(self):
         ] + self.config.driver.safe_files
         files = util.os_walk(self.ephemeral_directory, "*")
         for f in files:
-            if not any(sf for sf in safe_files if fnmatch.fnmatch(f, sf)):
-                os.remove(f)
+            try:
+                if not any(sf for sf in safe_files if fnmatch.fnmatch(f, sf)):
```
webknjaz, line 132: Nit: looks like this condition is actually
```suggestion
                if not any(fnmatch.fnmatch(f, sf) for sf in safe_files):
```

PR #3200, opened 2021-07-21
Bump jinja2 minimal version

ssbarnea:
This should ensure that features like `is boolean` are available in jinja2 used by molecule.

Related: https://github.com/ansible-community/molecule-vagrant/pull/112

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,7 +73,7 @@ install_requires =
     cookiecutter >= 1.7.3  # dependency issues in older versions
     dataclasses; python_version<"3.7"
     enrich >= 1.2.5
-    Jinja2 >= 2.10.1
+    Jinja2 >= 2.11.3
     packaging
     paramiko >= 2.5.0, < 3
     pluggy >= 0.7.1, < 1.0


apatard (approved):
``is boolean`` is a nice thing, so good to have it and we must avoid using jinja2 3.x for now imho

ssbarnea:
@apatard Why do you think we should avoid 3.x, in fact we use 3.x during testing as we did not cap its version. I did consider bumping the requirements to v3 but that version is quite new and I do not see any reason for doing that yet. Bumping should happen when we have reasons.

apatard:
Nothing really _serious_. I'm only thinking of:
- from semver point of view, it's a major update (2.X vs 3.X) so it theory, bad breakages may occur
- some distro may not have 3.X. For instance next debian will have 2.11.3 iirc

setup.cfg
```diff
@@ -73,7 +73,7 @@ install_requires =
     cookiecutter >= 1.7.3  # dependency issues in older versions
     dataclasses; python_version<"3.7"
     enrich >= 1.2.5
-    Jinja2 >= 2.10.1
+    Jinja2 >= 2.11.3
```
tadeboro, line 76: We should not cap the Jinja version here because Ansible users will expect things to work with 2.11 and 3.x. After all, Ansible works with those versions.

tadeboro (approved):

setup.cfg
apatard: fwiw, my remarks were about ``Jinja2 >= 2.11.3`` vs ``Jinja2 >= 3.0.0``.

tadeboro: Ahh, OK. I misunderstood. Whenever someone talks about dependencies and semver, the upper bound is what comes to my mind.

ssbarnea: Probably an upper bound like `<4.0` would make sense but on the other hand experience told me that upper constraints should be added only when you have knowledge of something already breaking you. Otherwise is like lottery, in half of cases you prevent a bug, in the other half you introduce a time-bomb conflict which is not real and you endup having to move the bound up and make a hotfix.

PR #3211, opened 2021-08-26
Allow updating dependencies from any platform

ssbarnea:
Adds workaround for forcing pip-compile to produce identical results regardless if is called on linux or macos.


ssbarnea:
This approach does not work as pip-compile output will still be different between macos and linux due to the selinux, regardless this change.

PR #3214, opened 2021-08-27
Capture sanity check warnings and display them

ssbarnea:
Captures runtime warnings raise by driver.sanity_check() and
display them at the end in case of failure.

Related: https://github.com/ansible-community/molecule-podman/pull/76

### DIFF
--- a/src/molecule/api.py
+++ b/src/molecule/api.py
@@ -36,6 +36,14 @@ class UserListMap(UserList):
         super(UserListMap, self).append(item)
 
 
+class MoleculeRuntimeWarning(RuntimeWarning):
+    """A runtime warning used by Molecule and its plugins."""
+
+
+class IncompatibleMoleculeRuntimeWarning(MoleculeRuntimeWarning):
+    """A warning noting an unsupported runtime environment."""
+
+
 @lru_cache()
 def drivers(config=None) -> UserListMap:
     """Return list of active drivers."""
--- a/src/molecule/provisioner/ansible_playbook.py
+++ b/src/molecule/provisioner/ansible_playbook.py
@@ -20,8 +20,10 @@
 """Ansible-Playbook Provisioner Module."""
 
 import logging
+import warnings
 
 from molecule import util
+from molecule.api import MoleculeRuntimeWarning
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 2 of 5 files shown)


src/molecule/util.py
```diff
@@ -96,7 +96,10 @@ def sysexit(code: int = 1) -> NoReturn:
 
 
 def sysexit_with_message(
-    msg: str, code: int = 1, detail: Optional[MutableMapping] = None
+    msg: str,
+    code: int = 1,
+    detail: Optional[MutableMapping] = None,
+    warns: Optional[List[Warning]] = None,
```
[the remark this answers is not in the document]
ssbarnea: I think list is better because I suspect we will add to it later, when we find other places that can raise warnings. Having an iterable does not allow us to add to it.

src/molecule/util.py
```diff
@@ -108,6 +111,9 @@ def sysexit_with_message(
             detail_str = str(detail)
         print(detail_str)
     LOG.critical(msg)
+    if warns:
+        for warn in warns:
+            LOG.warning(warn.__dict__["message"].args[0])
```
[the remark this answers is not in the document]
ssbarnea: I had to use the debugger to extract the exact string of the warning. If you try to use `warn.message` you will see that linters will report that `message` attribute is not present. In fact `.message` works mole like `warn.__repr__()` and renders something like `UserWarning('...')` which is not what I wanted.

I do know that the construct I come up with is very ugly but I was not able to find a nicer one. Maybe you can find one? Try it in debugger until you get it right. I was bit surprised about that.

src/molecule/api.py
```diff
@@ -36,6 +36,12 @@ def append(self, item) -> None:
         super(UserListMap, self).append(item)
 
 
+class MoleculeRuntimeWarning(RuntimeWarning):
+    """A runtime warning used by Molecule and its plugins."""
+
+    ...
```
webknjaz, line 42: This is unnecessary because the docstring is already present and it's the first statement in the class body.
```suggestion
```

src/molecule/api.py
```diff
@@ -36,6 +36,12 @@ def append(self, item) -> None:
         super(UserListMap, self).append(item)
 
 
+class MoleculeRuntimeWarning(RuntimeWarning):
```
webknjaz, line 39: I like this. I think we may want to also explore having some more specific warnings in the hierarchy. Like `IncompatibleMoleculeRuntimeWarning(MoleculeRuntimeWarning)` maybe.

src/molecule/provisioner/ansible_playbook.py
```diff
@@ -103,12 +105,16 @@ def execute(self):
             LOG.warning("Skipping, %s action has no playbook." % self._config.action)
             return
 
-        self._config.driver.sanity_checks()
-        result = util.run_command(self._ansible_command, debug=self._config.debug)
+        with warnings.catch_warnings(record=True) as warns:
+            warnings.filterwarnings("default", category=MoleculeRuntimeWarning)
+            self._config.driver.sanity_checks()
```
webknjaz, line 110: I think we should only catch warnings coming from this invocation.

ssbarnea: That is what is happening with record, when the context is destroyed the filter is also reset. At least that is what docs are saying.

src/molecule/api.py
ssbarnea: I sense some over-engineering odour here... but whatever, I will do it.

tadeboro (approved):

src/molecule/provisioner/ansible_playbook.py
webknjaz: I meant that we shouldn't suppress those warnings in `run_command()`

Issue #3219, opened 2021-08-31
Remove runtime dependency on setuptools


ssbarnea:
At this moment we do rely on `setuptools.pkg_resources` in order to retrieve version of molecule package itself and of its plugins. 

As stated on https://setuptools.readthedocs.io/en/latest/pkg_resources.html  use of it is discouraged in favour of newer `importlib_metadata`.

Success of removing this dependency would also make it easier to distribute molecule using system packaging like rpm/deb as setuptools use from system pythons is... a permanent source of surprises.

Related: https://github.com/ansible-community/molecule/pull/2900

@webknjaz @pabelanger

PR #3221, opened 2021-08-31
Remove dependency on setuptools

ssbarnea:
Fixes: #3219

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,6 +73,7 @@ install_requires =
     cookiecutter >= 1.7.3  # dependency issues in older versions
     dataclasses; python_version<"3.7"
     enrich >= 1.2.5
+    importlib-metadata<2; python_version<"3.8"
     Jinja2 >= 2.11.3
     packaging
     paramiko >= 2.5.0, < 3
@@ -80,7 +81,6 @@ install_requires =
     PyYAML >= 5.1, < 6
     rich >= 9.5.1
     subprocess-tee >= 0.3.2
-    setuptools >= 42 # for pkg_resources
     # selinux python module is needed as least by ansible-docker/podman modules
     # and allows us of isolated (default) virtualenvs. It does not avoid need
     # to install the system selinux libraries but it will provide a clear
--- a/src/molecule/__init__.py
+++ b/src/molecule/__init__.py
@@ -23,10 +23,11 @@
 from __future__ import absolute_import, division, print_function
 
 __metaclass__ = type
-
 try:
-    import pkg_resources
+    # py38+
+    from importlib.metadata import version  # type: ignore
+except ImportError:
+    # py36-py37
+    from importlib_metadata import version
 
-    __version__ = pkg_resources.get_distribution("molecule").version
-except Exception:
-    __version__ = "unknown"
+__version__ = version("molecule")
--- a/mypy.ini
+++ b/mypy.ini
@@ -31,9 +31,6 @@ ignore_missing_imports = True
 [mypy-ruamel]
 ignore_missing_imports = True
 
-[mypy-setuptools]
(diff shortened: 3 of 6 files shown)


setup.cfg
```diff
@@ -73,14 +73,15 @@ install_requires =
     cookiecutter >= 1.7.3  # dependency issues in older versions
     dataclasses; python_version<"3.7"
     enrich >= 1.2.5
+    importlib-metadata<2; python_version<="3.7"
```
webknjaz, line 76: Why isn't this listed among constraints?

src/molecule/__init__.py
```diff
@@ -25,8 +25,8 @@
 __metaclass__ = type
 
 try:
-    import pkg_resources
+    import importlib_metadata
```
webknjaz, line 28: This shouldn't be a part of try-except because it's clearly listed as a dependency.

src/molecule/__init__.py
```diff
@@ -25,8 +25,8 @@
...
 
-    __version__ = pkg_resources.get_distribution("molecule").version
+    __version__ = importlib_metadata.version("molecule")  # type: ignore
 except Exception:
```
webknjaz, line 31: I don't think that with importlib-metadata we should be having a catch-all handler. Let's do
```suggestion
except importlib_metadata.PackageNotFoundError:
```

src/molecule/__init__.py
```diff
@@ -25,8 +25,8 @@
 __metaclass__ = type
 
 try:
```
webknjaz, line 27:
```suggestion
import importlib_metadata


try:
```

setup.cfg
ssbarnea: That is because pip-tools bugs/limitations. I am considering removing the constraints soon unless someone has time to split them python version. Even if we split them per python versions, we will also need one for linux and non linux.

Fixing the deps/constraints is not as easy as it seams. Happy to discuss that but is outside the scope of that PR.

src/molecule/__init__.py
```diff
@@ -25,8 +25,8 @@
...
-    import pkg_resources
+    import importlib_metadata
 
-    __version__ = pkg_resources.get_distribution("molecule").version
+    __version__ = importlib_metadata.version("molecule")  # type: ignore
```
webknjaz, line 30: Plz use specific ignores rather than ignoring everything.
```suggestion
    __version__ = importlib_metadata.version("molecule")  # type: ignore[a-specific-problem-id]
```

src/molecule/__init__.py
ssbarnea: I do not fully remember all the reasons why we had this approach but I think that it was something related to running unittest without installing the package. I will try to remove and see what happens.

constraints.txt
```diff
@@ -26,6 +26,7 @@ distro==1.6.0
 enrich==1.2.6
 execnet==1.9.0
 idna==3.2
+importlib-metadata==4.8.1 ; python_version > "3.7"
```
webknjaz, line 29:
```suggestion
importlib-metadata==2.1.1 ; python_version <= "3.7"
importlib-metadata==4.8.1 ; python_version > "3.7"
```

ssbarnea: This file is not editable by hand, the change that you proposed will break the linting as pip-compile will remove your change.

setup.cfg
mwhahaha: This appears to cause problems when trying to install ansible-core + molecule + an openstack component (ex. os-net-config)  via constraints when installing under centos8 (e.g. 3.6).  importlib-metadata 4.8 works, what is the specific reason to lock it to 2?

PR #3222, opened 2021-08-31
docs: add hint for verify playbook

ltalirz:
As discussed in https://github.com/ansible-community/molecule/issues/3045, if the `verify.yml` playbook does not run the tested role, the `library` and `module_utils` provided by the tested role are not available there [1].

Usually, one does not want to re-run the role in the `verify.yml` playbook since it was already run twice before during the converge and idempotence stages, but testing the configuration of the server may well involve some of the modules provided by the role.
One workaround [suggested by @tadeboro](https://github.com/ansible-community/molecule/issues/3045#issuecomment-835571183) is to add an empty task list to the role and then to restrict tasks of the role to this file when including it in the `verify.yml` playbook.
This PR adds a note regarding this to the docs.

[1]  The `library` actually *is* currently available, but [this is an oversight](https://github.com/ansible-community/molecule/issues/3045#issuecomment-835518824); the `module_utils` are not available.

#### PR Type

- Docs Pull Request

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -84,7 +84,7 @@ directories:
   
 .. note::
 
-  Since the `verify.yml` playbook does not call your role, the `library` and `module_utils` 
+  If the `verify.yml` playbook does not explicitly `include_role` your role, the `library` and `module_utils` 
   provided by your role are not available in the playbook by default.
   If you need those for testing but would like to avoid re-running your role, consider adding 
   an empty task file `init.yml` to your role and use `tasks_from` to include your role in the `verify.yml` playbook::
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -81,27 +81,26 @@ directories:
 * ``verify.yml`` is the Ansible file used for testing as Ansible is the default :ref:`verifier`. This
   allows you to write specific tests against the state of the container after
   your role has finished executing. Other verifier tools are available (Note that :std:doc:`TestInfra <testinfra:index>` was the default verifier prior to molecule version 3).
-  
+
 .. note::
 
-    If the `verify.yml` playbook does not explicitly `include_role` your role, the `library` and `module_utils`
+    If the ``verify.yml`` playbook does not explicitly ``include_role`` your role, the ``library`` and ``module_utils``
     provided by your role are not available in the playbook by default.
     If you need those for testing but would like to avoid re-running your role, consider adding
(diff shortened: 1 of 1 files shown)


docs/getting-started.rst
```diff
@@ -81,6 +81,27 @@ directories:
 * ``verify.yml`` is the Ansible file used for testing as Ansible is the default :ref:`verifier`. This
   allows you to write specific tests against the state of the container after
   your role has finished executing. Other verifier tools are available (Note that :std:doc:`TestInfra <testinfra:index>` was the default verifier prior to molecule version 3).
+  
+.. note::
+
+  Since the `verify.yml` playbook does not call your role, the `library` and `module_utils`
```
ltalirz, line 87:
```suggestion
  If the `verify.yml` playbook does not explicitly `include_role` your role, the `library` and `module_utils` 
```

ssbarnea (approved):
I am ok to add this even if it is something module developers should already know. In fact it will soon become even more problematic as molecule will stop altering ansible include paths.

While I do think some people would be upset and say testing worked before and now is no longer working, we will say that our goal is to test how this code will run on production. On production we do not expect mounting random folder just to help ansible report an error. Being more strict and will less magic will help molecule provide better testing, like detecting things that are not installed correctly.

ssbarnea:
Also please fix reported issues.

docs/getting-started.rst
```diff
@@ -81,6 +81,28 @@ directories:
 * ``verify.yml`` is the Ansible file used for testing as Ansible is the default :ref:`verifier`. This
   allows you to write specific tests against the state of the container after
   your role has finished executing. Other verifier tools are available (Note that :std:doc:`TestInfra <testinfra:index>` was the default verifier prior to molecule version 3).
+  
+.. note::
+
+    If the `verify.yml` playbook does not explicitly `include_role` your role, the `library` and `module_utils`
```
ssbarnea, line 87: RST required double backquotes, single ones render as italic instead. See https://molecule--3222.org.readthedocs.build/en/3222/getting-started.html

ltalirz: thanks - yeah, I tried it with the "Edit on Github" feature but in the end writing clean RST that way isn't all too easy...

I now installed the pre-commit hooks locally and built the docs; the issues should be fixed.

tadeboro (approved):

PR #3230, opened 2021-09-05
Ensure SystemExit is captured on driver initialization

ssbarnea:
Prevent a broken driver from affecting usability of molecule by catching even SystemExit.

Related: https://github.com/ansible-community/molecule-podman/issues/81

### DIFF
--- a/src/molecule/api.py
+++ b/src/molecule/api.py
@@ -51,14 +51,14 @@ def drivers(config=None) -> UserListMap:
     pm = pluggy.PluginManager("molecule.driver")
     try:
         pm.load_setuptools_entrypoints("molecule.driver")
-    except Exception:
+    except (Exception, SystemExit):
         # These are not fatal because a broken driver should not make the entire
         # tool unusable.
         LOG.error("Failed to load driver entry point %s", traceback.format_exc())
     for p in pm.get_plugins():
         try:
             plugins.append(p(config))
-        except Exception as e:
+        except (Exception, SystemExit) as e:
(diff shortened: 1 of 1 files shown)


tadeboro (approved):

src/molecule/api.py
```diff
@@ -51,14 +51,14 @@ def drivers(config=None) -> UserListMap:
     pm = pluggy.PluginManager("molecule.driver")
     try:
         pm.load_setuptools_entrypoints("molecule.driver")
-    except Exception:
+    except (Exception, SystemExit):
```
webknjaz, line 54: @ssbarnea this doesn't catch *everything*, though. Is it intentional that you didn't use `BaseException`?

ssbarnea: I wasn't sure what kind of other exceptions would Base Exception catch so I went for that approach. Shortly I don't know if current code is better than just BaseException or not.

webknjaz: You could catch all the exceptions excluding `KeyboardInterrupt`, I think.

webknjaz: The whole hierarchy is documented here: https://docs.python.org/3/library/exceptions.html#exception-hierarchy

webknjaz: OTOH Molecule could also define own exceptions for communicating misconfiguration clearly. In this case, you'd need something like
```python
try:
    ...
except KeyboardInterrupt:
    raise
except MoleculeMisconfiguredException:
    report_exc()
except BaseException:
    process_exc()
```

PR #3237, opened 2021-09-13
Ensure subprocesses are run via Runtime()

ssbarnea:
Instead of directly calling subprocess, ensure we use Runtime
api methods for running various commands, most importantly
ansible commands.

This also removes subprocess-tee dependency as a desired action.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,7 +73,6 @@ install_requires =
     pluggy >= 0.7.1, < 2.0
     PyYAML >= 5.1
     rich >= 9.5.1
-    subprocess-tee >= 0.3.5
 
 [options.extras_require]
 # ansible extra is deprecated, will be removed in next version
--- a/src/molecule/app.py
+++ b/src/molecule/app.py
@@ -1,0 +1,13 @@
+"""Molecule Application Module."""
+from ansible_compat.runtime import Runtime
+
+
+class App:
+    """App class that keep runtime status."""
+
+    def __init__(self) -> None:
+        """Create a new app instance."""
+        self.runtime = Runtime(isolated=True)
+
+
+app = App()
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,7 +72,6 @@ repos:
           - packaging
           - rich
           - ruamel.yaml>=0.17.10
-          - subprocess-tee>=0.2.0
           - types-PyYAML
           - types-dataclasses
           - types-filelock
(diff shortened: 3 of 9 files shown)


webknjaz:
Idea: we could forbid calling subprocess via pylint.

ssbarnea:
> Idea: we could forbid calling subprocess via pylint.
We can do this in follow-up, I am trying to make this PR as small as possible.

src/molecule/app.py
```diff
@@ -0,0 +1,13 @@
...
+
+class App:
+    """App class that keep runtime status."""
+
+    def __init__(self) -> None:
+        """Create a new app instance."""
+        self.runtime = Runtime(isolated=True)
+
+
+app = App()
```
webknjaz, line 13: Is this an attempt to make a mutable singleton? Why not initialize it during the startup sequence and then pass around DI-style?

setup.cfg
```diff
@@ -59,7 +59,7 @@ zip_safe = False
 
 # These are required in actual runtime:
 install_requires =
-    ansible-compat >= 0.5.0
+    ansible-compat >= 1.0.0a0
```
webknjaz, line 62: Let's be stricter until the API is stable:
```suggestion
    ansible-compat ~= 1.0.0a0
```

src/molecule/app.py
ssbarnea: Molecule multi-layered structure makes it extremely inconvenient to pass such an object. I considered that approach but it would be a total mess. We can aim to achieve that at some point but it would need several major version changes and existence of plugins is a major limiting factor.

webknjaz: Affecting the import time is also messy. There's nothing wrong with multilayered bypasses, though.

ssbarnea:
@webknjaz @ganeshrn: I do not plan to merge this until I release 1.0.0 or ansible-compat, but the release of compat is waiting for the approval of this PR as I want to avoid having to patch it in case we miss something.

Please review it and let me know if I can release compat.

src/molecule/app.py
ssbarnea: I will not do that as part of this PR. If you want you can lead this challenge but I warn you that you will unlikely to succeed as you will have to change API of plugins and 15+ plugins. I am also worried that this may add delays which may interfere with `AAP` inclusion goals.

tadeboro (approved):

PR #3238, opened 2021-09-17
Allows to take a newer pluggy release

cjeanner:
The pluggy module was added back in 2019 - there are newer releases now,
and we can expect other projects to depend on the 1.0 one - it would
then conflict with molecule, preventing developers to run smoothly there
unit tests.


bshephar:
This will also need to be updated in the constraints file:
https://github.com/ansible-community/molecule/blob/main/constraints.txt#L37

cjeanner:
hm, need to add a label - lemme close and reopen it properly.

PR #3239, opened 2021-09-17
Relax pluggy version dependency

cjeanner:
The pluggy module was added back in 2019 - there are newer releases now,
and we can expect other projects to depend on the 1.0 one - it would
then conflict with molecule, preventing developers to run smoothly there
unit tests.

#### PR Type
- Feature Pull Request

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -77,7 +77,7 @@ install_requires =
     Jinja2 >= 2.11.3
     packaging
     paramiko >= 2.5.0, < 3
-    pluggy >= 0.7.1, < 1.0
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):
Correct fix is to make it < 2.0 as we already know that 1.0 did not break it and we do not expect any 1.x version to break it. It would be a more future proof dependency.

cjeanner:
Oh, Thanks Sorin! I should have pinged you earlier :).

tadeboro (approved):

PR #3244, opened 2021-09-23
Raise upper constraint for importlib-metadata

mwhahaha:
The current version of importlib-metadata is 4.8.1 which works fine even
when running python <3.8.  By restricting to <2, we're limiting the
co-installability of molecule with projects that require higher
importlib-metadata. 

The <2 limit was added as part of #3221, but it's not clear why this version
was selected.

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,7 +73,7 @@ install_requires =
     cookiecutter >= 1.7.3  # dependency issues in older versions
     dataclasses; python_version<"3.7"
     enrich >= 1.2.5
-    importlib-metadata<2; python_version<"3.8"
+    importlib-metadata; python_version<"3.8"
     Jinja2 >= 2.11.3
     packaging
     paramiko >= 2.5.0, < 3
--- a/tox.ini
+++ b/tox.ini
@@ -50,7 +50,7 @@ deps =
     devel: git+https://github.com/ansible-community/pytest-molecule#egg=pytest-molecule
     dockerfile: ansible>=2.10
     selinux
-    py{36,37}: importlib-metadata<2,>=0.12
+    py{36,37}: importlib-metadata>=0.12
     py: ansible-base
 extras =
     docker


tadeboro:
I must admit that I see no good reason for setting the upper version bound at all. Since the `importlib-metadata` package is upstream of the stdlib's implementation, I would not expect things to break often. But since I do not know the details about the version constraint, this will have to wait for other maintainers to pitch in with the information.

mwhahaha:
I can drop it, however I saw commentary about preventing issues with future versions which is the only reason why I set it. It's likely that we'd just hit a problem in the future when v5 is released

tadeboro:
> I can drop it, however I saw commentary about preventing issues with future versions which is the only reason why I set it. It's likely that we'd just hit a problem in the future when v5 is released

If there is one thing certain in life, it is the fact that things will break ;) If we add upper bound, things will fail because we will forget to bump it and something will start conflicting with that version. If we do not add the bound, things will fail once the importlib library breaks backward compatibility. I just happen to prefer the last failure because I am forced to keep code working with the latest stable versions of library if at all possible.

mwhahaha:
Ditto. I'll drop the upper limit and if someone wants it back I'll add it. I'd rather only limit when necesary and this only applies to <3.8 anyway.

hswong3i:
With https://github.com/ansible-community/molecule/blob/c6a456b61f0913d07a991b4d579c864d41ca4d82/.pre-commit-config.yaml#L60 we have `importlib-metadata>=4.6.1`, reported from https://github.com/ansible-community/molecule/search?q=importlib-metadata&type=code

I would recommend synchronize both `/setup.cfg ` and `/tox.ini` with same standard?

hswong3i:
I temporarily modify the line for my DEB/RPM packaging needs, and it works with `importlib-metadata>=4.6.1`:
* https://github.com/alvistack/ansible-community-molecule/commit/7167a7c626f86bcce881a98931803f72c949a2b1#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52R74
* https://build.opensuse.org/package/show/home:alvistack/ansible-community-molecule-3.5.1

ssbarnea (approved):

ssbarnea:
Based on my experience of almost a decade with python package dependencies, setting upper bounds created by far more problems than preventing. If dependency follows semver proves to not matter.

The only exception is for packages that we know to be prone to break, so basically ceiling seems to be effective only when a particular dependency broke the project more than once.

@mwhahaha thanks for making this.

PR #3248, opened 2021-09-30
Fix Sphinx config for `sphinx-ansible-theme` v0.8+

webknjaz:
The setup currently present on `main`, has an unconstrained dependency on `sphinx-ansible-theme`. As a result, when the theme was refactored, the RTD site pulled in a newer version that needed minor tweaks to work properly. The builds succeeded on RTD but it went unnoticed that the CSS files were no longer included. Note that the previous `latest` build (mapped to `master` / `main` branch) happened around 9 weeks ago — I don't know why, maybe a part of the automation is disabled.

This patch adds an upper boundary for the dependency and adjusts the Sphinx configuration according to how the theme is supposed to be used.

#### PR Type

- Bugfix Pull Request
- Docs Pull Request


gundalow (approved):
Thanks for digging into this.
+1 to merge once RTD has build has completed and looks OK.

ssbarnea (approved):
I still see that latest/stable link what does not work on RTD build but I suppose that is expected.

webknjaz:
> I still see that latest/stable link what does not work on RTD build but I suppose that is expected.

Yes, I think these are provided by RTD on the build. And PR builds are slightly different because they are not meant to build multiple branches.

webknjaz:
Urgh... I've been meaning to merge with a good commit message :(

PR #3288, opened 2021-11-04
[util] prevent run_command stdout from using rich

relrod:
Change:
- When importing molecule.console, we indirectly import 'enrich'.
  The 'enrich' library (at least, how we use it) will overwrite
  Python's sys.stdout and sys.stderr with a FileProxy from 'rich'.

  This causes problems when handling data with arbitrary escape codes
  that are used for various purposes such as ansible-runner's
  'awx_display' callback plugin, which hides base64 data with escape
  sequences, because 'rich' will drop any escape codes that are unknown
  to it.

  In order to be able to call into molecule's CLI via ansible-runner
  and not get a bunch of base64-encoded data mixed in the response, we
  need those escape codes to not be dropped. To work around this, we
  save a copy of sys.stdout before importing molecule.console, and
  swap sys.stdout to the original value (and back later) in
  run_command before we actually run the command.

Test Plan:
- Lots and lots of local testing and debugging.

Signed-off-by: Rick Elrod <rick@elrod.me>

#### PR Type

- Bugfix Pull Request

### DIFF
--- a/src/molecule/util.py
+++ b/src/molecule/util.py
@@ -38,13 +38,13 @@ from ansible_compat.ports import cache
 from rich.syntax import Syntax
 from subprocess_tee import run
 
+from molecule.constants import MOLECULE_HEADER
+
 # This must be BEFORE molecule.console is imported below.
 SYS_STDOUT = sys.stdout
-from molecule.console import console
+from molecule.console import console  # pylint: disable=wrong-import-position
 
 CONSOLE_STDOUT = sys.stdout
-
-from molecule.constants import MOLECULE_HEADER
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


webknjaz:
This feels like a dangerous band-aid. If that lib controls the stout, it may as well buffer things. We're in danger of injecting our output in between buffer dumps from `rich`. This may create a flaky hard to catch problem.

Do we have any comments from the library upstream?

relrod:
> Do we have any comments from the library upstream?

The upstream library that is taking control of `sys.stdout` is enrich (which is @ssbarnea's library), not rich itself. I have not asked, but I suspect the rich developers will say "don't do that" since it's not supported in rich itself.

So, no, I haven't talked to the rich devs yet, because I don't know that this is necessarily a rich problem.

PR #3292, opened 2021-11-09
Adopt check pipeline job

ssbarnea:


.github/workflows/tox.yml
```diff
@@ -84,3 +86,14 @@ jobs:
         run: |
           echo "${{ matrix.PREFIX }} tox -e ${{ matrix.tox_env }}"
           ${{ matrix.PREFIX }} tox -e ${{ matrix.tox_env }}
+        continue-on-error: ${{ matrix.experimental }}
+
+  check:
```
webknjaz, line 91: Apparently, this needs to be more complex to play well with branch protection: https://github.com/pyca/cryptography/pull/6512#discussion_r757996981.

ssbarnea: I took a look at it and ended up with a big headache. I believe it works but the fact that you have to write a custom action of few hundreds lines for just a simple things like this is not ok at all, for GHA in general. Lets try to nag their product managers to sort the problem w/o us having to stretch like this.

webknjaz: No, you don't have to. I have a one-line example. It's just a bit too long, but really one line :)
Anyway, I've just published a new action. All those many lines are mostly for decoration and readability — https://github.com/marketplace/actions/alls-green.

PR #3295, opened 2021-11-09
Deprecate ansible[-base] extras

ssbarnea:
* removes documentation about installing ansible as pip extra
* deprecates ansible and ansible-base extras
* updates our testing to make use of ansible-core instead of old
  ansible-base, especially as we already have two major versions
  using ansible-core package name.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -48,7 +48,7 @@ Pip
 
 .. warning::
 
-  The old `ansible` and `ansible-base` pip extras are now deprecated and will
+  The old ``ansible`` and ``ansible-base`` pip extras are now deprecated and will
   be removed in molecule 3.6.0. If you use them, please switch to explicit
   package mention to avoid problem with newer versions of molecule.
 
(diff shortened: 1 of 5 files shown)


ganeshrn (approved):

tadeboro (changes requested):
I think this deprecation needs to be documented in the user-facing documentation. Just deleting the part of the docs will not help users that currently use extras since they will only notice things changed when the installation will fail.

ssbarnea:
> I think this deprecation needs to be documented in the user-facing documentation. Just deleting the part of the docs will not help users that currently use extras since they will only notice things changed when the installation will fail.

Indeed the deprecation in a file almost nobody reads was not very useful. I updated the pr to mention it.

docs/installation.rst
```diff
@@ -40,13 +40,17 @@ Pip
...
+  using your distribution package installer.
 
   .. code-block:: bash
 
-      $ python3 -m pip install "molecule[ansible]"  # or molecule[ansible-base]
+      $ python3 -m pip install molecule ansible-core
+
+.. warning::
+
+  The old `ansible` and `ansible-base` pip extras are now deprecated and will
```
webknjaz, line 51:
```suggestion
  The old ``ansible`` and ``ansible-base`` pip extras are now deprecated and will
```

PR #3298, opened 2021-11-10
Enable doc8

ssbarnea:
- add doc8 linter
- fixes newlines in rst
- converts table to definition list to make it easier to read on
  narrow screens and also easier to maintain.


docs/ci.rst
```diff
@@ -193,15 +199,27 @@ Azure Pipelines
...
+The ``export PATH`` is required to ensure you can use the
+``molecule``/``ansible`` shell scripts. Azure doesn't add these by default.
 
 
 Jenkins Pipeline
 ^^^^^^^^^^^^^^^^
 
-`Jenkins`_ projects can also be defined in a file, by default named `Jenkinsfile` in the top folder of a repository. Two syntax are available, Declarative and Scripted. Here is an example using the declarative syntax, setting up a virtualenv and testing an Ansible role via Molecule.
+`Jenkins`_ projects can also be defined in a file, by default named
+`Jenkinsfile` in the top folder of a repository. Two syntax are available,
```
webknjaz, line 220:
```suggestion
``Jenkinsfile`` in the top folder of a repository. Two syntax are available,
```

docs/configuration.rst
```diff
@@ -35,28 +35,41 @@ Variable Substitution
...
-MOLECULE_DEPENDENCY_NAME          Dependency type name, usually 'galaxy'
-MOLECULE_DRIVER_NAME              Name of the molecule scenario driver
-MOLECULE_PROVISIONER_NAME         Name of the provisioner tool (usually 'ansible')
-MOLECULE_REPORT                   Name of HTML file where to dump execution report.
-MOLECULE_SCENARIO_NAME            Name of the scenario
-MOLECULE_VERBOSITY                Determine Ansible verbosity level.
-MOLECULE_VERIFIER_NAME            Name of the verifier tool (usually 'ansible')
-MOLECULE_VERIFIER_TEST_DIRECTORY  ?
-================================  ===========================================================
+MOLECULE_DEBUG
```
webknjaz, line 38: why reformat it like this?

tadeboro: IIRC, rST docs suggest using definition lists for documenting variables. So this change does make sense. @ssbarnea Did doc8 suggest that change?

ssbarnea: Not really but when I reached the very long lines of that table, I found myself forced to fix that table bug.

doc8 is not as smart to do that.

setup.cfg
```diff
@@ -159,3 +159,7 @@ max-line-length = 88
 per-file-ignores =
   docs/conf.py: D
   src/molecule/test/*: D100,D103,D104
+
+[doc8]
+# see https://github.com/PyCQA/doc8/issues/75
+ignore-path = .eggs,.tox,constraints.txt,build,molecule.egg-info,src/molecule.egg-info
```
webknjaz, line 165: 1) I'd multiline this
2) Isn't there a better config then package metadata?

ssbarnea: No multiline support, probably my comment should give you a hint that there is no much better place. FYI, flake is still inside setup.cfg for the same reasons.

Issue #3291, opened 2021-11-09
molecule is incompatible with ansible-devel (2.13) on py310


ssbarnea:
This is a real bug that seems to be reproducible where performance goes down exponentially and testing times out. Likely they would pass in 1-2 hours instead of the expected 10 minutes.

It was found while adding https://github.com/ansible-community/molecule/pull/3260/files and it seems to happen only with `ansible@devel` branch (2.13).

I am wondering if 2.13 is trying to do something fancy with stdout/stderr streams which may interfere with out own shady-magic from enrich, where we proxy stderr/stdout with a wrapper.

This seems specific to python 3.10, all other version seem to work fine. If someone can investigate this it would be great.

konstruktoid:
I can't reproduce this issue locally on
 - Ubuntu Focal Vagrant box (`vagrant init ubuntu/focal64` w/o modifications)
 - Darwin Kernel Version 20.6.0: MacBookPro14,1 (with the above box running)

Sure, 33 minutes is a long time but it was when using the ~1G RAM and 1 CPU Vagrant box.

```sh
$ rm -rf .tox/
$ time tox -e py310-devel
[...]
```

```sh
=========================================================================================== test session starts ============================================================================================
platform darwin -- Python 3.10.0, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
cachedir: .tox/py310-devel/.pytest_cache
plugins: html-3.1.1, mock-3.6.1, metadata-1.11.0, xdist-2.4.0, plus-0.2, molecule-1.5.1.dev3, forked-1.3.0, testinfra-6.4.0, cov-3.0.0
[...]
---------- coverage: platform darwin, python 3.10.0-final-0 ----------

=========================================================================================== slowest 10 durations ===========================================================================================
30.42s call     src/molecule/test/functional/test_command.py::test_command_dependency[galaxy]
28.06s call     src/molecule/test/functional/test_command.py::test_command_dependency[shell]
20.25s call     src/molecule/test/functional/test_command.py::test_command_idempotence[driver/delegated-delegated-default]
18.04s call     src/molecule/test/functional/test_command.py::test_command_verify[driver/delegated-delegated-default]
17.27s call     src/molecule/test/functional/test_command.py::test_command_dependency[shell]
16.57s call     src/molecule/test/functional/test_command.py::test_command_test[driver/delegated-delegated-default]
16.02s call     src/molecule/test/functional/test_command.py::test_command_check[driver/delegated-delegated-default]
[...]

PR #3304, opened 2021-11-18
Add Vagrant configuration to mimic GitHub runners

konstruktoid:
This PR:
- adds `tox_env: py39-devel` to the GitHub workflow
- adds Vagrant configuration to mimic GitHub runners for local testing
- is related to #3291 

#### PR Type

- Feature Pull Request

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -44,6 +44,8 @@ jobs:
           - tox_env: py38-devel
             PREFIX: PYTEST_REQPASS=449
             python-version: 3.8
+          - tox_env: py39-devel
+            PREFIX: PYTEST_REQPASS=449
           - tox_env: py310-devel
             PREFIX: PYTEST_REQPASS=449
             python-version: "3.10"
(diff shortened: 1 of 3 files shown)


ssbarnea:
@konstruktoid What is really the purpose of this change? Is it to enable reproduction of platform specific errors when running from another platform?

I am bit reluctant to add a lot of code that is not really tested under CI but on the other hand we could enable zuul to run it if we want as we already use zuul on few other molecule plugins.

konstruktoid:
The only actual CI change is the `tox` update, the Vagrant things is just about adding a easy (easier?) way to test on various platforms.

I was trying to catch something similar to https://github.com/ansible-community/molecule/issues/3291#issuecomment-971514289 but to no avail so far.

If this is too bloated, then feel free to close or let it exist so I (and more) can use the branch for testing.

ssbarnea:
@tadeboro WDYT about this, should it go in or not?

webknjaz:
I don't think it's reasonable for us to have this in the repository:
1) Having multiple copies of the same automation is hard to maintain/keep up-to-date
2) Adding such things means more maintenance burden because the contributors often don't maintain them
3) I've seen people adding extra ways of running tests as they like. The ones I saw in the past were Makefile and Dockerfile, and now it's Vagrantfile. The reality is that maintainers would only care about tox but still get "bug reports" about others that they may even know nothing about

konstruktoid:
Reasonable input . Closing.

PR #3324, opened 2021-11-29
Fix sdist file inclusion to just pack the Git-tracked files

webknjaz:
The current setup mixes up incompatible technologies — `setuptools-scm` and `check-manifest`. The former uses a setuptools hook to auto-include all the Git-tracked files but following the recommendations from the latter adds extra non-tracked rubbish from various build processes on top.

Some big blobs inflate the sdist too much. Strictly speaking, they shouldn't even be in Git but it is what it is. I've originally fixed this about 2.5 years ago by adding a single exclusion via https://github.com/ansible/molecule/pull/1924.
Later on, that setup has been broken by https://github.com/ansible-community/molecule/pull/2943 at which point it allowed random build artifacts to slip through the misconfiguration.

This change does not attempt to exclude small text files because they do not affect the size of the sdist too much (less than 1KB). Adding extra entries introduces undesired maintenance burden and is therefore not a commonly accepted practice when it comes to setuptools-scm managed projects. The considerations are explained in https://github.com/ansible-community/molecule/pull/3314#issuecomment-981618612 and https://github.com/pypa/setuptools_scm/issues/190#issuecomment-492339946.

See the commit messages for more specific details.

Supersedes #3314.

#### PR Type

- Bugfix Pull Request


relrod (approved):

ssbarnea (changes requested):
I manually checked the output and it looks mostly ok. The only regression that I did spot was that it started to include `.github` and `.zuul.d` folders and I see no use for these in sdist.

webknjaz:
This is not a regression but an intended outcome as proven by the linked discussions and linked explanations. Did you read my comments?

cidrblock:
+1

PR #3331, opened 2021-12-08
Add packit rpm building to molecule

ssbarnea:


tools/molecule.spec
```diff
@@ -0,0 +1,53 @@
...
+Summary: ...
+%forgemeta
+URL:       %{forgeurl}
+Source:    %{forgesource}
+License:   BSD
+BuildArch: noarch
+
+BuildRequires: python3-devel
+BuildRequires: python3-wheel
+BuildRequires: python3-tox-current-env
```
hroncok, line 18: Neither wheel nor tox-current-env should be explicitly BuildRequired: They are both generated with `%pyproject_buildrequires -t`.

tools/molecule.spec
```diff
@@ -0,0 +1,53 @@
...
+%global _description %{expand:
+...
+}
+
+%description %{_description}
+
+%package -n python3-%{srcname}
+Summary: %summary
+
+%py_provides python3-%{srcname}
```
hroncok, line 29: This should also not be needed.

tools/molecule.spec
```diff
@@ -0,0 +1,53 @@
...
+%build
+%{pyproject_wheel}
+
+%install
+%{pyproject_install}
+
+%pyproject_save_files %{pypiname}
+
+%check
+%{tox} -e lint,py
```
hroncok, line 48: Linting in Fedora might not be desired. Linting packages especially on Rawhide tend to get updated. If this was an official Fedora specfile, this would even be against the guidelines. https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_linters

.packit.yaml
```diff
@@ -0,0 +1,49 @@
...
+dist_git_namespace: rpms
+downstream_package_name: python3-molecule
+jobs:
+  - job: copr_build
+    metadata:
+      targets:
+        # Unable to build centos-stream due to missing python3-setuptools_scm_git_archive
+        # which is part of EPEL.
+        # - centos-stream-x86_64
+        - epel-8-x86_64
```
hroncok, line 31: pyproject-rpm-macros are unfortunately not EL8 compatible, as they use RPM features not yet available there.
If you wish for EPEL 8 compatible spec file, you need to follow https://docs.fedoraproject.org/en-US/packaging-guidelines/Python_201x/ Unfortunately packaging for a distro from the previous decade means following the guidelines from the previous decade as well.

tools/molecule.spec
hroncok: Also, if you desire to use specific toxenvs, you might need to pass the toxenvs to `%pyproject_buildrequires -e` to get all the relevant dependencies.

hroncok (commented):
A few RPM packaging notes.

apatard:
Since I've seen this PR, I've a stupid question: What about rpm signature ?

ssbarnea:
The purpose of this is not to produce a signed rpm, only to validate that we can produce rpms directly upstream. We will not produce the final rpms ourselves.

.containerignore
```diff
@@ -0,0 +1,16 @@
+.ansible-test
```
webknjaz, line 1: Maybe symlink this against `.gitignore`?

Containerfile
```diff
@@ -0,0 +1,8 @@
+FROM quay.io/packit/packit
+LABEL maintainer="Ansible <info@ansible.com>"
+
+# AS packit seems to not install BuildRequires by default:
+RUN dnf install -y --setopt=install_weak_deps=False python3-wheel python3-setuptools_scm+toml python3-tox-current-env tox
```
webknjaz, line 5: Why not publish an image with this pre-cached?

tools/molecule.spec
```diff
@@ -0,0 +1,59 @@
+%global srcname molecule
```
webknjaz, line 1: I saw several projects using `packaging/rpm/` dir as a convention for the spec file. Can we use the same location?

Containerfile
hroncok: As for how RPM build is supposed to work, installing the packages directly to the environment seems quite weird.  Packit offers a Copr-based service where it builds the packages for various different Fedoras/ELs. The BuildRequires are supposed to be installed for that chroot (in Copr). This seems to be a hybrid setup that tries to use packit to build the RPM package in dcoker. It might work, but it will not serve as a check whether this is packageable for Fedora, because in Fedora, we don't build packages like this.

tools/molecule.spec
```diff
@@ -0,0 +1,59 @@
...
+%global forgeurl https://github.com/ansible-community/%{srcname}
+
+Name:    %{pkgname}
+Version: 0.0.0
+Release: 1%{?dist}
+Summary: ...
+%forgemeta
+URL:       %{forgeurl}
+Source:    %{forgesource}
+License:   BSD
```
webknjaz, line 13: https://github.com/ansible-community/molecule/blob/main/LICENSE
```suggestion
License:   MIT
```

tools/molecule.spec
```diff
@@ -0,0 +1,59 @@
...
+%global pkgname python3-%{srcname}
+%global forgeurl https://github.com/ansible-community/%{srcname}
+
+Name:    %{pkgname}
+Version: 0.0.0
+Release: 1%{?dist}
+Summary: ...
+%forgemeta
+URL:       %{forgeurl}
+Source:    %{forgesource}
```
webknjaz, line 12: It's common for downstreams to use PyPI-published sdists as the source of truth, they differ from Git archives, especially since untagged Git versions will not have enough versioning metadata + they may have differing contents.

Instead, it's best to use the `%{pypi_source}` macro:
```suggestion
Source0:    %{pypi_source}
```

- https://docs.fedoraproject.org/en-US/packaging-guidelines/SourceURL/#_python_packages_pypi
- https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_source_files_from_pypi


@hroncok am I missing anything?

tools/molecule.spec
```diff
@@ -0,0 +1,59 @@
...
+License:   BSD
+BuildArch: noarch
+
+BuildRequires: python3-devel
+BuildRequires: python3-wheel
+BuildRequires: python3-tox
+# BuildRequires: python3-tox-current-env
```
webknjaz, line 19: I think the only thing truly necessary here is PEP 517 aware macros. Others are pulled in automatically.

tools/molecule.spec
hroncok: You are missing the packit bit. Packit patches the source in, so the value might not even matter.

As for whether using sdist tarball or github archive is better -- neither one is better, they are different. Both have advantages and disadvantages.

webknjaz: But in case we enable Packit to send PRs to distgits of downstreams, would it matter?

tools/molecule.spec
hroncok: The generate_buildrequires section is not used here anymore, so the buildrequires are needed. However, using the pyproject-rpm-macros like this is not really supported.

tools/molecule.spec
```diff
@@ -0,0 +1,59 @@
...
+%global _description %{expand:
+...
+}
+
+%description %{_description}
+
+%py_provides python3-%{srcname}
+
+%prep
+%autosetup -p1 -n %{srcname}-%{version}
```
webknjaz, line 30: Hm.. I thought this macro exists mostly for make-based projects. Need to check the docs.

tools/molecule.spec
```diff
@@ -0,0 +1,59 @@
...
+
+# generate_buildrequires chokes on tox deps = --editable...
+# also that one chokes too:
+# pyproject_buildrequires -t
```
hroncok, line 34: If pyproject_buildrequires chokes on `-t`, use at least `-r`.

tools/molecule.spec
hroncok: I guess packit PRs replace the downstream spec file with the upstream one, so it would. But the name of the spec file here differs from the Fedora spec file name, so it might not even work. I have very little experience with keeping spec files upstream (I consider it bad practice).

tools/molecule.spec
hroncok: [No](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_autosetup)

tox.ini
```diff
@@ -167,13 +168,33 @@ commands =
...
 commands =
     sh -c "type snapcraft && snapcraft build"
 
+[testenv:system]
+description = Runs tests using system packages and w/o installing extra deps
+skip_install = true
+sitepackages = true
+deps =
+commands =
+    pip check
```
webknjaz, line 195: This could go into `commands_pre`.

Containerfile
webknjaz: Oh, I missed this point. Fair enough. This container is unnecessary then and the base could just be used with a volume mount.

tools/molecule.spec
hroncok: Beware the RPM docs in https://rpm-software-management.github.io/rpm/manual/autosetup.html -- it has a mistake https://github.com/rpm-software-management/rpm/pull/1860

.containerignore
ssbarnea: If i remember correctly they are not fully compatible.

webknjaz: I don't remember any obvious incompatibilities but to the extent of things used here, it probably doesn't matter but it'd be easier to maintain the copies of exclusions.

tools/molecule.spec
webknjaz: Well, the whole point of Packit is to allow upstream testing and keep it as close as possible to downstream. I don't see a problem to rename this file to match what Fedora uses.

ssbarnea:
Closing in favour of #3456 which is using already known to work recipe from ansible-lint.

PR #3353, opened 2022-01-04
Made pip-compile output silent

ssbarnea:
Made pip-compile output silent while enabling --show-diff-on-failure which provides a much easier to read output.


ssbarnea:
I noted the bug with os.remove and fixing it in follow-up, as it was unrelated/random.

PR #3355, opened 2022-01-04
Make pip-compile work from macos

ssbarnea:
Ensure we produce the same constraints.txt regardless if user runs it from a linux or macos box.


webknjaz (approved):
Oh boi

PR #3375, opened 2022-01-05
Tune dependabot configuration

ssbarnea:
Related: https://github.com/dependabot/dependabot-core/issues/4062


.github/dependabot.yml
```diff
@@ -9,3 +9,5 @@ updates:
       - dependencies
       # as dependabot bumps only the pinned test dependencies, is safe to:
       - skip-changelog
+    open-pull-requests-limit: 3
+    rebase-strategy: disabled
```
webknjaz, line 13: Why? Is rather keep it on.

webknjaz (changes requested):

.github/dependabot.yml
ssbarnea: I prefer to retrigger rebasing manually because auto-rebasing is hammering the CI as each merged change will request all the other PRs to rebuild. I already spend almost the entire day playing whack-a-mole with this repository and the prospect of using dependabot on all repos seems now like a nightmare full-time job.

If we would have a PR limit of 1 it would not be a big problem to use auto-rebase, but that is not the case yet.

webknjaz: It's not a problem with auto-merge, though. FWIW this is not related to the main goal of the PR.

ssbarnea: I removed that line but I am not keen on activating auto-merge, especially as we require reviews, so it will not help much lowering the maintenance.

webknjaz: I suppose coupling that with auto-approving should help.

webknjaz (approved):

PR #3386, opened 2022-01-10
Delete PULL_REQUEST_TEMPLATE.md

ssbarnea:
That file is more of a inconvenience as we already have labels that determine the PR type. Also this does not play well with `gh` cli as it will append the template to the commit message, so someone will need to remove the template from the pull-request.


tadeboro (approved):
Always +1 when it comes to deleting lines of code ;)

webknjaz:
Well, it's only inconvenient when you don't use it. I'd just put the expectations there instead of having to explain the same things in reviews.

Issue #3381, opened 2022-01-07
Getting Started Docs are Insufficient for Getting Started


justin-f-perez:
# Issue Type

- Bug report

# Molecule and Ansible details

```
❯ python3 --version && ansible --version && molecule --version
Python 3.9.9
ansible [core 2.12.1]
  config file = None
  configured module search path = ['/Users/me/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/me/path/to/ansibroken/venv/lib/python3.9/site-packages/ansible
  ansible collection location = /Users/me/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/me/path/to/ansibroken/venv/bin/ansible
  python version = 3.9.9 (main, Nov 21 2021, 03:16:13) [Clang 13.0.0 (clang-1300.0.29.3)]
  jinja version = 3.0.3
  libyaml = True
molecule 3.5.2 using python 3.9
    ansible:2.12.1
    delegated:3.5.2 from molecule
    docker:1.1.0 from molecule_docker requiring collections: community.docker>=1.9.1
```

Molecule installation method (one of):

- pip (as recommended by the getting started docs)

Ansible installation method (one of):

- pip (as recommended by the getting started docs)

Detail any linters or test runners used:

None

# Desired Behavior

Just trying out molecule, having never used it before. The commands specified in the getting started docs aren't sufficient for someone who is- literally- just getting started. I expect onboarding documentation (especially for a testing package) to run without (unintentional) errors.

Please give some details of the feature being requested or what
should happen if providing a bug report.

# Actual Behaviour

Script:
```
#!/bin/sh
set -x

mkdir ansibroken && cd ansibroken
# create a venv with python >=3.6 https://molecule.readthedocs.io/en/latest/installation.html#requirements
# python3.9 is supported https://github.com/ansible-community/molecule/blob/main/setup.cfg
python3.9 -m venv ./venv && source ./venv/bin/activate
# make sure setuptools is up to date https://molecule.readthedocs.io/en/latest/installation.html#pip
python3 -m pip install --upgrade setuptools
# install ansible, molecule with the docker driver https://molecule.readthedocs.io/en/latest/installation.html#install
python3 -m pip install 'molecule[docker,lint]' 'ansible-core>=2.8'
# do the things the newbie docs tell me to https://molecule.readthedocs.io/en/latest/getting-started.html#creating-a-new-role
[...]

PR #3393, opened 2022-01-11
Improve documentation of molecule init role

ssbarnea:
Current implementation knows to add the namespace from the command line to `meta/main.yml` but we had few places where we forgot to update the examples.

Fixes: #3381


apatard:
yeah, it's matching the test I did the other day. While looking at the commit, found out that the ``verify.yml`` example may need the same kind of treatment: s/name: my_role/name: acme.my_new_role/ ? Maybe in a different PR ?

apatard (approved):

Issue #3285, opened 2021-11-01
`ansible-base` should be switched to `ansible-core`


gotmax23:
Hi Molecule developers,

## Issue Type

- Bug report

## Molecule and Ansible details

``` console
$ ansible --version && molecule --version
ansible [core 2.11.6] 
  config file = /home/gotmax/.ansible.cfg
  configured module search path = ['/home/gotmax/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.10/site-packages/ansible
  ansible collection location = /home/gotmax/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.0 (default, Oct  4 2021, 00:00:00) [GCC 11.2.1 20210728 (Red Hat 11.2.1-1)]
  jinja version = 3.0.2
  libyaml = True
molecule 3.5.2 using python 3.10 
    ansible:2.11.6
    delegated:3.5.2 from molecule
    docker:1.1.0 from molecule_docker requiring collections: community.docker>=1.9.1
    podman:1.0.1 from molecule_podman requiring collections: containers.podman>=1.7.0 ansible.posix>=1.3.0
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- OS package

[Non applicable questions removed]

## Description

I think we should switch the `ansible-base` extra dependency to `ansible-core` (or at least add another extra called `ansible-core`). After `ansible-base` 2.10, the name of the minimal ansible package was changed to `ansible-core`. Currently, `pip install molecule[ansible] --user` installs ansible (the community package that just includes collections) and its dependency `ansible-core<2.12,>=2.11.6` (the new version of the minimal Ansible package), but `pip install molecule[ansible-base] --user` installs a different version of the minimal Ansible package.

I know that `ansible` is an optional dependency and it is possible to just run `pip3 install molecule ansible-core`, but I still think we should switch it to avoid user confusion and prevent conflicting dependencies. What do you think?

Related: https://github.com/thegeeklab/ansible-later/issues/242

Thanks,
Maxwell

mafalb:
It is valid to want ansible-base installed (if you want to test explicitly against 2.10), so I would I would suggest not to switch but add another extra for ansible-core, then one could use either molecule[ansible] xor molecule[ansible-base] xor molecule[ansible-core].

```
ansible =
    ansible >= 2.10
ansible-base =
    ansible-base ~= 2.10.0
[...]

PR #3395, opened 2022-01-11
Remove `ansible` and `ansible-base` extras

ssbarnea:
In 4.0.0 we officially remove the deprecated extras so users can manually install the blend of ansible they want.

Previous commit to main started v4 development and now we have a `stable/3.6` created for backporting purposes.

Fixes: #3285

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -48,9 +48,9 @@ Pip
 
 .. warning::
 
-  The old ``ansible`` and ``ansible-base`` pip extras are now deprecated and will
-  be removed in molecule 3.6.0. If you use them, please switch to explicit
-  package mention to avoid problem with newer versions of molecule.
+  ``ansible`` and ``ansible-base`` pip **extras** were removed in molecule
+  ``4.0.0``. If you used them, please switch to explicit package mention to
+  avoid problem with newer versions of molecule.
 
 Keep in mind that on selinux supporting systems, if you install into a virtual
 environment, you may face :gh:`issue <ansible/ansible/issues/34340>` even
--- a/setup.cfg
+++ b/setup.cfg
@@ -71,12 +71,6 @@ install_requires =
     rich >= 9.5.1
 
 [options.extras_require]
-# ansible extra is deprecated, will be removed in next version
-ansible =
-    ansible >= 2.10
-# ansible-base extra is deprecated, will be removed in next version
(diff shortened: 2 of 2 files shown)


tadeboro (changes requested):
I do not think we should remove deprecated functionality in the minor version. And this is doubly true for deprecations we introduced to the repo only a few months ago. In my opinion, we should wait for Molecule 4 before we can remove this.

And I am aware that we documented the removal version, but that warning message should contain version 4 from the get-go.

docs/installation.rst
```diff
@@ -46,12 +46,6 @@ Pip
 
       $ python3 -m pip install molecule ansible-core
 
-.. warning::
```
webknjaz, line 49: Add a `.. versionremoved:: 3.6.0` so that the absence is documented.

ssbarnea: I am not going to add a new sphinx plugin for this but I kept it as a standard warning.

cidrblock (approved):

davedittrich:
@ssbarnea Is there an official and definitive installation guide for `ansible`, `ansible-lint`, `molecule` (with `docker` driver), and `testinfra`? I have been in a circular dependency hell for the past few weeks trying to make sure that `ansible-compat` cache is correctly populated, `molecule lint` invokes `ansible-lint` via `tox` the same way (it apparently does not) that `ansible-lint` from the command line works. It is really frustrating to track down some GitHub issue that was summarily closed before a definitive working installation is shown.

There are literally dozens of issues that include bits and pieces of what I am seeing in error messages, but when I try a suggestion from one it breaks some other component and I am back to square one. It would take a significant amount of time to adequately document all the various version combinations and attempts to fix things on separate branches that last long enough to get feedback in issues.

I had a functional `molecule` test environment with my own hack of symbolic links to an isolated collections directory that I placed in environment variables, but all of that was broken by the addition of `ansible-compat` into the mix (which doesn't seem to be documented well enough to even understand the side-effects, without reading through the source code of multiple Ansible products.) Right now, it isn't populating the cache directory with my collection any more and I'm back into another round of searching.

ssbarnea:
Stop using lint feature in molecule and do the linting separately. Its use was already discourage and will be removed in a future version anyway. I really cannot provide much information about testinfra and I am not using myself.

PR #3396, opened 2022-01-12
Refactor coverage implementation

ssbarnea:
Adopts the same coverage configuration we use on ansible-compat, which includes:

- avoid using pytest-cov
- integration with codecov.io
- avoid noisy messages from codecov on pull requests, as   we have a GHA check to report its status

Related: https://github.com/ansible-community/ansible-lint/pull/1810

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -29,26 +29,33 @@ jobs:
           - tox_env: py36
             PREFIX: PYTEST_REQPASS=448
             python-version: 3.6
+            cover: true
           - tox_env: py37
             PREFIX: PYTEST_REQPASS=448
             python-version: 3.7
+            cover: true
           - tox_env: py38
             PREFIX: PYTEST_REQPASS=448
             python-version: 3.8
+            cover: true
           - tox_env: py39
             PREFIX: PYTEST_REQPASS=448
             python-version: 3.9
+            cover: true
           - tox_env: py310
             PREFIX: PYTEST_REQPASS=448
             python-version: "3.10"
+            cover: true
           - tox_env: py38-devel
             PREFIX: PYTEST_REQPASS=448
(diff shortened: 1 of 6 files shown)


webknjaz (changes requested):
I don't think this is justified. Also, why do you mix up changes to Codecov reporting with the  coverage collection — they are separate.

ssbarnea:
> I don't think this is justified. Also, why do you mix up changes to Codecov reporting with the coverage collection — they are separate.

If you want to keep pytest-cov, you are welcome to make pull-request that adds codecov integration in parallel. It would be interesting to compare their results. This one reported 43% on https://app.codecov.io/gh/ansible-community/molecule/compare/3396

cidrblock:
I know this is just a draft, so feel free to ignore.  (and my review wasn't requested) Why run coverage separate from pytest?  Just curious as we might do something similar elsewhere.....

ssbarnea:
@webknjaz I moved this as ready for review and I am looking forward to merge it unless you can find a way to demonstrate that is either bad or there is another better change that achieve the same goal: having codecov.io report coverage on https://app.codecov.io/gh/ansible-community so we can prevent regressions on incoming PRs.

webknjaz:
> having codecov.io report coverage on [app.codecov.io/gh/ansible-community](https://app.codecov.io/gh/ansible-community) so we can prevent regressions on incoming PRs.

This usually works for me (and I use parallel mode in almost every project). What's the problem?

ssbarnea:
It seems that coverage.py by itself was not able to capture all coverage we do, likely failed on our subprocesses or xdist as it reported about half of what pytest-cov did. In their docs it does say that it should work, at least if you pass that magic env variable. Anyway, as pytest-cov worked i decided to keep it.

PR #3397, opened 2022-01-13
docs: molecule doesn't support interactive passwd

justin-f-perez:
I just recently started using ansible, ansible-vault, and molecule. Today I created my first vault, and added `ask_vault_pass = True` to my ansible.cfg. I spent a little too long trying to figure out what I did wrong- but it turns out this feature simply isn't supported by molecule (and doesn't seem to be documented outside of the issue tracker.)

Hopefully this short note will help future newbies 😄

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -142,3 +142,12 @@ approach like this in your role's ``meta/main.yml``:
   dependencies:
     - role: <your-dependee-role>
       when: lookup('env', 'MOLECULE_FILE')
+
+Can I use ASK_VAULT_PASS with molecule?
+=======================================
+
+No, molecule does not support interactive password prompts. If you need to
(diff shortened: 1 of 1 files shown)


justin-f-perez:
PS: I wonder whether the question should be more generalized, e.g.,

> Can I use interactive password prompts with molecule?

also, it might help indexing/search-ability to explicitly mention each flag, its underscored `.cfg` variant, and its all-caps env. var variant, e.g.

> This includes, but isn't necessarily limited to:
> * CLI args: --ask-vault-pass, --ask-su-pass, --ask-become-pass, --ask-sudo-pass
> * the corresponding config variables: ask_vault_pass, ...
> * the corresponding environment variables: ASK_VAULT_PASS, ...

justin-f-perez:
nevermind. I keep forgetting molecule ignores ansible.cfg ...

PR #3398, opened 2022-01-15
Enable codecov.io integration

ssbarnea:
Apparently pytest-cov does a [better job](https://app.codecov.io/gh/ansible-community/molecule/pulls?page=1&state=open&order=-pullid) measuring coverage than just coverage.py as it reports 93% instead of 43% via #3396

Closes: #3396


webknjaz:
It does. But the implementation in this PR is quite wrong.

Issue #3406, opened 2022-01-21
Remove the mol command alias


mperry2:
Issue #2737 should be reverted. We should not pollute the path with duplicate executables. This is **_exactly_** what aliases exist for, and users can easily create an alias if they wish to use mol to call molecule.

Maybe add a deprecation warning for one year and then remove it for good after that.

mperry2:
I submitted a PR, but I'm not sure if I put the deprecation in the correct place. I copied the format from what Ansible uses in order to be consistent.

ssbarnea:
As the author of the `mol` feature I have mixed feelings but I will wait for other molecule users to [vote up/down on that proposal](https://github.com/ansible-community/molecule/issues/3406#issue-1109986272) before making a decision.

Removal should not be a problem because is a recent feature and I doubt we have many users using it, especially in production (scripts). Considering that a major version of molecule is already baking I could even backport the deprecation on `3.5.x` and make a small patch, so we could include full removal in upcoming `3.6.0`. Still, i want feedback before acting on it.

tadeboro:
I am +1 on removing the mol alias, but not in the minor release.

ssbarnea:
@tadeboro The idea was to deprecate it in minor release and remove in major. Or you mean that you want to change the versioning pattern on molecule? So far, `3.0`,... `3.5` all counted as major releases, all introducing some breaking changes. Is that your call for `4.0`?

tadeboro:
I was convinced Molecule respects semver, but I guess I was wrong (and lucky that none of the projects I use molecule on broke during the version bumps). I was targeting 4.0 with my comment, but my argument is completely invalid if Molecule breaks things in "minor" releases.

Anyway, I think a deprecation warning should be in place for at least a few months before we can remove the mol entry point.

greg-hellings:
I agree with this proposal in principle. I use several `mol...` aliases and would use one for the full executable if I wanted to shorten it as well.

ziegenberg:
The `mol` command alias was removed in #3457 (and documentation was fixed in #3458). Therefor I guess we can close this issue.

PR #3407, opened 2022-01-21
Add deprecation warning for the `mol` command

mperry2:
The `mol` command can be replaced with a shell alias for those users that need
it. Output a deprecation warning when the `mol` command is run that informs the
user that the command will be removed in a future release of Molecule.

Addresses issue #3406

### DIFF
--- a/src/molecule/__main__.py
+++ b/src/molecule/__main__.py
@@ -20,6 +20,14 @@
 """Molecule CLI main entry point."""
 
 from molecule.shell import main
+import warnings
+
+def mol_deprecation_notice():
+    warnings.warn(
+        "The `mol` command is deprecated and will be removed in a future version of Molecule. If you wish to continue using it, please create an alias that calls `molecule`.",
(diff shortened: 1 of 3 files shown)


src/molecule/__main__.py
```diff
@@ -21,5 +21,17 @@
 
 from molecule.shell import main
 
+
+def mol_deprecation_notice():
+    print(
```
tadeboro, line 26: We should probably use the [warnings](https://docs.python.org/3/library/warnings.html#module-warnings) module here since there is already support for deprecating things.

tadeboro (changes requested):

mperry2:
I've updated the commit to use the warnings module. I set the category to `UserWarning` because output wasn't produced when I used `DeprecationWarning`, although the latter seems like it would be correct. Let me know if I need to change that.

tadeboro (approved):

ssbarnea (approved):

Issue #3404, opened 2022-01-18
ANSIBLE_ROLES_PATH not set correctly?


sdarwin:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version
```
5.1.0 and 3.5.1

Molecule installation method (one of):

- source
- pip

Have tried both methods.

Ansible installation method (one of):

- source
- pip
- OS package

pip

# Desired Behavior

Molecule tests that have been passing, should continue to be passing.  When running in CI, if roles are installed, the ANSIBLE_ROLES_PATH points to the roles.

# Actual Behaviour

Running molecule ansible tests in CircleCI.

Certain tests that succeeded one month ago are failing now, even though nothing apparently changed.  It's not impossible, that this is being caused by _something else entirely_ such as Ansible itself or another package.   However at the moment it looks like molecule so your feedback would be helpful to point in the right direction.  

----

Case 1:
https://app.circleci.com/pipelines/github/cloudalchemy/ansible-prometheus?branch=master&filter=all

When molecule ran on Dec 27, 2021, it succeeded.  
Here is a interesting snippet from the log output:

```
INFO     Added ANSIBLE_ROLES_PATH=/home/circleci/.cache/ansible-compat/dda4ac/roles:/home/circleci/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles
INFO     Using /home/circleci/.ansible/roles/cloudalchemy.prometheus symlink to current repository in order to enable Ansible to find the role using its expected full name.
```

Notice that it's creating a symlink.

Now,

```
INFO     Set ANSIBLE_ROLES_PATH=/home/circleci/.cache/ansible-compat/dda4ac/roles:/home/circleci/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles
```

The wording changed from "Added" to "Set".   So, something changed.  And it does not mention the symlink anymore.  Probably no symlink is being created. Because of that the test fails, during  "molecule syntax".  It can't find the role.

----

Case 2:

This repo used to succeed, about a month ago.
https://app.circleci.com/pipelines/github/prometheus/demo-site/833/workflows/39439543-5173-4027-8be0-3d1817db7b7c/jobs/3397

Now it's failing:
https://app.circleci.com/pipelines/github/prometheus/demo-site/857/workflows/82accec5-9ae4-4c17-ba50-9c25d35b550e/jobs/3516

When does it fail? During the tests, it gets to this step:

```
[...]

PR #3409, opened 2022-01-22
Fix prepare_environment call for role installs

scruffaluff:
Fixes local role installations with `ansible-compat` version 1.0.0 as discussed in issue #3404.

### DIFF
--- a/src/molecule/command/base.py
+++ b/src/molecule/command/base.py
@@ -108,7 +108,7 @@ def execute_cmdline_scenarios(scenario_name, args, command_args, ansible_args=()
 
         if scenario.config.config["prerun"]:
             LOG.info("Performing prerun...")
-            scenario.config.runtime.prepare_environment()
+            scenario.config.runtime.prepare_environment(install_local=True)
 
         if command_args.get("subcommand") == "reset":
             LOG.info("Removing %s", scenario.ephemeral_directory)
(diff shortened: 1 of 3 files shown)


ssbarnea:
That fix has a serious problem: it does not include a test so we would not see similar regression again.

artis3n:
Is there a way we can get this PR in a good state to merge? All of my molecule pipelines are currently broken due to this. (the temporary fix referenced through the commits in this PR is an intermediary bypass).

scruffaluff:
I've added an assertion to one of the tests to check that the role symbolic link is successfully created. Let me know if there are any other changes you want me to make.

src/molecule/command/base.py
```diff
@@ -108,7 +108,7 @@ def execute_cmdline_scenarios(scenario_name, args, command_args, ansible_args=()
 
         if scenario.config.config["prerun"]:
             LOG.info("Performing prerun...")
-            scenario.config.runtime.prepare_environment()
+            scenario.config.runtime.prepare_environment(install_local=True)
```
mohsinhijazee, line 111: Cannot see a use case yet but maybe this can com from a flag. Thank you for your fix, I was able to test it locally and it works.

ssbarnea (approved):

PR #3414, opened 2022-01-24
Update installation.rst

mkunin-nvidia:
It is not clear that `community.docker` should be installed. The error message shows this, however, I think, it is worth mentioning in the installation guide this step.


ssbarnea:
Molecule by itself does not require docker collection. That is either a local requirement that you have or caused by the use of molecule-docker package. Latest versions of molecule and plugins are also able to automatically installed required collections at runtime, when the module is needed.

I am going to close this but let us know if we missed anything. We always try to improve the docs.

PR #3416, opened 2022-01-26
Add support for MOLECULE_PLATFORM_NAME

### LINKED ISSUE #3415: Support --platform parameter
Support to test only one platform image
```
export MOLECULE_PLATFORM_NAME=ubuntu-18.04
molecule test
```

### Discussed in https://github.com/ansible-community/molecule/discussions/3410

<div type='discussions-op-text'>

<sup>Originally posted by **zhan9san** January 24, 2022</sup>
Support `--platform` for `molecule test` command.

```
--platform TEXT        Name of the platform to test
```

e.g.

Only test role on `centos-7` platform in `default` scenario, if no platform found, raise error.

```
molecule test --platform centos-7
```

Q: Why is this required?
A: Assume there are 10 platform docker images defined in `molecule.yml` and only one image(centos-7) are changed.

- With `--platform`, only one changed docker image would be tested via `molecule test --platform centos-7`
- Without `--platform`, all 10 platform docker image would be test while the other 9 are redundant.

More specifically, the `run: molecule test` can ben changed to `run: molecule test --platform ${{ matrix.package }}`.
https://github.com/zhan9san/systemd/blob/main/.github/workflows/ci.yml#L70</div>

zhan9san:
Fix #3415 

There are no environment variable `MOLECULE_PLATFORM_NAME` set by default.

How to use

1. Test one image

  Assume platform name is `ubuntu-18.04`
  ```bash
  molecule test --platform=ubuntu-18.04
  ```
  or
  ```bash
  export MOLECULE_PLATFORM_NAME=ubuntu-18.04
  molecule test
  ```

2. Test all images

  ```bash
  unset MOLECULE_PLATFORM_NAME
  molecule test
  ```

### DIFF
--- a/src/molecule/command/test.py
+++ b/src/molecule/command/test.py
@@ -31,6 +31,7 @@ from molecule.config import DEFAULT_DRIVER
 
 LOG = logging.getLogger(__name__)
 MOLECULE_PARALLEL = os.environ.get("MOLECULE_PARALLEL", False)
+MOLECULE_PLATFORM_NAME = os.environ.get("MOLECULE_PLATFORM_NAME", None)
 
 
 class Test(base.Base):
@@ -116,6 +117,12 @@ class Test(base.Base):
     help=f"Name of the scenario to target. ({base.MOLECULE_DEFAULT_SCENARIO_NAME})",
 )
 @click.option(
+    "--platform-name",
+    "-p",
+    default=MOLECULE_PLATFORM_NAME,
+    help="Name of the platform to target only. Default is None",
+)
+@click.option(
     "--driver-name",
     "-d",
     type=click.Choice([str(s) for s in drivers()]),
@@ -140,7 +147,14 @@ class Test(base.Base):
 )
 @click.argument("ansible_args", nargs=-1, type=click.UNPROCESSED)
 def test(
-    ctx, scenario_name, driver_name, __all, destroy, parallel, ansible_args
+    ctx,
+    scenario_name,
+    driver_name,
+    __all,
+    destroy,
+    parallel,
+    ansible_args,
+    platform_name,
 ):  # pragma: no cover
     """Test (dependency, lint, cleanup, destroy, syntax, create, prepare, converge, idempotence, side_effect, verify, cleanup, destroy)."""
     args = ctx.obj.get("args")
@@ -150,6 +164,7 @@ def test(
         "destroy": destroy,
         "subcommand": subcommand,
         "driver_name": driver_name,
+        "platform_name": platform_name,
     }
 
     if __all:
--- a/src/molecule/platforms.py
+++ b/src/molecule/platforms.py
@@ -68,13 +68,19 @@ class Platforms(object):
               - child_group1
     """
 
-    def __init__(self, config, parallelize_platforms=False):
+    def __init__(self, config, parallelize_platforms=False, platform_name=None):
         """
         Initialize a new platform class and returns None.
 
         :param config: An instance of a Molecule config.
+        :param parallelize_platforms: Parallel mode. Default is False.
+        :param platform_name: One platform to target only, defaults to None.
         :return: None
         """
+        if platform_name:
+            config.config["platforms"] = util._filter_platforms(
+                config.config, platform_name
+            )
         if parallelize_platforms:
             config.config["platforms"] = util._parallelize_platforms(
                 config.config, config._run_uuid
--- a/src/molecule/test/functional/test_command.py
+++ b/src/molecule/test/functional/test_command.py
@@ -57,6 +57,14 @@ def driver_name(request: FixtureRequest) -> Optional[str]:
         return None
 
 
+@pytest.fixture
+def platform_name(request):
+    try:
+        return request.param
+    except AttributeError:
+        return None
+
+
 @pytest.mark.extensive
 @pytest.mark.parametrize(
     "scenario_to_test, driver_name, scenario_name",
@@ -291,6 +299,42 @@ def test_command_syntax(scenario_to_test, with_scenario, scenario_name):
 )
 def test_command_test(scenario_to_test, with_scenario, scenario_name, driver_name):
     run_test(driver_name, scenario_name)
+
+
+def run_test_with_platform_name(
+    driver_name, platform_name, scenario_name="default", parallel=False
+):
+    cmd = [
+        "molecule",
+        "-vvv",
+        "--debug",
+        "test",
+        "--scenario-name",
+        scenario_name,
+        "--platform-name",
+        platform_name,
+    ]
+    if driver_name != "delegated":
+        if scenario_name is None:
+            cmd.append("--all")
+        if parallel:
+            cmd.append("--parallel")
+
+    assert run_command(cmd).returncode == 0
+
+
+@pytest.mark.serial
+@pytest.mark.parametrize(
+    ("scenario_to_test", "driver_name", "scenario_name", "platform_name"),
+    [
+        ("driver/delegated", "delegated", "default", "instance"),
+    ],
+    indirect=["scenario_to_test", "driver_name", "scenario_name", "platform_name"],
+)
+def test_command_test_with_platform_name(
+    scenario_to_test, with_scenario, scenario_name, driver_name, platform_name
+):
+    run_test_with_platform_name(driver_name, platform_name, scenario_name)
 
 
 @pytest.mark.serial
--- a/src/molecule/test/unit/test_platforms.py
+++ b/src/molecule/test/unit/test_platforms.py
@@ -35,3 +35,26 @@ def test_instances_property(_instance):
     ]
 
     assert x == _instance.instances
+
+
+@pytest.fixture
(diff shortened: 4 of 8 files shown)


src/molecule/command/test.py
```diff
@@ -115,6 +116,12 @@ def execute(self):
     default=base.MOLECULE_DEFAULT_SCENARIO_NAME,
     help=f"Name of the scenario to target. ({base.MOLECULE_DEFAULT_SCENARIO_NAME})",
 )
+@click.option(
+    "--platform-name",
+    "-p",
+    default=MOLECULE_PLATFORM_NAME,
+    help=f"Name of the platform to target only. ({MOLECULE_PLATFORM_NAME} means all)",
```
webknjaz, line 123: This will say "{whatever is in env} means all" which is probably not what's intended. Am I reading this correctly? Should this say something like "XXX by default"?

src/molecule/platforms.py
```diff
@@ -68,13 +68,19 @@ class Platforms(object):
               - child_group1
     """
 
-    def __init__(self, config, parallelize_platforms=False):
+    def __init__(self, config, parallelize_platforms=False, platform_name=None):
         """
         Initialize a new platform class and returns None.
 
         :param config: An instance of a Molecule config.
+        :param parallelize_platforms:
```
webknjaz, line 76: Since you're adding this (although I'd argue that it's probably out of the scope), could you fill out the text description for this param too?

src/molecule/test/functional/test_command.py
```diff
@@ -287,6 +296,20 @@ def test_command_test(scenario_to_test, with_scenario, scenario_name, driver_nam
...
+    "scenario_to_test, driver_name, scenario_name, platform_name",
+    [
+        ("driver/delegated", "delegated", "default", "instance"),
+    ],
+    indirect=["scenario_to_test", "driver_name", "scenario_name", "platform_name"],
+)
+def test_command_test_with_platform_name(
+    scenario_to_test, with_scenario, scenario_name, driver_name, platform_name
+):
+    run_test_with_platform_name(driver_name, platform_name, scenario_name)
```
webknjaz, line 310: Since this function is used as a name, I'd say that it should arguably be implemented right here, not in conftest. Conftest exists for test helpers, not test logic. Would you move it, please?

src/molecule/test/functional/test_command.py
```diff
@@ -287,6 +296,20 @@ def test_command_test(scenario_to_test, with_scenario, scenario_name, driver_nam
     run_test(driver_name, scenario_name)
 
 
+@pytest.mark.serial
+@pytest.mark.parametrize(
+    "scenario_to_test, driver_name, scenario_name, platform_name",
```
webknjaz, line 301: Let's use distinct values for readability:
```suggestion
    ("scenario_to_test", "driver_name", "scenario_name", "platform_name"),
```

src/molecule/test/unit/test_platforms.py
```diff
@@ -28,10 +28,23 @@ def _instance(config_instance):
...
         {"groups": ["foo", "bar"], "name": "instance-1", "children": ["child1"]},
         {"groups": ["baz", "foo"], "name": "instance-2", "children": ["child2"]},
     ]
 
     assert x == _instance.instances
+
+
+def test_instances_property_with_platform_name(_instance_with_platform_name):
+    x = [
+        {"groups": ["foo", "bar"], "name": "instance-1", "children": ["child1"]},
```
webknjaz, line 47: `instance-1` is hardcoded in the test body and the detached fixture. It is best to keep such coupled things in the same place. I think that having an extra fixture is not warranted in this case. But if it was, this should've definitely gone to "parameterize" where it could be evident/visible that these values are connected.

src/molecule/command/test.py
zhan9san: Thanks for your attention. You're right.

The description is modified in new commit.

src/molecule/platforms.py
zhan9san: It's fine. The description for parallelize_platforms is added in new commit

src/molecule/test/functional/test_command.py
zhan9san: Thank you. I agree with you.

This function is moved from `conftest` into `test_command`.

But I noticed there are 2 function still in `conftest`, [run_test and verify](https://github.com/ansible-community/molecule/blob/4f8fd055ab183a1928b3460ccb39e94ac6162ebf/src/molecule/test/functional/test_command.py#L31-L32)

Do we need to move them either?

src/molecule/test/functional/test_command.py
zhan9san: The value has been changed to a tuple.

But I noticed a string separated by comma is used in all other functions as well as [parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html)

Could you give us some more details on this difference?

src/molecule/test/unit/test_platforms.py
zhan9san: Thank you for pointing it out.

Two parametrizing test functions are added.

`test_instances_property_with_platform_name_instance_1` and `test_instances_property_with_platform_name_instance_2` is to avoid conditional assertions.

Let me know if you have any concerns.

ssbarnea:
I am wondering if it would not be better to implement it the same way ansible does:
```
  -l SUBSET, --limit SUBSET
                        further limit selected hosts to an additional pattern
```

zhan9san:
@ssbarnea 
Thanks for your attention.
I prefer to use '--platform' as `platform` is self-explanatory and used extensively in `molecule` project.

But if you insist, I don't mind changing it to 'subset'.

ssbarnea:
Any chance to refresh this?

zhan9san:
The PR is rebased on main.
Could you kindly review it?

ssbarnea:
@zhan9san Did you ever tried to run the tests locally because I never seen them passing on that change.

zhan9san:
Hi @ssbarnea 
Thanks for your time.

> I never seen them passing on that change.

Actually, it did pass the checks once.

As the main branch is growing faster, and it is easy to get the expected [PYTEST_REQPASS](https://github.com/ansible-community/molecule/blob/main/.github/workflows/tox.yml#L30) mismatched in PR.

I am wonder would it be possible to remove the `PYTEST_REQPASS` in Github action. Maybe it will avoid some unnecessary warnings.

> I will mark it as draft to avoid being noisy and we should get it out of draft only when the CI is passing on it.

Here is [a test PR](https://github.com/zhan9san/molecule/pull/2) in my forked repository, [the pipeline failed](https://github.com/zhan9san/molecule/runs/5354481724?check_suite_focus=true) at uploading coverage data because of the authentication issue.

I am leaving this PR status as draft. Could you please review it and trigger the pipeline?

Thanks for your amazing work maintaining this project.

ssbarnea:
If you change the number of tests you are expected to also update the `PYTEST_REQPASS` values. That variable is used only as a fail-safe mechanism for protecting us from making changes that might stop all/some tests from being run.

zhan9san:
Hi @ssbarnea 

Thank you for running the pipeline. The CI failed at `sudo apt-get update`.

> If you change the number of tests you are expected to also update the PYTEST_REQPASS values.

Please allow me to describe the following scenario. One PR may resolve conflicts many times if it is not merged in time.

Time 0: PYTEST_REQPASS is 500 in main branch
Time 1: 3 tests are added in `PR-1` and `PYTEST_REQPASS` is changed to 503, and CI for PR is successful. PR-1 is ready to merge.
Time 2: 4 tests are added in `PR-2` and `PYTEST_REQPASS` is changed to 504, and CI for PR is successful. PR-2 is ready to merge.
Time 3: `PR-1` is merged into main branch, which results in at least one conflict in PR-2 about `PYTEST_REQPASS`.
Time 4: Rebase PR-2 branch on main to resolve the conflict, and the `PYTEST_REQPASS` becomes 507 in PR-2 branch.

Due to some reasons, PR-2 is still not merged into master and PR-3 is created.

Time 5: `PR-3` is created and 5 tests are added and merged into main. The `PYTEST_REQPASS` becomes 508, which will also introduce at least one conflict in PR-2.

zhan9san:
Hi @ssbarnea 

The code is rebased on main, and the conflict `PYTEST_REQPASS` is fixed as well.

Would you mind checking it again?

ssbarnea (approved):

gardar:
@zhan9san @ssbarnea 
Shouldn't this option exist for other operations as well? Such as `molecule converge`, `molecule lint`, etc. From a users perspective I kind of expected the option to exist for all of the other commands that accept the `-s` flag.
It would be great to be able to quickly test a single platform during development with `molecule converge -s scenaro -p platform` and it would also be beneficial to be able to run the molecule lint/syntax check on a single platform in a CI since there is really no need to run the linting multiple times.

(Perhaps I should open a separate issue to discuss this)

zhan9san:
@gardar 

It would be great if you can submit a PR as well.

Issue #2540, opened 2020-02-07
molecule fails to reset itself when driver is changes


ssbarnea:
I observed that molecule gets confused when a driver is swithced inside a scenario, like moving from `delegated` to `docker` casing confusing errors that do not give a hint about what caused them.
```
--> Action: 'destroy'
ERROR! the playbook: None could not be found

An error occurred during the test sequence action: 'destroy'. Cleaning up.
ERROR:
--> Scenario: 'foo'
--> Action: 'cleanup'
Skipping, cleanup playbook not configured.
--> Scenario: 'foo'
--> Action: 'destroy'
ERROR! the playbook: None could not be found
```

Funny part is that this scenario is not supposed to have these playbooks. 

Removing temp files of molecule does solve the issue, but most users do not even know where these are located.

I am not yet sure how we can decide when we are in a tainted environment and to suggest performing a reset as we do want to do it by accident. Maybe timestamping molecule.yml file would be enough? When this is changed we can throw a clearly visible warning which tells user that config was tainted since last run and that they may want to do a reset if somethign fails.

Daniel-Sanchez-Fabregas:
Thanks @ssbarnea, removed temporal files (in my case  ~/.cache/molecule/) as workaround and everything ok.

ssbarnea:
There is now the `molecule reset` command that can do this but is not run automatically.

If anyone wants to implement this it would be really nice first feature.

PR #3429, opened 2022-02-06
Added state-file driver validation

superstes:
It seems the driver pulled from existing state-file(s) was not yet validated against the available drivers.

This was problematic for me as I previously used 'vagrant' and changed it to use 'docker'.
That simple check (as already used in the molecule.yml schema-validation) should fix this behaviour.

I pulled my hair out yesterday trying to understand why the configured 'docker' driver was not used by molecule.
There was no transparent way to know why molecule did not use my configuration. (as an user)


ssbarnea:
@superstes Is my impression or this should fix https://github.com/ansible-community/molecule/issues/2540 ?

src/molecule/config.py
```diff
@@ -230,7 +230,7 @@ def _get_driver_name(self):
         driver_from_state_file = self.state.driver
         driver_from_cli = self.command_args.get("driver_name")
 
-        if driver_from_state_file:
+        if driver_from_state_file and driver_from_state_file in api.drivers():
```
apatard, line 233: maybe I'm seeing problems where there are none because I've not understood what it's doing... I'm wondering if the check is enough. For instance, the scenario is configured to use driver "foo", the user installs driver "bar" and modify the configuration to use driver "bar" but doesn't uninstall the driver "foo", will the driver change get noticed ?

tadeboro: I think those are two different problems. "We are missing a driver that initialized scenario is using" is one and the other is "Scenario and its initialized instance diverged."

Current PR is trying to solve the first issue (making sure we can still converge, etc., against an initialized scenario) and does not tackle the second issue.

I am not sure if it is even possible to solve the second problem without making some severe changes to Molecule's internals. Why? It seems to me that once we initialize the scenario, Molecule should ignore changes to the source configuration and always use configuration from the initialized instance. But currently, Molecule does not distinguish this and always uses source config as a fallback configuration.

tadeboro, line 233: We should move that check to the end of the function and fail the Molecule run if the selected driver is missing.

tadeboro (changes requested):

superstes:
@[ssbarnea](https://github.com/ssbarnea) 

This issue is related to #2540, yes.

I completely agree with you that the user should at least get a warning when the cache might be 'invalid'.

I'll give my input if I've found a good way of solving this.

> Removing temp files of molecule does solve the issue, most users do not even know where these are located.

So true. I only found it after looking into the code.

src/molecule/config.py
superstes: Done.

I added the check at the end:
`if driver_from_state_file and not driver_name in api.drivers()`

One could just check `if driver_name in api.drivers()` but it seems to me that the other cases (_input via shell argument and config_) already get validated.
Also - this way the user get's to know that the state-file is the information-source and where it is located.

ssbarnea: I wonder if we should just prompt the user to clean the cache in that case with an interactive prompt. WDYT? (obviously that this would appear only on tty sessions).

ssbarnea:
@superstes If you find the approval annoying please find a very simple fix, like a typo and propose it. This approval appears only for those that never contributed and cannot be disabled.

superstes:
@ssbarnea Sorry for the delay - updated the test.

src/molecule/config.py
superstes: Yeah - I believe CI wouldn't run into those problems as the cache/runner instance has a short lifetime.

May I ask completely unbiased => what is the state file even used for?
As far as I've seen its only reference inside molecule's source is this one.
`config.py -> self.state -> self.state.driver`
Have I overlooked something?

ssbarnea (approved):

PR #3446, opened 2022-02-14
docs: ansible-lint is not a molecule dep

### LINKED ISSUE #3445: remnants of ansible-lint in molecule docs
# Issue Type

- Bug report

Note: the python was installed via conda, but the python packages are installed using pip, not conda.
```
ansible [core 2.12.2]
  python version = 3.9.10 | packaged by conda-forge | (main, Feb  1 2022, 21:25:34) [Clang 11.1.0 ]
  jinja version = 3.0.3
  libyaml = True
  ansible python module location = /opt/homebrew/Caskroom/miniconda/base/envs/tmptest/lib/python3.9/site-packages/ansible
molecule 3.6.1 using python 3.9
    ansible:2.12.2
    delegated:3.6.1 from molecule
```

- pip

- pip

ansible-lint

# Desired Behavior

Assuming ansible-lint was removed for good reason and shouldn't be added back into the `molecule[lint]` extra:

I think this would make clearer ansible-lint is never installed by molecule:
* add note here that ansible-lint is not installed by molecule, even if you installed the `[lint]` extra
 https://molecule.readthedocs.io/en/latest/configuration.html?highlight=ansible-lint#lint
* remove misleading comment in setup.cfg https://github.com/ansible-community/molecule/blob/c33c205b570cd95d599d16afa8772fabba51dd40/setup.cfg#L112
* change to actual default (yamllint) https://github.com/ansible-community/molecule/blob/c7ae6a27bed9ba6423d6dfe11d8e0d5c54da094f/src/molecule/command/init/scenario.py#L165

# Actual Behaviour

I can't say for sure (don't know project history well enough) but I *think* ansible-lint was once available through the `molecule[lint]` extra, but was subsequently promoted to a regular dependency, and eventually removed from molecule deps altogether. It appears there were a few spots in docs that got missed and may mislead a new user (see: me) into thinking ansible-lint will be installed with `pip install molecule[lint]` when in fact it is not.

justin-f-perez:
resolves #3445


justin-f-perez:
A few notes:
this PR assumes ansible-lint was removed from molecule altogether as a dependency so users who would like it need to install it separately


in docs/configuration.rst I changed 'is' to 'was' to clarify that this code is not literally reused from ansible-lint in the sense that it's a dependency, but rather that it was copied

the change in setup.cfg is, in my opinion, the most important. this is the comment that led me to believe `pip install molecule[lint]` (or even `pip install molecule`) would give me `ansible-lint` too.

For the docs, I updated in the 3 places I thought most relevant: 
1. someone installing molecule (could be a newbie, could be a vet with a foggy memory regarding extras)
2. someone just getting started for the first time
3. someone updating their lint config

ssbarnea (approved):

docs/installation.rst
```diff
@@ -96,6 +96,13 @@ Install Molecule:
 
     $ python3 -m pip install --user "molecule[lint]"
 
+Molecule does not include ansible-lint (nor does the lint extra), but
+is easily installed separately:
+
+.. code-block:: bash
+
+    $ python3 -m pip install --user "molecule[lint] ansible-lint"
```
ssbarnea, line 104: Remove lint extra from here, it is used only internally.

docs/getting-started.rst
```diff
@@ -122,7 +122,8 @@ keys represent the high level components that Molecule provides. These are:
   the driver to delegate the task of creating instances.
 
 * The :ref:`lint` command. Molecule can call external commands to ensure
-  that best practices are encouraged.
+  that best practices are encouraged. Note: `ansible-lint` is not included with
+  molecule or molecule[lint].
```
ssbarnea, line 126: Remove mention of extra.

docs/installation.rst
justin-f-perez: Ahh... maybe this is where all of the confusion originally stems from? The [current documentation](https://molecule.readthedocs.io/en/latest/installation.html) instructs users to install via:
```bash
$ python3 -m pip install --user "molecule[lint]"
```

If I understand correctly, `molecule[lint]` shouldn't be referenced in the user documentation at all. Is this correct? If so, I'll double check that there are no other references to it in the user docs and fix them all in this PR.

ssbarnea: You are right, that's mostly for internal use. There are no direct imports from molecule itself to any stuff inside lint exta.

PR #3454, opened 2022-03-01
Remove paramiko dependency

ssbarnea:
As paramiko is not used by molecule core we should not
list it as a dependency.


briantist (approved):
The changes make sense to me, but I don't understand why that test is failing.

ssbarnea:
> The changes make sense to me, but I don't understand why that test is failing.

That test failure is random and #3455 will fix it.

PR #3455, opened 2022-03-01
Bump ansible-compat

ssbarnea:
v2.0.0 fixes some concurrency bugs that are affecting our testing from time to time. Still there is no pressure to bump minimal required version as older ones should still be usable.


briantist:
is there a link to the changelog for ansible-compat?

ssbarnea:
> is there a link to the changelog for ansible-compat?

https://github.com/ansible-community/ansible-compat/releases but it would be useful to include that in main project readme and docs page, to make it easier to discover.

PR #3467, opened 2022-03-14
[feat] add args for verify and side_effect actions

amarao:
This feature allows to provide alternative location for
tests or side effects playbook in test sequence like this:
```
  test_sequence:
    - create
    - converge
    - verify
    - side_effect side_effect1.yml
    - verify tests_for_effect1/
    - side_effect side_effect2.yml
    - verify tests_for_effect2/
```

This allows to create more comprehensive tests for
resulting system, allowing, in paticular, to check
if it's works as indendent:
* after server reboot
* after applying system changes and reverting them back
* after introducing external stress

Classic apprach with one-side-effect per scenario can create
high overhead for both converge operations (if they are long),
and maintenance (because each scenario directory need to be
kept up-to-date).

This patch keep original behavior intact. If no argiments
are provided to verify and side_effect actions, they uses
values from molecule.yml. If there is action argument, they
* replace side effect playbook with list of playbooks after
  `side_effect` action (in molecule.yml), relative to molecule.yml file.
* replace tests (and `additional_files_or_dirs`) with list of
  locations (relative to `molecule.yml`), while keeping
  all other test settings intact. This is supported for both
  verifiers (testinfra and ansible).


amarao:
I have no permissions to set tags on PR. It's ready for review, tests are green.

briantist:
wow, I really like this!

ssbarnea (approved):

nkakouros:
This is not in the docs, right?

apatard:
@nkakouros unfortunately, it would not be the first time updating the doc is forgotten. If you have time, a PR with a doc update would be nice or at least, please open a bug report to make sure that it's not forgotten / lost ?

Issue #3405, opened 2022-01-20
Code examples in documentation are not rendered


radek-sprta:
# Issue Type

- Bug report

# Desired Behavior

Documentation at https://molecule.readthedocs.io/en/latest/configuration.html is properly rendered.

# Actual Behaviour

The code examples at https://molecule.readthedocs.io/en/latest/configuration.html (and other parts of documentation) are missing.

steven-cd:
I had a lot of trouble understanding what exactly molecule was for and how it worked until I found this issue and realised the docs were broken and had several parts missing.

This seems to be because the builds on Readthedocs are raising several warnings (although they still pass) along the lines of:

```
WARNING: autodoc: failed to import class 'config.Config' from module 'molecule'; the following exception was raised:
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/molecule/envs/latest/lib/python3.7/site-packages/sphinx/ext/autodoc/importer.py", line 70, in import_module
    return importlib.import_module(modname)
  File "/home/docs/checkouts/readthedocs.org/user_builds/molecule/envs/latest/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/docs/checkouts/readthedocs.org/user_builds/molecule/envs/latest/lib/python3.7/site-packages/molecule/config.py", line 32, in <module>
    from molecule import api, interpolation, platforms, scenario, state, util
  File "/home/docs/checkouts/readthedocs.org/user_builds/molecule/envs/latest/lib/python3.7/site-packages/molecule/api.py", line 11, in <module>
    from molecule.verifier.base import Verifier  # noqa
  File "/home/docs/checkouts/readthedocs.org/user_builds/molecule/envs/latest/lib/python3.7/site-packages/molecule/verifier/base.py", line 26, in <module>
    from molecule import util
[...]

PR #3474, opened 2022-03-21
Adapt the readthedocs config to be similar to what ansible-lint is using

ziegenberg:
This commit changes the way the python dependencies for readthedocs are configured and it mirrors the way `ansible-lint` is doing it. The `docs` extra_requires is dropped and replaced with an `/docs/requirements.in` file. The `tox.ini` configuration is adapted to this change.

`sphinx.fail_on_warning` is set to `true` for readthedocs config. This turns warnings into errors (`-W` and `--keep-going options`). This means the build fails if there is a warning and exits with exit status 1. This mirrors the current configuration in `tox.ini`. So if building the docs locally fails, it now also fails with readthedocs.

A simple `Readme.md` for building documentation is added.

Also the pre-commit `pip-compile` steps are adapted, a `pip-compile-docs` and `pip-compile-docs-upgrade` hook is added.

Fixes: #3405


ziegenberg:
@ssbarnea this is ready for review and it fixes building the docs at readthedocs.org  :)

ziegenberg:
I'll try to come up with an alternative approach for this.

ziegenberg:
I'll try to come up with an alternative approach for this.

zollo:
I strongly suggest we get this in as soon as possible, the lack of documentation is making it difficult for us to use Molecule. Happy to lend my support in developing/reviewing this PR!

zollo:
@ziegenberg thanks so much for your work on this!

requirements.txt
```diff
@@ -2,15 +2,12 @@
 # This file is autogenerated by pip-compile with python 3.8
 # To update, run:
 #
-#    pip-compile --extra=docs --extra=test --extra=yamllint --no-annotate --output-file=requirements.txt --strip-extras setup.cfg tools/opts.txt
+#    pip-compile --extra=test --extra=yamllint --no-annotate --output-file=requirements.txt --strip-extras setup.cfg tools/opts.txt
```
ziegenberg, line 5: @ssbarnea I guess this should be changed?
```suggestion
#    pip-compile --extra=test --extra=lint --no-annotate --output-file=requirements.txt --strip-extras setup.cfg tools/opts.txt
```

ziegenberg:
py-devel keeps failing and I'm pretty sure I did not change anything related... :(

Issue #2759, opened 2020-07-22
Cannot use "inventory.links.group_vars" together "inventory.host_vars"


y-hashida:
# Issue type

* Feature request
* Question

# molecule and ansible version

```
$ molecule --version
molecule 3.0.6
   ansible==2.9.11 python==3.8

$ ansible --version
ansible 2.9.11
```

# Desired Behavior

I would like to configure the following settings in "provisioner.inventory" of  "molecule.yml".

```yaml
provisioner:
  ... snip ...
  inventory:
    host_vars:
      ${MOLECULE_DISTRO}-${ANSIBLE_VER}-1:
        ipv4_address: "192.168.1.11"
      ${MOLECULE_DISTRO}-${ANSIBLE_VER}-2:
        ipv4_address: "192.168.1.12"
    links:
      group_vars: ${GROUP_VARS}
  ... snip ...
```

As you can see from this snippet, here is a summary of what I want to set up.

* Specify the link indicated by ${GROUP_VARS} in group_vars
* Include host_vars in molecule.yml.

# Actual Behaviour

However, in practice, only the setting of group_vars:${GROUP_VARS} in inventory.links applies. inventory.host_vars will be ignored.

Conversely, if you remove inventory.links, host_vars is not ignored.

```yaml
provisioner:
  ... snip ...
  inventory:
    host_vars:
      ${MOLECULE_DISTRO}-${ANSIBLE_VER}-1:
        ipv4_address: "192.168.1.11"
      ${MOLECULE_DISTRO}-${ANSIBLE_VER}-2:
        ipv4_address: "192.168.1.12"
    #links:
    #   group_vars: ${GROUP_VARS}
  ... snip ...
```

In referring to the molecule source code, it seems that the conditional branching is based only on the presence or absence of inventory.links.

https://github.com/ansible-community/molecule/blob/17c5da0e831a36f226769029d6452d8d0d692fd7/molecule/provisioner/ansible.py#L767-L778

If we can't use inventory.**links.group_vars** and inventory.**group_vars** together, then that's understandable, but if you can't use inventory.**links.group_vars** and inventory.**host_vars** together, then very inconvenient.

I would therefore like to request the following features.

* The feature to use inventory.**links.group_vars** and inventory.**host_vars** together.
* The feature to use inventory.**links.host_vars** and inventory.**group_vars** together.

Alternatively, I would like to know if there is a way to make this work as you wish without modifying the source code.

Best regards.

# My molecule.yml and converge.yml

[...]

PR #3501, opened 2022-04-08
Allow mixing links and inventory variables

cinek810:
Previous logic allowed either link or direct specification in molecule.yml, while it's very convenient to link to group_vars containing some common values and set other (specific to molecule test) as host_vars.

Fixes: #2759


cinek810:
This is to fix the issue: https://github.com/ansible-community/molecule/issues/2759

amarao:
Hello. It looks amazing and it fixing one of weak points of Molecule! I'm very excited to see it fixed.

Can you, please, add a few tests? It can quickly become messy (there is a lot of corner cases), and having some tests will helps to cover them all.

cinek810:
Sure. Can you point me to an example commit I can follow preparing the tests?

ssbarnea:
Can someone step in here to add some tests, without any tests I will likely close it.

cinek810:
I someone can point me to a commit (or external link) where similar tests where added I'll work on that. I just don't have relevant experience.

ssbarnea:
Can you point me to a minimal scenario that was failing before this change and that is working after it? ... that would be a test by itself.

ssbarnea:
Closing this due to lack of updates.

lifedanceman:
Hello! Are there any changes about the pull request?

jub-ioki:
@ssbarnea , could we re-open this please? 
As for a minimal case where molecule fails without this code, here is one...
```
provisioner:
  name: ansible
  inventory:
    links:
      group_vars: ../../../../group_vars/
   host_vars:
      molecule-host1:
        enable_thing: true
platforms:
  - name: molecule-host1
```
In this case molecule will silently ignore host host_vars and `enable_thing` will not be `true` on `molecule-host1`.  This behavior *is* documented, the documentation has the following statement:

> You can use either hosts/group_vars/host_vars sections of inventory OR links. If you use both, links will win.

But that statement is buried pretty deeply in the documention, and there is no runtime error or warning.  We could add a warning for it, but better yet to allow both, which is exactly what this PR does.

PR #3510, opened 2022-04-13
Disable latex output

ssbarnea:
This might fix RTD build failure on main branch, something that does not happen on pull requests.

Related: https://twitter.com/readthedocs/status/1507388916013314048
Related: https://github.com/sphinx-doc/sphinx/issues/10332

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,7 @@
 ---
 version: 2
 
-formats: all
+formats: []
 
 build:
   image: latest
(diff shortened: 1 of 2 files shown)


cblegare:
@ssbarnea 

The place in Sphinx 4.5.0 where your build fails looks like this: https://github.com/sphinx-doc/sphinx/blob/4.5.x/sphinx/builders/latex/__init__.py#L266.  This variable comes directly from `self.document_data` which is set by `latex_documents` which is a 6-tuple https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=latex_documents#confval-latex_documents

You had a 5-tuple:

```python
 latex_documents = [(master_doc, "Molecule.tex", "Molecule Documentation", "manual")]
 ```

PR #3511, opened 2022-04-14
Set permissions for GitHub actions

naveensrinivasan:
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

 Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/ack.yml
+++ b/.github/workflows/ack.yml
@@ -4,6 +4,11 @@ on:
   pull_request_target:
     types: [opened, labeled, unlabeled, synchronize]
 
+permissions:
+  contents: read
+
 jobs:
   ack:
+    permissions:
+      contents: none
     uses: ansible-community/devtools/.github/workflows/ack.yml@main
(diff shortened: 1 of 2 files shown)


ssbarnea:
@ansible-community/devtools @gundalow I am sure if I should take this pull request as something good or bad. While security matters, I do not fancy getting random/unrequested help that looks more like spam. There is nothing to be pawned in these workflows because we don't use any secrets in them.

That reminds me about "road to hell is paved with good intentions". I do still remember when others scripted proposing typo fixes to projects and started to spam project on github. They ended up with their accounts being blocked by github for breaching ToS.

Smells too much as using bots to advertise other projects/initiatives...

ssbarnea:
@

Issue #3513, opened 2022-04-14
Molecule is failing when refreshing Collections cache


MonolithProjects:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible [core 2.12.4]
  config file = None
  configured module search path = ['/home/mike/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/mike/.local/lib/python3.10/site-packages/ansible
  ansible collection location = /home/mike/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/mike/.local/bin/ansible
  python version = 3.10.2 (main, Jan 17 2022, 00:00:00) [GCC 11.2.1 20211203 (Red Hat 11.2.1-7)]
  jinja version = 3.0.3
  libyaml = True
molecule 3.6.1 using python 3.10 
    ansible:2.12.4
    delegated:3.6.1 from molecule
    docker:1.1.0 from molecule_docker requiring collections: community.docker>=1.9.1
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

Running locally on Fedora 35

# Desired Behavior

Second molecule command will not fail.

# Actual Behaviour

When i run molecule (any command) for Ansible Collection which has more than one dependencies in galaxy.yml for a first time, everything works fine. When i after that re-run molecule (the same or different molecule command), molecule fails when refreshing collections cache. 

So basically the molecule is failing when there is no change in the collections cache.

As a workaround for this issue i am forced to manually delete the ansible cache before i run the command. Second options is to temporally comment out the dependencies in galaxy.yml after the cache was created.

This error came with Molecule `3.6.1`. Older versions works fine.

## Example
```sh
❯ git clone git@github.com:lablabs/ansible-collection-wireguard.git
Cloning into 'ansible-collection-wireguard'...
Warning: Permanently added 'github.com' (ED25519) to the list of known hosts.
remote: Enumerating objects: 213, done.
remote: Counting objects: 100% (148/148), done.
remote: Compressing objects: 100% (75/75), done.
remote: Total 213 (delta 67), reused 116 (delta 49), pack-reused 65
Receiving objects: 100% (213/213), 51.35 KiB | 559.00 KiB/s, done.
Resolving deltas: 100% (78/78), done.

❯ cd ansible-collection-wireguard/

❯ molecule create
INFO     default scenario test matrix: dependency, create, prepare
INFO     Performing prerun...
[...]

PR #3514, opened 2022-04-16
Insert cache_dir into ANSIBLE_ROLES_PATH

zhan9san:
Fix #3495 and #3464 

Before:

There is a conflict if someone installs a role from Ansible Galaxy and want to use `molecule test` to debug the same role in another directory.

The first part of `ANSIBLE_ROLES_PATH` is `~/.ansible/roles`.
A role installed by `ansible-galaxy`, would be in `~/.ansible/roles` by default.
`molecule test` would try to create a symlink at `~/.ansible/roles` as well. 

Because the role fqrn is the same, the `molecule test` would fail, referring to https://github.com/ansible-community/molecule/issues/3495#issuecomment-1091390551

After:
The role installed by `ansible-galaxy` is still in the first default `ANSIBLE_ROLES_PATH`, `~/.ansible/roles`.
The role created by 'molecule test' would be in `cache_dir`, e.g. `/Users/xxx/.cache/ansible-compat/496064/roles`

And this `cache_dir` would be inserted into the roles search path, `ANSIBLE_ROLES_PATH`.

PR #3516, opened 2022-04-20
Fix wrong environment variables being passed to ansible compat (#3404)

marblenix:
You can find a  demo of this error here:
https://github.com/marblenix/molecule-bug-demo-3404/commits/main

In the body of each commit, details the commands that are needed to be run to reproduce the error with minimal setup, with the last one pushing a requirements.txt file to this branch. To see the error, it's recommended to be working on a machine that hasn't had `ansible-galaxy collection install` run on it before. If you are unable to reproduce the issue locally, try removing the cached roles and collections folders.

```sh
python3 -m venv .venv; source .venv/bin/activate
pip install pip wheel; pip install 'ansible-lint[core,yamllint]' 'molecule[docker,lint]'
molecule init role acme.my_new_role --driver-name docker
cd my_new_role

pip install --upgrade --force \
        git+https://github.com/ansible-community/molecule@main \
        git+https://github.com/ansible/ansible-compat@main \
        git+https://github.com/ansible/ansible@devel
molecule test
```

To test my changes, run:

```sh
pip install --upgrade --force git+https://github.com/marblenix/molecule@main
```

I ran the test suite by enabling actions on my fork. The tox parts failed when it needed the secrets, but the unit tests themselves still passed. https://github.com/marblenix/molecule/actions/runs/2193046395

Please note, this **_will not_** solve the issue with #3513. It does allow for a workaround to this bug by adding the following to a requirements.yml file:

```yaml
collections:
  - community.docker
```

--

The exact cause of this issue is this line here in ansible-compat: https://github.com/ansible/ansible-compat/blob/05d94d4d1b484b05c53c9b5844e4187d9872c258/src/ansible_compat/runtime.py#L515

The problem is ansible-compat is adding these environment variables to the runtime, and molecule is adding its own variables to the provisioner. Combining them seems to resolve the issue just fine.

To verify that this specific problem is resolved, run tests like so: `molecule --debug -v test`. The very last output of `DEBUG: SHELL REPLAY:` should include a line about ANSIBLE_COLLECTIONS_PATH being in `~/.cache/ansible-compat/DEADBEEF/collections:...`

### DIFF
--- a/src/molecule/provisioner/ansible_playbook.py
+++ b/src/molecule/provisioner/ansible_playbook.py
@@ -44,7 +44,9 @@ class AnsiblePlaybook(object):
         self._playbook = playbook
         self._config = config
         self._cli = {}
-        self._env = util.merge_dicts(self._config.provisioner.env, self._config.runtime.environ)
+        self._env = util.merge_dicts(
+            self._config.provisioner.env, self._config.runtime.environ
+        )
 
     def bake(self):
         """
--- a/src/molecule/provisioner/ansible_playbook.py
+++ b/src/molecule/provisioner/ansible_playbook.py
@@ -44,7 +44,7 @@ class AnsiblePlaybook(object):
         self._playbook = playbook
         self._config = config
         self._cli = {}
-        self._env = self._config.provisioner.env
+        self._env = util.merge_dicts(self._config.provisioner.env, self._config.runtime.environ)
 
     def bake(self):
         """


ssbarnea:
I am wondering how were you able to run these commands considering that latest version of ansible-lint removed both "core" and "yamllint" extras, so pip install with them would fail. Are you sure you used py38 or newer?

marblenix:
> Are you sure you used py38 or newer?

@ssbarnea 

Thank you for asking. I tested this with my companies internal galaxy roles, and it worked there. We host them in individual private git repos, and this bug hit us pretty hard when molecule stopped working with requirements.yml.

The versions of software that were tested are as follows:

```
molecule 3.6.2.dev38 using python 3.9.12
    ansible:2.14.0.dev0
    delegated:3.6.2.dev38 from molecule
    docker:1.1.0 from molecule_docker requiring collections: community.docker>=1.9.1
```

The Jenkins server we use to run the molecule testing uses python 3.8, and this worked with ansible stable 2.12.3.

To be sure I had a clean environment, I deleted the `.venv` directory and started from scratch with the following commands:

```shell
python3 -m venv .venv; source .venv/bin/activate
pip install --upgrade pip wheel; pip install --upgrade 'ansible-lint[core,yamllint]' 'molecule[docker,lint]'
pip install --upgrade \
    git+https://github.com/marblenix/molecule@main \
    git+https://github.com/ansible/ansible-compat@main \
    git+https://github.com/ansible/ansible@devel
```

I hope that answers your question.

ssbarnea:
Lets hope I will not have to yank that fix.

apatard:
This PR is actually regressing things (at least molecule-vagrant CI). From the logs, I can see:
```
ANSIBLE_LIBRARY: /Users/runner/work/molecule-vagrant/molecule-vagrant/.tox/py39/lib/python3.9/site-packages/molecule/provisioner/ansible/plugins/modules:/Users/runner/work/molecule-vagrant/molecule-vagrant/.tox/py39/lib/python3.9/site-packages/molecule_vagrant/modules:/Users/runner/.cache/molecule/XFRUV/test-scenario/library:/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_command_init_scenario0/XFRUV/library:/Users/runner/.ansible/plugins/modules:/usr/share/ansible/plugins/modules
```
vs
```
ANSIBLE_LIBRARY: /Users/runner/.cache/ansible-compat/1e0daf/modules:/Users/runner/.ansible/plugins/modules:/usr/share/ansible/plugins/modules
```
(https://github.com/ansible-community/molecule-vagrant/runs/6122171777?check_suite_focus=true)

Which means that the path to the molecule-vagrant module is lost, leading to the failures. I've yet to find time to try finding a fix.

sio:
Uhm, if I understand this PR correctly, when a key exists both in `self._config.provisioner.env` and in `self._config.runtime.environ` only the value from `runtime` gets used.

I'm not sure why such order of priority has been chosen. It seems to me that it should be the other way around (or maybe a even more complex merging logic is required) since a lot of Molecule code is touching `self._config.provisioner` and these edits are obviously not safe to drop en masse.

@marblenix, can you elaborate on why your suggested order of priority should be used?

marblenix:
I tested the changes I made on all of our companies projects, but if there are reported broken builds, that's definitely my fault for this assumption. I'll make a PR to switch the order today.

apatard:
Can't we merge variable instead of overwriting things ?

marblenix:
@apatard , @sio , @ssbarnea 

I've cleaned up this implementation and added a test. See https://github.com/ansible-community/molecule/pull/3529

zhan9san:
Hi @marblenix 

> To verify that this specific problem is resolved, run tests like so: molecule --debug -v test. The very last output of DEBUG: SHELL REPLAY: should include a line about ANSIBLE_COLLECTIONS_PATH being in ~/.cache/ansible-compat/DEADBEEF/collections:...

If the issue is `~/.cache/ansible-compat/DEADBEEF/collections:...` is not in `ANSIBLE_COLLECTIONS_PATH`

Could you please consider the following PRs?

* Insert cache_dir into ANSIBLE_ROLES_PATH and ANSIBLE_COLLECTIONS_PATH #3514 
* install role in cache dir  https://github.com/ansible/ansible-compat/pull/135

PR #3526, opened 2022-04-25
Remove dependency on cookiecutter

### LINKED ISSUE #3486: Drop cookiecutter from molecule
As the future of cookiecutter project seems to be not sure I would like to propose dropping it before it becomes a real issue, even a security one (yep current release 1.7.3 from pypi being affected by one command injection security issue).

I personally reached a point where I am ok with any of the following approaches:

* use either ansible playbooks for templating
* removal of init functionality (templating)

I am fully aware that some people might like this but I do not think that the maintenance burden justifies keeping that feature, especially as its interaction with molecule provisioner plugins makes it even more problematic to use and debug.

As this is likely to count as a major change, it should be planned for v4 which is already on main branch. If anyone is interesting in helping with this task, please let me know.

## References

* https://github.com/cookiecutter/cookiecutter/issues/1555
* https://github.com/cookiecutter/cookiecutter/issues/1636
* https://github.com/cookiecutter/cookiecutter/issues/1642

zhan9san:
Fix #3486 

@ssbarnea 

Both `cookiecutter` and `ansible template` are using `Jinja`. So I decide to continue to use `Jinja` as a template engine.

To avoid reinventing the wheel, the template render logic is migrated from `cookiecutter` with tailored function and parameters.

I'd like to confirm with you whether the above way is accepted.
If yes, I'll add more tests later and fix the lint issue in this PR.
If no, do you have any suggestions?

### To do

Do you have any concerns for to-do?

1. Fix CI issue in this PR
2. Add template directory with out `{{` and mark `{{` directory deprecated in another PR

### Backward compatible

Currently, the following commands work as before. It means, this change is backward-compatible, and for other drivers, too.

1. `delegated` by default

  ```shell
  $ molecule init role acme.my_new_role_docker --driver-name docker
  INFO     Initializing new role my_new_role_docker...
  No config file found; using defaults
  - Role my_new_role_docker was created successfully
  [WARNING]: No inventory was parsed, only implicit localhost is available
  localhost | CHANGED => {"backup": "","changed": true,"msg": "line added"}
  INFO     Initialized role in /Users/jackzhang/tmp/my_new_role_docker successfully.
  ```
4. `docker` driver

  ```shell
  molecule init role acme.my_new_role
  INFO     Initializing new role my_new_role...
  No config file found; using defaults
  - Role my_new_role was created successfully
  [WARNING]: No inventory was parsed, only implicit localhost is available
  localhost | CHANGED => {"backup": "","changed": true,"msg": "line added"}
  INFO     Initialized role in /Users/jackzhang/Downloads/tmp/roles/my_new_role successfully.
  ```

Related: https://github.com/cookiecutter/cookiecutter/issues/1555 https://github.com/cookiecutter/cookiecutter/issues/1636 https://github.com/cookiecutter/cookiecutter/issues/1642


ssbarnea (approved):
That looks awesome but I did not perform a deep review as I am still in vacation. Lets get help from at least one more reviewer before acting as we do want to be confident in it.

zhan9san:
@ssbarnea 

Enjoy yourself.

At least, you made me think I am on the right track.

ssbarnea:
@zhan9san Do not forget to fix the linting.

zhan9san:
@ssbarnea 

I am working on it.

src/molecule/generate.py
```diff
@@ -0,0 +1,359 @@
...
+        json_exc_message = str(e)
+        our_exc_message = (
+            f'JSON decoding error while loading "{full_fpath}".  Decoding'
+            f' error details: "{json_exc_message}"'
+        )
+        raise ContextDecodingException(our_exc_message)
+
+    # Add the Python object to the context dictionary
+    file_name = os.path.split(context_file)[1]
+    file_stem = file_name.split(".")[0]
```
apatard, line 99: why not using something like (untested) ? :
```
file_name = os.path.basename(context_file)
file_stem = file_name.splitext(file_name)[0]
```

apatard (commented):
I'm seeing no documentation change so I assume that it fully transparent to the user ?

src/molecule/generate.py
zhan9san: Hi @apatard 

Thanks for your review.

It is referred to [cookiecutter generater.py](https://github.com/cookiecutter/cookiecutter/blob/682fe854dca3ce735d656a6588e9ea9369c4ea47/cookiecutter/generate.py#L105-L107)

They do the same thing getting the filename without extension.

zhan9san: Hi @apatard 

`file_name` is generated by `os.path.basename(context_file)` is `string`, which does not have `splitext` function.

So I only change the `file_name` from `os.path.split(context_file)[1]` to `os.path.basename(context_file)`.

zhan9san:
@ssbarnea 

All CI, including `lint` jobs pass.

Would you kindly review this PR?

src/molecule/generate.py
felixfontein: I think @apatard meant `os.path.splitext(file_name)[0]` and not `file_name.splitext(file_name)[0]`. While I think that would be better, it is a potentially breaking change, since the current code splits at the *first* `.`, while `os.path.splitext` splits at the *last* `.`.

src/molecule/cookiecutter.py
```diff
@@ -0,0 +1,61 @@
+"""cookiecutter main function."""
```
felixfontein, line 1: I'm not a lawyer, but if this contains code that was taken from cookiecutter, doesn't it need to carry the [license](https://github.com/cookiecutter/cookiecutter/blob/master/LICENSE) and copyright from cookiecutter? (Same for other code taken from that project.)

src/molecule/generate.py
zhan9san: Hi @felixfontein 

Thanks for your review. You are right.

The `file_stem ` is generated by `os.path.splitext(file_name)[0]` in new commit.

For this specific case, as `cookiecutter.json` is used, the value of `file_stem` would be `cookiecutter`.

zhan9san:
> I'm seeing no documentation change so I assume that it fully transparent to the user ?


Yes, it's a backward compatible migration.

src/molecule/cookiecutter.py
zhan9san: @felixfontein 

Thanks for your reminder.

The cookiecutter [license](https://github.com/cookiecutter/cookiecutter/blob/master/LICENSE) and copyright are added in [new commit](https://github.com/ansible-community/molecule/pull/3526/commits/fd16041eb4c9b82e8c20fe77b694abf134f0dd87), referring to how `interpolation.py` did.
https://github.com/ansible-community/molecule/blob/04157c3684a46987ba00a6223261e3da30ff5b0c/src/molecule/interpolation.py#L1-L16

ssbarnea:
@apatard @felixfontein Any more concerns?

zhan9san:
> Cookiecutter is live again - we released 2.1.0 https://pypi.org/project/cookiecutter/2.1.0/

> _Originally posted by @jensens in https://github.com/ansible-community/molecule/issues/3486#issuecomment-1141455376_

If we continue to use `cookiecutter`, we can close this PR.

zhan9san:
Thanks.

I think it's not urgent. I'll try to make it work.

Besides, if `cookiecutter` works well, we don't want to reinvent the wheel.

PR #3529, opened 2022-04-26
Add more tests related to Ansible runtime path modifications

marblenix:
See: https://github.com/ansible-community/molecule/pull/3516


sio:
I like that you merge the values here instead of just overwriting one with another.

Yet the solution feels somewhat hacky: you introduce special handling only for ANSIBLE_ROLES_PATH and ANSIBLE_COLLECTIONS_PATH. How do we know that those are the only env vars that need to be merged? What about ANSIBLE_LIBRARY_PATH? There may be more such variables, how should we treat them and why?

Also, the code is not exactly DRY and has a potential to become even more repetitive. What if we decide that another variable needs to follow the same logic? I think that merging of colon-separated lists should be split off into a function.

marblenix:
Thanks for the feedback, I'll get to work on updating the pull request.

apatard:
> I like that you merge the values here instead of just overwriting one with another.
> 
> Yet the solution feels somewhat hacky: you introduce special handling only for ANSIBLE_ROLES_PATH and ANSIBLE_COLLECTIONS_PATH. How do we know that those are the only env vars that need to be merged? What about ANSIBLE_LIBRARY_PATH? There may be more such variables, how should we treat them and why?
> 
I was about to suggest writing a modified version of utils.merge_dicts() but I'm not sure how to make this function generic. I mean there's no way to know how to merge two variables without knowing what the env variables contain.

marblenix:
@sio
I hope this implementation is much cleaner. I've also added some tests.

ssbarnea (approved):

zhan9san:
Hi @ssbarnea 

The original issue is `~/.cache/ansible-compat/DEADBEEF/collections:...` is not in `ANSIBLE_COLLECTIONS_PATH`

Could you please consider the following implementation?

* Insert cache_dir into ANSIBLE_ROLES_PATH and ANSIBLE_COLLECTIONS_PATH #3514 
* install role in cache dir  https://github.com/ansible/ansible-compat/pull/135

@bonddim has verified they can fix #3404.

marblenix:
@webknjaz 

If you have the time today, could review this pr?
Thank you.

ssbarnea:
Can you please rebase this?

src/molecule/provisioner/ansible_playbook.py
```diff
@@ -53,6 +53,24 @@ def __init__(self, playbook, config, verify=False):
...
+        )
+        collections_paths = util.merge_paths(
+            self._env.get(config.ansible_collections_path),
+            self._config.runtime.environ.get(config.ansible_collections_path),
+        )
+
+        self._env["ANSIBLE_ROLES_PATH"] = roles_paths
+        self._env["ANSIBLE_LIBRARY"] = library_paths
+        self._env[config.ansible_collections_path] = collections_paths
+
```
zhan9san, line 73: If I understand it correctly, you want to append `cache_dir` created by `ansible-compat` in `ANSIBLE_ROLES_PATH`.

It has been inserted into `ANSIBLE_ROLES_PATH` at https://github.com/ansible-community/molecule/blob/d263e982dec4379dbd93a3f2addd6c5501b1e4ad/src/molecule/provisioner/ansible.py#L446-L468

Is it duplicate?

marblenix: Correct! This was fixed with this commit by Jack as of yesterday: https://github.com/ansible-community/molecule/commit/b9d8f7f30560acbe30cb4705e69d5b264d9ecca9

Since his got merged first, I'll remove the code changes I made, but keep the tests.

zhan9san (approved):

PR #3546, opened 2022-05-17
Use verifier env in ansible verifier

### LINKED ISSUE #3527: Ansible Verifier does not pass its env variables to the verify step
# Issue Type

- Bug report

```
ansible [core 2.12.3]
  config file = None
  configured module search path = ['/Users/g/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/g/.pyenv/versions/3.10.3/envs/ansible-avd/lib/python3.10/site-packages/ansible
  ansible collection location = /Users/g/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/g/.pyenv/versions/ansible-avd/bin/ansible
  python version = 3.10.3 (main, Mar 18 2022, 09:31:01) [Clang 13.0.0 (clang-1300.0.29.30)]
  jinja version = 3.0.3
  libyaml = False
molecule 3.4.1 using python 3.10
    ansible:2.12.3
    delegated:3.4.1 from molecule
    docker:1.1.0 from molecule_docker requiring collections: community.docker>=1.9.1

```

- pip

- pip

# Desired Behavior

env variables given to the verifier are usable in the verify playbook. i.e. in the output below the debug message should read `VERIFIER`

# Actual Behaviour

Please give some details of what is actually happening.
Include a [minimum complete verifiable example](https://stackoverflow.com/help/mcve) with
output of running `molecule --debug`.

The current Ansible Verifier is calling the Provisionner `verify` function: https://github.com/amarao/molecule/blob/master/src/molecule/verifier/ansible.py#L83. This means that env variables given to the verifier section are not used but rather the ones from the provisionner are:

```
g% cat molecule/default/molecule.yml
---
dependency:
  name: galaxy
driver:
  name: delegated
platforms:
  - name: instance
provisioner:
  name: ansible
  env:
    DEMO: "PROVISIONER"
verifier:
  name: ansible
  env:
    DEMO: "VERIFIER"
g% cat molecule/default/verify.yml
---
# This is an example playbook to execute Ansible tests.

- name: Verify
  hosts: all
  gather_facts: false
  tasks:
  - name: Example assertion
    debug:
      msg: "{{ lookup('env', 'DEMO') }}"

```

will result in:

```
g% molecule verify
[...]
PLAY [Verify] ******************************************************************

TASK [Example assertion] *******************************************************
ok: [instance] => {
    "msg": "PROVISIONER"
}
```

zhan9san:
Fix #3527 

@ssbarnea  Could you kindly review this PR?

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -27,23 +27,23 @@ jobs:
           - tox_env: docs
             python-version: 3.9
           - tox_env: py38
-            PREFIX: PYTEST_REQPASS=450
+            PREFIX: PYTEST_REQPASS=451
             python-version: 3.8
             cover: true
           - tox_env: py39
-            PREFIX: PYTEST_REQPASS=450
+            PREFIX: PYTEST_REQPASS=451
             python-version: 3.9
             cover: true
           - tox_env: py310
-            PREFIX: PYTEST_REQPASS=450
+            PREFIX: PYTEST_REQPASS=451
             python-version: "3.10"
             cover: true
           - tox_env: py38-devel
-            PREFIX: PYTEST_REQPASS=450
+            PREFIX: PYTEST_REQPASS=451
             python-version: 3.8
             cover: true
           - tox_env: py310-devel
-            PREFIX: PYTEST_REQPASS=450
+            PREFIX: PYTEST_REQPASS=451
             python-version: "3.10"
             # see https://github.com/ansible-community/molecule/issues/3291
             experimental: true
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -27,23 +27,23 @@ jobs:
           - tox_env: docs
             python-version: 3.9
           - tox_env: py38
-            PREFIX: PYTEST_REQPASS=451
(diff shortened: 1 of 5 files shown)


.github/workflows/tox.yml
```diff
@@ -27,23 +27,23 @@ jobs:
           - tox_env: docs
             python-version: 3.9
           - tox_env: py38
-            PREFIX: PYTEST_REQPASS=450
+            PREFIX: PYTEST_REQPASS=451
```
ziegenberg, line 30:
```suggestion
            PREFIX: PYTEST_REQPASS=453
```
Hi @zhan9san, you actually added three more tests, not just one. Can you adjust your commit accordingly (also below)?

ziegenberg (changes requested):

.github/workflows/tox.yml
zhan9san: Hi @ziegenberg 

Thanks for your reminder. It's done.

ssbarnea (approved):

PR #3571, opened 2022-06-01
doc: update doc on inventory/host section for provisioner

amarao:
1. Fix incorrect example, as 'hosts' should be two times, first
   time as a name for for 'inventory' section, and second, as
   normal Ansible `hosts` stanza in the group.

2. Add well-known fact about incompatibility of sections of
   inventory and links.

### DIFF
--- a/src/molecule/provisioner/ansible.py
+++ b/src/molecule/provisioner/ansible.py
@@ -19,6 +19,7 @@
 #  DEALINGS IN THE SOFTWARE.
 """Ansible Provisioner Module."""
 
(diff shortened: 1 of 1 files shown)


amarao:
Pylint is complaining on command option lines, I hadn't touch those.

ssbarnea (approved):

apatard (approved):
Too bad that there's a failure with pylint but since it's not related to this change... Still, I'm surprised that latest commits were getting no errors in the CI.

PR #3599, opened 2022-06-20
Avoid harassment from dependabot

ssbarnea:


ziegenberg:
I like the PR wording :)

PR #3600, opened 2022-06-20
Docs: fix undocumented env vars

gardar:
* Adds description for `MOLECULE_` env vars that was missing
* Adds default paths to all env vars that contain a path.

#### PR Type

    * Docs Pull Request


zhan9san (approved):
LGTM

PR #3639, opened 2022-08-19
Enable stale bot

zhan9san:
Do we need to enable stale bot?

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false


ssbarnea:
@zhan9san That is indeed a very good question. We do want to close some really old threads that are high-likely to no longer valid or where the original OP stopped responding.

Still, we do not want to just close tickets without warning, as we might have things that we need to keep open for years, so we do not forget about the.

Let me consult with @cidrblock on this and see which values are safe for a first trial. I also want to check if we could keep the configuration of stale bot centralized so it would apply to all our repositories.

Shortl, YES, we want the stable bot, but we need to be careful with what settings we use it, to avoid closing things too soon.

apatard:
While some kind of automatic treatment used to detect stale issues and clean things a little bit (well, 27 issues is not a lot...), I'm not sure how to properly configure the bot. We need  to be sure to have the new issues at least read by some people to ensure that the bot won't wrongly close things due to missing tags. I mean, there are some bugs needing to be dealt with later (eg low priority or things like that).

A stale bot may be a nice tool but can't help when the problem is more related to the lack of people looking at issues.

zhan9san:
If an issue is considered needing to be fixed later, a special label like 'pinned' can be added on the issue.

And the special label has to be added in 'exemptLabels' in stale.yml to make the issues never be considered stale

https://github.com/probot/stale


At the worst case, if an issue has never been added the special label for a long time like 60 days, it means either the lack of maintainer resource or no one cares about it anymore

ssbarnea:
Apparently stalebot does not also have an option to lock old closed issues, something that can prove useful. There is no week I do not get at least one message where someone makes an useless comment on something that should have being buried years ago. Maybe there is an alternative that can also lock? It would be better to investigate now, before we start investing more time on it.

Also, note that we will make the configuration inside the devtools project, not on each repository and use [extends](https://github.com/probot/probot/blob/c34f64d2f40ebebce8c3014758e5420ecfbb5994/docs/best-practices.md#store-configuration-in-the-repository).

But we can continue the discussion here until we decide which will be the first set of settings to enabled.

cognifloyd:
I really don't like the stale bot.

I'm always fighting with it across many projects, so I have to keep saying "yes this is still valid". So, unless there is a way for a non-maintainer to keep their issue open, I'm opposed to the stale bot.

I understands the impetus behind trying to close old stuff, but it still feels like a slap in the face every time another issue I care about gets closed in ansible/ansible, or when I find the bug I'm fighting with in an issue on other projects, but that is in a closed-as-stale issue.

When I search for a big I'm experiencing and find old bugs closed as stale, I take that as a signal that the project does not want that bug fixed, and does not care to leave an issue open as a record that it is a known issue, and therefore I don't work on a fix for it. "Just file a new issue" is a waste of effort because the maintainers probably won't chime in, and even if I make a PR, it'll probably end up closed as stale too.

If you want to mark something as stale without closing it, that is much less of a slap in the face. Still annoying, but acceptable.

ziegenberg:
> I really don't like the stale bot.
> 
> I'm always fighting with it across many projects, so I have to keep saying "yes this is still valid". So, unless there is a way for a non-maintainer to keep their issue open, I'm opposed to the stale bot.
> 
> I understands the impetus behind trying to close old stuff, but it still feels like a slap in the face every time another issue I care about gets closed in ansible/ansible, or when I find the bug I'm fighting with in an issue on other projects, but that is in a closed-as-stale issue.
> 
> When I search for a bug I'm experiencing and find old bugs closed as stale, I take that as a signal that the project does not want that bug fixed, and does not care to leave an issue open as a record that it is a known issue, and therefore I don't work on a fix for it. "Just file a new issue" is a waste of effort because the maintainers probably won't chime in, and even if I make a PR, it'll probably end up closed as stale too.
> 
> If you want to mark something as stale without closing it, that is much less of a slap in the face. Still annoying, but acceptable.

I second that!

zhan9san:
It makes maintainer feel awkward to close a stale issue which was indeed important a long time ago.
Because we don't know whether it is needed any more, or if someone still pay attention to it.

If an issue is not closed now, it's just postponed and is still hard to be closed in the future. The maintainers has to make a decision when it can be closed. So a specific stale time is introduced.

> There is no week I do not get at least one message where someone makes an useless comment on something that should have being buried years ago.

If we have issue locked, it may become `someone create an useless issue` because the issue which should have being buried years ago is locked.

So there may be still noise even we `lock old closed issues`.

> I'm always fighting with it across many projects, so I have to keep saying "yes this is still valid". So, unless there is a way for a non-maintainer to keep their issue open, I'm opposed to the stale bot.

If any maintainers has time to handle it, a special label like `pinned` can be added by the maintainer. Instead, if none of them think it is valuable, or it is a rare case for a specific user only, it would be close several days later.

> unless there is a way for a non-maintainer to keep their issue open

If a non-maintainer keep the issue open in a specific time, it's still hard for either non-maintainer or maintainer to recognize the issue which has been inactive for a long time is still needed in the future.

PR #3651, opened 2022-09-04
Reconfigure dependabot

ssbarnea:


zhan9san:
@ssbarnea 

Please allow me to ask a stupid question.

When/How do we bump the packages defined in `setup.py`?

They are defined in a range instead of a specific version.

AFAIK, dependabot can only help us bump the packages defined in `requirements.txt` which is generated by `pip-compile`.
And `requirements.txt` is considered as a constraint file.

ssbarnea:
@zhan9san Yes, our `requirements.txt` is only a constraints file. Usually we would name it constraints.txt but we keep using the former name because that is the only way to keep using dependabot with it, as dependabot has no option for custom filename.

When we make changes that require bumping the minimal version, we need to do it inside setup.cfg, manually. Still, we do this only when we know the the old version is not safe to use anymore.

zhan9san:
@ssbarnea 

Thanks for your time.

> When we make changes that require bumping the minimal version, we need to do it inside setup.cfg, manually. Still, we do this only when we know the the old version is not safe to use anymore.

And when/how do we know the old version is not safe to use anymore?

If I understand correctly, the `requirements.txt` is redundant, and can be left out.

I noticed the only reference of `requirements.txt` is in https://github.com/ansible-community/molecule/blob/b3a8f5ad0af54de479754b89906e393d948b0bbf/tox.ini#L45-L46

And `PIP_CONSTRAINT` would be applied to tox environment, [Depending on requirements.txt or defining constraints](
https://tox.wiki/en/latest/example/basic.html?highlight=PIP_CONSTRAINT#depending-on-requirements-txt-or-defining-constraints)

The trick here is that the `PIP_CONSTRAINT`(`requirements.txt`) is compiled from `setup.cfg`, but `tox` will install the dependencies defined in setup.cfg as well via https://github.com/ansible-community/molecule/blob/b3a8f5ad0af54de479754b89906e393d948b0bbf/tox.ini#L51-L52

The `requirements.txt` is user agnostic when one runs `pip install molecule` but `setup.cfg`. 

Upper limit of one dependency can be detected by scheduled CI task as it will always use the latest package. Once an incompatible packages is used, the CI job would fail. Then we can raise a PR, `FOO <= x.x.x` into `setup.cfg`.

## Test

I tried to delete the `requirements.txt` and remove the definition of `PIP_CONSTRAINT` in `tox.ini` and run `tox -r -e py10`. The test is successfully as well.

PR #3668, opened 2022-09-24
Adopt molecule config JSON schema

ssbarnea:
We move official location for Molecule Config JSON Schema inside molecule project itself. This will be followed by updates on JSON Schema Store and on vscode-ansible extension, and removal from ansible/schemas project. We already did the same with navigator and lint project, as it makes the maintenance easier.


src/molecule/data/molecule.json
```diff
@@ -0,0 +1,561 @@
...
+          "type": "string"
+        },
+        "options": {
+          "title": "Options",
+          "type": "object"
+        }
+      },
+      "title": "VerifierModel",
+      "type": "object"
+    },
```
davedittrich, line 492: The `VerifierModel` lacks the `directory` property, as noted in https://github.com/ansible-community/molecule/pull/3667. The last I looked, there was still code to process it, and it still appeared in documentation, so my assumption was the removal was an accident in the PR and it did not get caught in review.

Was this supposed to be removed? If so, could someone point me to the discussion about it and reasoning? It broke `testinfra` verification and I had to add it back on a fork to get unblocked. That would seem to indicate there is also a lack of tests to catch the regression. Could someone look into that?  Thanks!

Issue #3481, opened 2022-03-30
'molecule init' doesn't comply with documented 'molecule lint' configuration


Fixmetal:
# Issue Type

- Bug report

# Molecule and Ansible details

```
$ ansible --version
ansible [core 2.12.2]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/enomis/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.10/site-packages/ansible
  ansible collection location = /home/enomis/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.3 (main, Mar 18 2022, 00:00:00) [GCC 11.2.1 20220127 (Red Hat 11.2.1-9)]
  jinja version = 3.0.1
  libyaml = True

$ molecule --version
molecule 3.6.1 using python 3.10 
    ansible:2.12.2
    delegated:3.6.1 from molecule
    podman:1.1.0 from molecule_podman requiring collections: containers.podman>=1.7.0 ansible.posix>=1.3.0

```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

- yamllint
- ansible-lint
- flake8

# Desired Behavior

When creating a new role I would expect molecule to create a consistent skeleton for itself that complies with FQCN at the very least. This is because, as stated by doc:
```
Molecule supports only the latest two major versions of Ansible (N/N-1), meaning that if the latest version is 2.9.x, we will also test our code with 2.8.x.
```

So as of today, I'm expecting it to support these:
- ansible-core 2.12
- ansible-core 2.11
- ansible 5.5
- ansible 5.4

After applying as suggested about linting by the https://github.com/ansible-community/molecule/blob/v3.6.1/docs/configuration.rst#lint and populating proper info to `meta/main.yml` linter shouldn't complain at all, like:

```
$ cat molecule/default/molecule.yml 
---
dependency:
  name: galaxy
driver:
  name: delegated
platforms:
  - name: instance
provisioner:
  name: ansible
verifier:
  name: ansible
lint: |
  set -e
  yamllint .
  ansible-lint
  flake8

$ molecule lint
INFO     default scenario test matrix: dependency, lint
INFO     Performing prerun...
[...]

PR #3680, opened 2022-10-12
Use FQCN at any point I found

muellerbe:
Fixes: #3481

I found some places without FQCN, here they are ;)


zhan9san:
@muellerbe 

Good job

PR #3681, opened 2022-10-13
Support ANSIBLE_COLLECTIONS_PATH environment var

nre-ableton:
This commit adds supports for reading ANSIBLE_COLLECTIONS_PATH from the environment, in a similar manner to how ANSIBLE_ROLES_PATH is handled.

—---

Fixes https://github.com/ansible-community/molecule/issues/3663

### DIFF
--- a/src/molecule/provisioner/ansible.py
+++ b/src/molecule/provisioner/ansible.py
@@ -465,6 +465,15 @@ class Ansible(base.Base):
             ]
         )
 
+        if os.environ.get("ANSIBLE_COLLECTIONS_PATH", ""):
+            collections_path_list.extend(
+                list(
+                    map(
+                        util.abs_path, os.environ["ANSIBLE_COLLECTIONS_PATH"].split(":")
+                    )
+                )
+            )
+
         roles_path_list = [
             util.abs_path(
                 os.path.join(self._config.scenario.config.runtime.cache_dir, "roles")


zhan9san:
Hi @nre-ableton 

Thanks for your work.

IMHO, even if you extend the collection paths, the molecule will also install dependencies into ~/.cache/ansible-compat/hash-vaule/, which is an isolated cache path defined in ansible-compat package.

You can find more details in  ansible-compat package.

Are you sure the test result is correct?

nre-ableton:
@zhan9san Yes, molecule behaves the same with roles. I wish that molecule had proper offline support, but that is a separate discussion (see https://github.com/ansible/galaxy/issues/2030).

For this to work, `prerun: false` must also be set in the scenario's `molecule.yml` file. With that in place, molecule will install collections to `ANSIBLE_COLLECTIONS_PATH`.

zhan9san:
I wonder if we can introduce an `offline` mode, no need to install dependencies when run `molecule`. But instead, we can install them in a cache path on CI server in advance, and let `molecule` respect this cache path.

```yaml
---
dependency:
  name: galaxy
  offline: true
  role-file: requirements.yml
  requirements-file: collections.yml
```

Please allow me to explain it in detail.

## Internet access

Currently, there are two kinds of dependencies need to be downloaded from internet when run `molecule dependency`.

1. role dependencies which are defined in `requirements.yml`
2. dependencies for testing the role in different scenarios, which are defined in `role-file: requirements.yml` and `requirements-file: collections.yml`, referring to [molecule.dependency.ansible_galaxy.AnsibleGalaxy](https://molecule.readthedocs.io/en/latest/configuration.html#molecule.dependency.ansible_galaxy.AnsibleGalaxy)

## Installation location

Assume the my role name is `my_role_2`.

For the 1st, the dependencies will be installed by `ansible-compat` in `/Users/jackzhang/.cache/ansible-compat/0afefd/roles` and `/Users/jackzhang/.cache/ansible-compat/0afefd/collections`.

For the 2nd, the dependencies for scenario test will be installed by `molecule` in `/Users/jackzhang/.cache/molecule/my_role_2/default/roles` and `/Users/jackzhang/.cache/molecule/my_role_2/default/collections`

Then, the installed location would be as below.
1. `/Users/jackzhang/.cache/ansible-compat/0afefd/roles`
2. `/Users/jackzhang/.cache/ansible-compat/0afefd/collections`
3. `/Users/jackzhang/.cache/molecule/my_role_2/default/roles`
4. `/Users/jackzhang/.cache/molecule/my_role_2/default/collections`
5. `/Users/jackzhang/.cache/molecule/my_role_2/scenario-2/roles`
6. `/Users/jackzhang/.cache/molecule/my_role_2/scenario-2/collections`
7. `/Users/jackzhang/.cache/molecule/my_role_2/scenario-3/roles`
8. `/Users/jackzhang/.cache/molecule/my_role_2/scenario-3/collections`
9. other scenaios, etc.

## Unified location

As we can see, there are many different locations. If we install them in one unified location, and ensure `molecule` respects it, this issue would be solved. 

Let me know what you think.

nre-ableton:
@zhan9san I think that this would be a great idea, and I wouldn't mind to implement something like this. Offline mode would be very useful for us. However, I also think that it is a bit beyond the scope of this PR. 😁 I'll go ahead and un-draft this PR, unless you have any strong objections.

zhan9san:
LGTM.
Go ahead please

nre-ableton:
@zhan9san Thanks for the merge!

PR #3706, opened 2022-10-26
Re-add credentials to Docker registry schema

TheLastProject:
This allows logging into a registry again like Molecule 4.0.1 allowed

I noticed that upgrading from molecule 4.0.1 to 4.0.2 at work broke our ability to use containers on a private Docker registry (in our case, a private GitLab registry).

The specific error message we started receiving was:
```
["Additional properties are not allowed ('credentials' was unexpected)"]
```

I came to the conclusion this was most likely broken by accident given:
1. The option does work (removing the option from our `molecule.yml` gave us permission errors)
2. There was nothing in the CHANGELOG about Docker registries
3. Comparing release 4.0.1 to 4.0.2 showed nothing related to Docker registries either: https://github.com/ansible-community/molecule/compare/v4.0.1...v4.0.2
4. The old schema on release 4.0.1 contained these options: https://github.com/ansible-community/molecule/blob/v4.0.1/src/molecule/model/schema_v3.py#L245-L258
5. This feature previously received fixes: https://github.com/ansible-community/molecule/issues/1331


zhan9san:
@TheLastProject 

Thanks for your contribution.
It looks good.

Would you mind taking a look at failed ci test?

It's not related to your change.

cs35-owncloud:
It seems to work with molecule 4.0.4 !

PR #3709, opened 2022-10-26
Update test dependencies

ssbarnea:


zhan9san:
@ssbarnea 

`dependabot` will update the `requirements.txt`.

Could we remove the change in `requirements.txt` and just keep the change in `pyproject.toml`?

ssbarnea:
@zhan9san No, they have different scopes. Our requirements.txt acts as a constraint file for testing, while the requirements are listed in pyproject.toml. Dependabot does not update the later, only the first.

We should update pyproject.toml requirements manually only when we know that a particular version is no longer compatible, basically bumping the minimum version required. I hope this explains it.

zhan9san:
I see. This is the same as what I say in https://github.com/ansible-community/molecule/issues/3702#issuecomment-1291453593.

I mean in this specific PR, we just need to add 'ignore xxx' in `pyproject.toml`, with which the CI would succeed to make the PR smaller.

We don't need to make any changes in `requirements.txt` in this specific PR.

And I don't want to remove `requirements.txt` in this project.

PR #3713, opened 2022-10-30
Help messages when config does not match state file

scottalreadytaken:
This is a proposal to add messages that help users understand when changes to the molecule.yml file might not take effect.

It expands on the ideas in #3429 among others. 

For example issues occurred with the below:
1. I start off using the podman driver. 
2. I switch to libvirt and attempt to converge but there is an error - By chance libvirt on my system is not setup correctly
3. I don't want to fix libvirt right now so I change back to using the podman
4. I attempt to converge again but I'm still receiving errors relating to libvirt. Why? The state file is holding the libvirt setting but as a new user I don't necessarily know this. It took some time to discover the cache files and remove them but now the molecule reset command is available this solution is easy to apply if you know about it. 
5. Since we can easily compare the settings in each location it's easy to warn about any discrepancy. 

In this case the warning is:
`Driver 'libvirt' is currently in use but the scenario config has changed and now defines 'podman'. To change drivers, run 'molecule destroy' for converged scenarios or 'molecule reset' otherwise.`

I've also adopted the idea from another discussion thread that a warning could be displayed if the date modified on the molecule.yml has changed since the state file was first written. 
The generic warning is:
`The scenario config file ('/home/example/molecule.yml') has been modified since the scenario was created. If recent changes are important, reset the scenario with 'molecule destroy' to clean up created items or 'molecule reset' to clear current configuration.`

The driver comparison is probably in the right place because it's right after the check added for #3429 but I'm unsure if the check for the molecule.yml is the best check or in the best location. I'd be able to change it to whatever consensus suggests.


ssbarnea:
@scottalreadytaken Any reason why this change is still marked as draft? It looks ok to me.

ssbarnea (approved):

scottalreadytaken:
It's ready to go, I'm happy to incorporate any feedback.

PR #3724, opened 2022-11-03
Add "default" as a valid docker "network_mode" in molecule.json schema

fredflev:
As per discussion in: https://github.com/ansible-community/molecule/discussions/3723

Without the change in this PR, the molecule schema validation step fails and throw this "not so obvious" error: 

```console
molecule --debug syntax                                        
DEBUG    Validating schema /Users/My_User/ansible-role/molecule/default/molecule.yml.
CRITICAL Failed to validate /Users/My_User/ansible-role/molecule/default/molecule.yml

["'default' is not valid under any of the given schemas"]
```

And reference in: https://docs.ansible.com/ansible/latest/collections/community/docker/docker_container_module.html#parameter-network_mode

### DIFF
--- a/src/molecule/data/molecule.json
+++ b/src/molecule/data/molecule.json
@@ -225,7 +225,7 @@
         "network_mode": {
           "anyOf": [
             {
-              "enum": ["bridge", "host", "none"],
+              "enum": ["bridge", "default", "host", "none"],
               "type": "string"
             },
             {


src/molecule/data/molecule.json
```diff
@@ -225,7 +225,7 @@
         "network_mode": {
           "anyOf": [
             {
-              "enum": ["bridge", "host", "none"],
+              "enum": ["bridge", "default", "host", "none"],
```
felixfontein, line 228: If this is directly passed to docker, then this can be an arbitrary string (name of a docker network) and shouldn't be limited to a specific selection of networks that might or might not exist depending which networks the user created or deleted.

fredflev: I could be wrong but this is the mode and not the name of the network no?

fredflev: You are right so basically this should be used instead: 

```JSON
        "network_mode": {
          "title": "Network Mode",
          "type": "string"
        },
```

fredflev: I've raised https://github.com/ansible-community/molecule/pull/3725 as a fix.

PR #3725, opened 2022-11-03
Accept any string for "network_mode" in molecule.json schema

fredflev:
As mentioned by @felixfontein in https://github.com/ansible-community/molecule/pull/3724 `network_mode` can be any network name specified by a user and should not be limited to a specific set of strings or string patterns.

### DIFF
--- a/src/molecule/data/molecule.json
+++ b/src/molecule/data/molecule.json
@@ -223,8 +223,16 @@
           "type": "string"
         },
         "network_mode": {
-          "title": "Network Mode",
-          "type": "string"
+          "anyOf": [
+            {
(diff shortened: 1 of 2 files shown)


fredflev:
> * Remove regex checks, but not enun
> * Add another plain string option, after enum
> * Add a test inside https://github.com/ansible-community/molecule/tree/main/tools/test-schema that would check that "foo" string is allowed, so we do not endup with regressions in the future.
> 
> Enum should stay because it is the trick used to allow auto-complete to work on that field for most common values.

Can we assume that having `network_mode: service:vpn` in one of the existing test covers the `network_mode: foo` test case?
Cf:

ssbarnea (approved):

PR #3738, opened 2022-11-15
Add ansible-navigator support

anxjok:
Hey, I added ansible-navigator support for molecule. 
The ansible-navigator provisioner is a Derived class from the ansible provisioner class. 
To get the stdout output from ansible-navigator I parse the json artifacts from ansible navigator.

With the implementation of ansible-navigator, the roles or playbooks can be tested against a modified version of an awx-ee image, to ensure it works with the awx-ee image that is used in AAP or AWX.

I created a sample  [repository](https://github.com/anxjok/creator-ee) to modify the creator-ee to support docker from the host inside the ee.


src/molecule/config.py
```diff
@@ -221,6 +221,8 @@ def provisioner(self):
         provisioner_name = self.config["provisioner"]["name"]
         if provisioner_name == "ansible":
             return ansible.Ansible(self)
+        elif provisioner_name == "ansible-navigator":
+            return ansible_navigator.Ansible_Navigator(self)
```
zhan9san, line 225:
```suggestion
            return ansible_navigator.AnsibleNavigator(self)
```

zhan9san (changes requested):

src/molecule/config.py
anxjok: I applied these challenges in the last commit

bbaassssiiee:
branch has conflicts that must be resolved

ssbarnea:
I will close this because at this moment we don't want to use navigator from within molecule. Still, we will add integration but without making molecule aware about being run under navigator.

PR #3750, opened 2022-11-24
Add some FQCNs

muellerbe:
Fixes: #3481

I found some more places without FQCN


zhan9san:
@muellerbe 

Thanks for your work

Issue #3758, opened 2022-11-29
"Cannot autolaunch D-Bus without X11 $DISPLAY" while running tox test on dockerfile


zeitounator:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version
ansible [core 2.14.0]
  config file = None
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/Sources/molecule/.venv/lib/python3.9/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/user/Sources/molecule/.venv/bin/ansible
  python version = 3.9.5 (default, Nov 23 2021, 15:27:38) [GCC 9.3.0] (/home/user/Sources/molecule/.venv/bin/python)
  jinja version = 3.1.2
  libyaml = True
INFO     Found config file /home/user/Sources/molecule/.config/molecule/config.yml
molecule 4.0.4.dev28 using python 3.9 
    ansible:2.14.0
    delegated:4.0.4.dev28 from molecule
    docker:2.1.0 from molecule_docker requiring collections: community.docker>=3.0.2 ansible.posix>=1.4.0
```

Molecule installation method (one of):

- pip (editable inside a venv)

Ansible installation method (one of):

- pip

Detail any linters or test runners used:
tox as configured in the molecule source tree

# Desired Behavior

Running `tox` should run all tests and succeed (as long as current branch is ok). This includes dockerfile tests

# Actual Behaviour

All of the below happens by simply running `tox`

I have the following configuration for my docker client (i.e. only relevant part) in `~/.docker/config.json`
```json
{
        "auths": {},
        "credsStore": "secretservice"
}
```
This instruct docker to store the credentials in my local linux keystore whenever I issue a `docker login` rather that in clear text inside the json file.

For a reason I did not yet understand and although none of the relevant builds require credentials for a any docker registry while running tox dockerfile tests, all image builds fail with the following error message (abridged) regardless of the origin of their base image.
> TASK [Test Dockerfile template] ***********************************************************************<br/>
> An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible_collections.community.docker.plugins.module_utils._api.credentials.errors.StoreError: Credentials store docker-credential-secretservice exited with "Cannot autolaunch D-Bus without X11 $DISPLAY".

[...]

PR #3764, opened 2022-12-02
Fix running tox:dockerfile with docker secretservice enabled

zeitounator:
This PR fixes issue #3758 which was later migrated to discussion #3759.

As a short summary of the posts above.

The problem happens when you have configured `secretservice` as your `credstore` in your docker configfile (e.g. `~/.docker.config.json`)

running `tox -e dockerfile` launches the playbook `src/molecule/data/validate-dockerfile.yml` which then fails at image build taks with error 
> ansible_collections.community.docker.plugins.module_utils._api.credentials.errors.StoreError: Credentials store docker-credential-secretservice exited with "Cannot autolaunch D-Bus without X11 $DISPLAY"

Passing the `DISPLAY` environment variable to tox fixes the issue.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -26,6 +26,7 @@ skip_install = True
 passenv =
     CI
     CONTAINER_*
+    DISPLAY
     DOCKER_*
     GITHUB_*
     HOME


zhan9san:
Thanks for your contribution

PR #3770, opened 2022-12-05
Remove support for python3.8

ssbarnea:


ganeshrn (approved):

dalepotter:
Out of interest, what is the thinking behind this, given py37 is EOL in 2023-06-27, and py38 over a year later (2024-10)?

ssbarnea:
@dalepotter That is bringing molecule in sync with other projects such ansible-core and ansible-lint, which already require py39 or newer. Main reason for this is to ease maintenance, so we can focus on fixing bugs than covering older versions of python. Each extra versions we support adds considerable costs, human time in particular.

PS. When me mention supported, we mean fully-supported. A version that only gets security updates does not count as supported from our point of view. Think that you face a serious bug, nobody will help you merge a fix for it if that is not related to security for py37. For this reason alone, we cannot really say that we provide more support than whatever is provided by the upstream dependency (python in this case).

I also observe that you never contributed to molecule project, maybe if you can spend some time and become an active contributor, we might be able to support older pythons for longer.

While we do not use maintenance branches with molecule, that might change in the future, which will allow people to backport important fixed to older versions. Still, that need active community contributors in order to happen.

dalepotter:
@ssbarnea thanks for clarifying the thinking - I certainly agree that it would not make sense to maintain support for versions which themselves are not supported by `ansible-core`.

PR #3778, opened 2022-12-09
Reorder heading and reformat badges

mamercad:
Place the heading first ([MD041](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md041---first-line-in-a-file-should-be-a-top-level-heading)) and put the badges on a single line (so there's not such a large empty portion at the top of the `README.md`).

image

(Note the large amount of empty space)


briantist (approved):
I'm not a maintainer here but it looks good to me

PR #3802, opened 2023-01-05
Remove lint command

ssbarnea:


zhan9san:
It seems lint dependencies is missed.

https://github.com/ansible-community/molecule/blob/main/pyproject.toml#L68-L74

ssbarnea:
@zhan9san there are few missing bits as I just started but "lint" extra is not to be removed as its purpose is for internal use. What is funny is that even if we remove it, people will not observe because pip ignores any missing extras.

gardar:
Wouldn't it be a good idea to put up a deprecation warning for a few versions before removing the option entirely?

davedittrich:
Did this really need to be removed? Why can't a general `lint` phase remain, as it was, running a shell command? That command can be `tox` and any dependencies can be handled by the user. What is wrong with that?  That has worked fine for years. 

By removing the ability for `molecule` to perform linting each time tests are run, you require the user to add a `Makefile` to provide a dependent step (of linting).

gardar:
Agreed!
Rather than removing it it would have been better if the lint step was moved into a playbook, so that it could be user defined like converge.yml , prepare.yml,  cleanup.yml, etc.

I guess a possible workaround now, to be able to keep the linting within molecule and not relay on any external scripts, would be to add the linting into one of those playbooks.

dnorthup-ums:
@gardar @davedittrich My general experience is that once Sorin has put something into place that's the end of it (even if most of us on the outside are surprised), so we'll just have to figure out how to live with this change that breaks >95% of CI / CD workflows in the real world. In the end, not a big deal...just an unexpected major annoyance.

bbaassssiiee:
Sorin's changes have a process? Or is it just a destructive urge?

dnorthup-ums:
@madrover At $dayjob we'll likely be implementing an additional linting stage to the CI/CD flow that gets run on every push / merge request / and merge operation on our ansible repositories. Many places use git-hooks, both pre- and post-commit to enforce linting (but please read the manual before doing that, for the sake of your sanity). Additionally, many editors these days support on-the-fly linting via plugins (EG: VSCode and the dearly departed Atom). There are other viable options, this change was just rather unexpected outside of the core maintainer group.

ziegenberg:
For a short explanation of why lint was removed see here: https://github.com/ansible-community/molecule/discussions/3825#discussioncomment-4908366

madrover:
Thanks for the insights. I understand that this can be easily overcome on a CI/CD workflow and/or using other techniques, although it's annoying to have to tweak all the existing pipelines, but we also use Molecule intensively as a local development tool and having the linting capability built in was very nice.

I can understand the rationale described at https://github.com/ansible-community/molecule/discussions/3825#discussioncomment-4908366 but removing all the functionality altogether is probably not the most thoughtful approach for the users. IMHO, this functionality could have been disabled by default, thus solving the described issues, but left there for users that rely on it.

In a nutshell, while I don't disagree with the rationale of not running linting, I think that making it toggleable and disabled by default would have achieved the same result and would have been less onerous for users.

davedittrich:
> @gardar @davedittrich My general experience is that once Sorin has put something into place that's the end of it (even if most of us on the outside are surprised), so we'll just have to figure out how to live with this change that breaks >95% of CI / CD workflows in the real world. In the end, not a big deal...just an unexpected major annoyance.

From what I have seen in digging through code, I believe the linter worked perfectly fine until an attempt was made to produce fancy colored output, which didn't do encoding properly (hence the "c h a r a c t e r s  i n  t h e  o u t p u t" looked funny). Multiple attempts to fix this, or detect `stdin` being a `tty` or not, did not work because they weren't fully thought through, and attempts by outsiders to issue PRs that made the bugs clear were rejected. So instead of backing out all the buggy changes, the "solution" seems to be to remove the functionality entirely. That makes a developers life great, but results in a terrible user experience. Sure I can add a `Makefile` to perform the linting apart from `molecule` (which is designed to make such checks easy, ironically) and it will work, but not something everyone using the system will want to learn.

I also noticed that this discussion involved `pip` dependencies that were/were not present in various use cases. Switching to `poetry` would improve the situation, or using additional requirements files instead of just one. 

In summary, the pace of changes in this (and related) projects seems to be too high, since a non-trivial number of those changes produce regression problems, give the user no time to deal with deprecations, or introduce latent bugs that add to maintenance costs later on. I'd prefer fewer changes, more carefully designed and following a test-driven development methodology that exposes regression problems earlier, and working more closely with people contributing PRs or capable of multi-repo debugging.

fkuep:
@ssbarnea 

> _ No description provided. _

We are asking frequently.
At least put it in the faq https://ansible.readthedocs.io/projects/molecule/faq/ .

jimboolio:
I spent hours trying to debug my molecule installation before I landed here. There are still plenty of resources which hint that such feature exists and when searching for the official docs, it returns some results, from the count of which you may guess linting is not there anymore, but I feel like it would be very good to have something - anything confirming that this doesn't exist. This issue is already quite buried in github so I think finding this information out is difficult.

Just to provide context, I was quickly able to find some examples of configuring molecule linting from web search so it is totally reasonable someone else has experienced this frustration:
https://opensource.com/article/18/12/testing-ansible-roles-molecule
https://www.jeffgeerling.com/blog/2018/testing-your-ansible-roles-molecule

Also, on molecule version 6, if one goes and adds the `lint` step to their test sequence, molecule does not in any form indicate that this feature is removed. It crashes in such situations. Also, when running `molecule lint` it doesn't indicate that the command is deprecated.

I'm just getting started with molecule, while I do not like linting not being there anymore, I do understand the reasoning here. Additionally I feel like this thread was at least somewhat informational resource for finding a workaround, so it would provide value as linked to the docs if some mention is at some point added. Otherwise we have to pray for favorable SEO.

serpro69:
Molecule docs could really be improved to indicate that this has been deprecated. Searched the web for an hour or so before I landed here...
I guess one of the workarounds at the moment is to include linting via `dependency` with `shell` type. Ugly, but works (kind of). If only there was a way to configure retries... :facepalm:

PR #3814, opened 2023-01-17
Fix ansible-lint install command

koleo:
The documented command for installing `ansible-lint` was broken: the quotes consider `"molecule ansible-lint"` as a single package name. Anyway, we expect only `ansible-lint` in this document section.


ssbarnea:
Closing as it needs rebase after we switched to mkdocs. Feel free to reopen.

PR #3823, opened 2023-02-07
Remove use of pytest-html

ssbarnea:
Due to current maintenance status of pytest-html we stop using it.

Related: https://github.com/pytest-dev/pytest-html/pull/575
Related: https://github.com/pytest-dev/pytest-html/issues/570


cidrblock (approved):
html is over rated :)

PR #3882, opened 2023-04-20
Add `implicit-str-concat` & `inconsistent-return-statements` check

Ruchip16:
The `implicit-str-concat` rule ensures that the string concatenation is done implicitly in an assignment statement using an explicit operator, such as `+` or other methods like string-formatting, etc. This can improve code readability and reduce the likelihood of introducing errors related to string concatenation.

The `inconsistent-return-statements` rule raises an error when a function has multiple return statements, and some of them return an expression while others do not. This can lead to unexpected behavior and make the code harder to understand. The check ensures that the function has consistent return statements and that it always returns an expression or none of them return an expression, making the code more predictable and easier to understand.


src/molecule/provisioner/ansible.py
```diff
@@ -947,7 +947,7 @@ def _verify_inventory(self):
         :return: None
         """
         if not self.inventory:
-            msg = "Instances missing from the 'platform' " "section of molecule.yml."
```
ajinkyau, line 950: @Ruchip16 why do we need a concatenation here?
It can be a single string right?
Am I missing something?

src/molecule/provisioner/ansible.py
```diff
@@ -947,7 +947,7 @@ def _verify_inventory(self):
...
+            msg = "Instances missing from the 'platform' " + "section of molecule.yml."
```
ajinkyau, line 950: @Ruchip16  why do we need a concatenation here?
It can be a single string right?
Am I missing something?

Ruchip16: yeaa i thought they are two separate strings and hence I did not combine hem into one @ssbarnea shall we combine them into just one string?

PR #3885, opened 2023-04-21
Add `raise-missing-from` rule

Ruchip16:
The `raise-missing-from` error occurs when we catch an exception and raise a new exception with the original exception as the cause, but we do not explicitly use the `from` clause to include the original exception as the cause in the new exception. Hence to fix this, we can use the `from` clause when raising the new exception to explicitly specify the original exception as the cause. This would make the work easier to trace the original cause of the error when debugging.


ssbarnea (changes requested):
You included this file my mistake, remove it.

Ruchip16:
> You included this file my mistake, remove it.

ohh no yes removing it

ssbarnea (approved):

PR #3888, opened 2023-04-21
Add `no-value-for-parameter` check

Ruchip16:
The `no-value-for-parameter` check ensures that no function is called with missing or incomplete arguments.

### DIFF
--- a/src/molecule/__main__.py
+++ b/src/molecule/__main__.py
@@ -23,9 +23,19 @@ import click
 from molecule.shell import main
 
 
-def main(ctx):
-    if __name__ == "__main__":
-        # Creating a click context object
-        ctx = click.Context()
-        # Providing values for the required arguments when calling main() function
-        main(ctx=ctx, debug=True, verbose=False, base_config=None, env_file=None)
+def func(ctx: click.Context) -> None:
+    # Creating a click context object
+    click_ctx = click.Context(command=None)
+    # Providing values for the required arguments when calling main() function
+    main(
+        command=None,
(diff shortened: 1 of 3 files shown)


src/molecule/__main__.py
```diff
@@ -18,8 +18,14 @@
...
 """Molecule CLI main entry point."""
+import click
 
 from molecule.shell import main
 
-if __name__ == "__main__":
-    main()
+
+def main(ctx):
+    if __name__ == "__main__":
```
Ruchip16, line 27: @ssbarnea is this the correct way? coz its still giving some errors: Unexpected keyword argument `'debug'`, `verbose`, `base-config`, `env-file` in function call (unexpected-keyword-arg) & No value for argument 'command' in constructor call, when these are already declared

ssbarnea: Nope, this change makes no sense, `if __name__ == "__main__"` condition is never part of a function.

src/molecule/__main__.py
```diff
@@ -18,8 +18,24 @@
...
 #  DEALINGS IN THE SOFTWARE.
 """Molecule CLI main entry point."""
+import click
 
 from molecule.shell import main
 
+
+def func(ctx: click.Context) -> None:
+    # Creating a click context object
+    click_ctx = click.Context(command=None)
```
Ruchip16, line 28: @ssbarnea here what specific command should I pass, coz its giving me the error that argument command has incompatible type:none

PR #3898, opened 2023-04-26
Use api to get driver schema file

apatard:
Current code is assuming that the molecule driver are inside a molecule_$driver python resource, which is not the case. For instance, vagrant plugin is in molecule_plugins.

As a solution:
- extend driver class to provide a schema_file() member, allowing to return a path to the driver.json file,
- replace schema_v3 code to use molecule api and get the schema file path thanks to the new schema_file().

### DIFF
--- a/src/molecule/driver/base.py
+++ b/src/molecule/driver/base.py
@@ -271,6 +271,9 @@ class Driver:
             return p
         return None
 
+    def schema_file(self):
+        return None
+
     def modules_dir(self):
         """Return path to ansible modules included with driver."""
         p = os.path.join(self._path, "modules")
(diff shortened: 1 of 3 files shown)


apatard:
Opening the PR as draft, since I'd like testing and comments

ssbarnea (approved):

avanderm:
Started getting failures today with molecule-podman (molecule versoin 5.0.1):
```
FileNotFoundError: [Errno 2] No such file or directory: '/opt/hostedtoolcache/Python/3.11.3/x64/lib/python3.11/site-packages/molecule_podman/driver.json'
```

PR #3901, opened 2023-04-28
src/molecule/config.py: Ensure external drivers are installed

apatard:
If someone try installing plugins (say docker) with molecule[docker], nothing will get installed as it should be molecule-plugins or molecule-plugins[docker].
This leads to a python backtrace similar to the one found in https://github.com/ansible-community/molecule/issues/3883. This should be handled properly, so check that the driver is in api.drivers() before trying to get the value.


src/molecule/config.py
```diff
@@ -182,7 +182,12 @@ def driver(self):
         driver_name = self._get_driver_name()
         driver = None
 
-        driver = api.drivers(config=self)[driver_name]
+        api_drivers = api.drivers(config=self)
+        if driver_name not in api_drivers:
+            msg = f"Failed to find driver {driver_name}. Maybe missing 'molecule-plugins' ?\n"
```
ssbarnea, line 187: This might prove not enough info either because installing molecule-plugins does not install each driver own dependencies. We would need to do `pip install molecule-plugins[driver-name]` for the known drivers to ensure its installation with requirements.

How about making a list of known drivers inside molecule-plugins and display the exact line when the driver is missing?

PS. Also remove the space before question mark and the extra newlines.

apatard: My first message was about using ``molecule-plugins[driver-name]`` but the output was broken. I didn't look at it closely so removed it. Moreover, I'm not sure how good this message is. The majority of users is likely to use molecule-plugins but they may be using a driver not coming from there (hence the interrogative form).

ssbarnea (approved):

Issue #3851, opened 2023-03-18
wrong documentation for collections dependencies?


trickert76:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version
ansible [core 2.14.1]
  config file = /Users/trw/.ansible.cfg
  configured module search path = ['/Users/trw/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/trw/Library/Python/3.9/lib/python/site-packages/ansible
  ansible collection location = /Users/trw/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/trw/Library/Python/3.9/bin/ansible
  python version = 3.9.6 (default, Oct 18 2022, 12:41:40) [Clang 14.0.0 (clang-1400.0.29.202)] (/Library/Developer/CommandLineTools/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
molecule 4.0.4 using python 3.9 
    ansible:2.14.1
    delegated:4.0.4 from molecule
    docker:2.1.0 from molecule_docker requiring collections: community.docker>=3.0.2 ansible.posix>=1.4.0
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

# Desired Behavior

Molecule documentation should mention, that the collections.yml is parsed during dependency management.
And the output of galaxy dependency should change to

```
INFO     Running default > dependency
WARNING  Skipping, missing the requirements file.
WARNING  Skipping, missing the collections file.
```

# Actual Behaviour

I'd like add a collection to my molecule test, because there are filters in the collection, I need to use. See https://stackoverflow.com/questions/75765614/ansible-molecule-ignores-collection-requirements/75775152#75775152 for description. The role has no other dependencies then this collection.

I've added the collection as described here - https://molecule.readthedocs.io/en/latest/configuration/#dependency
According to that, I cannot use requirements.yml, because it must contains role only.

```
# molecule.yml
---
dependency:
  name: galaxy
  options:
    requirements-file: collections.yml
...
```

```
# collections.yml
---
collections:
  - name: https://github.com/acoby/ansible-collection.git
    type: git
    version: main
```

But when I run molecule test, I'm getting twice the warning and the test failed because of the missing filters from the collection.

```
INFO     Running default > dependency
WARNING  Skipping, missing the requirements file.
WARNING  Skipping, missing the requirements file.
```

[...]

PR #3916, opened 2023-05-16
Update dependency doc

zhan9san:
Related #3851

### DIFF
--- a/src/molecule/dependency/ansible_galaxy/__init__.py
+++ b/src/molecule/dependency/ansible_galaxy/__init__.py
@@ -12,7 +12,7 @@ class AnsibleGalaxy(Base):
     options dict.  Any option set in this section will override the defaults.
 
     The `role-file` and `requirements-file` search path is `<role-name>`
-    directory. The default value for `role-file` is `requirements.yml` and the
+    directory. The default value for `role-file` is `requirements.yml`, and the
     default value for `requirements-file` is `collections.yml`.
 
     1. If they are not defined in `options`, `molecule` will find them from
--- a/src/molecule/dependency/ansible_galaxy/__init__.py
+++ b/src/molecule/dependency/ansible_galaxy/__init__.py
@@ -11,6 +11,17 @@ class AnsibleGalaxy(Base):
     Additional options can be passed to ``ansible-galaxy install`` through the
     options dict.  Any option set in this section will override the defaults.
 
+    The `role-file` and `requirements-file` search path is `<role-name>`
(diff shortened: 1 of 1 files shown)


src/molecule/dependency/ansible_galaxy/__init__.py
```diff
@@ -11,6 +11,17 @@ class AnsibleGalaxy(Base):
     Additional options can be passed to ``ansible-galaxy install`` through the
     options dict.  Any option set in this section will override the defaults.
 
+    The `role-file` and `requirements-file` search path is `<role-name>`
+    directory. The default value for `role-file` is `requirements.yml` and the
```
ziegenberg, line 15:
```suggestion
    directory. The default value for `role-file` is `requirements.yml`, and the
```

src/molecule/dependency/ansible_galaxy/__init__.py
```diff
@@ -11,6 +11,17 @@ class AnsibleGalaxy(Base):
...
+    default value for `requirements-file` is `collections.yml`.
+
+    1. If they are not defined in `options`, `molecule` will find them from
```
ziegenberg, line 18:
```suggestion
    1. If they are not defined in `options`, `molecule` will find them from the
```

ziegenberg (commented):
Just a small change in wording and punctuation.

ziegenberg (approved):

src/molecule/dependency/ansible_galaxy/__init__.py
```diff
@@ -11,6 +11,17 @@ class AnsibleGalaxy(Base):
...
     options dict.  Any option set in this section will override the defaults.
 
+    The `role-file` and `requirements-file` search path is `<role-name>`
+    directory. The default value for `role-file` is `requirements.yml`, and the
+    default value for `requirements-file` is `collections.yml`.
+
+    1. If they are not defined in `options`, `molecule` will find them from the
+    `<role-name>` directory, e.g. `<role-name>/requirements.yml` and
+    `<role-name>/collections.yml`
+    2. If they are defined in `options`, `molecule` will find them from
```
zhan9san, line 21: @ziegenberg 

Thanks for your review.

Do we need add one `the` word here like above?

ziegenberg: I don't think so. In the above example, `directory` demanded a `the` in front. I also checked with Grammarly, and it's happy about it. :)

PR #3924, opened 2023-05-25
fix tests for commands `molecule init role` and `molecule init scenario`

ziegenberg:
Currently, the tests fail for `test_command_init_role` and `test_command_init_scneario`. The tests execute `ansible-lint` on the newly-created roles and the linter recently got more restrictive, so we now have to adapt.

There are still two failures with Python 3.11 and ansible-core devel. `ansible-compat` calls `ansible-galaxy collection list --format=json` internally (https://github.com/ansible/ansible-compat/blob/d30965cd023b4116f6e959ace0e4dbf219677952/src/ansible_compat/runtime.py#LL235C20-L235C20) and due to some unknown reason it is executed with elevated verbosity producing broken json output. The output begins with:
```
Using molecule/.ansible.cfg as config file
{"/home/daniel/.ansible/collections/ansible_collections": {"ansible.posix": {"version": "1.5.1"},

.
.
.
```

### DIFF
--- a/src/molecule/command/init/role.py
+++ b/src/molecule/command/init/role.py
@@ -90,7 +90,7 @@ class Role(base.Base):
 
         if namespace:
             # we need to inject namespace info into meta/main.yml
-            cmd = [
+            cmd_meta = [
                 "ansible",
                 "localhost",
                 "-o",  # one line output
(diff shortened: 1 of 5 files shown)


ziegenberg:
And @ssbarnea fixed the remaining issue in https://github.com/ansible/ansible-compat/pull/284 :)

ziegenberg:
This change is now included in PR #3904, so closing this one.

PR #3959, opened 2023-07-08
Reimplemented templating (init command)

ssbarnea:
- Replaced use of cookiecutter with ansible templates
- Removed 'init role' command  due to already being covered by ansible-galaxy
- Removed support for custom templating from drivers


theko2fi:
`Removed support for custom templating from drivers`....but WHY? What led to this decision? I would like to read about it if you guys documented it somewhere. Thanks

PR #3962, opened 2023-07-10
Disable isolation from ansible-compat Runtime

ssbarnea:
From now on, molecule will not rely on magic alteration of ansible
configuration for role paths, collections or plugins.


zhan9san:
If we disable isolation, `molecule` may overwrite the existing collection/role.

Here is a scenario.

The version1 of role1 have been installed in `~/.ansible/roles` for global use.

The version2 of role1 is defined in molecule's `role-file`. After `molecule` runs, the version1 would be overwritten.

PR #3966, opened 2023-07-12
Change current directory to scenario before running playbooks

ssbarnea:
From now on, instead of using the current working directory when
running as scenario, molecule will chdir to scenario folder.

This change will enable people to use scenario specific ansible.cfg
files. Follow-up change will enable running molecule scenarios
directly from within them. When user is inside a scenario, molecule
will recognize it and no longer require the --scenario argument.


cidrblock (approved):

zhan9san:
Would it be possible to add the related document as well?

ajinkyau (approved):

ajinkyau:
> Would it be possible to add the related document as well?

Just saw your msg, will be adding it in a separate PR

PR #3973, opened 2023-07-25
Revert "Remove testinfra verifier"

ajinkyau:
Reverts ansible-community/molecule#3964


33Fraise33:
Hi, is testinfra going to stay? Why the removal and afterwards the revert, anywhere I can read more about it?

l00ptr:
Hi,

the documentation about molecule says testinfra will be removed on molecule 7.  Do we have any roadmap about that change (remove) ? @33Fraise33 I guess that's why this commit was reverted, but maybe @ajinkyau can confirm that ?

Best regards,

PR #3975, opened 2023-07-25
Update documentation

ajinkyau:


ziegenberg:
So the removal of the testinfra driver is postponed?

ajinkyau:
Yes, as it would be a big breaking change, wanted the users to have some more time.
Maybe make `testinfra` as an optional dependency in v6 release and remove it in the next one.

ziegenberg:
So, if it's optional in `v6` and removed in the next release, we should probably put this in the docs?

PR #3980, opened 2023-07-31
Dependency installation directory is changed

zhan9san:
After #3962, the dependency installation directory is changed


cidrblock (approved):
Thanks for this

PR #3981, opened 2023-07-31
Remove constraints requirement files

zhan9san:
Could you kindly explain which value does these files(`requirements.txt` and `requirements-lock.txt`) bring?

## Without constraints

If someone report an issue related to a specific version of dependency and if this issue is fixed in a higher version, we can update the lower bound of the dependency in `.config/requirements.in`.

---

`Molecule`'s dependencies are defined in `.config/requirements.in`. IMHO, it is enough.

> A set of command line tools to help you keep your pip-based packages fresh, even when you've pinned them. You do pin them, right? (In building your Python application and its dependencies for production, you want to make sure that your builds are predictable and deterministic.)

> Capping dependencies has long term negative effects, especially for libraries, and should never be taken lightly. A library is not installed in isolation; it has to live with other libraries in a shared environment. Only add a cap if a dependency is known to be incompatible or there is a high (>75%) chance of it being incompatible in its next release. 

Reference

https://iscinumpy.dev/post/bound-version-constraints/#tldr
https://pypi.org/project/pip-tools/


ajinkyau:
@cidrblock any thoughts on this?

cidrblock (approved):

PR #3982, opened 2023-07-31
Make pytest-testinfra as an optional dependency

ajinkyau:
This PR makes `pytest-testinfra` as optional dependency for the upcoming v6 release.
If the user, wishes to include `pytest-testinfra` as a depedency, then use:
`pip install molecule[testinfra]`

testinfra will be removed in the v7 release.


.config/requirements-testinfra.txt
```diff
@@ -0,0 +1 @@
+pytest-testinfra==8.1.0
```
zhan9san, line 1: Could you please change `==` to `>=`?
```suggestion
pytest-testinfra>=8.1.0
```

Why `pytest-testinfra==8.1.0` is used in `requirements.txt`?
It's not an optional dependency. If we take a deep look at `requirements.txt`, it cannot be installed by any `pip install molecule[xxx]`.

ajinkyau: So, I removed it from `requirements.txt` and made it as an optional dependency which will allow it to be installed using `pip install molecule[testinfra]`

zhan9san:
@ajinkyau 

The lint job is not happy with this change.

Could you kindly review #3981 ? It may fix the lint issue.

PR #3985, opened 2023-08-03
Update documentation

ajinkyau:
Update documentation for molecule.


cidrblock:
Can we drop the use of docker here and maybe just have the getting started run against localhost?

This might necessitate the need for a follw on docs about "standing up infrastructure for a molecule scnearion" guide but it might be easier to demonstate the capabilities of molecule agains localhost

cidrblock (approved):
This is great.

There's a couple other things wI think that can be improved in the docs but let's get this good stuff in and build on it.

docs/getting-started.md
```diff
@@ -97,75 +85,98 @@ components that Molecule provides. These are:
...
-```bash
-$ docker run hello-world
-```
+One of the recommended ways to create a collection is to place it
+under the `collections/ansible_collections` directory.
 
-Now, we can tell Molecule to create an instance with:
+To create a collection, use the `ansible-galaxy` tool.
+`ansible-galaxy collection init <namespace>.<collection_name>`
+Example: `ansible-galaxy collection init foo.bar`
```
samccann, line 99: Consider an alternate way to write these sections. Each of these are steps in this procedure of creating a new collection and testing it with molecule.  So for this one, it could be written as:
1. Create a collection.
```
    `ansible-galaxy collection init foo.bar`
```

samccann: Also notice -this Getting Started is focused on molecule, so we can skip some of the background details about ..well anything that isn't molecule :-) like the naming of collections in this instance.

docs/getting-started.md
```diff
@@ -97,75 +85,98 @@ components that Molecule provides. These are:
...
-We can verify that Molecule has created the instance and they're up and
-running with:
+## Create a role inside the collection
 
-```bash
-$ molecule list
-```
+To initialize a role inside a collection we can again use `ansible-galaxy` tool under the `roles` directory.
+`ansible-galaxy role init <namespace>.<collection_name>`
+Example: `ansible-galaxy collection init my_role`
```
samccann, line 110: If we were to continue converting this into a series of steps in a procedure, this section would lose it title, and just become:

2. cd to the `roles` directory in your new collection.
3. Initialize a new role for this collection.
```
    `ansible-galaxy collection init my_role`
```
4. Add a task  under `my_role_tasks/main.yml`.

(and then keep the task as you have it below)

docs/getting-started.md
```diff
@@ -97,75 +85,98 @@ components that Molecule provides. These are:
...
 
-We can then tell Molecule to test our role against our instance with:
+## Create a playbook inside the collection
 
-```bash
-$ molecule converge
-```
+Under the collection, we can create a directory named `playbooks`
+and create a new playbook file under it.
+Example: `my_playbook.yml`
```
samccann, line 125: 5. Create a directory named `playbooks`.
6. Create a new playbook in this directory called `my_playbook.yml`
(and then keep the playbook you have below here)

docs/getting-started.md
```diff
@@ -97,75 +85,98 @@ components that Molecule provides. These are:
...
 
-We now have a free hand to experiment with the instance state.
+## Adding Molecule to the Collection
 
-Finally, we can exit the instance and destroy it with:
+Inside the collection, create a directory named `extensions`
+and initialize `molecule scenario` under this directory using
+the command: `molecule init scenario <scenario_name>`
+`scenario_name` is optional, if not provided, will create
+a `default` scenario.
```
samccann, line 145: I'd keep this as a section like it is, but convert the steps here to.. steps.  Something like:
1. Create a new directory in your collection called `extensions`.
2. Initialize a new default molecule scenario in this directory.
```
    `molecule init scenario`
```

docs/getting-started.md
```diff
@@ -97,75 +85,98 @@ components that Molecule provides. These are:
...
 
-```bash
-$ molecule destroy
-```
+Also, create an ansible configuration `ansible.cfg` file under the `extensions` directory.
+Add the `collections_path` to this configuration file.
```
samccann, line 149: 3. Create an ansible configuration...
4. Add the `collections_path`...

docs/getting-started.md
```diff
@@ -97,75 +85,98 @@ components that Molecule provides. These are:
...
+    ├── syntax
+    ├── create
+    ├── prepare
+    ├── converge
+    ├── idempotence
+    ├── side_effect
+    ├── verify
+    ├── cleanup
+    └── destroy
 ```
```
samccann, line 180: I find it a bit confusing between this section (running all tests) and the section below (running just playbook tests). Consider if it's better to just run the playbook tests first and then show the command to run everything? I'm guessing the everything test wouldn't run in this example because we've only had the reader modify converge.yml?

docs/getting-started.md
```diff
@@ -180,3 +191,44 @@ $ molecule test
...
+
+To test the playbook from a collection,
+let's add a task to our role under `tasks/main.yml` file like so:
+
+```
+---
+- name: Include a playbook from a collection
+  ansible.builtin.import_playbook: foo.bar.my_playbook
+
+```
```
samccann, line 213: At this point we haven't edited the default `converge.yml`  - is there a step missing or will this work with whatever was created by the init command?

samccann (commented):
Coming in with late comments. Consider using this as feedback for a way to in general, improve the getting started guide experience. It attempts to explain a different way of giving the reader information in a more step-by-step approach.

Issue #3987, opened 2023-08-07
Improve Scenario Detection


ajinkyau:
Molecule should be able to detect scenarios.
For example, maybe during `molecule list` commands and so on

cidrblock:
Closed by #3989

PR #3989, opened 2023-08-07
Allow molecule to list scenarios present under molecule directory within a collection

ajinkyau:
Related: #3987

### DIFF
--- a/.config/requirements.in
+++ b/.config/requirements.in
@@ -9,3 +9,4 @@ packaging
 pluggy >= 0.7.1, < 2.0
 PyYAML >= 5.1
 rich >= 9.5.1
+wcmatch>=8.1.2 # MIT
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,6 +72,7 @@ repos:
           - types-filelock
           - types-jsonschema
           - types-setuptools
+          - wcmatch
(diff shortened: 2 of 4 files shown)


ssbarnea (changes requested):
The fact this this passed packaging worries me because we added a new dependency which we failed to declare. That job should have failed. We better find a way to make it fail if we do this to avoid adding hidden deps. I am trying to figure-out how to do this properly...

cidrblock:
yeah, looks like it got pulled in from ansible-lint https://github.com/ansible/molecule/actions/runs/5785724537/job/15678953688#step:9:369

I think we should be able to do this without wcmatch though......or at least wait until we get a issue that can';t be resolved without it.

ssbarnea:
> 

I told @ajinkyau to use wcmatch, that is not an issue. The issue is that we did not declare it as a direct dependency. 

Second issue is why it got installed as I doubt we want to install ansible-lint in normal testing environments. I am still looking for a tool that can help us detect missing direct dependencies but i did not find one that works with modern packaging.

cidrblock:
ansible-lint is a dep for test........could we remove that?

I'll look into wcamtch, I'm not familiar enough with it to grok what it does above the std lib

cidrblock (approved):

Issue #3992, opened 2023-08-08
`molecule list` finds scenarios in virtual envrionments


cidrblock:
https://github.com/ansible/molecule/blob/6daea252e28383ab89b9e303491c755ed5c60e56/src/molecule/command/base.py#L194

We may want to comapre the glob result to the gitignore file.

Something like this may work:

https://github.com/ansible-community/pytest-ansible/pull/147/files#diff-02f455c445e2548f737624f7f6f482b28f0bd5f705b60f9d0dc3040293c4d0acR283

example here:

```
(venv) nuc ➜  ansible.scm git:(molecule2) ✗ molecule list 
WARNING  The scenario config file ('/home/bthornto/github/ansible.scm/venv/lib/python3.11/site-packages/molecule/test/scenarios/cleanup/molecule/default/molecule.yml') has been modified since the scenario was created. If recent changes are important, reset the scenario with 'molecule destroy' to clean up created items or 'molecule reset' to clear current configuration.
WARNING  The scenario config file ('/home/bthornto/github/ansible.scm/venv/lib/python3.11/site-packages/molecule/test/scenarios/driver/delegated/molecule/default/molecule.yml') has been modified since the scenario was created. If recent changes are important, reset the scenario with 'molecule destroy' to clean up created items or 'molecule reset' to clear current configuration.
WARNING  The scenario config file ('/home/bthornto/github/ansible.scm/venv/lib/python3.11/site-packages/molecule/test/scenarios/driver/delegated_invalid_role_name_with_role_name_check_equals_to_1/molecule/default/molecule.yml') has been modified since the scenario was created. If recent changes are important, reset the scenario with 'molecule destroy' to clean up created items or 'molecule reset' to clear current configuration.
WARNING  The scenario config file ('/home/bthornto/github/ansible.scm/venv/lib/python3.11/site-packages/molecule/test/scenarios/host_group_vars/molecule/default/molecule.yml') has been modified since the scenario was created. If recent changes are important, reset the scenario with 'molecule destroy' to clean up created items or 'molecule reset' to clear current configuration.
WARNING  The scenario config file ('/home/bthornto/github/ansible.scm/venv/lib/python3.11/site-packages/molecule/test/scenarios/interpolation/molecule/default/molecule.yml') has been modified since the scenario was created. If recent changes are important, reset the scenario with 'molecule destroy' to clean up created items or 'molecule reset' to clear current configuration.
[...]

PR #3996, opened 2023-08-10
Ignore scenarios in virtual envrionments

ajinkyau:
Ignore scenarios in virtual environments.
This only works when repository is initialized with git otherwise just returns all scenario files.

Fixes: #3992


ssbarnea (changes requested):
Please add test(s) for it.

cidrblock:
yeah, tests would be good here.  The only way I can think of testing this would be to touch a molecule.yml file in a scenario tree in a gitignored directory and ensure it isn't found.

If you move the filtering logic into it's own function, it might be easier to monkeypatch it in a test so you can turn it on and off to do both a psitive and negative test.

cidrblock (changes requested):
after our conversation today, I think both + and - test can be added w/o the overhead of a venv. LMK if you need anything along the way.

PR #3997, opened 2023-08-10
Scaffold future docs structure

cidrblock:
Scaffold out the future doc tree structure with placeholders to be populated as time permits.


docs/working/examples/example_1.md
```diff
@@ -0,0 +1 @@
+# Example 1
```
samccann, line 1: We should  watch these 'examples' carefully. Ideally, an example is part of a 'user story' or use case that is in the user guide. But might be clearer once the first example or two is created to see if it needs its own 'examples' section or is really part of a the user guide.

mkdocs.yml
```diff
@@ -67,6 +67,34 @@ nav:
...
+          - working/installation/install.md
+          - working/installation/upgrade.md
+      - Getting Started: working/getting_started/getting_started.md
+      - User Guide:
+          - working/user_guide/understanding.md
+          - working/user_guide/creating_testing_collection.md
+          - working/user_guide/testing_existing_collection.md
+          - working/user_guide/testing_existing_roles.md
+          - working/user_guid/ci_integration.md
+          - working/user_guid/using_tox-ansible.md
```
samccann, line 82: What's the theory here with having a user_guide and a user_guid directory?

mkdocs.yml
```diff
@@ -67,6 +67,34 @@ nav:
...
+      - References:
+          - working/references/cli.md
+          - working/references/configuration.md
+          - working/references/porting_change_log.md
+          - working/references/road_map.md
+      - Examples:
+          - working/examples/example_1.md
+          - working/examples/example_2.md
+          - working/examples/example_3.md
+      - Frequently asked questions: working/faq/faq.md
```
samccann, line 92: Consider not having a FAQ but including everything that is in the current FAQ into sections in the user guide.  FAQs tend to expand over time as an easy 'dumping ground' of new information. That said, y'all know your readers better so maybe some of the FAQs really are.. .frequently... repeatedly...asked questions :-)

docs/working/user_guide/understanding.md
```diff
@@ -0,0 +1 @@
+# Concepts
```
samccann, line 1:
```suggestion
# Understanding Molecule
```

samccann (commented):
Thanks for doing this! Looking good. Just a few nit comments/questions.

mkdocs.yml
cidrblock: lol, mistake :)

mkdocs.yml
cidrblock: We tend to use the faq as a place to put things that have come in as an issue or question that would only otherwise end up as "Note" or "warning" in the user guide. Users of downstream docs would necessarily have access to the upstream GH issues list so having the FAQ downstream can help.

That said, the point is good. Before adding a new ? to the faq, stop and ask would it be better in the user guide.

briantist:
was this meant to go live? The docs now have a lot of blank pages https://ansible.readthedocs.io/projects/molecule/working/references/cli/

samccann:
This PR attempts to fill some of the blanks, but it's stuck w CI problems I don't know how to fix - https://github.com/ansible/molecule/pull/4044

It still leaves holes for sure, though. The idea was to create the 'ideal docset' structure and then start filling in the blanks.

PR #4010, opened 2023-08-14
fix: remove `role` from `init` help info

chenrui333:
relates to #3959

relates to https://github.com/Homebrew/homebrew-core/pull/139499


cidrblock (approved):
TY!

PR #4044, opened 2023-09-14
Restructure docs

samccann:
This PR merges existing files with 'empty' future files into one doc set.  This is an alternative to the existing docsite which has a large but empty WIP documentation section.

NOTE: this PR still introduces a handful of empty files that readers may find confusing.


samccann:
This is how the updated left-hand navigation would look:

image

samccann:
heh that's what i get for doing a PR before running tox -e docs locally...:-P

samccann:
@ssbarnea - I might need some remedial help on this one. Basically, I moved a batch of files/renamed some etc so I could get a clean left-hand navigation. But this PR has a bunch of broken links problem similar to the other docs scaffolding PR - https://github.com/ansible/molecule/pull/3997

So I'm wondering what I need to fix since that other PR I was using to learn from was merged with the similar CI errors to this one?

samccann:
Thanks @ciecierski ! I tried that fix locally and yes it does remove the errors. So seems we have two ways foward:
1 - Merge (after I fix the conflicts) and ignore the tox errors since we know why they are there.
2 - Change the tox.ini to run only on the docs. This should be done in a separate PR imo if we want to go that route.

I've seen other projects use #1 so they aren't changing the default CI testing.  Who do we need to bring into the discussion to decide which way to go?

ciecierski:
> Thanks @ciecierski ! I tried that fix locally and yes it does remove the errors. So seems we have two ways foward: 1 - Merge (after I fix the conflicts) and ignore the tox errors since we know why they are there. 2 - Change the tox.ini to run only on the docs. This should be done in a separate PR imo if we want to go that route.
> 
> I've seen other projects use #1 so they aren't changing the default CI testing. Who do we need to bring into the discussion to decide which way to go?

I think we should change tox first to not ignore errors and then rebase this pull-request to fix conflicts and have fixed tox definition. I've already proposed fix to tox in https://github.com/ansible/molecule/pull/4093  

@ssbarnea Could you help us with making decision/reviews or providing your guidance on this matter?

ciecierski (approved):
done

ciecierski (approved):
Pull request looks good to me. All empty files removed and it just needs final rebase

ajinkyau:
@ssbarnea What needs to be done here?
Or
Can we close this as #4102 is related.

PR #4060, opened 2023-09-30
Docs: highlight importance of `ansible_collections` at root dir

felher:
Hey all,

thank you for providing molecule <3

I tried to follow the getting started guide but got a role-not-found-error. After some experimenting I found out that my mistake was that I didn't name my collection-root `ansible_collections`. Other people might make the same mistake I did. So I propose 2 changes:

- change "the `collections/ansible_collections`" to "a `collections/ansible_collections`", since it doesn't seem to matter where that directory lies and, at least on my system, the two ones that already exist are not writable by a normal user by default.
- Highlight that it is not only recommend, but important that the collections root is named `ansible_collections`. If you put your collections somewhere else, but follow the guide in every other regard, you will get a "role not found" error.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -6,7 +6,7 @@ familiar with the basics of how to use Molecule and what it can offer.
 
 1.  Create a collection
 
-    One of the recommended ways to create a collection is to place it under the `collections/ansible_collections` directory.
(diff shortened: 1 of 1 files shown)


felher:
Is the tox/docs error due to my changes? I searched the log for the getting-started, because I changed the `getting-started.md` file, but it came up short. At least for errors.

ssbarnea (approved):

PR #4100, opened 2023-12-11
Add macos to test matrix

cristianonicolai:


ssbarnea (changes requested):
Install docker on macos, look at vscode-ansible workflows to learn how to do it or you can add some code inside test_docker to not do any testing on mac.

Do not skip the test because PYTEST_REQPASS will cause the execution to fail if the number of passed test would be differnt. So keep in mind to modify the inside of the test.

Optimal is to enable docker, but the other option is also ok.

ssbarnea (approved):

cidrblock:
Work continued in #4191

PR #4132, opened 2024-02-07
GHA maintenance

ssbarnea:


.github/workflows/release.yml
```diff
@@ -8,36 +6,29 @@ on:
...
       - name: Check out src from Git
         uses: actions/checkout@v4
         with:
           fetch-depth: 0 # needed by setuptools-scm
-          submodules: true
-
       - name: Build dists
-        run: python -m tox
-
+        run: python -m tox -e pkg
```
webknjaz, line 32: FWIW dists **MUST NOT** be built in a job that has OIDC access. It must be a separate job with reduced privileges. This is due to the possibility of privilege elevation though build dep poisoning that would compromise accesses beyond just the PyPI.

Issue #3839, opened 2023-03-07
molecule ignores ANSIBLE_FILTER_PLUGINS env var


ultral:
# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible [core 2.11.12] 
  config file = None
  configured module search path = ['/home/jenkins_slave/workspace/st_all_roles_and_branches_master/library']
  ansible python module location = /home/jenkins_slave/workspace/st_all_roles_and_branches_master/.venv/lib64/python3.8/site-packages/ansible
  ansible collection location = /home/jenkins_slave/workspace/st_all_roles_and_branches_master/.cache/collections
  executable location = /home/jenkins_slave/workspace/st_all_roles_and_branches_master/.venv/bin/ansible
  python version = 3.8.13 (default, Nov  8 2022, 17:19:05) [GCC 8.5.0 20210514 (Red Hat 8.5.0-15)]
  jinja version = 3.0.3
  libyaml = True

  molecule 4.0.4 using python 3.8 
    ansible:2.11.12
    delegated:4.0.4 from molecule
    docker:2.1.0 from molecule_docker requiring collections: community.docker>=3.0.2 ansible.posix>=1.4.0
```

Molecule installation method (one of):

- pip

Ansible installation method (one of):

- pip

Detail any linters or test runners used:

# Desired Behavior

if set `ANSIBLE_FILTER_PLUGINS` is set to `/home/jenkins_slave/workspace/st_all_roles_and_branches_master/filter_plugins` then molecule pass it to ansible-playbook comand. 

# Actual Behaviour

`ANSIBLE_FILTER_PLUGINS` is configured to 
```
    /home/jenkins_slave/workspace/st_all_roles_and_branches_master/.venv/lib64/python3.8/site-packages/molecule/provisioner/ansible/plugins/filter
    /home/jenkins_slave/.cache/molecule/bashrc_vars/default/plugins/filter
    /home/jenkins_slave/workspace/st_all_roles_and_branches_master/roles/bashrc_vars/plugins/filter
    /home/jenkins_slave/.ansible/plugins/filter
    /usr/share/ansible/plugins/filter
```

this behaviour is coming from `.venv/lib/python3.8/site-packages/molecule/provisioner/ansible.py`. the paths are hardcoded
```
                "ANSIBLE_FILTER_PLUGINS": ":".join(
                    [
                        self._get_filter_plugin_directory(),
                        util.abs_path(
                            os.path.join(
[...]

PR #4135, opened 2024-02-09
molecule uses ANSIBLE_FILTER_PLUGINS

ultral:
if ANSIBLE_FILTER_PLUGINS set then molecule uses it instead of hardcoded paths. closes https://github.com/ansible/molecule/issues/3839


ssbarnea (changes requested):
If we do want to prevent a regression in the future, we also need a test for this. Can you please add one?

ultral:
@ssbarnea sure np. are there any guidelines?

ssbarnea:
> @ssbarnea sure np. are there any guidelines?

Not really, look at similar tests for inspiration. Codecov will complain if you fail to provide enough coverage. I will mark it as draft until we get CI to report green on all checks.

ultral:
pre-commit.ci autofix

ultral:
@ssbarnea  I've change slightly logic. I created the `_get_filter_plugins_directories` and covered it by tests. it's quite similiar to the `_get_modules_directories`. 

P.s. In general, I'm thinking it make sense to unify approaches for ANSIBLE_ROLES_PATH, ANSIBLE_LIBRARY, ANSIBLE_FILTER_PLUGINS. they looks quite simillar but have different defaults and rules.

ultral:
@ssbarnea @shatakshiiii @Qalthos  it looks like there was a bug on CI/CD side because of flapping tests. can we re-run it?

ssbarnea (approved):

PR #4155, opened 2024-03-21
Update configuration.md for role name check

AlexanderStohr:
* fixed links to other docs on the web (name standard for namespace and role)
* added hint for extra freedom with a computed fully qualified role name

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -32,9 +32,10 @@ If the computed fully qualified role name does not follow current galaxy
 requirements, you can ignore it by adding `role_name_check:1` inside the configuration file.
 
 It is strongly recommended to follow the name standard of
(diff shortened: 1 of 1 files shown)


AlexanderStohr:
if that were code i would add the label 'bug'.
but as its only documentation, i would rather add the label 'minor'.
as of now i cant see any "gear" symbol or edit button that would allow me to add any label at all.

AlexanderStohr:
@ssbarnea @audgirka @ziegenberg - are you interested in participating with this little change?

AlexanderStohr:
> if that were code i would add the label 'bug'. but as its only documentation, i would rather add the label 'minor'. as of now i cant see any "gear" symbol or edit button that would allow me to add any label at all.

only the label is still /open/.
please assist, so this fix can continue. thank you.

ssbarnea (approved):

PR #3972, opened 2023-07-22
Add example of integration with kubevirt

jangel97:
Greetings,

In our team, we handle important Ansible roles to configure various aspects of RHEL systems, such as sshd, firewalld, grub, crypto-policies, and more. Although we have implemented automated testing with Molecule and Podman, there are instances where manual testing on our RHEL hosts becomes necessary before we proceed with deployment. I propose the integration of Molecule with Kubevirt as a potential solution to this issue. Such an integration would allow us to quickly and effectively test our Ansible roles on actual VMs, enhancing our testing procedures and providing more reliable results.

Some benefits of testing on VMs:

1. **System-level realism**: Ephemeral VMs provide a complete, isolated guest OS environment, much like your production environment. Containers share the host's kernel and are not fully isolated. This difference can occasionally lead to inconsistencies between testing and production environments. With VM-based testing, you can ensure the roles will work as expected on the actual operating system.

2. **Broader Compatibility**: Not all applications or configurations are container-friendly, especially when they interact with the system at a low level. VMs provide broader compatibility for testing as they offer a full-fledged OS environment.

3. **Improved Debugging**: Since VMs provide an entire guest operating system, it is often easier to debug issues related to system services, kernel modules, and other low-level components.

4. **Greater Variety of Testing**: VMs can run different kernel versions, different operating systems, or different system-level configurations. In contrast, containers are somewhat limited by the features and configurations of the host kernel.

In this pull request, I have utilized ephemeral VMs as a part of my example to showcase the benefits of this approach. By harnessing the capabilities of ephemeral VMs, we can test our Ansible roles in a highly realistic environment that closely mirrors our actual production environment, taking full advantage of the benefits outlined above.

This example serves as a practical guide to illustrate how a real-world application of Molecule and Kubevirt integration can enhance our current testing methods, enabling more realistic testing.

PR #4158, opened 2024-03-28
dynamically get the nodeport host

yahoon:
Greetings,

This is an update to the example doc of [`Using Kubevirt`](https://ansible.readthedocs.io/projects/molecule/examples/kubevirt). The purpose is to **dynamically** get the NodePort host to replace the hardcoded hostname that is used for the VM SSH connection.

**The problem we solved**
By fetching the `hostname` dynamically, we avoid using a pre-defined host for the VM SSH connection, that will reduce the the potential failures by the connection problem with the host. In this update, the hostname for the SSH connection is fetched dynamically, which is the K8S node where the VM is running.

I get this tested in our environment with @jangel97

Thanks.

### DIFF
--- a/docs/examples/kubevirt.md
+++ b/docs/examples/kubevirt.md
@@ -40,6 +40,9 @@ rules:
   - apiGroups: [""]
     resources: ["services"]
     verbs: ["get", "list", "watch", "create", "delete", "patch", "edit"]
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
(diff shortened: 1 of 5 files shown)


yahoon:
I don't have the permission to add labels to this PR. Please help add `docs`, `enhancement` to it, thanks.

BTW, @jangel97 is the original author of this guide in [!3972](https://github.com/ansible/molecule/pull/3972). He tested this change with me together.

jangel97:
Hello,

I believe this would help streamline the experience for users trying to use ephemeral kubevirt VMs. Can someone provide feedback?

Thx!

Qalthos (approved):

cidrblock (approved):

PR #4169, opened 2024-04-22
Added more dynamic arguments to podman container creation playbook

felipe4334:
It can be a bit deceiving when [platforms:] changes inside of molecule.yml do not take effect when creating containers. I believe it would be better if create.yml example playbook accepted more dynamic arguments by default.


felipe4334:
> Label error. Requires exactly 1 of: bug, enhancement, major, minor, patch, skip-changelog. Found:

I cant add label I don't think but, its: docs, enhancement

PR #4160, opened 2024-04-04
Use MOLECULE_GLOB environment variable for list command as well

elara-leitstellentechnik:

Issue #4117, opened 2024-01-19
CRITICAL 'molecule/default/molecule.yml' glob failed.  Exiting - Unable to run molecule 


iamgini:
### Prerequisites

- [X] This was not already reported in the past (duplicate check)
- [X] It does reproduce it with code from main branch (latest unreleased version)
- [X] I include a minimal example for reproducing the bug
- [X] The bug is not trivial, as for those a direct pull-request is preferred
- [X] Running `pip check` does not report any conflicts
- [X] I was able to reproduce the issue on a different machine
- [X] The issue is not specific to any driver other than 'default' one

### Environment

```
$ ansible --version
ansible [core 2.16.2]
  config file = /home/iamgini/.ansible.cfg
  configured module search path = ['/home/iamgini/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.12/site-packages/ansible
  ansible collection location = /home/iamgini/collections:/home/iamgini/.ansible/collections/ansible_collections:/usr/share/ansible/collections/ansible_collections:/home/iamgini/ansible/ansible-real-life/ansible-molecule-demo/collections/ansible_collections
  executable location = /usr/bin/ansible
  python version = 3.12.1 (main, Dec 18 2023, 00:00:00) [GCC 13.2.1 20231205 (Red Hat 13.2.1-6)] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

### What happened

I have a collection with molecule configured and its working perfectly in one folder.
Now I am trying to copy the molecule setup for another collection but facing the below error.

```
$ molecule create
CRITICAL 'molecule/default/molecule.yml' glob failed.  Exiting.

$ molecule test
CRITICAL 'molecule/default/molecule.yml' glob failed.  Exiting.

$ molecule reset
CRITICAL 'molecule/default/molecule.yml' glob failed.  Exiting.

$ molecule destroy
CRITICAL 'molecule/default/molecule.yml' glob failed.  Exiting.
```

I tried:

1. re-copy the working extensions folder from another working collection dir but getting same result!
2. Freshly done `molecule init scenario`

Any idea?

priyamsahoo:
@iamgini, it is hard for us to reproduce the scenario. Is it possible for you to provide more info on the setup, or maybe if you could create a branch with the setups, we can reproduce the error?

shameoff:
I have the same error but maybe in little different case. 
[...]

Issue #4061, opened 2023-10-04
Molecule Reset overwrites projects in COLLECTIONS_PATHS


markfaine:
### Prerequisites

- [X] This was not already reported in the past (duplicate check)
- [X] It does reproduce it with code from main branch (latest unreleased version)
- [X] I include a minimal example for reproducing the bug
- [X] The bug is not trivial, as for those a direct pull-request is preferred
- [X] Running `pip check` does not report any conflicts
- [X] I was able to reproduce the issue on a different machine
- [X] The issue is not specific to any driver other than 'default' one

### Environment

molecule 6.0.2 using python 3.11
    ansible:2.15.1
    default:6.0.2 from molecule
Linux wsl 5.15.90.1-microsoft-standard-WSL2 #1 SMP Fri Jan 27 02:56:13 UTC 2023 x86_64 GNU/Linux
Ubuntu 22.04.2 LTS

### What happened

Running molecule reset can delete the entire project.  The role is in a collection and I think the issue is it tries to install the same collection from GitLab  using ansible-galaxy, since my collections path is the same as my development path it overwrites my project.   I've been using this same setup for years and this did not happen with previous versions of Molecule.

I am actively developing the collection so it is expected that I might need to run them.  When I run them they need to be on the collections path.

If I edit my ansible.cfg and comment out the collections_path, it works correctly:
```
molecule reset
WARNING  The scenario config file ('/home/ansible/ansible/development/collections/ansible_collections/nats/default/roles/aide/molecule/default/molecule.yml') has been modified since the scenario was created. If recent changes are important, reset the scenario with 'molecule destroy' to clean up created items or 'molecule reset' to clear current configuration.
INFO     default scenario test matrix:
INFO     Performing prerun with role_name_check=0...
INFO     Running from /home/ansible/ansible/development/collections/ansible_collections/nats/default/roles/aide : ansible-galaxy collection install -vvv --force ../..
INFO     Removing /home/ansible/.cache/molecule/aide/default
```

I could probably work around it by using a shell function or alias but I don't want to have to rely on a workaround.

### Reproducing example

```yml
To reproduce ensure your COLLECTIONS_PATHS contains the role that you are running molecule on and do `molecule reset` in any role.

For example:

ansible-config dump
...
COLLECTIONS_PATHS(/home/ansible/ansible/development/ansible.cfg) = ['/home/ansible/ansible/development/collections']
...
```

Change into a role that is under COLLECTIONS_PATHS and run molecule reset

```
[...]

Issue #4000, opened 2023-08-12
RFE: Make molecule collection aware


cidrblock:
As a user I want molecule to be aware that my scenarios are contained within an ansible collection.

Use cases:
- Provide the collection name to the user for use in ansible content, e.g. molecule_collection_name.  This may be used in task names, debugging, etc and will avaoid the need for every user to parse the galaxy.yml file
- Will allow for the introduction of shared resources between scenarios within the same collection.  See follow up RFEs

PR #4177, opened 2024-05-01
Fix: Update default globbing pattern for collection matching

trfore:
This update:
- restricts the glob matching for `molecule test`
- updates the default value for `MOLECULE_GLOB="**/molecule/*/molecule.yml"`

This allows:
- users to call molecule from the collection folder and molecule will use the collection name in the ephemeral cache, `.cache/molecule/COLLECTION_NAME`.

WARNING:
If working within the ANSIBLE_COLLECTIONS_PATH directory, e.g. `~/.ansible/collections/ansible_collections/COLLECTION_NAME` (default), using the new default setting `MOLECULE_GLOB="**/molecule/*/molecule.yml"` will globber clobber and delete the collection folder. This occurs in the presence of a `galaxy.yml` - as molecule calls `ansible_compat.runtime(isolate=False)` and will run `ansible_compat.runtime.prepare_environment`. However, if you delete the `galaxy.yml` OR chdir into `COLLECTION_NAME/extensions` and it will not delete the collection folder.

Currently (v24.2.1):
Calling `molecule list` will find all molecule.yml files within a collection folder that use a nested dir structure, i.e
`COLLECTION_NAME/extensions/molecule/SCENARIO_NAME/molecule.yml`. As it calls `base.get_configs` and sets the argument `glob=str"**/molecule/*/molecule.yml"`.

This does not occur with `molecule test`, as it calls to `base.execute_cmdline_scenarios` and uses
`MOLECULE_GLOB = os.environ.get("MOLECULE_GLOB", "molecule/*/molecule.yml")`.

Attempting to set an env var `MOLECULE_GLOB="**/molecule/*/molecule.yml"` will fail, as `test` attempts to insert the scenario name into every wild-card. But, this fuzzy matching acts as a safe guard against deleting the collection when working within the same ANSIBLE_COLLECTIONS_PATH directory:

	$ MOLECULE_GLOB="**/molecule/*/molecule.yml" molecule test
	CRITICAL 'defaultdefault/molecule/default/molecule.yml' glob failed.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -47,7 +47,7 @@ jobs:
       matrix: ${{ fromJson(needs.pre.outputs.matrix) }}
 
     env:
-      PYTEST_REQPASS: 453
+      PYTEST_REQPASS: 454
     environment: test
     steps:
       - uses: actions/checkout@v4
--- a/src/molecule/command/base.py
+++ b/src/molecule/command/base.py
@@ -40,7 +40,7 @@ from molecule import config, logger, text, util
 from molecule.console import should_do_markup
 
 LOG = logging.getLogger(__name__)
-MOLECULE_GLOB = os.environ.get("MOLECULE_GLOB", "molecule/*/molecule.yml")
+MOLECULE_GLOB = os.environ.get("MOLECULE_GLOB", "**/molecule/*/molecule.yml")
 MOLECULE_DEFAULT_SCENARIO_NAME = "default"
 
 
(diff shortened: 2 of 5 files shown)


trfore:
So I was curious why `molecule list` is collection aware and all the other commands are not. The current work-around for making molecule collection aware is to use one of the following approaches:
```sh
$ MOLECULE_GLOB="./extensions/molecule/*/molecule.yml" molecule test
$ MOLECULE_GLOB="$(pwd)/extensions/molecule/*/molecule.yml" molecule test
```
- This fix, partially addresses #4000 and #4117. 
- Is compatiable with PR #4160.

- Yet, it easily exposes one of the current bugs within molecule (#4061) if end-users are working within the same directory as `ANSIBLE_COLLECTIONS_PATH`.

PR #4181, opened 2024-05-14
Fix: MOLECULE_GLOB not respected in list command

MaxBidlingmaier:
Molecule allows to change the config folder by setting MOLECULE_GLOB env. This is working for most parts but not for the list command.


MaxBidlingmaier:
Doubles https://github.com/ansible/molecule/pull/4160

PR #4185, opened 2024-05-20
Avoid requests 2.32.0 during testing

ssbarnea:
Apparently requests==2.32.0 is incompatible with vendored versions of docker-py from inside community.docker collection. We do not have a direct dependency on that but one of our integration tests fails, so we only update our test dependencies.

Related: https://github.com/docker/docker-py/issues/3256
Related: https://github.com/ansible-collections/community.docker/issues/860
Related: https://github.com/ansible-collections/community.docker/pull/861


abhinabphy:
how to revert the requests back to 2.31.0 here

PR #4191, opened 2024-06-04
Update tox.yml

cidrblock:
Switch to shared tox.yml

A couple changes for the tests:

Skip the macos podman and docker test as we don't have podman or docker installed
The `test_command_dependency[dependency-default-shell]` fixture was changed to set the collection path for ansible to the directory where the collection is installed.

PYTEST_REQPASS will be reintroduced consistently across the board later

### DIFF
--- a/test/scenarios/dependency/molecule/shell/molecule.yml
+++ b/test/scenarios/dependency/molecule/shell/molecule.yml
@@ -13,9 +13,6 @@ platforms:
     image: ${TEST_BASE_IMAGE}
 provisioner:
   name: ansible
-  config_options:
-    defaults:
-      collections_paths: ${MOLECULE_EPHEMERAL_DIRECTORY}/collections
 scenario:
   name: shell
 verifier:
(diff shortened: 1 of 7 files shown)


cidrblock:
Related: https://github.com/ansible/ansible-compat/pull/387

This is why the collection can't be found

test/scenarios/dependency/molecule/shell/molecule.yml
```diff
@@ -3,7 +3,7 @@ dependency:
   name: shell
   command: >
     bash -c "
-    ansible-galaxy collection install -p '${MOLECULE_EPHEMERAL_DIRECTORY}/collections' community.molecule &&
+    ansible-galaxy collection install -p '${MOLECULE_EPHEMERAL_DIRECTORY}/collections' community.molecule --force &&
```
cidrblock, line 6: This was added to ensure the collection is installed because the environment is not isolated and the ANSIBLE_COLLECTIONS_PATH not managed when calling with bash.  This will force install the collection even if it exists in a stand location.

The following scenarios will have their collection path set to the ephmeral directory by default.

cidrblock:
no runtime changes, self approving to match the common tox

PR #4192, opened 2024-06-05
Respect ANSIBLE_COLLECTIONS_PATH

cidrblock:
Inform ansible compat that the ANSIBLE_COLLECTIONS_PATH does not need to be set.

The change in compat behaviour changed here: https://github.com/ansible/ansible-compat/commit/32ce03c9989698544353a24a5f8b7399c1a0b8a0#diff-52de7db791930eca228ba958a8fe0d53d2695de26a1d6e1aa1906da28f607892R395

and was released in 24.6.0

### DIFF
--- a/.config/requirements.in
+++ b/.config/requirements.in
@@ -1,4 +1,4 @@
-ansible-compat >= 4.1.8
+ansible-compat >= 24.6.0
 ansible-core >= 2.12.10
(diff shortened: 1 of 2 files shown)


cidrblock:
Self approving as the comapt release introduced a breaking change for molecule, release to follwo

PR #4197, opened 2024-06-06
Update mypy configuration

cidrblock:
- Switch to typical mypy configuration
- Add inline type ignores
- Update requriements.txt and update constraints as needed

This should allow for incremental typ hint addition w/o regression since the scope is per-error and not global for the project.


cidrblock:
No runtime changes, just CI clenaup

PR #4201, opened 2024-06-06
Update ruff configuration

cidrblock:
- Add inline noqa so the code does not regress with broad exclusions
- Add ruff to test extras and constraints


cidrblock:
insignificant coverage change

PR #4206, opened 2024-06-07
Restructure tests

cidrblock:
Rename `a_unit` tests to `unit`
Rename `b_fuctional` tests to `integration`
Move `resources` and `scenarios` into a test-named fixture folder
Update a few tests as needed, add doc string and types while there

(determining whenther or not each test is _really_ a unit or integration test was outside the scope of what Iw asafter here)

(I'm not sure that "integration" is really the best anme here, but at least it is consistent with most of other projects)


cidrblock:
ignoring insigificant coverage change

PR #4219, opened 2024-06-08
chore: Update pyproject.toml

cidrblock:
Update tmp retention policy for pytest

Note:  For some reason I recall this was an issue for molecule but I was not able to recreate it.  If it proves to be, let's remove it.


cidrblock:
ah, only on GH

PR #4221, opened 2024-06-10
Restore examples in docs

cidrblock:
As a result of the project directory restrucutre the examples in the docs were lost.

This should restore the examples.

If additional includes are needed in the future that fall outside of the specified directory it can be modified later.


cidrblock:
Fixing previous mistake

PR #4224, opened 2024-06-11
Unit test cleanup

cidrblock:
- Eliminate a bunch of unused fixtures
- Rename the test_ephemeral_dir_path fixture to test_cache path since it's being used for other purposes
- Add odcstring and type hints as needed in some places
- Test no longer write to the user's `~/.cache directory`
- Added session finish cleanup to remove the project .cache/.molecule/tests directory after a successful run
- Serial running of tests should no longer be necessary since there should be no overlap of temporary directories between tests

One runtime change, sinc ehte test cache directory may have a `[id]` in it, the glob would picxk that up.  So we will escape the directory name since it's a real directory and not user provided.


cidrblock:
just a bunch of test cleanup

PR #4236, opened 2024-06-25
Restore pydoclint, add class docstrings

cidrblock:
This reverts #4235 

Depends on #4237 for the milestone fix

Flake8 is necessary as it is the only way to get realtime feedback during development using VsCode.

In this PR, the flake8 and pydoclint configurations are restored to what they were.

The latest version of pydoclint added class atribute checking in the class docstrings.  Included are those docstrings such that precommit should pass the check.

This will also keep the pydoclint configuration consistent with some of the other team repos:

[pytest-ansible](https://github.com/ansible/pytest-ansible/blob/main/.flake8)
[tox-ansible](https://github.com/ansible/tox-ansible/blob/main/.flake8)
[ansible-creator](https://github.com/ansible/ansible-creator/blob/main/.flake8)
[ansible-dev-environment](https://github.com/ansible/ansible-dev-environment/blob/main/.flake8)
[ansible-dev-tools](https://github.com/ansible/ansible-dev-tools/blob/main/.flake8)

Note: an updater script was included if needed
Also: Having DOC as an external for ruff is not needed as there are no `noqa`s inline for DOC. It is unecessary configuration.


cidrblock:
`tox/tox/pkg` failure appears to be identified upstream with work happening on it: https://github.com/pypa/twine/pull/1124

.pre-commit-config.yaml
```diff
@@ -71,10 +71,13 @@ repos:
       - id: cspell
         name: Spell check with cspell
 
-  - repo: https://github.com/jsh9/pydoclint
-    rev: 0.5.1
+  - repo: https://github.com/pycqa/flake8
```
ssbarnea, line 74: This is the big problematic:
- no pinning and automatic update of pydoclint itself
- flake8 maintenance "style".
- name of hook make is very hard to call it directly (a lot to type)

ssbarnea (changes requested):

.pre-commit-config.yaml
cidrblock: I recognise the compromise here and am open to suggestions.

I do think it is important that what ever we do there is a corresponding extension for VsCode and feedback during development.  

I wanted to bring attention to this as it is somewhat related, there is an ongoing conversation withing the ruff repo related to pydoclint. Hopefully the configuation here is short lived and ruff gets parity. https://github.com/astral-sh/ruff/issues/458#issuecomment-2122773827

PR #4237, opened 2024-06-25
Update test for milestone 2.18

cidrblock:
For unknown reason, the behaviour of milestone differs such that the PWD environment variable is not updated to the directory of the process.

Because we are not running with a shell and it's the shell's responsibility to keep it updated, we'll switch instead to a lookup pipe to get the pwd and not rely on the env var.

After several hours of digging without success, this should accomplish what the test intended to do with a dependency on the shell.


cidrblock:
https://github.com/ansible/molecule/pull/4236#issuecomment-2189831770

ssbarnea:
pkg job failing due to new twine release, see https://github.com/pypa/twine/issues/977

ssbarnea (approved):

PR #4239, opened 2024-06-26
Prevent building docs with missing includes

ssbarnea:
Fixes: #4221


cidrblock:
I missed which docs are missing examples.

I do like the throw exception addition!

cidrblock (approved):

PR #4241, opened 2024-06-26
Add VsCode support for pydoclint

cidrblock:
Since flake8 has been removed, VsCode no longer provided feedback during development about new docstring errors.

This PR add a new task called `pydoclint` that runs pydoclint in the background.  Message from pydoclint are parsed with a regex and presented in the problems tab of VsCode by the task.

VsCode does not have native support for running a task on save, so a new extension is added to support that.

The flake8 extension is removed which was an oversight in #4235 

Related #4236

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,10 +2,10 @@
   "recommendations": [
     "charliermarsh.ruff",
     "esbenp.prettier-vscode",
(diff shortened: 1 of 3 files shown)


cidrblock:
Gonna merge this one so it can be carried over to other projects

PR #4263, opened 2024-07-27
Remove cookiecutter related remnants from unit tests

micwoj92:
Small clean up after https://github.com/ansible/molecule/commit/21dcd2bb7e8e9002be8bbc19de3e66ec3ce586f1

I am unsure whether file `tests/fixtures/resources/invalid_role_template/bad_format/README.md` also needs to be deleted or not.


Qalthos:
> I am unsure whether file tests/fixtures/resources/invalid_role_template/bad_format/README.md also needs to be deleted or not.

I t doesn't look like we use the invalid_role_template path at all, so you should be able to delete the entire directory :+1:

micwoj92:
Done

Issue #4043, opened 2023-09-13
Incorrect pip install documentation


anshulbehl:
### Prerequisites

- [X] This was not already reported in the past (duplicate check)
- [X] It does reproduce it with code from main branch (latest unreleased version)
- [X] I include a minimal example for reproducing the bug
- [x] The bug is not trivial, as for those a direct pull-request is preferred
- [X] Running `pip check` does not report any conflicts
- [X] I was able to reproduce the issue on a different machine
- [X] The issue is not specific to any driver other than 'default' one

### Environment

This is a documentation bug on the pip install of molecule

### What happened

https://ansible.readthedocs.io/projects/molecule/installation/#pip

The above docs mention ansible-core as not a direct dependency, but looks like it is if I do a pip instal or I check 

https://github.com/ansible/molecule/blob/v6.0.2/.config/requirements.in#L2

alisonlhart:
Fixed in #4279.

PR #4279, opened 2024-08-31
Fix: Incorrect pip install documentation

madalynbartman:
#4043

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -6,10 +6,12 @@ Molecule.
 
 ## Requirements
 
+- ansible-core >= 2.12
+
 Depending on the driver chosen, you may need to install additional OS
 packages.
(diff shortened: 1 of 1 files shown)


madalynbartman:
Correct me if I'm, but I don't believe I can add a label to the MR as a regular user https://stackoverflow.com/questions/67840274/how-to-add-label-to-an-issue-on-github

Believe the label should be [docs](https://github.com/ansible/molecule/labels/docs)

alisonlhart:
@madalynbartman Thank you for the contribution! You're correct, the label add permissions are limited to maintainers.  I've added the `bug` label as that's what we typically use for documentation adjustments.

alisonlhart (approved):

PR #4298, opened 2024-10-10
Two updates to util

Qalthos:
* Type merge_dicts with a bound Typevar so that it is clearer about what it accepts and that it returns the same thing it takes in.
* write_file was missed in the Path update. This updates the method to accept Paths and simplifies the method by using Path internally.

### DIFF
--- a/src/molecule/util.py
+++ b/src/molecule/util.py
@@ -29,7 +29,7 @@ import sys
 
 from pathlib import Path
 from subprocess import CalledProcessError, CompletedProcess
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, TypeVar
 
 import jinja2
 import yaml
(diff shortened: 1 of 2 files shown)


src/molecule/util.py
```diff
@@ -48,6 +48,10 @@
     from typing import Any, AnyStr, NoReturn
     from warnings import WarningMessage
 
+    Basic = str | float | bool
+    NestedDict = MutableMapping[str, Basic | None | list[Basic | "NestedDict"] | "NestedDict"]
+    _T = TypeVar("_T", bound=NestedDict)
```
Qalthos, line 53: This also works, I just like having a little more specificity on what's actually coming out of these methods, and there's a lot of things that realistically aren't going to be in one of these dictionaries.

```suggestion
    _T = TypeVar("_T", bound=MutableMapping[str, Any])
```

ssbarnea: Take a look at https://github.com/ansible/ansible-compat/blob/main/src/ansible_compat/types.py

ssbarnea (approved):

PR #4303, opened 2024-10-18
Type hint and document config.py

Qalthos:
Config touches a lot of other files, so there's quite a bit of related changes.


ssbarnea (changes requested):
Were you not able to avoid the `# type: ignore[union-attr]` bits? I really hate to see them?

If you plan to address them in a follow-up because it would increase the size of the PR too much, is ok and you can discard my request.

Qalthos:
> Were you not able to avoid the `# type: ignore[union-attr]` bits? I really hate to see them?
> 
> If you plan to address them in a follow-up because it would increase the size of the PR too much, is ok and you can discard my request.

Yeah, the `union-attr`s I'm _adding_ are because a lot of these are typed as `type | None`, and we make assumptions about them existing. They absolutely need to be fixed, but given the time it would take, I figured it's better to keep going, identify all the problem areas, and clean up all of them at once in a later PR.

ssbarnea (approved):

PR #4340, opened 2024-12-09
Make molecule collection-aware

Qalthos:
Fixes https://github.com/ansible/molecule/issues/4000

Two new Config properties named `collection` and `collection_path` will be populated if Molecule detects it is running inside a collection.

### DIFF
--- a/src/molecule/config.py
+++ b/src/molecule/config.py
@@ -279,11 +279,11 @@ class Config:
         important_keys = {"name", "namespace"}
         if missing_keys := important_keys.difference(galaxy_data.keys()):
             LOG.warning(
-                "The detected galaxy.yml file (%s) is incomplete, missing %s",
+                "The detected galaxy.yml file (%s) is invalid, missing mandatory field %s",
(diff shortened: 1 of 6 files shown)


src/molecule/config.py
```diff
@@ -240,6 +258,20 @@ def molecule_directory(self) -> str:
...
+        """Collection metadata sourced from galaxy.yml.
+
+        Returns:
+            A dictionary of information about the collection molecule is running inside, if any.
+        """
+        if not self.collection_directory:
+            return None
+
+        galaxy_file = self.collection_directory / "galaxy.yml"
+        galaxy_data = util.safe_load_file(galaxy_file)
```
cidrblock, line 273: If we don't at least have name and namespace we should fail fast and let the user know the galaxy.yml file needs these values.  I think the intent of the user here would be to be collection aware rather than not

cidrblock (changes requested):

cidrblock (approved):

Issue #4142, opened 2024-02-24
Molecule config file in {project}/.config/molecule/config.yml not found


erichoog:
### Prerequisites

- [X] This was not already reported in the past (duplicate check)
- [X] It does reproduce it with code from main branch (latest unreleased version)
- [X] I include a minimal example for reproducing the bug
- [X] The bug is not trivial, as for those a direct pull-request is preferred
- [X] Running `pip check` does not report any conflicts
- [X] I was able to reproduce the issue on a different machine
- [X] The issue is not specific to any driver other than 'default' one

### Environment

OS Release: Red Hat 8.9 
```
$molecule --version
molecule 6.0.3 using python 3.11 
    ansible:2.16.2
    azure:23.5.0 from molecule_plugins
    containers:23.5.0 from molecule_plugins requiring collections: ansible.posix>=1.3.0 community.docker>=1.9.1 containers.podman>=1.8.1
    default:6.0.3 from molecule
    docker:23.5.0 from molecule_plugins requiring collections: community.docker>=3.0.2 ansible.posix>=1.4.0
    ec2:23.5.0 from molecule_plugins
    gce:23.5.0 from molecule_plugins requiring collections: google.cloud>=1.0.2 community.crypto>=1.8.0
    podman:23.5.0 from molecule_plugins requiring collections: containers.podman>=1.7.0 ansible.posix>=1.3.0
    vagrant:23.5.0 from molecule_plugins```

### What happened

**ISSUE:**
Molecule settings put into `{PROJECT_DIR}/.config/molecule/config.yml` are not found or used, but they are if I move `.config/molecule/.config.yml` to my user home directory.

**EXPECTATIONS:**
Molecule [documentation](https://ansible.readthedocs.io/projects/molecule/configuration/#prerun) states that we can put configurations into `{PROJECT_DIR}/.config/molecule/config.yml` or into `~/.config/molecule/config.yml`   So both options should work.

**EXAMPLE:**
My example directory structure:
```
$ tree -a my_project/
my_project/
├── .config
│   └── molecule
│       └── config.yml
└── roles
    └── somerole
        └── molecule
            └── default
                ├── converge.yml
                ├── create.yml
                ├── destroy.yml
                └── molecule.yml
```

Here are the contents of `config.yml`:
```
---
driver:
  name: docker
```

Here's what I see if I configure using `~/.config/molecule/config.yml` - it finds the file and the settings are used, this is good:
```
$ molecule converge
[...]

PR #4373, opened 2025-01-22
Fix ansible#4142 - config file in proj dir (second attempt)

erichoog:
This resolves the issue raised in Issue #4142 where Molecule does not find the config.yml file the project directory `{project_dir}/.config/molecule/config.yml` but does find it in the home directory.

I ran tests by isolating the 2 functions being called when `LOCAL_CONFIG` variable is defined and verifying that the config file is found when in the project dir.  If there is no `{project_dir}/.config/molecule/config.yml`  then it looks in `~/.config/molecule/config.yml` and finds the config file.

I have not been able to test the full project with the `tox` command as my environment is having issues.  The `tox` tests are not completing even on a freshly cloned repo.

The code in the find_vcs_root function could likely be modified more as it does not appear to be called elsewhere and it searches for ".git", ".hg", ".svn" dirs as well for what appears to be no reason as far as I can tell since it is only used for locating `.config/molecule/config.yml`

Let me know if anything can be improved upon here with my delivery of this as it is my first PR with the project.

Fixes: #4142


Qalthos (approved):

erichoog:
> Missing tests and documentation, the `.config` folder is not realy a vcs.

I'll work on adding the tests hopefully this week.  

For the `.config` not being a vcs, I'll take a look at the code again and see if this should be implemented elsewhere or in a new function.

erichoog:
> Missing tests and documentation, the `.config` folder is not realy a vcs.

@ssbarnea After reviewing this further, I have determined that the issue I was seeing was because I was not using vcs like git in my test project dir.  Therefore, since there was no `.git` directory in my project directory the `find_vcs_root` function was not finding my `.config/molecule/config.yml` file.

I am now wondering if it would just make more sense to update the documentation to make it more clear that the `.config/molecule/config.yml` would only be found in the project directory if version control is being used?

Or would it still be good to pursue modifying the code to allow the `.config/molecule/config.yml` to be found in the project dir when source control is not being used?

PR #4375, opened 2025-01-22
Validate docker installation before running docker tests

ssbarnea:
Related: https://github.com/ansible-collections/community.docker/issues/660
Related: https://github.com/containers/podman/issues/16548"

### DIFF
--- a/tests/integration/test_command.py
+++ b/tests/integration/test_command.py
@@ -19,9 +19,12 @@
 #  DEALINGS IN THE SOFTWARE.
 from __future__ import annotations
 
+import os
+import shutil
 import subprocess
-import sys
+import warnings
 
+from functools import cache
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, TypedDict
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_command.py
```diff
@@ -50,6 +53,62 @@ def run(cmd: list[str], env: dict[str, str] | None = None) -> subprocess.Complet
...
+        # detect if docker is NOT using podman's backend (libpod)
+        result = run(
+            ["curl", "--silent", "--unix-socket", "/var/run/docker.sock", "http:/v1/_ping"],
+        )
+        if result.returncode != 0:
+            failure = "Docker does not seem to respond to default socket."
+        else:
+            result = run(
+                [
+                    "curl",
```
cidrblock, line 78: only because I was curious, this is neat: https://www.hackitu.de/urllib_uds/

cidrblock (approved):

tests/integration/test_command.py
ssbarnea: In fact I wanted to do something like this but decided to postpone. I bookmarked that code and will use it. Likely I will implement this as a fixture inside pytest-plus, so we can reuse it in multiple places.

PR #4380, opened 2025-01-27
Remove code that modifies ansible import paths

ssbarnea:
As part of new test-isolation strategy, molecule will no longer take
care itself about modification of:

- ANSIBLE_COLLECTIONS_PATH
- ANSIBLE_ROLES_PATH
- ANSIBLE_LIBRARY
- ANSIBLE_FILTER_PLUGINS

Test isolation is supposed to be covered transparently by ansible-compat runtime when code runs inside a virtual environment. Molecule is already using it for running all its external commands.

Related: https://ansible.readthedocs.io/projects/dev-tools/user-guide/test-isolation/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -178,3 +178,4 @@ tests/fixtures/resources/.extensions/
 
 # docs output
 _readthedocs/
+.ansible
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -93,7 +93,7 @@ alter them when needed.
 
 Yes, roles contained in a
 [monorepo](https://en.wikipedia.org/wiki/Monorepo) with other roles are
-automatically picked up and `ANSIBLE_ROLES_PATH` is set accordingly. See
+automatically picked up. See
 [this
 page](guides/monolith.md)
 for more information.
--- a/docs/guides/monolith.md
+++ b/docs/guides/monolith.md
@@ -41,9 +41,6 @@ $ molecule --debug test
 DEBUG: ANSIBLE ENVIRONMENT
 ---
 ANSIBLE_CONFIG: /private/tmp/monolith-repo/roles/baz/molecule/default/.molecule/ansible.cfg
-ANSIBLE_FILTER_PLUGINS: /Users/jodewey/.pyenv/versions/2.7.13/lib/python2.7/site-packages/molecule/provisioner/ansible/plugins/filters:/private/tmp/monolith-repo/roles/baz/plugins/filters:/private/tmp/monolith-repo/roles/baz/molecule/default/.molecule/plugins/filters
-ANSIBLE_LIBRARY: /Users/jodewey/.pyenv/versions/2.7.13/lib/python2.7/site-packages/molecule/provisioner/ansible/plugins/libraries:/private/tmp/monolith-repo/roles/baz/library:/private/tmp/monolith-repo/roles/baz/molecule/default/.molecule/library
-ANSIBLE_ROLES_PATH: /private/tmp/monolith-repo/roles:/private/tmp/monolith-repo/roles/baz/molecule/default/.molecule/roles
 ```
 
 Molecule can be customized any number of ways. Updating the
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,7 +93,7 @@ arg-type-hints-in-docstring = false
 baseline = ".config/pydoclint-baseline.txt"
 check-return-types = false
 check-yield-types = false
-exclude = '\.cache|\.git|\.tox|build|out|venv'
+exclude = '\.ansible|\.cache|\.git|\.tox|build|out|venv'
 quiet = true # no need to print out the files being checked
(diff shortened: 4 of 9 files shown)


jsf9k:
@ssbarnea - This breaks my `molecule` setup (and probably that of many other folks), and I'm not sure the best way to fix it.  See, e.g., the most recent GitHub Actions runs from cisagov/ansible-role-assessment-tool#62 where `molecule` can no longer find the role that is being tested (ansible-role-assessment-tool).  I'd like `molecule` to find and use the local version of the role that is in the code checked out from GitHub, since that is the version I want to be tested.  In this case, am I now expected to set an environment variable before running `molecule` to add the local project directory to the role path, since `molecule` no longer does this for me?

In cisagov/skeleton-ansible-role#217 I tried explicitly listing the role to be tested in `molecule.default/requirements.yml` as a solution, but if I do that _and_ modify the Ansible code for the role in a feature branch then running `molecule` from said feature branch will not pick up the local changes (since it will pull the default branch from the GitHub repo).  This is obviously subpar.

If there are more changes forthcoming that will fix this then I can pin to `<-25.1.0` for now, but I'm not sure if that is the case.

remod:
We fixed it by explicitly setting `ANSIBLE_COLLECTIONS_PATH=/path/to/collection` before invoking `molecule`.

jsf9k:
cisagov/skeleton-ansible-role#221 shows how I am intending to work around this.  Very similar to @remod's solution.

pierrehenrymuller:
Molecule and Ansible must behave in exactly the same way when searching for roles. Ansible has its own rules, and Molecule must test that Ansible will work properly in production. Changing this on the Molecule side no longer guarantees that when Ansible is running, everything will work properly.

Everything's broken with this new version, and a step backwards is imperative.

How does Molecule benefit from managing things differently?
What is the need to randomize functional testing?

Adding an environment variable to return to normal operation is not an efficient solution.

hswong3i:
Once I revert this PR my CI case working fine again:
* https://github.com/alvistack/ansible-community-molecule/commit/19a4df14eb8b1a1a596c7e0b4567186b36f9ed53
* https://build.opensuse.org/package/show/home:alvistack/ansible-community-molecule-25.2.0

alix-rm:
@ssbarnea would you care to comment on these changes?

From a user perspective, your commits through ansible-compat, molecule and even ansible-lint just broke a lot of people's workflows. The documentation is not sufficient for such drastic changes and unexpected for a minor version.

dgibbs64:
Can confirm testing of all my ansible roles is now broken since update to 25.2.0. Getting the following error. Is there a workaround or update planned to release asap?
```
ansible_compat.errors.AnsibleCommandError: Got 250 exit code while running: ansible-galaxy collection install -r /home/runner/work/ansible-role-nodejs/ansible-role-nodejs/requirements.yml
```

matthijs-oosterhoff:
These changes seem to break setups using `molecule-plugins[vagrant]` (and I suppose more), see https://github.com/ansible-community/molecule-plugins/issues/301. 

Setting the `ANSIBLE_LIBRARY` variable to something like `<your_venv>/lib/python<version>/site-packages/molecule_plugins/vagrant/modules` resolves the issue. But I don't think it's really user friendly to have to set this variable before running `molecule`.  Am I missing something? What is the advised way to manage these variables, now Molecule no longer takes care of them?

joerg:
For us this release also breaks all our roles and molecule tests... and I am not sure why. We have pretty simple and standard roles and molecule/ansible is set up via venv. I tried reading through all the linked documentation that I found and tested some approaches but none of them seem to make any difference.
I think it has to do with the test isolation linked in the original comment but if the role I am develping and testing is not automatically picked up by molecule then molecule is simply not doing its job. Probably I am just missing some smaller changes required on my side but these should be documented somewhere at least.

chriscroome:
FWIW when running Molecule in GitLab CI you can solve this issue by adding this at the top of the `.gitlab-ci.yml` file:

```yaml
variables:
    ANSIBLE_ROLES_PATH: "${CI_PROJECT_DIR}/..:~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles"
```

This might be worth adding to the docs.

dgibbs64:
@ssbarnea Are there plans to revert this change or provide any documentation on this? This change has broken a lot of peoples workflows with no warning or documentation on what we need to update. I feel like this change was badly planned and should be in a major release and not a minor one. 

In my (and other commenters) view, this change is a backwards step that makes using molecule more difficult to use . I am failing to see the benefit of such a chance.

MaKaNu:
@Exagone313 I am still not able to verify, but what I can say:

take a look into https://ansible.readthedocs.io/projects/dev-tools/

Yes, the molecule documentation is missing a description at the moment, but since the documentation above also explains molecule stuff, I think we will be able to achieve stuff this way.

What I know so far, this seems to be intended only for collections, so no solo roles anymore.

PR #4413, opened 2025-03-25
Replace internal calls to SystemExit with custom exception

Qalthos:


cidrblock (approved):
Looks good, approved including future changes to tests

PR #4416, opened 2025-03-31
Remove internal use of SystemExit part 2

Qalthos:


ssbarnea:
@Qalthos Is this getting closer to be merged?

PR #4492, opened 2025-08-01
Add kubevirt driver support to molecule.json

gokberkgunduzz:
**What this PR does / why we need it**:
This PR adds `"kubevirt"` to the list of supported drivers in `molecule.json`, allowing Molecule to properly validate and accept `driver.name: kubevirt` in scenario configurations.

This is required for users who wish to use the `molecule-kubevirt` driver (e.g., for testing Ansible roles in KubeVirt-based clusters). Without this addition, Molecule throws an error when parsing `molecule.yml`.

**Background context**:
- Molecule 25.x+ dropped legacy driver autodetection and enforces declared drivers from `molecule.json`
- This patch is based on `molecule==25.1.0`
- Driver is defined externally in the `molecule-kubevirt` plugin repo (see related PR)

**Release note**:
```release-note
Add "kubevirt" to allowed drivers in molecule.json to support external molecule-kubevirt driver

### DIFF
--- a/src/molecule/data/molecule.json
+++ b/src/molecule/data/molecule.json
@@ -87,7 +87,7 @@
                 "gce",
                 "libvirt",
                 "lxd",
-		"kubevirt"
+                "kubevirt"
               ]
             },
             {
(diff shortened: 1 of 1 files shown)


gokberkgunduzz:
Hi team — this PR is failing the label check, but I can't apply labels as a contributor.  
Could you please add the `enhancement` label so the workflow can pass?

Thanks!

PR #4553, opened 2025-09-15
Remove default `ansible.cfg` configuration option `ansible_managed` as it's deprecated

ziegenberg:
With a current ansible (2.19.2) we get the following deprecation warning:

```
[DEPRECATION WARNING]: DEFAULT_MANAGED_STR option. Reason: The
`ansible_managed` variable can be set just like any other variable, or a
different variable can be used.  Alternatives: Set the `ansible_managed`
variable, or use any custom variable in templates. This feature will be
removed from ansible-core version 2.23.
```


cidrblock:
thx

-brad

PR #4654, opened 2026-07-06
fix: remove outdated --driver-name reference

### LINKED ISSUE #4635: Documentation guide references removed option --driver
### Prerequisites

- [x] This was not already reported in the past (duplicate check)
- [x] It does reproduce it with code from main branch (latest unreleased version)
- [x] I include a minimal example for reproducing the bug
- [x] The bug is not trivial, as for those a direct pull-request is preferred
- [x] Running `pip check` does not report any conflicts
- [x] I was able to reproduce the issue on a different machine
- [x] The issue is not specific to any driver other than 'default' one

### Environment

```
molecule 26.4.0 using python 3.13 
    ansible:2.20.5
    containers:25.8.12 from molecule_plugins requiring collections: ansible.posix>=1.3.0 community.docker>=1.9.1 containers.podman>=1.8.1
    vagrant:25.8.12 from molecule_plugins
    ec2:25.8.12 from molecule_plugins
    podman:25.8.12 from molecule_plugins requiring collections: containers.podman>=1.7.0 ansible.posix>=1.3.0
    default:26.4.0 from molecule
    docker:25.8.12 from molecule_plugins requiring collections: community.docker>=3.10.2 ansible.posix>=1.4.0
    azure:25.8.12 from molecule_plugins
    gce:25.8.12 from molecule_plugins requiring collections: google.cloud>=1.0.2 community.crypto>=1.8.0
    openstack:25.8.12 from molecule_plugins requiring collections: openstack.cloud>=2.1.0
```

Image

### What happened

The docs at docs/guides/custom-image.md references an argument `--driver-name` which seems to no longer be supported (live version: [here](https://docs.ansible.com/projects/molecule/guides/custom-image/#:~:text=%2Dtime%20when-,%2D%2Ddriver%2Dname,-is%20docker.%20The)).

**Docs:**
Image

### Reproducing example

```yml
# Removed option, but still referenced in the docs
molecule init scenario --driver-name docker
```

Dotify71:
Fixes #4635 by removing the unsupported `--driver-name` argument reference from the custom image documentation.

## Summary by CodeRabbit

* **Documentation**
  * Updated the custom image guide with clearer instructions for creating and customizing scenario Docker image templates.
  * Added an example showing how to use the configured platform image and install dependencies.
  * Clarified that image building requires setting `pre_build_image` to `false`.

* **Tests**
  * Improved command output validation to accommodate formatting differences while confirming the expected platform information is present.


docs/guides/custom-image.md
```diff
@@ -19,7 +19,7 @@ items:
     >   >   scenario dir)
 
 The `Dockerfile.j2` template is generated at
-`molecule init scenario`-time when `--driver-name` is `docker`. The
+`molecule init scenario`-time. The
```
matthew-pakulski: > The `Dockerfile.j2` template is generated at `molecule init scenario`-time.

I don't think this is accurate anymore either. Only when `--driver-name` was supported did the `molecule init scenario` command create the `Dockerfile.j2` template file (presumably).

As a result, this documentation is lacking in instructing users how to use Dockerfile.j2 files as templates. An example would be ideal, with a useful or commonly used template substitution.

Dotify71: Thanks for pointing that out! You're completely right—since `--driver-name` was removed, the `Dockerfile.j2` template isn't generated automatically anymore. 

I've updated the documentation to remove that inaccurate statement. I also added a section explaining how users can create their own `Dockerfile.j2` template manually, along with a practical example demonstrating how to substitute the base image from `molecule.yml` (`{{ item.image }}`) and install basic dependencies. 

The changes have been pushed. Let me know if the example looks good!

matthew-pakulski (approved):

rockygeekz (approved):

PR #4663, opened 2026-07-21
feat: Support molecule login/molecule list for ansible-native scenarios

jirib:
## Summary

Ansible-native scenarios (no `platforms:` declared, instance lifecycle managed entirely by custom `create.yml`/`destroy.yml`) left `molecule login` and `molecule list` non-functional — both relied solely on the static `platforms:` list, which these scenarios don't have.

This adds a fallback: when no platforms are configured, both commands query the real inventory via `ansible-inventory`, reusing the exact same `--inventory` sources `ansible-playbook` itself is invoked with. No new molecule-specific config file is required — connection details already present in your inventory (e.g. written by `create.yml`) are used directly.

## Changes

- `driver/delegated.py` — resolve the same `--inventory` sources as `ansible-playbook`; add `ansible-inventory`-backed host discovery (`get_ansible_native_hosts`) and login-option resolution (`_get_inventory_login_options`), used as a fallback when `instance_config.yml` has no entry for the instance.
- `driver/base.py` — `status()` reports real host names from the inventory instead of a single blank placeholder when no platforms are configured.
- `command/login.py` — `execute()`/`_get_hostname()` use the same inventory fallback; clearer error messages when no hosts can be found by either mechanism.
- **Tests** — unit coverage for all of the above, including `ansible-inventory` subprocess success/failure handling.
- **Docs** — new "Logging Into / Listing Instances" section in `docs/ansible-native.md`, plus notes on the existing container/cloud examples flagging that dynamically-assigned connection details must be persisted to a `host_vars` file (not just registered as in-play facts) to be discoverable by `login`/`list`.

## Backward compatibility

`instance_config.yml` is still checked first if a create playbook populates it — the inventory query only kicks in as a fallback, so existing non-native scenarios are unaffected.

## Testing

Verified end-to-end against a real ansible-native scenario (custom libvirt-based instance lifecycle): confirmed `molecule list`/`molecule login` correctly resolve the real instance name and connect over SSH, both with and without `instance_config.yml` present. Full unit suite passes (`53` new/updated tests).

## Summary by CodeRabbit

* **New Features**
  * Added inventory-based host discovery for ansible-native flows without static platform instances, enabling `molecule login` and `molecule list` to use runtime-resolved connection details.
* **Bug Fixes**
  * Improved handling when hosts are unavailable or `--host` is specified.
  * Added clearer errors when SSH connection details cannot be determined.
* **Documentation**
  * Documented persisting runtime connection details and inventory lookup behavior.
* **Tests**
  * Expanded coverage for host discovery, inventory parsing, mapping, and error handling.

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -120,6 +120,7 @@ selectattr
 serviceaccount
 sessionfinish
 skipsdist
+sles
 ssbarnea
 stestr
 stopall
--- a/docs/ansible-native.md
+++ b/docs/ansible-native.md
@@ -355,7 +355,22 @@ The ansible-native approach leverages ansible collections for testing resource m
         host: "{% raw %}{{ hostvars[item].ansible_host }}{% endraw %}"
         delay: 5
       loop: "{% raw %}{{ groups['test_resources'] }}{% endraw %}"
+
+    - name: Persist connection details for molecule login/list
+      ansible.builtin.copy:
+        dest: "{% raw %}{{ molecule_scenario_directory }}{% endraw %}/inventory/host_vars/{% raw %}{{ item }}{% endraw %}/runtime.yml"
+        content: |
+          ansible_host: "{% raw %}{{ hostvars[item].ansible_host }}{% endraw %}"
+          ansible_port: "{% raw %}{{ hostvars[item].published_port }}{% endraw %}"
+      loop: "{% raw %}{{ groups['test_resources'] }}{% endraw %}"
 ```
+
+`podman_container` assigns the published port and IP at runtime - they only exist as
+in-memory facts for the duration of this play. `molecule login`/`molecule list` run
+`ansible-inventory` as a separate process later, so they only see connection details
+that were actually written to a file under the configured `--inventory` path, not
+facts that were merely registered or `add_host`-ed during `create.yml`. See
+[Logging Into / Listing Instances](#logging-into-listing-instances) below.
 
 ### Cloud Testing Resources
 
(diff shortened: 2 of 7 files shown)


tanwigeetika1618:
Thanks for this PR  @jirib. The design is clean and fills a real usability gap for ansible-native scenarios. The fallback logic is well-structured and preserves backward compatibility. A few things need addressing before we can merge:

  CI blockers:

  1. mypy — self._config.provisioner can be None (Ansible | None). Add a None guard before accessing .ansible_args (line 169) and .inventory_directory (line 183) in delegated.py.
  2. mypy — "host": None in test_login.py doesn't match the CommandArgs TypedDict which expects str for "host".
  3. cspell — "sles" is an unknown word (used in test hostname "sles15-efi"). Either add it to the project dictionary or rename the test hostname.
  4. pydoclint — _molecule_data_native fixture in test_delegated.py is missing a return section in its docstring.
  5. PR title — Needs a conventional commit prefix, e.g. feat: Support molecule login/molecule list for ansible-native scenarios.

  Suggestions (non-blocking):

  6. _molecule_data_native fixture is defined identically in both tests/unit/command/conftest.py and tests/unit/driver/test_delegated.py — consider deduplicating via a shared conftest.
  7. _get_inventory_login_options declares -> dict[str, str] but ansible_port from inventory can be an int — return type should be dict[str, str | int] or explicitly cast port to str.

tanwigeetika1618 (approved):

shvenkat-rh (approved):

